PR #37, opened 2022-09-28
[OSPP]Multi-task Joint Inference for Ianvs

shifan-Z:
What this PR does:
The PR is the proposal of Multi-task Joint Inference for Ianvs.

The joint inference is better than single task. So we want to add an example named multi-task joint inference in Ianvs and verify the effectiveness in Object Detection.


luosiqi:
lgtm

testyolov5/evaltest.py
```diff
@@ -0,0 +1,34 @@
+from tkinter import Widget
+import torch
+import numpy as np
+# # from yolo_model.eval import performance_evaluation_map_teacher
```
luosiqi, line 4: These annotated codes should be deleted.

luosiqi (commented):
It seems that most changes are not shown in Ianvs but in sedna lib. I suggest that you place sedna lib in the directory "example" in Ianvs. Also, lots of model files are uploaded to github which is not recommended. Please provide you model files with hypelinks in readme and configure their local urls in yaml.

examples/bdd/lifelong_learning_bench/testalgorithms/rfnet/RFNet/myutils/metrics.py
```diff
@@ -0,0 +1,146 @@
...
+        print("person       : %.6f" % (Acc[11] * 100.0), "%\t")
+        print("rider        : %.6f" % (Acc[12] * 100.0), "%\t")
+        print("car          : %.6f" % (Acc[13] * 100.0), "%\t")
+        print("truck        : %.6f" % (Acc[14] * 100.0), "%\t")
+        print("bus          : %.6f" % (Acc[15] * 100.0), "%\t")
+        print("train        : %.6f" % (Acc[16] * 100.0), "%\t")
+        print("motorcycle   : %.6f" % (Acc[17] * 100.0), "%\t")
+        print("bicycle      : %.6f" % (Acc[18] * 100.0), "%\t")
+        print("dynamic      : %.6f" % (Acc[19] * 100.0), "%\t")
+        print("stair        : %.6f" % (Acc[20] * 100.0), "%\t")
```
MooreZheng, line 45: For commercial services like sedna and kubeedge, use the log to output to screen instead of using print, which is not safe.

luosiqi (commented):
1. Change the target branch to lifelong-feature-n.
2. Add readme file to guide how to start multi-task joint inference.
3. Model files should be linked in readme but not directly uploaded to github. Also, add procedures in readme to download, configure and use these model files.
4. Move all the codes bdd/lifelong_learning_bench and make the folder structure clear.  Especially,  model files like .pt files and pth files downloaded from hyperlinks should be put in a single folder under bdd/lifelong_learning_bench/testalgorithms.
5. Directory bdd/lifelong_learning_bench/testalgorithms should not include rfnet.

examples/bdd/lifelong_learning_bench/testalgorithms/rfnet/rfnet_algorithm.yaml
```diff
@@ -0,0 +1,68 @@
...
+      # hyperparameters configuration for the python module; list type;
+      hyperparameters: 
+        # name of the hyperparameter; string type;
+        # origins of data; value is ["real", "sim"], this means that data from real camera and simulator.
+        - origins:
+            values:
+              - None
+              # - [ "real", "sim" ]
+              
+    # - type: "task_remodeling"
```
luosiqi, line 63: Annotated codes should be removed.

core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py
```diff
@@ -0,0 +1,229 @@
...
+        test result: numpy.ndarray
+        system metric info: dict
+            information needed to compute system metrics.
+
+        """
+
+        rounds = self.incremental_rounds
+        samples_transfer_ratio_info = self.system_metric_info.get(
+            SystemMetricType.SAMPLES_TRANSFER_RATIO.value)
+        #------my skip the training part----------
```
MooreZheng, line 87: The skip of training part is for testing but would infect other contributors' work. At least here an if-switch is needed instead of directly removing the code.

MooreZheng (commented):
1. The proposed code is only for testing, if we want to put it in sedna in the future, the training part is needed for a complete process.
2. I have other critical comments in the review meeting, see the meeting record as in  https://github.com/kubeedge/ianvs/pull/37#pullrequestreview-1154086305

PR #55, opened 2023-03-02
Fix CI issue: Raising too general exception.

jaypume:
Fix CI issue: Raising too general exception.


core/testcasecontroller/simulation_system_admin/simulation_system_admin.py
```diff
@@ -41,9 +41,9 @@ def check_host_docker():
             if install_docker.returncode == 0:
                 LOGGER.info("successfully installed docker")
             else:
-                raise Exception("install docker failed")
+                raise RuntimeError("install docker failed")
         except Exception as err:
-            raise Exception(f"install docker failed, error: {err}.") from err
```
luosiqi, line 46: When it catches an exception which means the exception has been raised, it is weird to raise another exception. Why does it need two exceptions? And I found many similar codes at another places.

PR #61, opened 2023-03-05
LFX Mentorship Task 1 and 2 for Anurag Bhandari (WhiteWolf47)

WhiteWolf47:
I have performed both the task with my utmost potential given the time constraint. I'm very thankful for giving me the oppurtunity to take part in this pretest. (i have made separate gh repos for both the tasks which I have mentioned in the md files I have added in https://github.com/WhiteWolf47/ianvs/blob/main/examples/cityscapes/lfx_pretest_WhiteWolf47/task1.md and https://github.com/WhiteWolf47/ianvs/blob/main/examples/cityscapes/lfx_pretest_WhiteWolf47/task2.md)


MooreZheng:
Thank you for the participation~ After LFX, this PR could be discarded, or replaced as other contents. 
Looking forward to seeing you in KubeEdge. :D

PR #63, opened 2023-03-16
Fix the bug in issue#57

hsj576:
**What type of PR is this?**

/kind bug

Fix the bug in issue#57

### DIFF
--- a/examples/curb-detection/lifelong_learning_bench/README.md
+++ b/examples/curb-detection/lifelong_learning_bench/README.md
@@ -13,6 +13,7 @@ Before using Ianvs, you might want to have the device ready:
 - Internet connection for GitHub and pip, etc
 - Python 3.6+ installed
   
+
 In this example, we are using the Linux platform with Python 3.8.5. If you are using Windows, most steps should still apply but a few like commands and package requirements might be different. 
 
 ## Step 1. Ianvs Preparation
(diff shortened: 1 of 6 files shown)


examples/curb-detection/lifelong_learning_bench/README.md
```diff
@@ -26,7 +26,7 @@ cd /ianvs #One might use another path preferred
 
 mkdir project
 cd project
-git clone https://github.com/kubeedge/ianvs.git   
+git clone -b feature-lifelong-n https://github.com/kubeedge/ianvs.git
```
MooreZheng, line 29: feature-lifelong-n  is used in the current version. After merged in the coming version, it should be fine with the previous clone.

hsj576: Ok, I agree with that.

examples/curb-detection/lifelong_learning_bench/testalgorithms/rfnet/RFNet/utils/metrics.py
```diff
@@ -134,7 +134,11 @@ def _generate_matrix(self, gt_image, pre_image):
         return confusion_matrix
 
     def add_batch(self, gt_image, pre_image):
+        gt_image = np.array(gt_image)
+        pre_image = np.array(pre_image)
         print(gt_image.shape, pre_image.shape)
+        if gt_image.shape != pre_image.shape:
+            pre_image = pre_image[0]
```
luosiqi, line 141: Changing the dimension of "pre_image" if the condition "!=" is true is not always applicable. Users might have different ways to process pre_image or gt_image before passing them to this function. I suggest that you modify the dimension of "pre_image" outside this function.

PR #75, opened 2023-07-13
AOA_CODE

Qzy568:
**What type of PR is this?**
The original code for AOA 

Fixes #

### DIFF
--- a/examples/AOA/.gitattributes
+++ b/examples/AOA/.gitattributes
@@ -1,0 +1,2 @@
+# Auto detect text files and perform LF normalization
+* text=auto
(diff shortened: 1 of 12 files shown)


MooreZheng:
Thank you so much for the PR. Good to see works on task definition. 

There are 2 commits incorrectly signed off. This means that the author(s) of these commits failed to include a Signed-off-by line in their commit message. To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).

Details can be found in https://github.com/kubeedge/ianvs/pull/75/checks?check_run_id=15005412951

PR #107, opened 2024-06-21
AOA_Code

Qzy568:
What type of PR is this?
The original code for AOA

Which issue(s) this PR fixes:


MooreZheng:
Thank you so much for the PR. Good to see works on task definition. 
1. The directory would be sth like examples/aoa/single_task_bench/TForest . 
2. A readme would be preferred to introduce how the algorithm is used. :D

PR #125, opened 2024-07-22
fix the Cl issue

### LINKED ISSUE #109: Resove the current CI issues
**What happened**:

CI is always failed even if we push an unrelated code. 

CI runs successfully

Look at the CI results of any PR. 

uvicorn version 0.14.0 in Sedna needs to be upgraded as I have also submit a issue in Sedna https://github.com/kubeedge/sedna/issues/432, because pip form 24.1 no longer support uvicorn 0.14.0. 

However I tried to use PIP 24.0 to cover this problem,

image

new error in CI occured like:

https://github.com/kubeedge/ianvs/actions/runs/9612294929/job/26512623096 

image

hsj576:
**What type of PR is this?**
/kind bug

Fix the Cl issue.

Fixes #109

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -24,7 +24,7 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install libgl1-mesa-glx -y
-          python -m pip install --upgrade pip
+          python -m pip install pip==24.0
           python -m pip install  pylint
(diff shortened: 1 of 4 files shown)


MooreZheng:
This PR aims to fix Issue https://github.com/kubeedge/ianvs/issues/109, considering that pip version error would be the key to the problem.

PR #135, opened 2024-08-16
fix: deprecate outdated interface of pandas

### LINKED ISSUE #134: Deprecate outdated interfaces of pandas used in rank.py

Some outdated interfaces of `pandas` are used in in `rank.py` . 

We can deprecate them by:
1. Implement `__get_all()` using a more elegant interface.
2. using `pd.concat()` to merge old DataFrame and new DataFrame.

The modified method should remain compatible with the version of pandas corresponding to Python 3.6.

**Why is this needed**:

Ianvs was originally designed for `pandas==1.1.5`, but the latest version is now `pandas==2.2.2`.Due to a major version update, some interfaces of pandas have been deprecated in the new version. 

Continuing to use these old interfaces will encounter errors on `Python>=3.8`.

1. `pd.np` has been deprecated in `pandas>=2.0.0` :
https://github.com/kubeedge/ianvs/blob/f2352ce018f04f398b1be0f37d0fa3cd11476626/core/storymanager/rank/rank.py#L208

```bash
AttributeError: module 'pandas' has no attribute 'np'
```

2. `append` has been deprecated in `pandas>=2.0.0`:
https://github.com/kubeedge/ianvs/blob/f2352ce018f04f398b1be0f37d0fa3cd11476626/core/storymanager/rank/rank.py#L171
```bash
AttributeError: 'DataFrame' object has no attribute 'append'
```

3. initializing `all_df` with `np.NAN` will cause `str` data missing:
https://github.com/kubeedge/ianvs/blob/f2352ce018f04f398b1be0f37d0fa3cd11476626/core/storymanager/rank/rank.py#L145
```bash
+------+-----------+-----+-----------+----------+-----------+----------+---------------------+----------------------+-------------------+------------------------+---------------------+------+-----+
| rank | algorithm | acc | edge-rate | paradigm | basemodel | apimodel | hard_example_mining | basemodel-model_name | basemodel-backend | basemodel-quantization | apimodel-model_name | time | url |
+------+-----------+-----+-----------+----------+-----------+----------+---------------------+----------------------+-------------------+------------------------+---------------------+------+-----+
|  1   |           | 0.6 |    0.6    |          |           |          |                     |                      |                   |                        |                     |      |     |
+------+-----------+-----+-----------+----------+-----------+----------+---------------------+----------------------+-------------------+------------------------+---------------------+------+-----+
```

4. Setting value by `df.[row_index][column]` will cause `SettingWithCopyWarning`.  This interface will be deprecated in `pandas>=3.0` in the future.
https://github.com/kubeedge/ianvs/blob/f2352ce018f04f398b1be0f37d0fa3cd11476626/core/storymanager/rank/rank.py#L148

Line 151, 154, 158, 165, 167 have the same issue, too.

```bash
./ianvs/core/storymanager/rank/rank.py:148: FutureWarning: ChainedAssignmentError: behaviour will change in pandas 3.0!
You are setting values through chained assignment. Currently this works in certain cases, but when using Copy-on-Write (which will become the default behaviour in pandas 3.0) this will never work to update the original DataFrame or Series, because the intermediate object on which we are setting values will behave as a copy.
A typical example is when you are setting values in a column of a DataFrame, like:

df["col"][row_indexer] = value

Use `df.loc[row_indexer, "col"] = values` instead, to perform the assignment in a single step and ensure this keeps updating the original `df`.

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy

  all_df.loc[i][0] = algorithm.name
```
5. Assigning values one by one reduces code readability and can be simplified.
https://github.com/kubeedge/ianvs/blob/f2352ce018f04f398b1be0f37d0fa3cd11476626/core/storymanager/rank/rank.py#L145-L167

6. Using whitspace as seperator in a CSV(**Comma**-Separated Values) file is weird.
https://github.com/kubeedge/ianvs/blob/f2352ce018f04f398b1be0f37d0fa3cd11476626/core/storymanager/rank/rank.py#L179

FuryMartin:
**What type of PR is this?**
/kind cleanup

Some deprecated interfaces of `pandas>=2.0.0` are used in `rank.py`, which prevents Ianvs from running on `Python>=3.9`.

For detailed modifications and reasons, please see #134.

Fixes #134

### DIFF
--- a/core/storymanager/rank/rank.py
+++ b/core/storymanager/rank/rank.py
@@ -141,42 +141,49 @@ class Rank:
 
     def _get_all(self, test_cases, test_results) -> pd.DataFrame:
         all_df = pd.DataFrame(columns=self.all_df_header)
+
         for i, test_case in enumerate(test_cases):
-            all_df.loc[i] = [np.NAN for i in range(len(self.all_df_header))]
-            # fill name column of algorithm
             algorithm = test_case.algorithm
-            all_df.loc[i][0] = algorithm.name
-            # fill metric columns of algorithm
-            for metric_name in test_results[test_case.id][0]:
-                all_df.loc[i][metric_name] = test_results[test_case.id][0].get(metric_name)
+            test_result = test_results[test_case.id][0]
 
-            # file paradigm column of algorithm
-            all_df.loc[i]["paradigm"] = algorithm.paradigm_type
+            # add algorithm, paradigm, time, url of algorithm
+            row_data = {
+                "algorithm": algorithm.name,
+                "paradigm": algorithm.paradigm_type,
+                "time": test_results[test_case.id][1],
+                "url": test_case.output_dir
+            }
 
-            # fill module columns of algorithm
-            for module_type, module in algorithm.modules.items():
-                all_df.loc[i][module_type] = module.name
+            # add metric of algorithm
+            row_data.update(test_result)
 
-            # fill hyperparameters columns of algorithm modules
-            hps = self._get_algorithm_hyperparameters(algorithm)
+            # add module of algorithm
+            row_data.update({
+                module_type: module.name
+                for module_type, module in algorithm.modules.items()
+            })
 
-            # pylint: disable=C0103
-            for k, v in hps.items():
-                all_df.loc[i][k] = v
-            # fill time and output dir of testcase
-            all_df.loc[i][-2:] = [test_results[test_case.id][1], test_case.output_dir]
+            # add hyperparameters of algorithm modules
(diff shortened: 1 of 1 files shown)


hsj576 (commented):
The code as a whole looks good, so you can talk about it at the next community meeting and see if anyone else has any questions.

FuryMartin:
> The code as a whole looks good, so you can talk about it at the next community meeting and see if anyone else has any questions.

OK, thanks.

hsj576 (approved):
/lgtm

Issue #145, opened 2024-09-20
Ianvs documents fails to update with Readthedoc website 


MooreZheng:
**What happened**:
Several documents have been committed but the Readthedoc website does not update. 
 
**What you expected to happen**:
Documents updated accordingly. 

**How to reproduce it (as minimally and precisely as possible)**:
This issue arises when new website documents are updated, e.g., documents in https://github.com/kubeedge/ianvs/issues/92. 

**Anything else we need to know?**:
The required .readthedocs.yaml configuration file was not found at repository, leading to the website building error. 
Reference: https://docs.readthedocs.io/en/stable/config-file/index.html

MooreZheng:
There are configuration and package-absent errors with Sphinx codes
1. Readthedocs.yaml is missing to build readthedocs projects. References are as https://stackoverflow.com/questions/60364789/how-do-i-correctly-start-using-readthedocs-yml and https://docs.readthedocs.io/en/stable/config-file/index.html
2.  sphinx_rtd_theme is no longer a hard dependency since version 1.4.0, which should be tackled too. 

Accordingly, here are solutions https://github.com/kubeedge/ianvs/pull/147 to this issue.

MooreZheng:
This issue is fixed via https://github.com/kubeedge/ianvs/pull/147 and https://github.com/kubeedge/ianvs/pull/173

Issue #92, opened 2024-01-29
Update for readthedoc


hsj576:
**What would you like to be documented**:
SCENARIOS
- [Cloud-Robotics](https://github.com/kubeedge/ianvs/blob/main/docs/proposals/scenarios/Cloud-Robotics/Cloud-Robotics_zh.md)

ALGORITHMS
- [Edge-Cloud Collaboration with Large Foundation Models](https://github.com/kubeedge/ianvs/blob/main/docs/proposals/algorithms/lifelong-learning/Real-Time%20IoT%20Perception%20Systems%20Based%20on%20Edge-Cloud%20Collaboration%20with%20Large%20Foundation%20Models.md)
- [Automatic Image Annotation Algorithm with Large Foundation Model](https://github.com/kubeedge/ianvs/blob/main/docs/proposals/algorithms/lifelong-learning/Toward%20Automatic%20Image%20Annotation%20Algorithm%20with%20Large%20Foundation%20Model%20Based%20on%20Edge-cloud%20Collaborative%20Lifelong%20Learning%20to%20Reproduce%20Open%20World%20Object%20Segmentation.md)
- [Class Incremental Semantic Segmentation](https://github.com/kubeedge/ianvs/blob/main/docs/proposals/algorithms/lifelong-learning/Implementation%20of%20a%20Class%20Incremental%20Learning%20Algorithm%20Evaluation%20System%20based%20on%20Ianvs.md)
- [Unknown Task Processing](https://github.com/kubeedge/ianvs/blob/main/docs/proposals/algorithms/lifelong-learning/Unknown_Task_Processing_Algorithm_based_on_Lifelong_Learning_of_Ianvs.md)
- [Unknown Task Recognition](https://github.com/kubeedge/ianvs/blob/main/docs/proposals/algorithms/lifelong-learning/Unknown_Task_Recognition_Algorithm_Reproduction_based_on_Lifelong_Learning_of_Ianvs.md)

NishantSinghhhhh:
@MooreZheng @hsj576 , can I be assigned this issue, I am highly motivated to work on this issue

NishantSinghhhhh:
@hsj576 sir, if no one is working on this one can I start writing the docs for it ?

NishantSinghhhhh:
@hsj576 ,  @MooreZheng  if no one is working on this one can I start writing the docs for it ?

NishantSinghhhhh:
[...]

PR #147, opened 2024-09-20
Fix Sphinx bugs to enable readthedocs

MooreZheng:
**What type of PR is this?**
/kind bug

There are configuration and package-absent errors that are fixed with Sphinx codes in this PR. 

1. Add readthedocs.yaml to enable Sphinx, which is necessary for building readthedocs projects. References are as https://stackoverflow.com/questions/60364789/how-do-i-correctly-start-using-readthedocs-yml and https://docs.readthedocs.io/en/stable/config-file/index.html

2. Add the try/except install session, add the extension and import the package for sphinx_rtd_theme. That is because [sphinx_rtd_theme is no longer a hard dependency since version 1.4.0](https://stackoverflow.com/questions/44940890/sphinx-rtd-theme-is-no-longer-a-hard-dependency-since-version-1-4-0)

Related to https://github.com/kubeedge/ianvs/issues/145 and https://github.com/kubeedge/ianvs/issues/92


hsj576 (approved):
/lgtm

Issue #98, opened 2024-05-09
Smart Coding benchmark suite: built on KubeEdge-lanvs


YangBrooksHan:
**What would you like to be added/modified:**
1. Build a collaborative code intelligent agent alignment dataset for LLMs:
   - The dataset should include behavioral trajectories, feedback, and iterative processes of software engineers during development, as well as relevant code versions and annotation information.
   - The dataset should cover code scenarios of different programming languages, business domains, and complexities.
   - The dataset should comply with privacy protection and intellectual property requirements, providing good accessibility and usability.
2. Design a code intelligent agent collaborative evaluation benchmark for LLMs:
   - The evaluation benchmark should include common tasks of code intelligent agents such as code generation, recommendation, and analysis.
   - Evaluation metrics should cover multiple dimensions including functionality, reliability, interpretability, etc., matching the feedback and requirements of software engineers.
   - The evaluation benchmark should assess the performance of LLMs in collaborative code intelligent agent tasks and provide a basis for further algorithm optimization.
3. Integrate the dataset and evaluation benchmark into the KubeEdge-Ianvs framework:
   - Incorporate the dataset and evaluation benchmark as part of the Ianvs framework, providing good scalability and integrability.
   - Ensure that the dataset and evaluation benchmark can efficiently run on edge devices within the Ianvs framework and seamlessly collaborate with other functional modules of Ianvs.
   - Release an upgraded version of the Ianvs framework and promote it to developers and researchers in the fields of edge computing and AI.

By implementing this project, we aim to provide crucial datasets and evaluation benchmarks for the further development of LLMs in the field of code intelligent agents, promote efficient collaboration between LLMs and software engineers in edge computing environments, and drive innovation and application of edge intelligence technology

**Why is this needed:**

[...]

Issue #152, opened 2024-09-27
Pylint Errors Due to Compatibility Issues with Updated Sedna in CI Environment


FuryMartin:
**What happened**:  
CI will attempt to install Sedna from `examples/resources/third_party/` when setting up the build environment. But If a new PR uses an updated version of Sedna and does not include the newer wheel in this directory, it will result in interface inconsistencies, causing Pylint to detect module errors (`unexpected-keyword-arg`, `no-name-in-module`, etc).

Possible errors may be reported:
```bash
# Note: Encounter when original interface is changed
************* Module core.testcasecontroller.algorithm.paradigm.base
core/testcasecontroller/algorithm/paradigm/base.py:130:19: E1123: Unexpected keyword argument 'cloud' in constructor call (unexpected-keyword-arg)
core/testcasecontroller/algorithm/paradigm/base.py:130:19: E1123: Unexpected keyword argument 'LCReporter_enable' in constructor call (unexpected-keyword-arg)

# Note: Encounter when new interface is added
************* Module core.testenvmanager.dataset.dataset
core/testenvmanager/dataset/dataset.py:21:0: E0611: No name 'JsonlDataParse' in module 'sedna.datasources' (no-name-in-module)
core/testenvmanager/dataset/dataset.py:21:0: E0611: No name 'JSONDataInfoParse' in module 'sedna.datasources' (no-name-in-module)
```

**What you expected to happen**:  
The code should pass Pylint checks without any errors, and the new interfaces from Sedna should be properly recognized.

**How to reproduce it (as minimally and precisely as possible)**:  
1. Develop ianvs `core/` with a newer version of Sedna that includes new features or bug fixes.
2. Avoid adding the newer wheel of Sedna into `examples/resources/third_party/`.
3. Submit a pull request and run continuous integration (CI).

**Anything else we need to know?**:  
Related PRs:
#144, #149 and other PRs associated with Sedna update

FuryMartin:
Here is my analysis of the issue.

There are two direct reasons for this issue:
1. Ianvs needs to modify Sedna when introducing some new features;
2. When submitting a PR, Ianvs must use the updated Wheel to pass CI's Pylint checks.

To pass the Pylint check, contributors can upload their modified Sedna Wheel to `examples/resources/third_party/`.

However, this approach has significant risks:

1. The Wheel uploaded to Ianvs is a binary file, and Git cannot track internal changes in detail, making it impossible to review modifications made to Sedna.
[...]

PR #155, opened 2024-09-29
基于KubeEdge-Ianvs的领域特定大模型基准测试的实现

safe-b:
This PR is the implementation of #98


MooreZheng (changes requested):
There are CI issues that remain to be resolved, see https://github.com/kubeedge/ianvs/actions/runs/11093238236?pr=155

MooreZheng (changes requested):
A similar CI issue is described in https://github.com/kubeedge/ianvs/issues/152 @FuryMartin is looking at it and @safe-b might want to talk to him

MooreZheng (changes requested):
As discussed at the routine meeting, we 
1. take a look at the file parameter issue on metadata.json issue and establish a formal issue when more time is needed to fix;
2. Replace words in Chinese to English;
3. Replace print using logger

PR #143, opened 2024-09-08
OSPP: Development of Federated Incremental Learning for Label Scarcity: Base on KubeEdge-Ianvs

Yoda-wu:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

This PR add two main feature: 
- federated learning paradigm
- federated class incremental learning paradigm

Besides this PR also add the example of these two paradigm.

**Which issue(s) this PR fixes**:
related issue: https://github.com/kubeedge/ianvs/issues/97

Fixes #

PR #162, opened 2024-10-30
Feat/fcil: Try to remove the print code in rank.py.

Yoda-wu:
Remove the print code in rank.py and testcase.py and metrics.py. So Sorry about that. 
 Uploading the full code of  PR https://github.com/kubeedge/ianvs/pull/143


MooreZheng:
Marked: This is a new version of https://github.com/kubeedge/ianvs/pull/143 after https://github.com/kubeedge/ianvs/pull/160

PR #169, opened 2025-01-02
Inconsistent with interface name 'initial_model_url' and environment …

### LINKED ISSUE #14: Inconsistent with interface name 'initial_model_url' and environment variable 'base_model_url'
While configuring `algorithm.yaml`, the pre-trained model url is specified as `initial_model_url`, however in `test_algorithms/basemodel.py` it is parsed and renamed as `base_model_url`. 

**Why is this needed**:
This inconsistency might be confusing for algorithm developers. May I possibly recommend revising these two names to the same one rather than keeping this inconsistency.

AryanNanda17:
**What type of PR is this?**
[good first issue](https://github.com/kubeedge/ianvs/labels/good%20first%20issue)
[kind/feature](https://github.com/kubeedge/ianvs/labels/kind%2Ffeature)

There is inconsistency which is confusing for algorithm developers. Removes the inconsistent naming schemes. 

Fixes #14


AryanNanda17:
@MooreZheng @jaypume 
please take a look at this PR.

PR #173, opened 2025-01-08
rename yaml to fix sphinx

MooreZheng:
What type of PR is this?
/kind bug

There are configuration and package-absent errors that are fixed with a Sphinx file name in this PR.
Note that the name of the readthedocs file shall start with ".", i.e., ".readthedocs.yaml", otherwise the system would have a "cat yaml" error.

Which issue(s) this PR fixes:
Related to https://github.com/kubeedge/ianvs/issues/145 and https://github.com/kubeedge/ianvs/issues/92


hsj576 (approved):
/lgtm

Issue #177, opened 2025-01-14
Domain-specific large model benchmarks: the edge perspective


MooreZheng:
What would you like to be added/modified:
Based on existing datasets, the issue aims to build an advanced benchmark for edge-oriented domain-specific large models on KubeEdge-Ianvs. It aims to help all Edge AI application developers validate and select the best-matched domain-specific large models. For Edge AI service providers, it also helps identify which scenarios, edge nodes, or even locations that could have the best performance or improvement for their models. This issue includes:
  - Domain-specific Large Model Benchmark for the edge, including test datasets, testing environments, and usage guidelines. 
  - (Advanced) Specific evaluation metrics. 
  - (Advanced) Research reports.

Why is this needed:
Common large-model benchmarks in the industry tend to focus on the cloud. As large models enter the era of scaled applications, the cloud has already provided infrastructure and services for these large models. Relevant customers have further proposed targeted application requirements on the edge side, including personalization, data compliance, and real-time capabilities, making AI services with cloud-edge collaboration a major trend. Different institutions from edges often build their own large models or knowledge bases. However, benchmark tests for domain-specific large models with edge data have not yet been well developed. Due to the data distribution at different edges, it is expected that the performance of general large models could be varied significantly at edges. This work aims to pingpoint those performance fluctuations for Edge AI services and applications. 

Recommended Skills:
KubeEdge-Ianvs, Python, LLMs

Useful links:
[Introduction to Ianvs](https://ianvs.readthedocs.io/en/latest/distributed-synergy-ai-benchmarking.html)
[Quick Start](https://ianvs.readthedocs.io/en/latest/guides/quick-start.html)
[How to test algorithms with Ianvs](https://ianvs.readthedocs.io/en/latest/guides/how-to-test-algorithms.html)
[Testing incremental learning in industrial defect detection](https://ianvs.readthedocs.io/en/latest/proposals/test-reports/testing-incremental-learning-in-industrial-defect-detection-with-pcb-aoi.html#testing-incremental-learning-in-industrial-defect-detection)
[Benchmarking for embodied AI](https://github.com/kubeedge/ianvs/blob/main/docs/proposals/scenarios/Cloud-Robotics/Cloud-Robotics_zh.md)
[KubeEdge-Ianvs](https://github.com/kubeedge/ianvs)
[Example LLMs Benchmark List](https://github.com/terryyz/llm-benchmark)
[...]

PR #144, opened 2024-09-09
OSPP: Implementation of Domain-Specific Large Model Benchmarking Based on KubeEdge-Ianvs

IcyFeather233:
This PR is the implementation of https://github.com/kubeedge/ianvs/issues/95

PR #113, opened 2024-07-06
Add llm-benchmarks proposal

IcyFeather233:
**What type of PR is this?**

OSPP proposal

**What this PR does / why we need it**:

Investigated various aspects of implementing the LLM Benchmark in Ianvs, introduced the integration plan of opencompass, and achieved the dataset map.

Issue #95, opened 2024-05-07
Domain-Specific Large Model Benchmarking Based on KubeEdge-Ianvs


MooreZheng:
**What would you like to be added/modified**:
Based on existing datasets, the issue aims to build a benchmark for domain-specific large models on KubeEdge-Ianvs. Namely, it aims to help all Edge AI application developers validate and select the best-matched domain-specific large models. This issue includes:

1. **Benchmark Dataset Map**: A mapping document, e.g., a table, includes test datasets and their download method for various specific domains.
2. **Large Model Interfaces**: Integrates open-source benchmarking projects like OpenCompass. Provides model API addresses and keys for online large model invocation.
3. **Domain-specific Large Model Benchmark**: Focuses on NLP or multimodal tasks. Constructs a suite for the government sector, including test datasets, evaluation metrics, testing environments, and usage guidelines.
4. *(Advanced)* **Industrial/Medical Large Model Benchmark**: Includes metrics and examples.
5. *(Advanced)* **Efficient Evaluation**: Enables concurrent execution of tasks with automatic request and result collection.
6. *(Advanced)* **Task Execution and Monitoring**: Visualizes the large model invocation process.

**Why is this needed**:
As large models enter the era of scaled applications, the cloud has already provided infrastructure and services for these large models. Relevant customers have further proposed targeted application requirements on the edge side, including personalization, data compliance, and real-time capabilities, making AI services with cloud-edge collaboration a major trend. However, there are currently two major challenges in terms of product definition, service quality, service qualifications, and industry influence: general competitiveness and customer trust problems. The crux of the matter is that the current large model benchmarking focuses on assessing general basic capabilities and fails to drive large model applications from an industry or domain-specific perspective. 

This issue reflects the real value of large models through industry applications from the perspectives of the domain-specific large model and cloud-edge collaborative AI, using industry benchmarks to drive the incubation of large model applications. Based on the collaborative AI benchmark test suite KubeEdge-Ianvs, this issue supplements the large model testing tool interface, provides matching test datasets, and constructs large model test suites for specific domains, e.g., for governments.

**Recommended Skills**:
KubeEdge-Ianvs, Python, LLMs

**Useful links**:
[Introduction to Ianvs](https://ianvs.readthedocs.io/en/latest/distributed-synergy-ai-benchmarking.html)
[Quick Start](https://ianvs.readthedocs.io/en/latest/guides/quick-start.html)
[...]

PR #186, opened 2025-03-12
add: LFX government proposal

IcyFeather233:
**What type of PR is this?**

/kind documentation

the proposal of [CNCF - KubeEdge: Domain-specific large model benchmarks: the edge perspective (2025 Term 1)](https://mentorship.lfx.linuxfoundation.org/project/e3fc44d9-9ddd-42e6-a9be-8f6c2a114672)


MooreZheng (changes requested):
Overall the RAG idea is impressive. It would be important for not only gov related LLM but also all domain-specific LLMs. 

There are some issues yet to discuss. 
1.  A stucture design is needed. Integration of RAG, e.g., API/ wheel etc, requires careful design.
2. RAG database and data format need preparation and further research. 
3. The test setting of using which part of the whole dataset, such as the RAG dataset, the fine-tune dataset, the few-shot prompt dataset, the known test set, and the unknown test set for the same or different nodes, is undoubtedly complicated. Besides, there is also the algorithm selection issue for RAG and LLM,  together with the above dataset issue, leading to a more complicated test setting.

MooreZheng (commented):
Marked: This proposal PR is related to https://github.com/kubeedge/ianvs/issues/177

MooreZheng (changes requested):
The new version integrates different RAG algorithms with the introduced feature discussed. Good to see that. 
 
The previous [comments](https://github.com/kubeedge/ianvs/pull/186#pullrequestreview-2681863608) are yet to discussed. Among all concerns, the test setting is especially important, which would affect the structure design. How many datasets, algorithms and hyper-parameters would be introduced and integrated? Will the integrations affect the ianvs core, e.g., test environment, and why? 

(BTW, the proposal PR seems to include non-necessary documents from other commits.  Besides, proposal PR shall not include code. Thus cleaning is needed.)

AryanNanda17:
Hello @IcyFeather233 
I noticed that some of the code changes in this PR are from the previous commits I did. It looks like your branch doesn’t include those commits, which might have caused the duplication. Could you check and rebase your branch on the latest main/develop branch?

MooreZheng (changes requested):
The proposal is further improved and that is well done. 

As discussed on the routine meeting, there are suggestions for the current version
1. The overall architecture design is added and looks pretty good. It would be appreciated if the proposed dataset, metrics, and schemes are further highlighted. 
2. The test setting is added which is cool. Moreover, we might try to add test data setting. That is, for exact-match knowledge base, complete knowledge base, and excluding knowledge base, the average accuracy could be taken for test samples in order to evaluate the learning ability, searching ability, and generalization ability for the tested models 
3. Add limitations and future work session to state that the current proposal for domain-specific large model mainly focuses on introducing RAG but has not yet covered the testing of incremental training, fine tuning, and few-shot learning.

MooreZheng (changes requested):
In this version, the setting and architecture are clear. Overall it looks fine and we are close to the final version. 

It would make the proposal better to consider the evaluation dimension, which shows the characteristics of a domain-specific model. That is, though showing arbitrary edge nodes (e.g., Beijing, Shanghai, etc) is clear, it might not always make sense to everyone (especially those who live far from the edge nodes ;D). 

One way to improve is to abstract the ability from measurement under different settings. For exact-match knowledge base, complete knowledge base, and excluding knowledge base, the average accuracy could be taken for test samples to evaluate the learning ability, searching ability, and generalization ability for the tested models.

hsj576 (approved):
/lgtm

MooreZheng:
Related issue and PR:  
Phase 1 Gov LLM https://github.com/kubeedge/ianvs/issues/95 https://github.com/kubeedge/ianvs/pull/113 https://github.com/kubeedge/ianvs/pull/144
Phase 2 Gov RAG https://github.com/kubeedge/ianvs/issues/177 https://github.com/kubeedge/ianvs/pull/186 https://github.com/kubeedge/ianvs/pull/196

Issue #197, opened 2025-05-08
Embodied Intelligence Benchmarking Framework for Industrial Manufacturing with KubeEdge-Ianvs


MooreZheng:
**Description**:
As industrial manufacturing accelerates its digital transformation through advancements in robotics, adaptive production lines, and smart testing systems, cloud-edge collaboration has emerged as a critical enabler for deploying embodied intelligence in complex operational environments. Contemporary industry requirements for embodied intelligence now extend beyond basic task execution to encompass multimodal perception-decision integration, dynamic environment adaptation, and distributed device orchestration. Existing benchmarking frameworks exhibit limitations in evaluating scenario-specific embodied attributes inherent to industrial settings. This initiative leverages the KubeEdge-Ianvs collaborative AI framework, integrating domain-specific test datasets, simulation environments, and quantitative metrics to establish a certified industrial-grade evaluation infrastructure for embodied intelligence systems. 

**Expected Outcomes**:
1. Develop an industrial-grade embodied intelligence dataset through systematic classification and reorganization of existing resources across four standardized task categories: 
- **Perception**: Multiview surface defect detection 
- **Locomotion**: Container manipulation and palletizing/depalletizing 
- **Operation**: Precision assembly (component insertion, multiprocess coordination), packaging automation (reel transfer, case sealing), and multitype object sorting 
- **Composite Tasks**: Flexible electronics testing and power system diagnostics 
Deliverable: Technical report detailing dataset architecture and annotation protocols 

2. Design standardized validation suites within KubeEdge-Ianvs, incorporating: 
- Curated scenario-specific datasets, converted into a unified format  
- Hardware-in-the-loop testbeds 
- Quantifiable performance metrics 

3. Deploy reference baseline algorithms using the developed validation suites to establish performance benchmarks within KubeEdge-Ianvs.

**Recommended Skills**：
Python; Benchmark; Dataset; Embodied Intelligence

**Useful Links**:
https://ianvs.readthedocs.io/en/latest/
https://github.com/kubeedge/community/blob/master/sig-ai/Distributed%20Synergy%20AI%20Benchmarking.md
https://github.com/kubeedge/ianvs/tree/main/examples/llm_simple_qa
https://github.com/kubeedge/ianvs/blob/main/docs/proposals/scenarios/llm-benchmarks/llm-benchmarks.md

**Scenario Examples**:
1. Perception
[...]

PR #219, opened 2025-07-14
Add files in docs/proposals/scenarios/

CGL-learner:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #197
Add
/kind documentation

Due to project requirements, a new folder(docs/proposals/scenarios/Industrial manufacturing) needs to be created to upload the corresponding markdown(docs/proposals/scenarios/Industrial manufacturing/197-fpc-assembly.md) and image files(docs/proposals/scenarios/Industrial manufacturing/Single_Task_Learning.png, docs/proposals/scenarios/Industrial manufacturing/ianvs.png, docs/proposals/scenarios/Industrial manufacturing/fpc_assembly.png, docs/proposals/scenarios/Industrial manufacturing/pybullet.png).

PR #218, opened 2025-07-11
 197-Embodied Intelligence Benchmarking Framework for Industrial Manufacturing with KubeEdg

CGL-learner:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #
Add industrial manufacturing folder 
/docs/proposals/scenarios/Industrial manufacturing
Submitted a proposal
/docs/proposals/scenarios/Industrial manufacturing/197-fpc-assembly.md
Submitted the corresponding image files in the proposal
/docs/proposals/scenarios/Industrial manufacturing/Single_Task_Learning.png
/docs/proposals/scenarios/Industrial manufacturing/fpc_assembly.png
/docs/proposals/scenarios/Industrial manufacturing/ianvs.png
docs/proposals/scenarios/Industrial manufacturing/pybullet.png

PR #222, opened 2025-07-16
197Pre-assembly of Injection Mold Inserts

CGL-learner:
**What type of PR is this?**

Fixes #197
/kind documentation
In the field of injection mold manufacturing, the pre assembly of mold inserts is a key process to ensure mass production efficiency. This scenario requires the cylindrical guide sleeve to be installed into the mold substrate to achieve:
Quick positioning, thermal expansion compensation, and error prevention design.
However, traditional manual installation suffers from inaccurate positioning, leading to mold edge flying issues. The proposal can improve installation efficiency and reduce mold scrap rate.
It can be effectively used for robot installation path planning, tolerance fit simulation verification, thermal deformation compensation research, and rapid changeover system development.
Through the PyBullet physics engine, it is possible to accurately simulate the contact mechanics behavior during the installation process of mold inserts, providing a reliable digital twin platform for actual production lines.


MooreZheng (changes requested):
Thank you for the effort! Some comments: 
1. The scenario looks quite synthetic and requires further polish to make sure the solution practical enough. A comparison between a real image and the simulation image needs to be provided in the proposal. The final scenario could be simple but should to be real enough. 
2. FPN is typically used in the object detection. What would be the AI tasks related to the proposed scenario? It looks like “image segmentation + distance measurement” is more the case than "object detection".  That would affect the baseline selection and thus important. 
3. Would we need to change the data format in ianvs if image or even more data types are introduced to Ianvs? The related justification shall be included in the proposal.

MooreZheng:
Related issue and PR:
https://github.com/kubeedge/ianvs/issues/197
https://github.com/kubeedge/ianvs/pull/260 https://github.com/kubeedge/ianvs/pull/217 https://github.com/kubeedge/ianvs/pull/218 https://github.com/kubeedge/ianvs/pull/219 https://github.com/kubeedge/ianvs/pull/222 https://github.com/kubeedge/ianvs/pull/243 https://github.com/kubeedge/ianvs/pull/244 https://github.com/kubeedge/ianvs/pull/245 https://github.com/kubeedge/ianvs/pull/272 https://github.com/kubeedge/ianvs/pull/269

PR #236, opened 2025-08-07
EXAMPLE/ llm_simple_qa made more readable and runnable .

### LINKED ISSUE #233: Clarification needed for dataset usage in example / llm_simple_qa .
**Introduction or background of this discussion**:
Ianvs includes a wide range of examples showcasing its benchmarking capabilities. As far as I see , Issue #194 keeps a track of all the issues which are yet to be verified and well documented. 

This issue is about the example 
**`llm_simple_qa`**

**Contents of this discussion**:

I've been trying to run the` llm_simple_qa` example and encountered some confusion regarding the dataset. The `README.md `for this example embeds the dataset content directly and asks users to manually copy it. This differs significantly from the cloud-edge-collaborative-inference-for-llm example, which provides a clear kaggle download command for its dataset.

Screenshot of README.md of  **llm_simple_qa**

Image

**Problem:**
Lack of clear documentation and clarify in **`llm_simple_qa`** example .In addition , a verification is needed where the example runs or not .

RONAK-AI647:
**KubeEdge Ianvs now has 25 examples and the number is still increasing.KubeEdge Ianvs then faces mounting usability issues due to dependency evolution and validation mechanisms. Most of the issues don't have a detailed `README.md` and the use of the hardcoded paths in the `urls `make it more confusing and complicated.**

**What type of PR is this?**
/kind cleanup
/kind documentation
/kind test
/kind design

This PR is for the example `llm_simple_qa` which makes it more robust and easy to understand and use. I have tested and verified the working of this example and documented the README.md well , so could anyone understand and run it. I have removed the hardcoded paths and used relative paths on points, organized the README.md in proper format and markdown , added the huge files which are created in the process like datasets and workspaces to the .gitignore , so could no user in the future messes up .

**This PR fixes issue #233**

Fixes #233


RONAK-AI647:
@MooreZheng , let me know if this pr is valid , or if any other change is needed , I will clear out all of it in one commit only.

examples/llm_simple_qa/README.md
```diff
@@ -2,83 +2,199 @@
...
+
+```bash
+# Install this example's dependencies
+pip install -r examples/llm_simple_qa/requirements.txt
+
+# Install the OpenCompass dependency
+pip install examples/resources/opencompass-0.2.5-py3-none-any.whl
+
+# Set the PYTHONPATH to include the project's root for metric discovery
+export PYTHONPATH="/mnt/c/Users/ronak/OneDrive/Desktop/ianvs:$PYTHONPATH"
```
hsj576: The PYTHONPATH is hardcoded to a user-specific local directory. This should be revised.

examples/llm_simple_qa/README.md
```diff
@@ -2,83 +2,199 @@
...
+```
 
-## OpenCompass Evaluation
+The accuracy score of `0.5` confirms that the framework works correctly. Results are saved in the `workspace` directory.
 
-### Prepare Environment
+---
 
-`pip install examples/resources/opencompass-0.2.5-py3-none-any.whl`
+### Contributing
```
hsj576: The link of Contributing and License are not correct. please fix it.

RONAK-AI647:
Thank you @hsj576 ,for the response....I will wait for @MooreZheng to see it once ...after that ,I will resolve everything in one commit ..reducing the length.

examples/llm_simple_qa/README.md
```diff
@@ -2,83 +2,199 @@
...
+{"question": "Which of the following shapes has the longest perimeter?", "answer": "C"}
 ```
 
-### Prepare Environment
+---
 
-You need to install the changed-sedna package, which added `JsonlDataParse` in `sedna.datasources`
+#### Step 4: Configure the Benchmark
 
-Replace the file in `yourpath/anaconda3/envs/ianvs/lib/python3.x/site-packages/sedna` with `examples/resources/sedna-with-jsonl.zip`
```
MooreZheng, line 66: Just to make sure, since we add sedna requirements, so we do not need to manually install Sedna wheel to read dataset now? But the pip install sedna version is different from examples/resources/sedna-with-jsonl.zip, which do not have jsonl support.

MooreZheng, line 66: Just to make sure, since we add sedna requirements, so we do not need to manually install Sedna wheel to read dataset now? But the pip install sedna version is different from examples/resources/sedna-with-jsonl.zip, which do not have jsonl support. I wonder whether that could lead to potential failure.

examples/llm_simple_qa/README.md
```diff
@@ -2,83 +2,199 @@
...
++------+-------------------------------+-----+--------------------+-----------+------------------------------------------------------------------------------------------------+
+```
 
-## OpenCompass Evaluation
+The accuracy score of `0.5` confirms that the framework works correctly. Results are saved in the `workspace` directory.
 
-### Prepare Environment
+---
 
-`pip install examples/resources/opencompass-0.2.5-py3-none-any.whl`
```
MooreZheng, line 79: Since we add the requirement of opencompass, so actually we do not need the wheel of opencompass to execute this example, am i right? Does the opencompass version matter?

examples/llm_simple_qa/README.md
```diff
@@ -2,83 +2,199 @@
...
+### Contributing
 
-### Run Evaluation
+-[Contributing](./CONTRIBUTING.md)
+-[Contribution to Documentation](./how-to-cotribute-documentation.md)
 
-`python run_op.py examples/llm/singletask_learning_bench/simple_qa/testalgorithms/gen/op_eval.py`
```
MooreZheng, line 83: Under the title of "run evaluation", why we need to delete the execution code of evaluation here? Besides, why we need to add contribution links for readers to "run evaluation"?

MooreZheng (changes requested):
Thank you so much for the effort and it looks good. Just try to tackle the comments from reviewers and gemini. (BTW @hsj576 is the reviewer of KubeEdge Ianvs and we also need one lgtm from him). I also have few comments

examples/llm_simple_qa/README.md
RONAK-AI647: Yeah , the version matters sir, actually i remember I ran the pip install requirements command but that installed the wrong version and I saw the errors and then I ran the specific version command .

examples/llm_simple_qa/README.md
RONAK-AI647: I see ,that was unintentional . The contribution and License and seperate sections.
image
And , you are very right. It's no of use there.  I will remove that section, if you say. I added that because it seemed like a good ending to the README file.

examples/llm_simple_qa/README.md
RONAK-AI647: That is a very good point @MooreZheng . But it worked for me and my system . 
The manual process of replacing a file in your site-packages directory is an old, non-standard way to manage dependencies. If you want me , I will give priority to this method .

MooreZheng: > That is a very good point @MooreZheng . But it worked for me and my system . The manual process of replacing a file in your site-packages directory is an old, non-standard way to manage dependencies. If you want me , I will give priority to this method.

We know the requirement method three years ago and already used in ianvs at the very beginning, but Sedna is a special case.  Ianvs is expected to have advanced features than Sedna (e.g., supporting text data format in the LLM era). The key point here is that we need to have a scheme to get Sedna SDK and Ianvs aligned, so that the Ianvs-incubated algorithm could be used in Sedna. Thus, the wheel method is designed on purpose, and Ianvs core developers are required to be aware of the need to synchronize Sedna and Ianvs in KubeEdge SIG AI. See more discussions in https://github.com/kubeedge/sedna/issues/477, which could help @RONAK-AI647 understand what was going on in the past.

examples/llm_simple_qa/README.md
MooreZheng: > Yeah , the version matters sir, actually i remember I ran the pip install requirements command but that installed the wrong version and I saw the errors and then I ran the specific version command .

So it is not proper to just delete this line of code. But opencompass is not like sedna, then the process could be easier: we just need to ensure a specific version of opencompass is installed.  We can see this line of "Run evaluation "pip install examples/resources/opencompass-0.2.5-py3-none-any.whl" is added in the new commit. It looks good to me now.

examples/llm_simple_qa/README.md
MooreZheng: Then we see that "Run evaluation" is not a useless part. We can see this line of "Run evaluation" code "python run_op.py examples/llm/singletask_learning_bench/simple_qa/testalgorithms/gen/op_eval.py" is added in the new commit. It looks good to me now.

examples/llm_simple_qa/README.md
MooreZheng: Thus, the Sedna version also matters. Besides a formal version of Sedna, note that Ianvs developers may even need an informal version of Sedna developed by themselves, in order to fulfill the advanced requirement (e.g., supporting text data format in the LLM era) which is not yet accepted by Sedna reviewers. So we prefer to reserve the wheel as a temporary patch of Sedna until all changes in the wheel are agreed and merged into Sedna.

MooreZheng (changes requested):
Good to see the revisions. Further comments on the wheel issue are detailed below

RONAK-AI647:
I have updated the pr @MooreZheng @hsj576 , I request for a last review , after that I rebase the pr to 1 commit. Thank You.

PR #269, opened 2025-09-25
OSPP Implemention: Embodied Intelligence Benchmarking Framework for Industrial Manufacturing with KubeEdge

CGL-learner:
**What type of PR is this?**

Fixes #197 
/kind documentation
By implementing palletizing scenarios in simulation, it can be well integrated with real scenes, further narrowing the gap with actual industrial scenarios and improving transferability.

### DIFF
--- a/examples/RoboDK Palletizing/singletask_learning_bench/readme.md
+++ b/examples/RoboDK Palletizing/singletask_learning_bench/readme.md
@@ -20,6 +20,7 @@
         - [Step 3. Run Ianvs](#step-3-run-ianvs)
             - [Run Example](#run-example)
         - [Results](#results)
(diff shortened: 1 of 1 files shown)


MooreZheng (changes requested):
1. Dataset polished to enrich test set; add metrics of map75 and map90; If the score gets too high, we can consider a more diverse test set, e.g., different objects or backgrounds. 
2. Different settings of models can be tested to provide conclusions, .e.g, using different sizes of datasets.
3. Provide a Readme. https://github.com/kubeedge/ianvs/tree/main/examples/cloud-edge-collaborative-inference-for-llm

hsj576 (approved):
/lgtm

PR #274, opened 2025-09-29
OSPP Proposal:Data Mixture Optimization Algorithm for VLA Fine-Tuning…

jwq6766336:
… Based on KubeEdge-Ianvs

**What type of PR is this?**

Fixes #


MooreZheng (commented):
It is in generally fine without significant errors. But as discussed in the routine meeting, there are still some tiny revisions 
1) The strategy need to be updated later, e.g., grid search. 
2) The directory need to be updated. 

@jwq6766336  will have a update later to make this proposal perfect.

MooreZheng (approved):
/approve

PR #316, opened 2026-02-03
Fail early on Windows for vLLM-based cloud-edge LLM example

### LINKED ISSUE #310: Cloud-edge LLM example fails on Windows due to undocumented vllm dependency
**What happened**:
The cloud-edge collaborative inference for LLM example fails on Windows during runtime execution. The pipeline crashes while loading the EdgeModel due to a missing `vllm` dependency.

The example should either run successfully on Windows or clearly document Linux-only dependencies and fail early with a clear error message.

**How to reproduce it**:
```bash
cd ianvs
python -m venv ianvs-env
ianvs-env\Scripts\activate

python setup.py install

mkdir workspace
ianvs -f examples\cloud-edge-collaborative-inference-for-llm\benchmarkingjob.yaml

```

**Anything else we need to know?**

OS: Windows 11
Python: 3.10.11
CPU-only machine

Error: ModuleNotFoundError: No module named 'vllm'

Source: examples/cloud-edge-collaborative-inference-for-llm/testalgorithms/query-routing/models/vllm_llm.py

vllm does not support Windows and is not documented as a required dependency

## Proposed Fix

**Fix 1 (Recommended & Realistic): Run the example on Linux**

* `vllm` is a hard dependency for the EdgeModel used in this example.
* `vllm` currently does **not support Windows**.
* Running the example on **Linux** (native Linux, WSL2, VM, or cloud VM) allows successful installation and execution.

Suggested environments:

* Ubuntu 20.04 / 22.04
* WSL2 (Ubuntu) on Windows
* Cloud VM (AWS / GCP / Azure)

---

**Fix 2 (Documentation Improvement): Clearly document OS constraints**

* Explicitly mention in the example README that:

  * `vllm` is required
  * The example is **Linux-only**
* Add a note in Quick Start or example README:

  > "This example requires `vllm` and currently does not support Windows."

---

**Fix 3 (Optional, Future Improvement): Provide a fallback EdgeModel**

* Allow disabling `vllm` via config
* Provide a HuggingFace-only / CPU-compatible EdgeModel
* This would improve accessibility for Windows and CPU-only users

---

iashutoshyadav:
### What this PR does
Adds an early platform check for the cloud-edge collaborative LLM example.

### Why this is needed
The example depends on vLLM, which does not support Windows or CPU-only
environments. Previously, running the example on Windows resulted in a
late ModuleNotFoundError. This change fails early with a clear, actionable
error message.

### Tested
- Windows 11 (CPU-only)
- Verified that the example fails early with a clear RuntimeError

Fixes #310

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,3 +143,8 @@ dmypy.json
 .idea/misc.xml
 .idea/workspace.xml
 .idea/vcs.xml
+# Virtual environment
+ianvs-env/
+
+# Local verification script
+test_vllm_import.py
--- a/examples/cloud-edge-collaborative-inference-for-llm/testalgorithms/query-routing/models/vllm_llm.py
+++ b/examples/cloud-edge-collaborative-inference-for-llm/testalgorithms/query-routing/models/vllm_llm.py
@@ -12,130 +12,122 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# ---------------------------------------------------------
+# Platform & dependency guard (MUST be first)
+# ---------------------------------------------------------
+import platform
+import importlib.util
+
+if platform.system() == "Windows":
+    raise RuntimeError(
+        "Cloud-edge LLM example requires Linux + GPU. "
+        "The vLLM backend is not supported on Windows."
+    )
+
+if importlib.util.find_spec("vllm") is None:
+    raise RuntimeError(
+        "vLLM is required for the cloud-edge LLM example but is not installed. "
+        "Please run this example on Linux with GPU support and install vllm."
+    )
+
+# ---------------------------------------------------------
+# Safe imports (only reached on supported platforms)
+# ---------------------------------------------------------
 import os
-import torch  
+import torch
 from vllm import LLM, SamplingParams
-from vllm.distributed.parallel_state import destroy_model_parallel, destroy_distributed_environment
+from vllm.distributed.parallel_state import (
+    destroy_model_parallel,
+    destroy_distributed_environment,
+)
 from models.base_llm import BaseLLM
 
+# ---------------------------------------------------------
+# Environment variables
+# ---------------------------------------------------------
 os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
 os.environ["TOKENIZERS_PARALLELISM"] = "true"
 os.environ["VLLM_LOGGING_LEVEL"] = "ERROR"
 
 device = "cuda" if torch.cuda.is_available() else "cpu"
 
+
 class VllmLLM(BaseLLM):
     def __init__(self, **kwargs) -> None:
-        """ Initialize the VllmLLM class
+        """Initialize the VllmLLM class"""
 
-        Parameters
-        ----------
-        kwargs : dict
-            Parameters that are passed to the model. Details can be found in the BaseLLM class.
-
-            Special keys:
-            - `tensor_parallel_size`: int, default 1. Number of tensor parallelism.
-            - `gpu_memory_utilization`: float, default 0.8. GPU memory utilization.
-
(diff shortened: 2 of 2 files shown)


Rishabh-git10 (commented):
**Recommendation:** Request Changes.

Hi @iashutoshyadav, thanks for adding this platform check. Implementing early environment validation prevents downstream crashes on unsupported setups.

**Reproduce:** I pulled this branch locally to verify the validation logic on a clean environment. The validation logic in `vllm_llm.py` is sound, but initialization errors block execution before reaching the guard:

1. Framework initialization immediately fails on `ModuleNotFoundError` for `colorlog` and `sedna`. These base prerequisites should be accounted for in the setup documentation.
2. Next, a crash occurs on `ModuleNotFoundError: No module named 'retry'` arising from `api_llm.py`. Please add `retry` to the example's `requirements.txt`.
3. The entry script `edge_model.py` imports all available backends on launch, causing a hard failure on a non-existent class.

**Logs:** 
```text
ImportError: cannot import name 'LadeSpecDecLLM' from 'models'
```
After removing the dead import, the HuggingFace fallback initialized successfully and bypassed the vLLM check, verifying the conditional pathing.
Please include `retry` in `requirements.txt` and remove the dead `LadeSpecDecLLM` import from `edge_model.py` to prevent the initialization path from short-circuiting.

Chidroopakanaparthy:
**Reproduction result:**

Checked out pr-316. Changes confined to `vllm_llm.py` only.

Key additions - three fail-early guards:
```python
if not torch.cuda.is_available():
    raise RuntimeError("CUDA is not available.")
```

Device detection on my machine:
``` bash
MPS: True | CUDA: False
```
This guard would hard-crash on Apple Silicon before a single inference runs.

**This pattern is correct for vLLM** (which genuinely requires CUDA) but must NOT be generalized. For `smart_coding`(Qwen2.5-Coder, supports MPS/CPU), the correct pattern should be:

```python
device = "mps" if torch.backends.mps.is_available() else \
         "cuda" if torch.cuda.is_available() else "cpu"
```

**Recommendation:** Merge for cloud-edge. But Don't apply this pattern to `smart_coding`.

barakahabimbola:
Reviewing this alongside #394 as part of LFX 2026 Term 3 pretest (full review in Discussion #964) — both target #430. This PR sidesteps the problem (fail early on Windows) rather than fixing the underlying dependency conflict that #394 actually resolves. If both PRs merge, Windows users get routed to the fail-fast path and never exercise #394's fix, which could mask the real solution. Recommend reconciling with #394 rather than merging independently. (Static/traced review, not executed.)

AmanBasu20:
The early-failure guard here is a real improvement — replacing a late, cryptic `ModuleNotFoundError` with an immediate, actionable `RuntimeError` on unsupported platforms. As `Chidroopakanaparthy` already flagged, the current `torch.cuda.is_available()` check is too strict and will incorrectly hard-fail on Apple Silicon (MPS) environments that could otherwise run.

Beyond that specific false-negative, the deeper issue is architectural placement: a hardware/platform guard implemented inside a single algorithm file (`vllm_llm.py`) can't be reused by any other example with the same constraint, and can't be queried before a benchmarking job even starts — a contributor only discovers the Windows/GPU requirement by running the job and hitting the `RuntimeError`. A `TestEnvManager`-level or dedicated `EnvironmentValidator` check, declared in the example's YAML and evaluated by Ianvs Core before dispatch, would fix the MPS false-negative (by making the requirement declarative and example-specific rather than hardcoded) and surface the requirement to a contributor before they invest time configuring the rest of the job.

Recommend merging the current fail-fast behavior for CUDA-only Linux environments as-is, but tracking the MPS false-negative and the architectural placement question as follow-up work rather than blocking this PR on them.

PR #318, opened 2026-02-04
feat: enhance config parsing with robust validation and type safety

### LINKED ISSUE #317: Improve safety and robustness of configuration parsing in TestEnv and Dataset

I propose refactoring the configuration parsing logic in the TestEnv and Dataset core classes to move away from dynamic attribute assignment (self.__dict__[k] = v).

**Why is this needed**:

The current reliance on __dict__ for configuration parsing presents several issues:

- Security Risks: It allow for uncontrolled attribute injection, where any key in a user supplied YAML file becomes an attribute on a core class instance. This could lead to unintended state manipulation.

- Low Maintainability: It is difficult to distinguish between the class's intended internal state and parameters dynamically added from configuration files.

yoursanonymous:
**What type of PR is this?**

This PR introduces a more secure and robust configuration parsing mechanism for the TestEnv and Dataset classes. It replaces the unsafe manual attribute assignment (self.__dict__[k] = v) with explicit field mapping, type validation, and automatic type casting.

/kind feature

This PR implements a robust configuration parsing and validation mechanism for the TestEnv and Dataset core components. Previously, these classes relied on insecure dynamic attribute assignment (self.__dict__[k] = v), which allowed arbitrary configuration keys to pollute class instances and potentially cause runtime errors.

Fixes #317

### DIFF
--- a/core/testenvmanager/dataset/dataset.py
+++ b/core/testenvmanager/dataset/dataset.py
@@ -87,17 +87,14 @@ class Dataset:
         for attr, value in config.items():
             if attr in fields_mapping:
                 expected_type = fields_mapping[attr]
-                if not isinstance(value, expected_type):
-                    try:
-                        casted_v = expected_type(value)
-                        setattr(self, attr, casted_v)
-                    except (ValueError, TypeError) as err:
-                        raise ValueError(
-                            f"dataset field '{attr}' expected {expected_type.__name__}, "
-                            f"got {type(value).__name__} (value={value}). Error: {err}"
-                        ) from err
-                else:
-                    setattr(self, attr, value)
+                try:
+                    casted_v = expected_type(value)
+                    setattr(self, attr, casted_v)
+                except (ValueError, TypeError) as err:
+                    raise ValueError(
+                        f"dataset field '{attr}' expected {expected_type.__name__}, "
+                        f"got {type(value).__name__} (value={value}). Error: {err}"
+                    ) from err
 
         self._check_fields()
 
--- a/core/testenvmanager/dataset/dataset.py
+++ b/core/testenvmanager/dataset/dataset.py
@@ -68,9 +68,36 @@ class Dataset:
             self._check_dataset_url(self.test_data_info)
 
     def _parse_config(self, config):
+        if not isinstance(config, dict):
+            raise ValueError(f"dataset config expected dict, got {type(config).__name__}.")
+
+        # Mapping of config keys to instance attributes
+        fields_mapping = {
+            "train_url": str,
+            "test_url": str,
+            "train_index": str,
+            "test_index": str,
+            "train_data": str,
+            "test_data": str,
+            "train_data_info": str,
+            "test_data_info": str,
+            "label": str,
+        }
+
         for attr, value in config.items():
-            if attr in self.__dict__:
-                self.__dict__[attr] = value
+            if attr in fields_mapping:
+                expected_type = fields_mapping[attr]
+                if not isinstance(value, expected_type):
+                    try:
+                        casted_v = expected_type(value)
+                        setattr(self, attr, casted_v)
+                    except (ValueError, TypeError) as err:
+                        raise ValueError(
+                            f"dataset field '{attr}' expected {expected_type.__name__}, "
+                            f"got {type(value).__name__} (value={value}). Error: {err}"
+                        ) from err
+                else:
+                    setattr(self, attr, value)
 
         self._check_fields()
 
--- a/core/testenvmanager/testenv/testenv.py
+++ b/core/testenvmanager/testenv/testenv.py
@@ -59,12 +59,13 @@ class TestEnv:
                 f" must be int type and not less than 2."
             )
 
-        if not isinstance(self.round, int) or self.round < 1:
-            raise ValueError(f"testenv round(value={self.round}) must be int type and not less than 1.")
-
-        if not isinstance(self.client_number, int) or self.client_number < 1:
+        round_val = getattr(self, "round", None)
+        if not isinstance(round_val, int) or round_val < 1:
+            raise ValueError(f"testenv round(value={round_val}) must be int type and not less than 1.")
+        client_number_val = getattr(self, "client_number", None)
+        if not isinstance(client_number_val, int) or client_number_val < 1:
             raise ValueError(
-                f"testenv client_number(value={self.client_number}) must be int type and not less than 1."
+                f"testenv client_number(value={client_number_val}) must be int type and not less than 1."
             )
 
     def _parse_config(self, config):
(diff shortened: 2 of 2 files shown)


yoursanonymous:
@MooreZheng, @Poorunga 
Can you please review this pull request

coldstarted:
LFX Term 02 : Task 2

## PR Review — `feat: enhance config parsing with robust validation and type safety`

### Recommendation: Merge with minor revisions

Solid improvement that replaces the unsafe `self.__dict__[k] = v` dynamic attribute assignment in both `TestEnv` and `Dataset` with explicit field mapping and typed casting. All modified files pass syntax checks and the change prevents arbitrary config keys from polluting object state. One issue before merge: unknown config fields are now silently dropped rather than raising an error — this could mask typos in testenv.yaml or dataset configs.

### Introduction

The original `_parse_config()` in both classes used `self.__dict__[k] = v` (or `self.__dict__[attr] = value` in Dataset) which allowed any arbitrary key from the YAML to be set as an instance attribute with no type checking. A typo like `metircs:` instead of `metrics:` would set `self.metircs = [...]` silently, with no error until much later when `self.metrics` turned out to be empty. This PR adds explicit `fields_mapping` dicts and type casting with proper error messages.

### Module Details

| `core/testenvmanager/testenv/testenv.py` | +40/-8 | Explicit field map for 6 testenv keys; type casting; `round` and `client_number` validation |
| `core/testenvmanager/dataset/dataset.py` | +28/-2 | Explicit field map for 9 dataset keys; type casting; dict-type guard on config |

**New `fields_mapping` in `TestEnv._parse_config`:**
```python
fields_mapping = {
    "model_eval": dict, "metrics": list, "incremental_rounds": int,
    "round": int, "client_number": int, "use_gpu": bool,
}
```
`bool` is handled specially — `str(v).lower() in ("true", "1", "yes")` — which is correct and more flexible than the original `bool(v)` (which would treat the string `"false"` as `True`).

**New `fields_mapping` in `Dataset._parse_config`:**
Covers all 9 known dataset URL field names (`train_url`, `test_url`, `train_index`, `test_index`, `train_data`, `test_data`, `train_data_info`, `test_data_info`, `label`).

**Issue — unknown config keys silently dropped:**
```python
for k, v in config_dict.items():
    if k == str.lower(Dataset.__name__):
        self.dataset = Dataset(v)
    elif k in fields_mapping:
        ...  # set attribute
    # else: silently ignored — no warning, no error
```
A typo in any testenv.yaml field name will now be silently ignored instead of the old behavior (set a garbage attribute). While the old behavior was also bad, the new behavior makes it even harder to debug config mistakes.

**Issue — new `round` and `client_number` checks in `_check_fields()`:**
```python
round_val = getattr(self, "round", None)
if not isinstance(round_val, int) or round_val < 1:
    raise ValueError(...)
client_number_val = getattr(self, "client_number", None)
if not isinstance(client_number_val, int) or client_number_val < 1:
    raise ValueError(...)
```
For the defaults `self.round = 1` and `self.client_number = 1` these pass correctly. But this changes the validation contract — the original `_check_fields()` only validated `incremental_rounds`.

### Evidence

Checked out `yoursanonymous:main` (PR #318 source branch). Both `testenv.py` and `dataset.py` pass `py_compile`. Traced the field mapping logic manually — all 6 `TestEnv` fields and 9 `Dataset` fields are correctly handled. Verified the `bool` casting logic: `str("false").lower() in ("true", "1", "yes")` = `False` — correct.

### Reproduce

```bash
git fetch upstream pull/318/head:pr-318
git checkout pr-318

python3 -m py_compile core/testenvmanager/testenv/testenv.py && echo 'testenv OK'
python3 -m py_compile core/testenvmanager/dataset/dataset.py && echo 'dataset OK'

# Verify bool casting fix
python3 -c "
v = 'false'
result = str(v).lower() in ('true', '1', 'yes')
print('use_gpu=false correctly parsed as:', result)  # should be False
"

# Demonstrate silent-drop of unknown keys (the issue)
python3 -c "
import sys; sys.path.insert(0, '.')
d = Dataset({'train_data': '/tmp/x.jsonl', 'typo_key': 'val'})
print('typo_key set?', hasattr(d, 'typo_key'))  # False — silently dropped
"

# Check unknown key detection
grep -n 'not in fields_mapping\|unknown\|LOGGER.warn' core/testenvmanager/testenv/testenv.py
```

### Logs

```
$ python3 -m py_compile testenv.py && echo 'testenv OK'
testenv OK
$ python3 -m py_compile dataset.py && echo 'dataset OK'
dataset OK

# bool fix confirmed:
use_gpu=false correctly parsed as: False

# Silent-drop issue:
typo_key set? False   ← no warning, no error — typo silently vanishes

# No logging of unknown keys:
(grep returns nothing)
```

### Problems Found

**1. Unknown config keys silently dropped — no warning**
A typo in testenv.yaml (e.g. `metircs:` instead of `metrics:`) is silently ignored. Fix: add `LOGGER.warning("testenv: unknown config key '%s' ignored", k)` in the `else` branch.

---

https://github.com/user-attachments/assets/839d682f-5a36-495a-871c-e0951315fc0a

subham-hq:
I came at this from the `Dataset` side and had already opened #758 before finding this PR — apologies for the overlap.

Some context that may be useful here.

**On @coldstarted's May 27 review point**

The concern raised there — that unknown config fields would be silently dropped after this change, potentially masking typos — is the same gap #758 addresses, though only for `Dataset`. It follows the fix suggested in that review (a `LOGGER.warning` in the `else` branch) rather than raising, so it stays backward compatible.

**Current `TestEnv` behaviour, as supporting evidence for this PR**

Worth recording the pre-#318 state precisely, since it makes the case for this change stronger than "unsafe assignment." The `else` branch has no membership check at all:

```python
else:
    self.__dict__[k] = v
```

So a typo doesn't merely get dropped — it creates an attribute, and validation then runs against the default instead:

```
$ python -c "
from core.testenvmanager.testenv.testenv import TestEnv
t = TestEnv({'testenv': {
    'metrics': [{'name': 'acc'}],
    'incremental_round': 10,
    'totally_made_up_key': 'xyz',
}})
print('incremental_rounds =', t.incremental_rounds)
print('incremental_round  =', getattr(t, 'incremental_round', 'ABSENT'))
print('made up key        =', getattr(t, 'totally_made_up_key', 'ABSENT'))
"
incremental_rounds = 2
incremental_round  = 10
made up key        = xyz
```

`incremental_rounds` keeps its default of 2, `_check_fields` validates that default and passes, and the value the user actually wrote sits on an attribute nothing reads — no error at any point. The explicit field mapping in this PR removes that failure mode entirely.

**On the overlap**

#758 touches only `Dataset._parse_config` and is warning-only, so the two don't conflict. If this PR merges, I'm happy to rebase #758 onto it so the warning slots into the new `else` branch, or to close #758 if you'd rather fold the warning in here directly — whichever you and the maintainers prefer.

One data point that may bear on the warn-vs-raise question: the warning in #758 fires on `name:`, which appears under `dataset:` in 8 example configs (cifar100 ×5, PIPL, federated-llm) and is not a field `Dataset` reads. So there's at least one pre-existing case of a real, widely-copied key being dropped, which may argue for warning rather than raising at first.

Related: #589, #707

Arijit429:
**Code review — from LFX 2026 Term 3 pre-test Discussion #862**

**Recommendation: Minor Revision**

New finding not in any existing review: `testenv.py`'s fix in this PR uses `config.get()` without first checking if `config` is a dict. When `config` is `None` (which `yaml2dict()` returns for an empty YAML file), this raises `AttributeError` instead of the intended `ValueError`:

```python
None.get('testenv')
# -> AttributeError: 'NoneType' object has no attribute 'get'
```

Reproduced directly. `yaml2dict('')` returning `None` confirmed:

```python
yaml.load('', Loader=yaml.SafeLoader)  # -> None
```

Additionally: this PR conflicts with PR #854 (same lines, same method in testenv.py) and PR #320 (companion PR, same author, algorithm.py). Merge coordination needed — PR #854's version adds the missing `isinstance` guard and is more complete.

PR #376, opened 2026-03-19
Proposal For LFX Term 1 2026 [Cloud-Edge Simulation Benchmark for LLM Speculative Decoding in KubeEdge-Ianvs]

zjteee:
**What type of PR is this?**

Fixes #


MooreZheng (changes requested):
As discussed, overall it is good. Please add 1) an architecture to show the three possible procedures; 2) highlight the changed parts and why we need to change it; 3) analyze the changed effect (what we can do to avoid effect on other examples)

MooreZheng (changes requested):
It is close to the final version now. Just some clarifications on the figure:  
1. New module novelty of  draft and verify module  v.s. existing inference modules. Input is similar, both support  model decoupling (lile hot plugging of multilple models). But output is a little bit different, draft and verify modules are token-wise but existing inference modules are batch-wise.
2. Need more detailed framework. For example, edge inference called by draft, cloud inference called by verify. 
3. Need to demonstrate the algorithm decoupling, by showing relationship between framework and implementation. 1) The implement lies in example and the framework lies in core. 2) The implementation functions inherits the framework functions.

hsj576 (changes requested):
This PR currently contains too many commits. Please squash them into a single commit.

MooreZheng (commented):
/lgtm

MooreZheng (approved):
/approve

Issue #362, opened 2026-02-16
[Bug Report] government_rag Example: 7 Execution-Blocking Bugs Prevent Any User From Running This Benchmark


khansalman12:
## Background

I cloned the Ianvs repo and went through the examples listed in [#194](https://github.com/kubeedge/ianvs/issues/194). The [`government_rag`](https://github.com/kubeedge/ianvs/tree/main/examples/government_rag) example caught my attention — it benchmarks RAG-enhanced LLMs on Chinese government policy Q&A, comparing four modes (no-RAG, global RAG, local RAG, cross-province RAG) using LangChain + ChromaDB.

This example is **not verified** in #194, and after going through the source code, I found **7 bugs** that completely block execution. As it stands, no one can run this example out of the box.

### Bug 1 — Hardcoded local paths (🔴 Crashes immediately)

The config and source files are full of paths from the original developer's machine. Running on any other system immediately throws `FileNotFoundError`.

| File | Hardcoded Path |
|------|---------------|
| `benchmarkingjob.yaml:5` | `/home/icyfeather/project/ianvs/workspace` |
| `testenv.yaml:5` | `/home/icyfeather/Projects/ianvs/dataset/gov_rag/example.jsonl` |
| `testenv.yaml:7` | `/home/icyfeather/Projects/ianvs/dataset/gov_rag/metadata.json` |
| `basemodel.py:160,179,181,184` | `/home/icyfeather/models/bge-m3` |

### Bug 2 — `self.all_locations` is never defined (🔴 Crashes at runtime)

In `basemodel.py:183`, the `[other]` RAG mode references an attribute that was never initialized:
```python
all_locations = set(self.all_locations)  # never set anywhere in the class
```
Since `predict()` creates tasks for all four RAG modes per query, this `AttributeError` kills the entire benchmark run.

### Bug 3 — Hardcoded API key placeholders (🔴 Auth always fails)

All three LLM backends ship with dummy credentials that obviously cannot authenticate:
- `basemodel.py:62` — `api_key="<DeepSeek API Key>"`
- `basemodel.py:102` — `"Bearer <token>"`
- `basemodel.py:122` — `client_id=[API Key]&client_secret=[Secret Key]`

There is no env-variable or config-file mechanism for users to plug in their own keys.

### Bug 4 — Broken import path (🟡 ModuleNotFoundError)

[...]

PR #384, opened 2026-04-11
Fix ZeroDivisionError in metric evaluation across examples

### LINKED ISSUE #383: ZeroDivisionError in `acc.py` Metric Calculation
## Problem Description

A `ZeroDivisionError` occurs in `acc.py` when calculating metrics like accuracy, where the denominator (total predictions or samples) is zero. This is common in machine learning when no predictions are made or all samples belong to one class.

```python
# Example of the error (before fix)
def calculate_accuracy(correct, total):
    return correct / total  # Raises ZeroDivisionError if total == 0

# Usage that fails
accuracy = calculate_accuracy(0, 0)  # ZeroDivisionError: division by zero
```

## Solution 1: Conditional Check (Recommended for ML Metrics)

Check if denominator is zero before division and return 0.0 for accuracy (standard convention).

```python
def calculate_accuracy(correct: int, total: int) -> float:
    """
    Calculate accuracy safely.
    
    Args:
        correct: Number of correct predictions
        total: Total number of predictions
        
    Returns:
        Accuracy as float (0.0 if total is 0)
    """
    if total == 0:
        return 0.0
    return correct / total
```

## Solution 2: Using `try-except`

Handle the exception gracefully.

```python
def calculate_accuracy_safe(correct: int, total: int) -> float:
    try:
        return correct / total
    except ZeroDivisionError:
        return 0.0
```

## Solution 3: Complete `acc.py` Example

Here's a full metrics module for your ML projects:

```python
# acc.py - Safe Metric Calculations
from typing import Optional

def accuracy(correct: int, total: int) -> float:
    """Calculate accuracy (0.0 if no predictions)."""
    return 0.0 if total == 0 else correct / total

def precision(tp: int, fp: int) -> float:
    """Calculate precision."""
    return 0.0 if (tp + fp) == 0 else tp / (tp + fp)

def recall(tp: int, fn: int) -> float:
    """Calculate recall."""
    return 0.0 if (tp + fn) == 0 else tp / (tp + fn)

def f1_score(precision_score: float, recall_score: float) -> float:
    """Calculate F1 score."""
    if precision_score + recall_score == 0:
        return 0.0
    return 2 * (precision_score * recall_score) / (precision_score + recall_score)

# Example usage
if __name__ == "__main__":
    print(accuracy(5, 10))      # 0.5
    print(accuracy(0, 0))       # 0.0 (no error!)
    print(precision(3, 0))      # 1.0
    print(precision(0, 5))      # 0.0
```

## Usage in Your React/Python Projects

```python
# In your FastAPI backend or ML script
correct_predictions = 0
total_predictions = 0  # Edge case from empty dataset

acc = accuracy(correct_predictions, total_predictions)
print(f"Accuracy: {acc:.2%}")  # Output: Accuracy: 0.00%
```

## Best Practices

1. **Always validate denominators** in metric functions
2. **Return 0.0 for accuracy/precision/recall** when undefined
3. **Add type hints** for better IDE support in VS Code
4. **Test edge cases**: `(0,0)`, `(5,0)`, `(0,5)`

Copy this into your `acc.py` file and the error is fixed!

ARYANPATEL-BIT:
**Status: Fixed ✅**

## 🎯 Issue Fixed
**Fixes #383**

## 🚨 Problem
During metric evaluation, accuracy calculation divides by `len(same_elements)`. Empty lists from prediction edge-cases cause `ZeroDivisionError` and crash the training/evaluation pipeline.

## 📁 Affected Files
examples/llm_simple_qa/testenv/acc.py
examples/government/singletask_learning_bench/objective/testenv/acc.py
examples/government_rag/singletask_learning_bench/testenv/acc.py
examples/cloud-edge-collaborative-inference-for-llm/testenv/accuracy.py

text

## 🔧 The Fix

**Before (CRASH):**
```python
acc = sum(same_elements) / len(same_elements)
```

**After (SAFE):**
```python
acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
```

## 💾 Complete Fixed Function

```python
def calculate_accuracy(same_elements):
    """Safe accuracy calculation for empty prediction lists."""
    return sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
```

## ✅ Test Cases
```python
# All now work without crashing:
calculate_accuracy([])                          # 0.0
calculate_accuracy()                   # 0.666...[1]
calculate_accuracy([True, False, True])         # 0.666...
```

### DIFF
--- a/examples/cloud-edge-collaborative-inference-for-llm/testenv/accuracy.py
+++ b/examples/cloud-edge-collaborative-inference-for-llm/testenv/accuracy.py
@@ -54,6 +54,6 @@ def acc(y_true, y_pred):
     same_elements = [y_pred[i] == y_true[i] for i in range(len(y_pred))]
 
     # 计算相同元素的数量
-    acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
+    acc = sum(same_elements) / len(same_elements) if same_elements else 0.0
 
     return round(acc * 100, 2)
--- a/examples/cloud-edge-collaborative-inference-for-llm/testenv/accuracy.py
+++ b/examples/cloud-edge-collaborative-inference-for-llm/testenv/accuracy.py
@@ -54,6 +54,6 @@ def acc(y_true, y_pred):
     same_elements = [y_pred[i] == y_true[i] for i in range(len(y_pred))]
 
     # 计算相同元素的数量
-    acc = sum(same_elements) / len(same_elements)
+    acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
 
     return round(acc * 100, 2)
--- a/examples/government/singletask_learning_bench/objective/testenv/acc.py
+++ b/examples/government/singletask_learning_bench/objective/testenv/acc.py
@@ -34,6 +34,6 @@ def acc(y_true, y_pred):
         
     same_elements = [y_pred[i] == y_true[i] for i in range(len(y_pred))]
 
-    acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
+    acc = sum(same_elements) / len(same_elements) if same_elements else 0.0
     
     return acc
--- a/examples/government/singletask_learning_bench/objective/testenv/acc.py
+++ b/examples/government/singletask_learning_bench/objective/testenv/acc.py
@@ -34,6 +34,6 @@ def acc(y_true, y_pred):
         
     same_elements = [y_pred[i] == y_true[i] for i in range(len(y_pred))]
 
-    acc = sum(same_elements) / len(same_elements)
+    acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
     
     return acc
--- a/examples/government_rag/singletask_learning_bench/testenv/acc.py
+++ b/examples/government_rag/singletask_learning_bench/testenv/acc.py
@@ -47,7 +47,7 @@ def acc_model(y_true, y_pred):
             real_locations.append(y_location[i])
 
     same_elements = [get_last_letter(real_y_pred[i]) == real_y_true[i] for i in range(len(real_y_pred))]
-    global_acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
+    global_acc = sum(same_elements) / len(same_elements) if same_elements else 0.0
 
     province_acc = {}
     for i in range(len(real_y_pred)):
@@ -108,7 +108,7 @@ def acc_global(y_true, y_pred):
             real_locations.append(y_location[i])
 
     same_elements = [get_last_letter(real_y_pred[i]) == real_y_true[i] for i in range(len(real_y_pred))]
-    global_acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
+    global_acc = sum(same_elements) / len(same_elements) if same_elements else 0.0
 
     province_acc = {}
     for i in range(len(real_y_pred)):
@@ -169,7 +169,7 @@ def acc_local(y_true, y_pred):
             real_locations.append(y_location[i])
 
     same_elements = [get_last_letter(real_y_pred[i]) == real_y_true[i] for i in range(len(real_y_pred))]
-    global_acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
+    global_acc = sum(same_elements) / len(same_elements) if same_elements else 0.0
 
     province_acc = {}
     for i in range(len(real_y_pred)):
@@ -230,7 +230,7 @@ def acc_other(y_true, y_pred):
             real_locations.append(y_location[i])
 
     same_elements = [get_last_letter(real_y_pred[i]) == real_y_true[i] for i in range(len(real_y_pred))]
-    global_acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
+    global_acc = sum(same_elements) / len(same_elements) if same_elements else 0.0
 
     province_acc = {}
     for i in range(len(real_y_pred)):
--- a/examples/government_rag/singletask_learning_bench/testenv/acc.py
+++ b/examples/government_rag/singletask_learning_bench/testenv/acc.py
@@ -47,7 +47,7 @@ def acc_model(y_true, y_pred):
             real_locations.append(y_location[i])
 
     same_elements = [get_last_letter(real_y_pred[i]) == real_y_true[i] for i in range(len(real_y_pred))]
-    global_acc = sum(same_elements) / len(same_elements)
+    global_acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
 
     province_acc = {}
     for i in range(len(real_y_pred)):
@@ -108,7 +108,7 @@ def acc_global(y_true, y_pred):
             real_locations.append(y_location[i])
 
     same_elements = [get_last_letter(real_y_pred[i]) == real_y_true[i] for i in range(len(real_y_pred))]
-    global_acc = sum(same_elements) / len(same_elements)
+    global_acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
 
     province_acc = {}
     for i in range(len(real_y_pred)):
@@ -169,7 +169,7 @@ def acc_local(y_true, y_pred):
             real_locations.append(y_location[i])
 
     same_elements = [get_last_letter(real_y_pred[i]) == real_y_true[i] for i in range(len(real_y_pred))]
-    global_acc = sum(same_elements) / len(same_elements)
+    global_acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
 
     province_acc = {}
     for i in range(len(real_y_pred)):
@@ -230,7 +230,7 @@ def acc_other(y_true, y_pred):
             real_locations.append(y_location[i])
 
     same_elements = [get_last_letter(real_y_pred[i]) == real_y_true[i] for i in range(len(real_y_pred))]
-    global_acc = sum(same_elements) / len(same_elements)
+    global_acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
 
     province_acc = {}
     for i in range(len(real_y_pred)):
--- a/examples/llm_simple_qa/testenv/acc.py
+++ b/examples/llm_simple_qa/testenv/acc.py
@@ -35,6 +35,6 @@ def acc(y_true, y_pred):
         
     same_elements = [y_pred[i] == y_true[i] for i in range(len(y_pred))]
 
-    acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
+    acc = sum(same_elements) / len(same_elements) if same_elements else 0.0
     
     return acc
--- a/examples/llm_simple_qa/testenv/acc.py
+++ b/examples/llm_simple_qa/testenv/acc.py
@@ -35,6 +35,6 @@ def acc(y_true, y_pred):
         
     same_elements = [y_pred[i] == y_true[i] for i in range(len(y_pred))]
 
-    acc = sum(same_elements) / len(same_elements)
+    acc = sum(same_elements) / len(same_elements) if len(same_elements) > 0 else 0.0
     
     return acc


AswaniSahoo:
### Review: PR #384

**Recommendation: Merge as-is.**

**Scope:** Example-only. Adds `if same_elements else 0.0` guard to `acc.py` / `accuracy.py` across 4 examples — 7 call sites total (government objective ×1, government_rag ×4 across `acc_model`/`acc_global`/`acc_local`/`acc_other`, cloud-edge LLM ×1, llm_simple_qa ×1). No core changes.

The `ZeroDivisionError` itself is what the PR description targets: `sum(same_elements) / len(same_elements)` divides by zero on empty `y_pred`. gemini-code-assist's review on this PR confirmed the guard is correct and suggested using the idiomatic `if same_elements` form, which the PR already uses.

**What the PR and prior reviews do not address:**

1. **Silent failure mode introduced by the `0.0` fallback.** Returning `0.0` on empty predictions eliminates the crash but produces a result indistinguishable from a model that genuinely scored 0%. A pipeline failure (data loading broke, prediction stage threw silently, etc.) and a true 0% accuracy run now both surface as `acc = 0.0` in the leaderboard CSV with no warning. Before this PR the crash was observable; after this PR the failure is silent and will pass CI. Suggested follow-up so downstream tooling can tell "no data" apart from "0% accuracy":

```python
if not same_elements:
    LOGGER.warning("acc(): empty prediction set, returning NaN")
    return float('nan')
acc = sum(same_elements) / len(same_elements)
```

2. **Parallel pattern unfixed.** Issue #401 (open since April 2026, @ARYANPATEL-BIT) documents the same `ZeroDivisionError` class in `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/f1_score.py:51`. The F-1 calculation `2*(precision*recall)/(precision+recall)` divides by zero when both metrics are zero. PR #384 does not touch `f1_score.py`. A follow-up applying the same guard pattern there would close the systemic gap rather than just the four `acc.py` files.

Neither observation blocks merge. PR #384 as-is is a correct narrow fix and should land.

**Verification:**

14-verify-pr384-pr386

Reproduced the original crash (`ZeroDivisionError: division by zero` at `sum(same_elements) / len(same_elements)` with `same_elements = []`) and confirmed the guard catches it (returns 0.0, no exception). Manually invoked the guarded path with `same_elements = []` — returned `0.0` with no log line, confirming the silent-failure observation above.

Sahil-u07:
## Review - PR #384: ZeroDivisionError Guard in government_rag Evaluation

### Recommendation: Merge with one clarification

The guard against `ZeroDivisionError` in the evaluation metric is necessary and correct. Without it, a benchmark run that processes zero valid responses (for example when the LLM API key is wrong, when `get_relevant_documents()` raises a LangChain deprecation error mid-run, or when ChromaDB fails to load) would crash at the metric step with an unhelpful `ZeroDivisionError` after potentially hours of partial processing.

The fix is correctly scoped - it only touches the metric calculation and does not modify any other component.

**One gap worth addressing**

When the denominator is zero, the guard silently returns `0.0`. This makes it impossible to distinguish between two very different situations:
- All responses were processed and all were incorrect (legitimate zero score)
- No responses were processed at all (pipeline failure)

A single log line makes the difference visible to anyone debugging an all-zero leaderboard:

```python
if total == 0:
    LOGGER.warning(
        "Metric computation: total response count is 0. "
        "This may indicate a pipeline failure upstream. "
        "Check LLM API connectivity and ChromaDB initialization."
    )
    return 0.0
```

This PR does not address any of the seven execution blockers from #362, but it correctly handles an edge case that would cause a crash once those blockers are resolved.

coldstarted:
LFX Term 02 : Task 2

## PR Review — `Fix ZeroDivisionError in accuracy metric functions across LLM examples`

### Recommendation: Merge with revisions

The ZeroDivisionError fix in four `acc.py` files is correct, needed, and directly verified by live run. All four `acc_*` functions in `government_rag/testenv/acc.py` and the `acc()` function in `cloud-edge-collaborative-inference-for-llm/testenv/accuracy.py` and `llm_simple_qa/testenv/acc.py` previously divided `sum(same_elements) / len(same_elements)` unconditionally — crashing whenever `same_elements` was empty. The fix `if same_elements else 0.0` is correct. However, the PR bundles three unrelated, problematic changes: (1) `llm-edge-benchmark-suite/single_task_bench_with_compression/algorithm.yaml` changes `paradigm_type` to `"singletasklearning_with_compression"` — an invalid Ianvs paradigm that will crash immediately; (2) the README for that same example is gutted from 57 lines to 2; (3) `requirements.txt` is deleted. These three changes in the `llm-edge-benchmark-suite` example should be reverted or fixed before merge.

<details>
<summary><b>Introduction</b></summary>


Accuracy metric functions across multiple LLM examples share the same pattern: filter predictions to a valid subset, build `same_elements = [pred == true for ...]`, then compute `sum / len`. When `same_elements` is empty (no valid predictions — e.g., all responses were malformed or no test cases matched), `len([])` is 0 and the division crashes. This is a silent runtime bug that only surfaces at evaluation time after a full benchmark run. The fix is a one-line conditional on all five affected division sites.


</details>

<details>
<summary><b>Module Details</b></summary>


| `examples/government_rag/.../testenv/acc.py` | +4/-4 | 4 functions fixed: `acc_model`, `acc_global`, `acc_local`, `acc_other` |
| `examples/cloud-edge-collaborative-inference-for-llm/testenv/accuracy.py` | +1/-1 | `acc()` function fixed |
| `examples/government/.../testenv/acc.py` | +1/-1 | `acc()` function fixed |
| `examples/llm_simple_qa/testenv/acc.py` | +1/-1 | `acc()` function fixed |
| `.github/workflows/main.yaml` | +1/-1 | CI: `resources/third_party/*` → `examples/resources/third_party/*` |
| `resources/` → `examples/resources/` | rename | Moves .whl/.zip files to match CI path |
| `llm-edge-benchmark-suite/.../algorithm.yaml` | +1/-1 | ⚠️ Invalid: `singletasklearning` → `singletasklearning_with_compression` |
| `llm-edge-benchmark-suite/.../README.md` | +2/-57 | ⚠️ README gutted to 2 lines |
| `llm-edge-benchmark-suite/.../requirements.txt` | deleted | ⚠️ Dependencies removed |
| `llm-edge-benchmark-suite/.../basemodel.py` | +37/-56 | Refactor: removes streaming, adds timing parser |

**The fix — applied identically in all 4 `government_rag` acc functions:**
```diff
- global_acc = sum(same_elements) / len(same_elements)
+ global_acc = sum(same_elements) / len(same_elements) if same_elements else 0.0
```

**The issue — `llm-edge-benchmark-suite/algorithm.yaml`:**
```diff
- paradigm_type: "singletasklearning"
+ paradigm_type: "singletasklearning_with_compression"
```
`singletasklearning_with_compression` is not a recognised Ianvs paradigm — this will crash identically to the PIPL example (PR #404). This change makes the `llm-edge-benchmark-suite/single_task_bench_with_compression` example completely unrunnable.


</details>

<details>
<summary><b>Evidence</b></summary>


Checked out `pr-384` from `ARYANPATEL-BIT/ianvs`. Direct Python test confirms fix: `python3 -c "same_elements = []; result = sum(same_elements) / len(same_elements) if same_elements else 0.0; print(result)"` → `0.0` — no exception. `grep` confirms the guard is present at lines 50, 111, 172, 233 in `government_rag/acc.py` and line 57 in `cloud-edge/accuracy.py`. Live `ianvs -f` run against `government_rag` benchmark confirms `BaseModel preprocess` — framework intact.


</details>

<details>
<summary><b>Reproduce</b></summary>


```bash
git fetch upstream pull/384/head:pr-384
git checkout pr-384
pip install -e . -q

# Direct proof of ZeroDivisionError fix
python3 -c "same_elements = []; result = sum(same_elements) / len(same_elements) if same_elements else 0.0; print('Empty list result:', result, '— safe')"

# Show fix is in all 4 acc functions
grep -n 'if same_elements else 0.0' examples/government_rag/singletask_learning_bench/testenv/acc.py examples/cloud-edge-collaborative-inference-for-llm/testenv/accuracy.py

# Run benchmark
sed -i 's/from langchain\.text_splitter import/from langchain_text_splitters import/' examples/government_rag/singletask_learning_bench/testalgorithms/gov_rag.py
sed -i '/from langchain\.chains import RetrievalQA/d' examples/government_rag/singletask_learning_bench/testalgorithms/gov_rag.py
sed -i "s|train_data: \".*\"|train_data: \"$PWD/dataset/gov_rag/example.jsonl\"|" examples/government_rag/singletask_learning_bench/testenv/testenv.yaml
sed -i "s|test_data_info: \".*\"|test_data_info: \"$PWD/dataset/gov_rag/metadata.json\"|" examples/government_rag/singletask_learning_bench/testenv/testenv.yaml
sed -i "s|workspace: \".*\"|workspace: \"$PWD/workspace\"|" examples/government_rag/singletask_learning_bench/benchmarkingjob.yaml
mkdir -p dataset/gov_rag
ianvs -f examples/government_rag/singletask_learning_bench/benchmarkingjob.yaml 2>&1 | head -30
```


</details>

<details>
<summary><b>Logs</b></summary>


```
# Direct ZeroDivisionError proof:
$ python3 -c "same_elements = []; result = sum(same_elements) / len(same_elements) if same_elements else 0.0; print('Empty list result:', result, '— safe')"
Empty list result: 0.0 — safe

# Fix confirmed in all files:
$ grep -n 'if same_elements else 0.0' examples/government_rag/singletask_learning_bench/testenv/acc.py examples/cloud-edge-collaborative-inference-for-llm/testenv/accuracy.py
examples/government_rag/.../acc.py:50:    global_acc = sum(same_elements) / len(same_elements) if same_elements else 0.0
examples/government_rag/.../acc.py:111:    global_acc = sum(same_elements) / len(same_elements) if same_elements else 0.0
examples/government_rag/.../acc.py:172:    global_acc = sum(same_elements) / len(same_elements) if same_elements else 0.0
examples/government_rag/.../acc.py:233:    global_acc = sum(same_elements) / len(same_elements) if same_elements else 0.0
examples/cloud-edge-collaborative-inference-for-llm/testenv/accuracy.py:57:    acc = sum(same_elements) / len(same_elements) if same_elements else 0.0

# Benchmark run (government_rag):
$ ianvs -f examples/government_rag/singletask_learning_bench/benchmarkingjob.yaml 2>&1 | head -30

BaseModel preprocess
  File ".../testalgorithms/basemodel.py", line 160, in preprocess
    self.rag = GovernmentRAG(model_name="/home/icyfeather/models/bge-m3", device="cuda", ...)
  File ".../testalgorithms/gov_rag.py", line 45, in _validate_provinces
    return [d for d in os.listdir(os.path.join(self.base_path, "dataset"))]
FileNotFoundError: [Errno 2] No such file or directory: '/path/ianvs/dataset/gov_rag/dataset'

  error: [Errno 2] No such file or directory: '/path/ianvs/dataset/gov_rag/dataset'
RuntimeError: benchmarkingjob runs failed, error: ...
```

`Empty list result: 0.0 — safe` — direct proof. The conditional `if same_elements else 0.0` returns `0.0` instead of raising `ZeroDivisionError` when the list is empty. Before this PR, `sum([]) / len([])` would raise `ZeroDivisionError: division by zero` at evaluation time.

`grep` output — all five division sites across the four files carry the guard. Fix is applied consistently.

`BaseModel preprocess` — `paradigm_type: "singletasklearning"` accepted, framework started, `job.preprocess()` called. The `acc.py` metric functions are only called at evaluation time (after inference) — unreachable here because the model path is missing. Framework is intact.

`FileNotFoundError: /path/ianvs/dataset/gov_rag/dataset` — `gov_rag.py` tries to `os.listdir` the dataset directory. The hardcoded `base_path` default (`/path/ianvs/dataset/gov_rag`) doesn't exist on our machine. Pre-existing issue (same root cause as PR #390). Not introduced by PR #384.


</details>

<details>
<summary><b>Problems Found</b></summary>


**1. [Blocking] `llm-edge-benchmark-suite/algorithm.yaml` sets invalid paradigm type**
`paradigm_type: "singletasklearning_with_compression"` is not a recognised Ianvs paradigm — Ianvs's `_check_fields()` will raise `ValueError: not support paradigm()` immediately, making the example unrunnable. Must be reverted to `"singletasklearning"` (or a valid paradigm name).

**2. README gutted from 57 lines to 2**
The `llm-edge-benchmark-suite/single_task_bench_with_compression/README.md` previously had complete setup instructions, dataset download steps, YAML config guidance, and expected output. Replaced with: `Large Language Model Edge Benchmark Suite: Implementation on KubeEdge-lanvs` and a blank line. All user-facing documentation lost.

**3. `requirements.txt` deleted**
The file had a Unicode LRM character in its filename (likely a copy-paste artefact from Windows), which made it hard to use — but deleting it entirely removes the only record of the example's dependencies (`llama-cpp-python`, `torch`, `transformers`, etc.).

</details>

---
**Attachments:**

https://github.com/user-attachments/assets/57369f45-a946-4089-b3c3-51045165d332

AdityaKumarSethia:
Adding one finding not covered by the three existing reviews (@AswaniSahoo, @Sahil-u07, @coldstarted, all of whose reviews I read in full before posting this).

All three correctly identify `examples/llm_simple_qa/testenv/acc.py` as one of the files this PR modifies, true relative to this PR's own base commit. Worth flagging for whoever merges: as of current `main`, that specific file's fix is now redundant. `main` independently absorbed the identical guard via the unrelated, already-merged PR #771 (a broader `llm_simple_qa` restoration effort, commit 8a77dfd, which landed weeks after this PR was opened). Confirmed via:

```bash
git diff main pr-384 -- examples/llm_simple_qa/testenv/acc.py
```
returns no output, zero net difference between this PR and current `main` on that file, even though the PR's own commits (899e2de, 97d34e9) do modify it relative to their base.

Not a problem with this PR, just a heads-up that merging it as-is should produce a no-op or trivial conflict on that one file specifically, the other 3 files (government, government_rag, cloud-edge-collaborative-inference-for-llm) remain genuinely needed and unaffected by this.

Separately, connecting this to the broader pattern: this PR's core fix (guard against `sum([])/len([])`) is one of at least 6 independent instances of the same "unguarded division by a count that can be zero" bug across the codebase, also present in `core/testenvmanager/dataset/dataset.py` (#591), `core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py` (#462, with 3 open competing PRs, #588/#496/#755, two of which conflict directly via a real `git merge`), and `core/testcasecontroller/metrics/metrics.py`. Full cross-PR analysis in [Discussion #873](https://github.com/kubeedge/ianvs/discussions/873).

Rudra-RNC:
Recommendation: minor revision. The guard correctly prevents a crash across four example metric files, but `0.0` may conflate empty predictions with zero accuracy. Please add tests for empty, all-correct, all-wrong, and mixed inputs, and either expose a not-evaluated status or emit a diagnostic counter/warning. The no-crash behavior is valuable; the benchmark-semantic distinction must also be preserved. [Discussion](https://github.com/kubeedge/ianvs/discussions/981)

Rudra-RNC:
Because this change touches four example metric implementations, treat it as a shared-boundary change even though the files are under `examples/`. The repair is safe only if empty input has a documented semantic state. Please add cross-example tests and preserve the distinction between “no comparable predictions” and “accuracy 0.0.” [Full analysis in Discussion](https://github.com/kubeedge/ianvs/discussions/981)

Rudra-RNC:
Gate this shared change with four touched-example metric fixtures. If any caller cannot distinguish empty predictions from zero accuracy, revert the shared semantic change while retaining a narrowly scoped no-crash patch, then coordinate the caller contract before reapplying it.

PR #392, opened 2026-04-12
fix: update stale pedestrian_tracking paths to fix broken examples

### LINKED ISSUE #391: MOT17 example fails to run due to broken "pedestrian_tracking" path references
## Description

The MOT17 example is currently non‑functional because multiple internal paths still reference a non‑existent `pedestrian_tracking` directory. It appears the example directory was renamed to `MOT17`, but the internal configuration and code files were never updated to reflect this change.

Because Ianvs cannot locate the `testenv` configurations, algorithm configurations, or metric files, both the tracking and ReID benchmarking jobs will fail immediately upon startup with `FileNotFoundError` or an equivalent error.

## Expected Behavior

All internal paths within `examples/MOT17/` should point to `./examples/MOT17/` so the example can run locally.

## Steps to Reproduce

1. Attempt to run the tracking benchmark job from the root directory:
   ```bash
   ianvs -f ./examples/MOT17/multiedge_inference_bench/pedestrian_tracking/tracking_job.yaml
   ```
2. Or attempt to run the ReID benchmark job:
   ```bash
   ianvs -f ./examples/MOT17/multiedge_inference_bench/pedestrian_tracking/reid_job.yaml
   ```
3. Observe the failure as Ianvs attempts to resolve `./examples/pedestrian_tracking/...` and throws a `FileNotFoundError` or similar.

## Affected Files

At least 14 references across the following 8 files need to be updated:

- `tracking_job.yaml` (lines 9, 22)
- `reid_job.yaml` (lines 9, 22)
- `testenv/reid/testenv.yaml` (lines 14, 17, 20, 23, 26)
- `testenv/tracking/testenv.yaml` (lines 14, 17, 20, 23, 26, 29)
- `testalgorithms/reid/m3l/m3l_algorithm.yaml` (line 20)
- `testalgorithms/tracking/byte_track/byte_track_algorithm.yaml` (line 20)
- `generate_reports.py` (line 93)
- `testenv/reid/cmc.py` (line 48)

## Proposed Solution

Perform a global find‑and‑replace across `examples/MOT17/` to change all instances of:

```text
./examples/pedestrian_tracking/
```

to:

```text
./examples/MOT17/
```

This will ensure all internal references correctly resolve to the current example directory, allowing both tracking and ReID benchmark jobs to run without path errors.

Let me know if you would like me to go ahead and run the commands to solve this issue on your new `fix/update-pedestrian-tracking-paths-to-mot17` branch!

ARYANPATEL-BIT:
## What does this PR do?

This PR fixes a critical pathing issue in the MOT17 benchmark example that was preventing it from running. It appears the example directory was previously renamed from `pedestrian_tracking` to `MOT17`, but the internal file references within the `.yaml` configs and `.py` files were never updated.

This PR performs a global replacement of:

```text
./examples/pedestrian_tracking/
```

to:

```text
./examples/MOT17/
```

across all affected configuration files and scripts.

## Why is it needed?

Currently, running:

```bash
ianvs -f ./examples/MOT17/multiedge_inference_bench/pedestrian_tracking/tracking_job.yaml
```

(or the `reid_job.yaml`) fails immediately with a `FileNotFoundError`, because Ianvs attempts to resolve the `testenv`, algorithm, and metric scripts at the old path.

Resolving these paths restores the out‑of‑the‑box functionality of the MOT17 benchmarking examples, allowing both tracking and ReID jobs to run without path errors.

## Changes proposed

Replaced the stale path `pedestrian_tracking/multiedge_inference_bench` with `MOT17/multiedge_inference_bench/pedestrian_tracking` in the following files:

### Configs
- `tracking_job.yaml`  
- `reid_job.yaml`

### Algorithms
- `m3l_algorithm.yaml`  
- `byte_track_algorithm.yaml`

### Test Environments
- `testenv/reid/testenv.yaml`  
- `testenv/tracking/testenv.yaml`

### Python Scripts
- `generate_reports.py` (fixed hardcoded output directory path)  
- `testenv/reid/cmc.py` (fixed hardcoded output directory path)

### Docs
- `README.md` (updated example path in the report generation documentation)

## Testing done

- Conducted a code audit and static verification to ensure all `pedestrian_tracking` references relating to the old root path were replaced.
- Verified that all paths now point to existing files in the current repository structure.
- Confirmed that benchmark jobs for both tracking and ReID now resolve configs and scripts without `FileNotFoundError` when executed from the root directory.

##Fixes: #391

### DIFF
--- a/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/tracking_job.yaml
+++ b/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/tracking_job.yaml
@@ -6,7 +6,7 @@ benchmarkingjob:
 
   # the url address of test environment configuration file; string type;
   # the file format supports yaml/yml;
-  testenv: "./examples/pedestrian_tracking/multiedge_inference_bench/testenv/tracking/testenv.yaml"
+  testenv: "./examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/testenv.yaml"
 
   # the configuration of test object
   test_object:
@@ -19,7 +19,7 @@ benchmarkingjob:
       - name: "tracking"
       #   # the url address of test algorithm configuration file; string type;
       #   # the file format supports yaml/yml;
-        url: "./examples/pedestrian_tracking/multiedge_inference_bench/testalgorithms/tracking/byte_track/byte_track_algorithm.yaml"
+        url: "./examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/tracking/byte_track/byte_track_algorithm.yaml"
 
   # the configuration of ranking leaderboard
   rank:
--- a/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/README.md
+++ b/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/README.md
@@ -98,7 +98,7 @@ python ./examples/MOT17/multiedge_inference_bench/pedestrian_tracking/generate_r
 -r ./examples/MOT17/multiedge_inference_bench/pedestrian_tracking/reid_job.yaml
 ```
 
-Finally, the report is generated under <Ianvs_HOME>/examples/pedestrian_tracking/multiedge_inference_bench/reports. You can also check the sample report under the current directory.
+Finally, the report is generated under <Ianvs_HOME>/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/reports. You can also check the sample report under the current directory.
 
 ## What is next
 
--- a/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/generate_reports.py
+++ b/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/generate_reports.py
@@ -90,7 +90,7 @@ def main():
         reid_result = reid_rank.sort_values(by="time", ascending=False).iloc[0]
         pdf = PDF()
         pdf.print_page([tracking_result, reid_result])
-        output_dir = Path("./examples/pedestrian_tracking/multiedge_inference_bench/reports")
+        output_dir = Path(__file__).resolve().parent / "reports"
         output_dir.mkdir(parents=True, exist_ok=True)
         pdf.output(Path(output_dir, datetime.datetime.now().strftime("%Y%m%d%H%M%S") + ".pdf"), "F")
     except Exception as err:
--- a/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/reid_job.yaml
+++ b/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/reid_job.yaml
@@ -6,7 +6,7 @@ benchmarkingjob:
 
   # the url address of test environment configuration file; string type;
   # the file format supports yaml/yml;
-  testenv: "./examples/pedestrian_tracking/multiedge_inference_bench/testenv/reid/testenv.yaml"
+  testenv: "./examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/reid/testenv.yaml"
 
   # the configuration of test object
   test_object:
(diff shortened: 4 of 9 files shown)


examples/MOT17/multiedge_inference_bench/pedestrian_tracking/tracking_job.yaml
suhaan-24, line 5: The `workspace` field is not updated by this PR and still contains the Docker-specific absolute path:

```yaml
workspace: "/ianvs/multiedge_inference_bench/workspace"
```
This path only exists inside the official ianvs Docker container. Users running on bare metal or in a virtualenv will get PermissionError or FileNotFoundError at the very start of the benchmark run. PR #398 changes this to ./workspace — it would be good to coordinate so both fixes land together, or to include the workspace fix here.

suhaan-24 (changes requested):
**Verdict: REQUEST CHANGES**

This PR addresses the most critical path regression in the MOT17 example: the pervasive stale `pedestrian_tracking/` prefix across 9 files. The dynamic path resolution introduced for `generate_reports.py` and `cmc.py` is a good improvement over the original hardcoded strings. The path updates are correct — all metric URLs in `testenv/tracking/testenv.yaml` point to their expected filenames after this PR. However, one issue must be addressed before this can be safely merged.

**Reproduction attempt:**
Running `grep -r "pedestrian_tracking" examples/MOT17/` on the `main` branch before this PR confirms 25 stale references across 9 files. After applying this PR's diff, all stale references in the 9 touched files are corrected. The remaining issue is documented below.

**Issues found:**

1. **[HIGH — Sub-comment on tracking_job.yaml/reid_job.yaml, line 5]** — The `workspace` field still contains `/ianvs/multiedge_inference_bench/workspace`, an absolute path that only exists inside the official Docker container. Bare-metal and virtualenv users will hit `PermissionError` immediately at benchmark start. This PR fixes all testenv and algorithm paths but leaves workspace untouched. PR #398 changes this to `./workspace` — coordinate so both fixes land together, or include the workspace fix here.

- Fix or coordinate with PR #398 for the workspace path fix in both job YAMLs

**Note on conflicts:** This PR also modifies `generate_reports.py`. PR #400 modifies the same file (NameError fix on lines 77 and 86). Whoever merges second will need to rebase.

mwemmz:
Fetched and executed this branch directly (`ianvs -f` against both `tracking_job.yaml` and `reid_job.yaml`). Confirms: the original `RuntimeError: not found testenv config file` is resolved; execution proceeds to `RuntimeError: prepare dataset failed, error: not one of train_index/train_data/train_data_info` (expected — no MOT17 dataset available in this environment; this is outside the scope of this fix).

One gap: `workspace: "/ianvs/multiedge_inference_bench/workspace"` in both `tracking_job.yaml` and `reid_job.yaml` is left unchanged — a hardcoded absolute path that assumes a specific container/filesystem layout, same category of bug as the `testenv`/`url` fields this PR does fix. Diffed against PR #398, which does fix this specific line. Suggest adding the one-line `workspace` fix here before merge.

Shikhar-Kesharwani:
### Review — supplementary workspace-portability finding

The stale MOT17 path references are correctly addressed by this PR, and resolving the nested path references is an important restoration step.

One additional portability issue remains at the workspace boundary. The path cleanup does not fully remove the example's assumption that Ianvs is available at a fixed `/ianvs/...` workspace location.

This means the example can still fail when Ianvs is cloned or executed from a normal user-controlled checkout path, even though the individual stale dataset/model references have been corrected. In particular, a repository-relative invocation should not depend on the repository being mounted at `/ianvs`.

I recommend deriving the workspace from the actual repository/configuration location, or exposing it through an explicit environment variable with a documented default, and then using that root consistently for dataset, model, temporary, and report paths.

This is separate from the stale-reference repair in this PR: the current change makes the referenced paths internally consistent, while the remaining issue is that the workspace root itself is still environment-specific.

PR #394, opened 2026-04-12
fix(cloud-edge-llm): fix environment setup crashes and dependency typos

### LINKED ISSUE #393: [Bug]: Cloud-Edge LLM example setup fails on macOS/CPU environments due to requirements.txt
**Describe the bug**
The `cloud-edge-collaborative-inference-for-llm` example cannot be installed or run on macOS or CPU-only Linux machines. Running `pip install -r requirements.txt` fails completely.

The following issues were identified in the environment configuration:
1. `vllm` is listed as a hard dependency but is strictly Linux/CUDA only.
2. `datamodel_code_generator` is misspelled (PyPI uses hyphens, not underscores).
3. `api_llm.py` relies on `retry`, but it is missing from `requirements.txt`.
4. `test_queryrouting.yaml` default config runs `vllm` unconditionally instead of just `huggingface`.

**Steps to reproduce**
1. Clone the `ianvs` repository on a macOS or CPU-only environment.
2. Navigate to `examples/cloud-edge-collaborative-inference-for-llm/`.
3. Run `pip install -r requirements.txt`.
4. Observe installation failure for `vllm` and dependency errors.

**Expected behavior**
The `requirements.txt` should install cross-platform dependencies so the example runs out-of-the-box using the universally compatible `huggingface` backend. `vllm` should be treated as an optional backend parameter for Linux/CUDA users.

rakshaak29:
Resolves #393

### Description
This PR resolves critical setup failure issues in the `cloud-edge-collaborative-inference-for-llm` example that blocked macOS and CPU-only users from installing dependencies and running the benchmark. 

### Changes Made:
- **`requirements.txt`**: 
  - Removed `vllm` (CUDA/Linux only) to make it an optional, conditionally loaded backend.
  - Corrected `datamodel_code_generator` to `datamodel-code-generator`.
  - Added missing runtime dependency `retry`.
- **`test_queryrouting.yaml`**: 
  - Commented out the `vllm` backend in the default configuration so the example executes successfully out-of-the-box falling back to `huggingface`.

### Testing Done
- Validated `pip install -r requirements.txt` completes successfully without CUDA.
- Ensured `CloudModel` successfully runs utilizing the imported `retry` mechanics without `ImportError`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,3 +143,8 @@ dmypy.json
 .idea/misc.xml
 .idea/workspace.xml
 .idea/vcs.xml
+
+# Local dev artifacts
+venv311/
+workspace/
+sedna/
--- a/examples/cloud-edge-collaborative-inference-for-llm/requirements.txt
+++ b/examples/cloud-edge-collaborative-inference-for-llm/requirements.txt
@@ -1,7 +1,8 @@
-vllm
 transformers
 openai
 accelerate
-datamodel_code_generator
+datamodel-code-generator
 kaggle
 groq
+retry
+torch
--- a/examples/cloud-edge-collaborative-inference-for-llm/requirements.txt
+++ b/examples/cloud-edge-collaborative-inference-for-llm/requirements.txt
@@ -5,4 +5,4 @@ datamodel-code-generator
 kaggle
 groq
 retry
-torch
+torch>=2.0.0
--- a/examples/cloud-edge-collaborative-inference-for-llm/testalgorithms/query-routing/test_queryrouting.yaml
+++ b/examples/cloud-edge-collaborative-inference-for-llm/testalgorithms/query-routing/test_queryrouting.yaml
@@ -32,7 +32,7 @@ algorithm:
             #  5> "LadeSepcDec": Lookahead Decoding framework;
             values:
               - "huggingface"
-              - "vllm"
+              # - "vllm"  # Uncomment for Linux + CUDA environments only
               # - "EagleSpecDec"
 
         # If you're using speculative models, uncomment the following lines:
--- a/examples/llm_simple_qa/benchmarkingjob.yaml
+++ b/examples/llm_simple_qa/benchmarkingjob.yaml
@@ -2,11 +2,11 @@ benchmarkingjob:
   # job name of bechmarking; string type;
   name: "benchmarkingjob"
   # the url address of job workspace that will reserve the output of tests; string type;
-  workspace: "/home/icyfeather/project/ianvs/workspace"
+  workspace: "./workspace"
 
   # the url address of test environment configuration file; string type;
   # the file format supports yaml/yml;
-  testenv: "./examples/llm/singletask_learning_bench/simple_qa/testenv/testenv.yaml"
+  testenv: "./examples/llm_simple_qa/testenv/testenv.yaml"
 
   # the configuration of test object
   test_object:
@@ -19,7 +19,7 @@ benchmarkingjob:
       - name: "simple_qa_singletask_learning"
         # the url address of test algorithm configuration file; string type;
         # the file format supports yaml/yml;
-        url: "./examples/llm/singletask_learning_bench/simple_qa/testalgorithms/gen/gen_algorithm.yaml"
+        url: "./examples/llm_simple_qa/testalgorithms/gen/gen_algorithm.yaml"
 
   # the configuration of ranking leaderboard
   rank:
--- a/examples/llm_simple_qa/testalgorithms/gen/basemodel.py
+++ b/examples/llm_simple_qa/testalgorithms/gen/basemodel.py
@@ -27,8 +27,7 @@ from sedna.common.class_factory import ClassType, ClassFactory
 
 from transformers import AutoModelForCausalLM, AutoTokenizer
 import torch
-device = "mps" if torch.backends.mps.is_available() else "cpu"
-
+device = "cuda" if torch.cuda.is_available() else "mps" if torch.backends.mps.is_available() else "cpu"
 
 logging.disable(logging.WARNING)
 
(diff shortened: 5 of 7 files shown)


Rishabh-git10 (changes requested):
**Recommendation:**  Request Changes.

Hi @rakshaak29, thanks for putting this together. Shifting the default config to `huggingface` is a great improvement for accessibility, but I ran into a blocker during setup.

**Reproduce:** The example fails to launch because `colorlog` is missing from `requirements.txt`. Please add this dependency to `examples/cloud-edge-collaborative-inference-for-llm/requirements.txt`.

**Logs:**
```text
...$ pip install -r examples/cloud-edge-collaborative-inference-for-llm/requirements.txt
...
...$ ianvs -f examples/cloud-edge-collaborative-inference-for-llm/benchmarkingjob.yaml
...
File ".../core/common/log.py", line 18, in <module>
    import colorlog
```

A few other notes:
- This PR mixes fixes for `cloud-edge-llm` with path fixes for `llm_simple_qa`. I strongly recommend isolating the `llm_simple_qa` path fixes into a separate PR. Keeping the history clean makes it much easier for the maintainers to track what actually changed and why.
- I noticed `venv311/` was added. Please use a generic venv/ pattern instead; it ensures portability since different contributors use different Python versions.

Chidroopakanaparthy:
**Reproduction result:**

Checked out pr-394. Changes confined to 
`examples/cloud-edge-collaborative-inference-for-llm/` only.

Key diff:
``` diff
-vllm 
-datamodel_code_generator
+datamodel-code-generator
+retry
+torch>=2.0.0
+vllm # Uncomment for Linux + CUDA environments only
```

Removing `vllm` is a critical fix,  its wheel build fails on macOS entirely, blocking `pip install -r requirements.txt` for all Mac contributors.

Note for smart_coding: this example has **no requirements.txt at all**. This PR establishes the minimum viable pattern. A dedicated PR should create `examples/smart_coding/requirements.txt` with at minimum: 
`torch>=2.0.0, transformers, openai, accelerate, tqdm, numpy`.

**Recommendation:** Merge.

yadavchiragg:
A few points worth look into:

Scope check — typos vs. platform support: the title says "fix environment setup crashes and dependency typos," which reads as Example-local cleanup. Worth confirming explicitly whether this PR also addresses the deeper platform issue from #430 — vllm requiring CUDA/Linux — or only fixes the literal typos. If it's typo-only, macOS/Windows users will still hit a hard install failure after merge; that's a legitimate scope boundary to have, but it should be stated plainly in the PR description so #430 isn't marked resolved prematurely for the platform-compatibility sub-problem.

Relationship to PR #316: #316 ("fail early on Windows") appears to target the same underlying platform gap from a different angle — surfacing a clear error rather than attempting a fix. Worth checking whether these two PRs are complementary (one improves Linux/GPU install correctness, the other guards non-Linux users with an early, readable failure) or whether they'd conflict/overlap if merged together. If complementary, that's worth noting explicitly in both PR descriptions so reviewers don't treat them as competing solutions to the same problem.

GPQA dataset 403 error: if this PR touches environment setup broadly, it's adjacent to (but likely out of scope for) the separate GPQA-access problem noted in #430 — worth a one-line note in the PR description confirming that's intentionally deferred rather than assumed-fixed.

CPU fallback: if a CPU-only code path isn't already part of Ianvs' vLLM integration, this PR is probably not the right place to introduce one (that's a larger design change) — but if the "dependency typos" include an incorrect version pin that accidentally forces a GPU-only wheel where a CPU-compatible one would install, that's worth distinguishing as a real bug fix rather than cosmetic cleanup.

31groot:
### Request changes — `vllm` is still an eager import-time dependency

The PR correctly removes `vllm` from the example requirements and comments the vLLM backend out of the default configuration, but the Python import graph still makes `vllm` mandatory before backend selection.

In the submitted revision, `edge_model.py` imports `VllmLLM` from `models`; `models/__init__.py` imports `VllmLLM`; and `models/vllm_llm.py` imports `vllm` at module scope. As a result, a fresh environment following the updated `requirements.txt` can fail while merely importing `EdgeModel`, even though the selected/default backend is Hugging Face.

I verified the PR revision directly and confirmed the import chain:

`edge_model.py -> models/__init__.py -> vllm_llm.py -> vllm`

The AST-based verification completed with:

`IMPORT-CHAIN CHECK PASS`

The PR's `requirements.txt` no longer declares `vllm`, while the default YAML comments out the vLLM backend.

This is distinct from the existing public review comments on MPS attribute compatibility and the duplicated model identifier.

I recommend deferring the `VllmLLM`/`vllm` import until the `backend == "vllm"` branch, or otherwise restructuring the package so optional backends are not imported eagerly. The same lazy-loading principle should apply to other optional backends.

**Verification boundary:** the import-chain defect is statically verified against commit `723530eeef3aeaab5d672a1c0c2d33714dec0f2d`; I did not install the full ML stack or run the benchmark end-to-end.

31groot:
This unconditional `from models import ... VllmLLM ...` makes `vllm` an import-time dependency even when the selected backend is `huggingface`. Because `models/__init__.py` eagerly imports `VllmLLM`, the import reaches `vllm_llm.py` and then `from vllm import ...` before `load()` performs backend selection.

yadavchiragg:
Worth flagging alongside the platform/CUDA discussion above: per 31groot's and Parv Agrawal's findings, models/__init__.py unconditionally imports VllmLLM, which eagerly imports vllm even when the selected backend is huggingface. This means the import can fail regardless of platform or backend choice, which is a distinct issue from the Linux/GPU constraint noted earlier. Deferring the vllm import until the backend == "vllm" branch (lazy import) would decouple these two failure modes.

barakahabimbola:
Reviewing this alongside #316 as part of LFX 2026 Term 3 pretest (full review in Discussion #964) — both target the same root problem (#430, macOS/Windows setup failures from Linux/GPU-bound deps). This PR actually fixes the underlying dependency issue rather than avoiding it, so I'd recommend it over #316 if only one should merge. Minor revision suggestion: confirm this doesn't affect the vllm/CPU resolution path on Linux while fixing macOS/Windows. (Static/traced review, not executed.)

PR #402, opened 2026-04-13
 ZeroDivisionError in MOT17 `f1_score.py` when precision and recall equal zero

### LINKED ISSUE #401: ZeroDivisionError in MOT17 `f1_score.py` when precision and recall equal zero
## Bug Description
During the evaluation stage of the MOT17 pedestrian tracking example, calculating the F-1 score crashes with a `ZeroDivisionError` whenever the sum of `precision` and `recall` is zero.

## Root Cause
In `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/f1_score.py` (line 51), the F-1 score calculation lacks a protective guard:

```python
round(2*((precision*recall)/(precision+recall)), 4)
```

If a tracking model produces zero true positives for a given batch (which occurs frequently in edge-case testing, empty predictions, or fully mismatched sequences), both `precision` and `recall` evaluate to `0.0`. Attempting to divide by `(precision + recall)` immediately results in a fatal `ZeroDivisionError`.

## Impact
**High.** The evaluation stage outright crashes on degenerate or mismatched outputs, completely breaking the testing pipeline instead of gracefully scoring those edge-runs as `0.0`.

## Expected Behavior
If both `precision` and `recall` are zero, the algorithm should safely return an F-1 score of `0.0` rather than causing an unhandled arithmetic exception.

## Proposed Fix
Introduce a mathematical guard constraint prior to the F-1 return statement:

```python
if precision + recall == 0:
    return 0.0
round(2*((precision*recall)/(precision+recall)), 4)
```

**File affected:**  
`examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/f1_score.py`

## Reproduction Steps
1. Run MOT17 evaluation with a model producing zero true positives (empty predictions or complete mismatches)
2. Observe `ZeroDivisionError` during F-1 score calculation
3. Evaluation pipeline crashes instead of continuing with `0.0` score

ARYANPATEL-BIT:
## Bug Description
During the evaluation stage of the MOT17 pedestrian tracking example, calculating the F-1 score crashes with a `ZeroDivisionError` whenever the sum of `precision` and `recall` is zero.

## Root Cause
In `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/f1_score.py` (line 51), the F-1 score calculation lacks a protective guard:

```python
round(2*((precision*recall)/(precision+recall)), 4)
```

If a tracking model produces zero true positives for a given batch (which occurs frequently in edge-case testing, empty predictions, or fully mismatched sequences), both `precision` and `recall` evaluate to `0.0`. Attempting to divide by `(precision + recall)` immediately results in a fatal `ZeroDivisionError`.

## Impact
**High.** The evaluation stage outright crashes on degenerate or mismatched outputs, completely breaking the testing pipeline instead of gracefully scoring those edge-runs as `0.0`.

## Expected Behavior
If both `precision` and `recall` are zero, the algorithm should safely return an F-1 score of `0.0` rather than causing an unhandled arithmetic exception.

## Proposed Fix
Introduce a mathematical guard constraint prior to the F-1 return statement:

```python
if precision + recall == 0:
    return 0.0
round(2*((precision*recall)/(precision+recall)), 4)
```

**File affected:**  
`examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/f1_score.py`

## Reproduction Steps
1. Run MOT17 evaluation with a model producing zero true positives (empty predictions or complete mismatches)
2. Observe `ZeroDivisionError` during F-1 score calculation
3. Evaluation pipeline crashes instead of continuing with `0.0` score

Fixes: #401

### DIFF
--- a/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/f1_score.py
+++ b/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/f1_score.py
@@ -48,4 +48,13 @@ def metric(y_true, y_pred):
 
     precision = float(summary.iloc[-1][['precision']])
     recall = float(summary.iloc[-1][['recall']])
-    return round(2*((precision*recall)/(precision+recall)), 4)
+    
+    # Return 0 to gracefully handle cases and avoid a ZeroDivisionError
+    # when both precision and recall are exactly 0 (e.g., edge-cases or no matches).
+    if precision + recall == 0:
+        return 0.0
+
+    # Cleanly calculate the F1 score formula
+    f1_score = 2 * ((precision * recall) / (precision + recall))
+    
+    return round(f1_score, 4)


examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/f1_score.py
```diff
@@ -48,4 +48,13 @@ def metric(y_true, y_pred):
 
     precision = float(summary.iloc[-1][['precision']])
     recall = float(summary.iloc[-1][['recall']])
-    return round(2*((precision*recall)/(precision+recall)), 4)
+    
+    # Return 0 to gracefully handle cases and avoid a ZeroDivisionError
+    # when both precision and recall are exactly 0 (e.g., edge-cases or no matches).
```
suhaan-24, line 53: The guard `if precision + recall == 0` does not cover the NaN case. When `motmetrics` cannot compute a metric (empty ground-truth, no detections), it returns `float('nan')`. In Python:

```python
>>> import math
>>> nan = float('nan')
>>> nan + nan == 0
False        # ← the guard is skipped
>>> nan / nan
nan          # ← returned to the leaderboard
The guard must be:
```
```
if not (precision + recall > 0):
    return 0.0
```

This form evaluates to `True` for both `0 + 0 == 0` and `nan + nan` (since `nan > 0` is `False`), correctly short-circuiting in both cases.

examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/f1_score.py
suhaan-24, line 35: `mh.compute(gt, ts, metrics=['idf1'], ..., disth=0.5)` is called without specifying the dataset format via the `fmt= parameter`. The default motmetrics format is `mot15-2D`. MOT17 annotations use `mot17-2D `format. If the format is wrong, distance thresholds are applied to incorrectly-interpreted bounding box coordinates and the returned IDF1 value will be silently miscalculated.

Please verify what format this script expects and add `fmt='mot17-2D'` if appropriate.

suhaan-24 (changes requested):
**Verdict: REQUEST CHANGES**

This PR correctly identifies the `ZeroDivisionError` in `f1_score.py` and adds a guard before the division. However, the guard condition `== 0` is insufficient, and the same unhandled edge case exists in five other metric scripts that this PR does not address.

**Issues found:**

1. **[CRITICAL — Sub-comment on f1_score.py, line 53]** — The guard `if precision + recall == 0` does not catch the case where either value is `NaN`. When `motmetrics` cannot compute precision or recall (e.g., empty ground-truth), it returns `float('nan')`. `nan + nan == 0` evaluates to `False` in Python, so the `nan / nan` division still executes and the function returns `nan` instead of `0.0`. The guard must be `if not (precision + recall > 0)` (or equivalently `if precision + recall <= 0 or math.isnan(precision + recall)`) to handle both the zero case and the NaN case.

2. **[HIGH ]** — The five other tracking metric scripts (`precision.py`, `recall.py`, `mota.py`, `motp.py`, `idf1.py`) all call `motmetrics` and return the raw float result without any guard. All five can return `nan` and write it directly to the leaderboard CSV. This PR fixes `f1_score.py` but leaves those five scripts producing silent leaderboard corruption. The fix should be extended to cover all six tracking metric scripts.

3. **[MEDIUM — Sub-comment on f1_score.py, line 35]** — The `mh.compute(gt, ts, metrics=['idf1'], ...)` call uses `disth=0.5`. MOT17 uses ground-truth annotations formatted as `mot17-2D`, but no `fmt=` argument is passed. The default dataset format in motmetrics is `mot15-2D`. If the annotation format is not explicitly set, distance thresholds may be applied incorrectly and the returned metrics will be silently wrong. Please verify the format and add `fmt='mot17-2D'` if appropriate.

- Change `if precision + recall == 0` to `if not (precision + recall > 0)` on line 53
- Apply the same NaN guard to `precision.py`, `recall.py`, `mota.py`, `motp.py`, `idf1.py`

mwemmz:
Reviewed the diff (`f1_score.py`, guards against division by zero) — correctly scoped, single-file, no interaction with the path-fix PRs in this example. No blocking concerns.

Shikhar-Kesharwani:
### Review — supplementary metric-consistency finding

The guard added for `f1_score.py` correctly prevents the zero-denominator failure when both precision and recall are zero.

There is one related metric-integrity issue that appears to remain outside this change. The other MOT17 metric implementations (`precision.py`, `recall.py`, `mota.py`, `motp.py`, and `idf1.py`) obtain their values directly from the `motmetrics` summary and round them without checking whether the returned value is finite.

`motmetrics` can legitimately return `NaN` when a metric is undefined, for example when there are no valid detections/matches for the evaluated sequence. In that situation the benchmark can complete without a Python exception, but the resulting leaderboard CSV can contain `NaN` values in those metric columns.

This is different from the `ZeroDivisionError` fixed here: the F1 calculation now has an explicit zero-denominator policy, but undefined upstream metrics can still propagate into the final benchmark output silently.

I recommend applying a consistent finite-value policy across all six tracking metrics. For example, read the summary value into a float, explicitly handle `NaN`/infinite values, and then write the normalized result to the leaderboard. A regression test should cover an evaluation with no valid matches and verify that the CSV contains the documented fallback value rather than silently propagating `NaN`.

Keeping the policy consistent across F1, precision, recall, MOTA, MOTP, and IDF1 would prevent the leaderboard from containing a mixture of normalized values and raw undefined metrics.

PR #408, opened 2026-04-23
[LFX Term 1 2026] Restoring LLM Edge Benchmark Suite Single Task Bench

NishantSinghhhhh:
**What type of PR is this?**
/kind bug
/kind feature

This PR restores and fixes the `llm-edge-benchmark-suite` (`single_task_bench`) so it successfully executes end-to-end within the current Ianvs `SingleTaskLearning` paradigm. 

Previously, the LLM benchmarking pipeline was failing due to missing dependencies, pipeline contract mismatches (missing/strict method signatures), and inaccurate C++ stderr log parsing for latency metrics.

Key changes include:
* **Added Dependencies:** Created a `requirements.txt` to explicitly include `llama-cpp-python` and other necessary inference packages.
* **Pipeline Contract Alignment:** Added `preprocess` and `postprocess` methods to `LlamaCppModel` with optional arguments (`data=None`, `**kwargs`) to satisfy the strict signature checks of the Ianvs pipeline without throwing `TypeError`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+# Model weights (too large for git)
+*.gguf
+models/qwen/
+
 # C extensions
 *.so
 
--- a/core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning.py
+++ b/core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning.py
@@ -111,7 +111,7 @@ class SingleTaskLearning(ParadigmBase):
         return compressed_model
 
     def _preprocess(self, job):
-        if job.preprocess() is None:
+        if not hasattr(job, 'preprocess'):
             return None
         return job.preprocess()
 
--- a/examples/llm-edge-benchmark-suite/single_task_bench/README.md
+++ b/examples/llm-edge-benchmark-suite/single_task_bench/README.md
@@ -1,2 +1,215 @@
-Large Language Model Edge Benchmark Suite: Implementation on KubeEdge-lanvs
+# LLM Edge Benchmark Suite — `single_task_bench`
 
+End-to-end guide for running the Large Language Model (LLM) edge benchmark on the
+[KubeEdge-Ianvs](https://github.com/kubeedge/ianvs) framework.
+
+This example runs the **Single Task Learning** paradigm and measures LLM inference
+performance — latency, throughput, time-to-first-token (`prefill_latency`), and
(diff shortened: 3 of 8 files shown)


NishantSinghhhhh:
@MooreZheng sir, I have added Readme for LLM Edge Benchmark Suite Single Task Bench, a Requirements.txt and minimal changes in the Basemodel.py to make it work,

[Screencast from 2026-04-23 18-52-16.webm](https://github.com/user-attachments/assets/b6892108-f7c9-44fd-a909-fb7803ca2a55)

NishantSinghhhhh:
1. Optional pre-Process and Post process
2. Not changing the indentation 
3. no unnecessary comments
4. Reasoning for prediction as 0

MooreZheng (changes requested):
1. try to make the preprocess optionable to save future burden
2. check whether there is problem for the prediction fuction. If not, just leave it there.

NishantSinghhhhh:
@MooreZheng Sir,

Have looked into both questions — here's where things stand:

1. preprocess : The function needs to stay. The framework calls it unconditionally — SingleTaskLearning invokes job.preprocess() with zero arguments at singletask_learning.py:114, so removing it causes an AttributeError before the pipeline even starts. I've kept it as a minimal pass-through (data=None, **kwargs) so the body is as lightweight as possible. Making it truly optional would require a change in Ianvs core itself.

2. predict : No structural issues here. Streaming/token accumulation, latency timing, memory capture, and hyperparameters are all working correctly. The one actual bug in the predict path was in _parse_timings: the regex was matching against llama_print_timings:, but modern llama-cpp-python emits llama_perf_context_print: instead — so prefill_latency was silently returning 0. I've widened the regex to cover both patterns; predict itself is otherwise untouched.

Two minor items I left as-is for now:
• data = data[:10] — the hardcoded 10-prompt cap appears intentional
• mem_usage is sampled at end-of-prompt (steady-state), not at peak

NishantSinghhhhh:
[Screencast from 2026-05-20 16-35-27.webm](https://github.com/user-attachments/assets/dd5440ac-1fb7-40e0-ad63-52ac0fd3807f)

The example runs without the pre-process function !!

MooreZheng (commented):
Good! Very close to the final version. Just squash the commit into one and this PR is good to go.

NishantSinghhhhh:
@MooreZheng Sir @hsj576 Sir squashed the commits !!

hsj576 (approved):
/lgtm

MooreZheng (approved):
/approve

PR #19, opened 2022-08-11
[OSPP]Multi-task Joint Inference for Ianvs

shifan-Z:
**What this PR does**:
The PR is the proposal of Multi-task Joint Inference for Ianvs.

The joint inference is better than single task. So we want to add an example named multi-task joint inference in Ianvs and verify the effectiveness in Object Detection.

PR #424, opened 2026-05-10
fix(core): enable fault-tolerant execution and incremental result persistence

### LINKED ISSUE #423: [BUG] Framework Crash on Downstream Configuration Fails to Persist Prior Successful Results
**What happened**:
When executing multi-configuration benchmarks, a failure in a downstream configuration causes a fatal framework crash (`RuntimeError`), erasing the results of all previously successful configurations from memory. 

A user reported this in the KubeEdge Slack ([Reference](https://cloud-native.slack.com/archives/C066UJZJKQE/p1771607930442969)): running the `cloud-edge-collaborative-inference-for-llm` benchmark on a CPU-only machine succeeded for the first configuration (`huggingface`, running for 8.5 hours) but crashed on the second (`vllm`) due to missing hardware. The exception terminated `run_testcases`, preventing Ianvs from writing the successful execution data to the `Rank` CSVs.

The framework must be fault-tolerant and persist data incrementally:
1. Exceptions in individual test cases should be caught, logged, and isolated so the execution loop can continue safely.
2. Successful test case results should be saved to the `Rank` CSVs immediately after completion, rather than waiting for the entire loop to finish.

1. Run `ianvs -f examples/cloud-edge-collaborative-inference-for-llm/benchmarkingjob.yaml` on a machine without a GPU.
2. Ensure the algorithm list includes a CPU-compatible backend followed by a GPU-dependent backend (e.g., `vllm`).
3. The first configuration will succeed, but the second will throw a hardware/dependency exception.
4. Check the workspace output. `all_rank.csv` remains unmodified because `Rank.save()` is never reached.

I have identified the bottleneck in `testcasecontroller.py` and `benchmarkingjob.py`. I have a PR ready that introduces an `incremental_save_cb` callback to persist results dynamically and isolates exceptions to prevent system failure.

Rishabh-git10:
**What type of PR is this?**
/kind bug

Enables fault-tolerant execution and incremental data persistence to prevent data loss during multi-configuration benchmarks. 

Previously, a downstream configuration failure caused a fatal `RuntimeError`, terminating the execution loop and erasing all prior successful results from memory.

**Changes:**
1. **Exception Isolation:** Modified `run_testcases` in `testcasecontroller.py` to catch, log, and isolate individual test exceptions, allowing the loop to continue safely.
2. **Incremental Persistence:** Introduced an `incremental_save_cb` callback in `benchmarkingjob.py` connected to the `Rank` module's concatenation logic. Successful results are now written to disk immediately after each test completes.

Fixes #423

### DIFF
--- a/core/cmd/obj/benchmarkingjob.py
+++ b/core/cmd/obj/benchmarkingjob.py
@@ -94,7 +94,9 @@ class BenchmarkingJob:
         def incremental_save(testcase, res, time):
             self.rank.save([testcase], {testcase.id: (res, time)}, output_dir=self.workspace)
 
-        succeed_testcases, test_results = self.testcase_controller.run_testcases(self.workspace)
+        succeed_testcases, test_results = self.testcase_controller.run_testcases(
+            self.workspace, incremental_save_cb=incremental_save
+            )
 
         if test_results:
             self.rank.save(succeed_testcases, test_results, output_dir=self.workspace)
--- a/core/cmd/obj/benchmarkingjob.py
+++ b/core/cmd/obj/benchmarkingjob.py
@@ -94,12 +94,10 @@ class BenchmarkingJob:
         def incremental_save(testcase, res, time):
             self.rank.save([testcase], {testcase.id: (res, time)}, output_dir=self.workspace)
 
-        _, test_results = self.testcase_controller.run_testcases(
-            self.workspace,
-            incremental_save_cb=incremental_save
-        )
+        succeed_testcases, test_results = self.testcase_controller.run_testcases(self.workspace)
 
         if test_results:
+            self.rank.save(succeed_testcases, test_results, output_dir=self.workspace)
             self.rank.plot()
 
     def _parse_config(self, config: dict):
--- a/core/cmd/obj/benchmarkingjob.py
+++ b/core/cmd/obj/benchmarkingjob.py
@@ -91,10 +91,15 @@ class BenchmarkingJob:
         self.testcase_controller.build_testcases(test_env=self.test_env,
                                                  test_object=self.test_object)
 
-        succeed_testcases, test_results = self.testcase_controller.run_testcases(self.workspace)
+        def incremental_save(testcase, res, time):
+            self.rank.save([testcase], {testcase.id: (res, time)}, output_dir=self.workspace)
+
+        _, test_results = self.testcase_controller.run_testcases(
+            self.workspace,
+            incremental_save_cb=incremental_save
+        )
 
         if test_results:
-            self.rank.save(succeed_testcases, test_results, output_dir=self.workspace)
             self.rank.plot()
 
     def _parse_config(self, config: dict):
--- a/core/testcasecontroller/testcasecontroller.py
+++ b/core/testcasecontroller/testcasecontroller.py
@@ -18,6 +18,7 @@ import copy
 
 from core.common import utils
 from core.common.constant import TestObjectType
+from core.common.log import LOGGER
 from core.testcasecontroller.algorithm import Algorithm
 from core.testcasecontroller.testcase import TestCase
 
@@ -43,7 +44,7 @@ class TestCaseController:
             for algorithm in algorithms:
                 self.test_cases.append(TestCase(test_env, algorithm))
 
-    def run_testcases(self, workspace):
+    def run_testcases(self, workspace, incremental_save_cb=None):
         """
         Run all test cases.
         """
@@ -52,11 +53,15 @@ class TestCaseController:
         for testcase in self.test_cases:
             try:
                 res, time = (testcase.run(workspace), utils.get_local_time())
-            except Exception as err:
-                raise RuntimeError(f"testcase(id={testcase.id}) runs failed, error: {err}") from err
+            except Exception as err: # pylint: disable=broad-exception-caught
+                LOGGER.error("testcase(id=%s) runs failed, error: %s", testcase.id, err)
+                continue
 
             succeed_results[testcase.id] = (res, time)
             succeed_testcases.append(testcase)
+
+            if incremental_save_cb:
+                incremental_save_cb(testcase, res, time)
 
         return succeed_testcases, succeed_results
 
--- a/core/testcasecontroller/testcasecontroller.py
+++ b/core/testcasecontroller/testcasecontroller.py
@@ -60,9 +60,6 @@ class TestCaseController:
             succeed_results[testcase.id] = (res, time)
             succeed_testcases.append(testcase)
 
-            if incremental_save_cb:
-                incremental_save_cb(testcase, res, time)
-
         return succeed_testcases, succeed_results
 
     @classmethod


Chidroopakanaparthy:
**Reproduction result:**

Checked out pr-424. Key changes in core:

`testcasecontroller.py`:
```diff
-  raise RuntimeError(f"testcase(id={testcase.id}) runs failed, error: {err}") from err
+  LOGGER.error("testcase(id=%s) runs failed, error: %s", testcase.id, err)
+  continue
```

`benchmarkingjob.py` -  adds incremental save callback:
```diff
+  def incremental_save(testcase, res, time):
+      self.rank.save([testcase], {testcase.id: (res, time)}, output_dir=self.workspace)
```

**Direct relevance to smart_coding:** This fixes Bug #19 from Issue #440 - "No error handling for API failure, inference results lost." On main, when DeepSeek returns a 401, the exception propagates and crashes the benchmark, discarding 25 seconds of CPU inference results. With this PR, the failed testcase is logged and skipped, all other testcases continue and results already computed are saved incrementally.

**Recommendation:** Merge. This is a core improvement that benefits all examples.

core/cmd/obj/benchmarkingjob.py
```diff
@@ -91,7 +91,12 @@ def run(self):
         self.testcase_controller.build_testcases(test_env=self.test_env,
                                                  test_object=self.test_object)
 
-        succeed_testcases, test_results = self.testcase_controller.run_testcases(self.workspace)
+        def incremental_save(testcase, res, time):
```
Rishabh-git10, line 94: Enables direct runtime file system persistence to guarantee historical benchmark results are saved if trailing configurations fail.

Rishabh-git10 (commented):
**Recommendation:** Approve.

**Reproduce:** The implementation introduces a try-except isolation block and an incremental results persistence callback to ensure partial downstream configuration crashes do not discard earlier successful benchmark data.

**Logs:**
```bash
...ianvs$ sed -n '53,58p' core/testcasecontroller/testcasecontroller.py
# Verified: try-except wrapper isolates execution failures per testcase sequence loop iteration.
# Verified: incremental_save_cb commits successful iterations instantly to disk storage.
```

coldstarted:
LFX Term 02 : Task 2

## PR Review — `fix(core): enable fault-tolerant execution and incremental result persistence`

### Recommendation: Merge with minor revisions

The incremental save feature is genuinely useful: if ianvs crashes mid-run with 9 of 10 test cases complete, today all results are lost. This PR saves each successful result immediately. However the fault-tolerant execution (swallowing exceptions with `continue`) is a double-edged sword — it lets a multi-config benchmark run to completion even when one config fails, but it makes failures easy to miss. One issue before merge: the final run summary does not report how many test cases failed.

### Introduction

Previously, a single `RuntimeError` from any test case in `run_testcases()` immediately re-raised and terminated the entire benchmark loop. Any test cases that had already succeeded still had their results discarded because `rank.save()` was only called after all test cases completed. This PR adds (1) a `continue` on exception (so the loop keeps running) and (2) an `incremental_save_cb` that writes each successful result to disk immediately.

### Module Details

| `core/testcasecontroller/testcasecontroller.py` | +5/-3 | `raise RuntimeError` → `LOGGER.error + continue`; `incremental_save_cb` parameter |
| `core/cmd/obj/benchmarkingjob.py` | +6/-1 | `incremental_save` closure passed as callback |

**The change:**
```diff
-    def run_testcases(self, workspace):
+    def run_testcases(self, workspace, incremental_save_cb=None):
         for testcase in self.test_cases:
             try:
                 res, time = (testcase.run(workspace), utils.get_local_time())
-            except Exception as err:
-                raise RuntimeError(f"testcase(id={testcase.id}) runs failed, error: {err}") from err
+            except Exception as err:
+                LOGGER.error("testcase(id=%s) runs failed, error: %s", testcase.id, err)
+                continue
             succeed_results[testcase.id] = (res, time)
             succeed_testcases.append(testcase)
+            if incremental_save_cb:
+                incremental_save_cb(testcase, res, time)
```

**Incremental save closure:**
```python
def incremental_save(testcase, res, time):
    self.rank.save([testcase], {testcase.id: (res, time)}, output_dir=self.workspace)

succeed_testcases, test_results = self.testcase_controller.run_testcases(
    self.workspace, incremental_save_cb=incremental_save
)
```
This writes a partial leaderboard CSV after each test case completes. The final `rank.save(succeed_testcases, test_results, ...)` call still writes the full leaderboard at the end.

**Issue — no failure summary:**
When test cases fail, `LOGGER.error` is called but the final leaderboard only shows successes. A user running a 10-algorithm benchmark won't know 3 of them failed unless they read the log carefully. At minimum, print a summary like `"3 of 10 test cases failed — check logs above"` after the loop.

**Issue — broad exception swallowing:**
`except Exception` catches all exceptions including `MemoryError`, `RecursionError`, `KeyboardInterrupt` (via `Exception` in Python 3 — actually `KeyboardInterrupt` inherits from `BaseException`, not `Exception`, so it's fine). The scope is appropriate.

### Evidence

Checked out `fix/intermediate-result-persistence` from `Rishabh-git10/ianvs`. Both files pass `py_compile`. Traced the `incremental_save_cb` call chain — the closure correctly calls `self.rank.save()` with a single-element list, producing a valid (partial) leaderboard entry. The final `rank.save()` still produces the full leaderboard. No double-write bug — partial saves write individual rows and are overwritten by the final full save.

### Reproduce

```bash
git fetch upstream pull/424/head:pr-424
git checkout pr-424

find core/ -name '*.py' | xargs -I{} python3 -m py_compile {} && echo 'core OK'

# Confirm the new signature
grep -n 'def run_testcases\|incremental_save' core/testcasecontroller/testcasecontroller.py

# Confirm the callback is wired up
grep -n 'incremental_save' core/cmd/obj/benchmarkingjob.py
```

### Logs

```
$ ianvs -f examples/government_rag/singletask_learning_bench/benchmarkingjob.yaml 2>&1 | head -60

[2026-05-26 23:42:44,572] testcasecontroller.py(57) [ERROR] - testcase(id=7c2604a6-...) runs failed,
  error: (paradigm=singletasklearning) pipeline runs failed,
  error: [Errno 2] No such file or directory: '/path/ianvs/dataset/gov_rag/dataset'
BaseModel preprocess
```

**What the logs tell us — and why this is the exact proof we needed:**

Compare the output of PR #336 (blocker), PR #309 (passing), and this PR side by side:

- PR #309 (passing): `BaseModel preprocess` → full crash traceback with RuntimeError chain
- PR #336 (blocker): instant `ValueError` — never reached preprocess
- PR #424 (this PR): `[ERROR] testcase runs failed` → `BaseModel preprocess` — no crash, no traceback

The `[ERROR]` line appearing without a crash traceback is the proof. On `main`, the same `FileNotFoundError` from `GovernmentRAG` would have propagated up as a `RuntimeError` chain and killed the entire benchmark. On `pr-424`, it is caught by the new `except Exception` block, logged with `LOGGER.error()` at `testcasecontroller.py line 57`, and execution continues — which is exactly what the `continue` keyword in the PR does.

Why `BaseModel preprocess` appears after the `[ERROR]` line — the print inside `basemodel.py` writes to stdout (buffered), while `LOGGER.error()` writes to the log handler (flushed immediately). When piped through `head`, stdout flushes after the logger output, making `BaseModel preprocess` appear last even though it was executed first. This is stdout buffering — not a bug.

What this run confirms:
- Framework starts, config parses, paradigm dispatches — nothing broken
- Test case failure is logged as `[ERROR]`, not raised as a crashing `RuntimeError`
- `continue` works — execution did not stop at the error
- Program exits cleanly with no multi-level RuntimeError traceback chain
- No failure summary printed at the end — this is the one issue flagged in the review

### Problems Found

**1. No failure summary at end of run**
Failed test cases are logged via `LOGGER.error` but nothing summarises failures after the loop. Fix: track `failed_count` in the loop and print `LOGGER.warning("%d of %d test cases failed", failed_count, total)` before returning.

---

https://github.com/user-attachments/assets/4802c037-d3d1-44b0-b40a-cd109dbfe7b2

PRANABKUMARNEOGI:
Hi @Rishabh-git10  , thanks for putting this together! This is a much-needed boundary to prevent the framework from crashing due to downstream configuration failures.

I reviewed this PR as part of my LFX Mentorship evaluation. While analyzing the code path, I noticed a critical hardware regression regarding Resource Starvation that we should address before merging.

### **Observation**
The new try-except block successfully catches software exceptions in the orchestrator loop and issues a continue. However, there is no mechanism to clear the hardware state (GPU VRAM) before yielding control to the next task.

### **Impact**
If a benchmarking algorithm crashes mid-execution after tensors have already been allocated, this except block will safely swallow the error, but those tensors will remain permanently locked in GPU memory. As a result, the next sequential task in the multi-stage benchmark is almost guaranteed to fail with a false-negative OutOfMemoryError (OOM).

### **Recommendation**
To make this execution boundary truly fault-tolerant, the framework needs to tear down the orphaned hardware state before continuing the software loop.

Adding explicit garbage collection inside the except block would guarantee a pristine memory pool for the next task. For example:

```python
except Exception as e:
    logger.error(f"Task failed with error: {e}")
    
    # 1. Force Python garbage collection
    import gc
    gc.collect()
    
    # 2. Flush orphaned tensors from GPU VRAM
    import torch
    if torch.cuda.is_available():
        torch.cuda.empty_cache()
        
    continue
```

Would you be open to incorporating a hardware teardown step like this before the continue statement? Let me know what you think!

Talha8899:
## Review of PR #424

PR #424 addresses an important framework-level problem: when one testcase fails during a multi-configuration benchmark, Ianvs should continue executing the remaining testcases instead of terminating the entire benchmark and losing previously completed results.

The implementation changes `TestCaseController.run_testcases()` so testcase exceptions are logged and execution continues, and adds incremental persistence so successful testcase results can be written while the benchmark is running.

I reproduced the PR locally using:

```text
ianvs -f examples\government_rag\singletask_learning_bench\benchmarkingjob.yaml
```

The testcase failed because of a missing dependency:

```text
[ERROR] testcase(id=53e2a2f6-...) runs failed,
error: ... No module named 'langchain_community'.
```

However, execution continued and the process subsequently reported:

```text
[INFO] benchmarking.py(39) - benchmarkingjob runs successfully.
```

This confirms that the new exception-isolation behavior works: the testcase failure was logged instead of propagating as the previous fatal `RuntimeError`.

### Finding

The framework can now continue after an individual testcase failure, but the overall benchmark still reports `benchmarkingjob runs successfully` even when a testcase has failed.

This creates a misleading success state. A user or automation system can interpret the benchmark as successful even though one or more testcases were not executed successfully.

This is different from simply lacking a failure summary: the issue is that the top-level execution status itself can indicate success while the benchmark contains a failed testcase.

### Impact

This can produce false-positive CI/automation results and make an incomplete benchmark appear successful. It is especially important for example restoration because many example failures are environment-, dependency-, or configuration-related; the framework should preserve successful results without hiding the existence of failed configurations.

### Recommendation

Retain the fault-tolerant execution and incremental persistence design, but make the final benchmark status reflect testcase failures. The runner should track failed testcases and expose a non-success status or equivalent explicit failure state when one or more testcases fail, while still preserving successful results.

Existing reviews already discuss the missing failure summary, incremental-save performance, callback failure handling, and GPU-resource cleanup. I am not repeating those findings here; this review focuses on the mismatch between testcase failure and the final benchmark success status.

**Verification boundary:** the failure-isolation behavior was reproduced locally on PR #424. The testcase itself could not succeed because the environment lacked `langchain_community`; this dependency failure is used only to exercise the Core failure-handling path.

Issue #353, opened 2026-02-13
Government Single-Task : [LFX] Fix Critical Issues in Government Single-Task Learning Benchmark Example


Mshahnawaz1:
# LFX Mentorship Proposal: Fix `government/singletask_learning` Example in KubeEdge Ianvs (Task 1)

---
Applicant Information:
- **Name:** Mohammad Shahnawaz
- **Email:** shahnawaz919956@gmail.com
- **GitHub:** [Mshahnawaz1](https://github.com/Mshahnawaz1)
- **Self link**: [Link](https://github.com/kubeedge/ianvs/issues/353)

_Submitted for: LFX Term 1 2026 Mentorship — KubeEdge Ianvs_

### Overview
The `government/singletask_learning` example in [KubeEdge Ianvs](https://github.com/kubeedge/ianvs) demonstrates edge AI benchmarking applied to government-domain question-answering tasks. It provides two evaluation paradigms:

- **Objective benchmark** — Model outputs are assessed against ground-truth answers over multiple-choice questions with programmatic accuracy scoring.
- **Subjective benchmark** — Responses are evaluated using LLM-as-judge or rubric-based scoring across multiple evaluation dimensions (modality, domain, task type).

This example serves researchers and practitioners who want to evaluate large language models on structured government-domain datasets using Ianvs's distributed, edge-aware benchmarking infrastructure. However, **the example is currently non-functional** due to two distinct categories of bugs.

---

### Issue Category 1: Setup Failures (Overlaps with Issue #340)

Like several other examples in the repository (discussed in issue #340), `government/singletask_learning` suffers from infrastructure-level failures:

1. **Hardcoded file paths** — Absolute paths left by previous contributors, causing immediate `FileNotFoundError` on other system.
2. **Missing `requirements.txt`** — No modules list to getting started causing users `ModuleNotFoundError`.
3. **Inadequate documentation** — The existing README provides inaccurate schema and structure of the datasets and no proper way for benchmark setup.
4. **Hardcoded device assignment** — The code assumes CUDA device 0 is available and fails ungracefully on CPU-only systems.
5. **Hardcoded LLM model** — The `Qwen/Qwen2-0.5B-Instruct` model is hardcoded making it harder to test other language models.

These issues are frustrating but simpler to fix. They share similarities with issues discussed in other examples, and can be fixed fairly easily.

---
### Issue Category 2: Core Data Loading Bug — `JSONMetaDataParse` Incompatibility (Unique to This Example)
[...]

PR #426, opened 2026-05-12
fix(government): restore singletask_learning_bench example

### LINKED ISSUE #425: bug: government/singletask_learning_bench example broken by stale paths, missing deps, and dataset layout changes
**Description**
The examples/government/singletask_learning_bench example cannot currently be run by a new contributor. Multiple issues block execution from the very first step.

**Issues found**

1. YAML files contain hardcoded /home/icyfeather/... absolute paths for workspace, dataset, and model locations.
2. No per-example requirements.txt exists. Contributors have no reliable install path.
3. README instructs users to manually patch sedna in site-packages.
4. Model loading hardcoded to /home/icyfeather/models/Qwen2-0.5B-Instruct with no env-based fallback.
5. Code forced cuda, failing on CPU-only machines with no fallback.
6. GovAff Kaggle dataset layout changed to "government/multi-choice questions" and "government/subjective questions" — no longer matches expected paths in YAML.
7. Objective rows use question/options/answer but Sedna metadata parser expects query/response.
8. Subjective train data uses query/response but Sedna JSONL train parser expects question/answer.
9. Subjective metric looked for 'Overall Points' but dataset judge prompts request 综合得分.

**Expected behavior**
Clone Ianvs, install dependencies, download GovAff dataset, run objective and subjective benchmark jobs successfully.

**Actual behavior**
Setup fails at multiple points before any benchmark runs. A new contributor cannot complete this example without manually debugging source code.

**Environment**
- Python 3.9 / 3.10
- Dataset: https://www.kaggle.com/datasets/kubeedgeianvs/the-government-affairs-dataset-govaff

**Additional context**
A fix addressing all of the above is being submitted in a separate PR.

Sahil-u07:
Restores the government single-task learning benchmark so contributors can run it on Python 3.9/3.10 without manual debugging.

## Changes

- Add pinned per-example requirements.txt
- Replace all hardcoded /home/icyfeather/... paths in YAML files with repo-relative paths
- Add prepare_dataset.py to convert current GovAff Kaggle layout into Ianvs-ready objective/subjective folders
- Make model path configurable via GOVERNMENT_BENCH_MODEL env var with Qwen/Qwen2-0.5B-Instruct as default
- Add CUDA/CPU fallback in both basemodel.py files
- Fix subjective score extraction regex to handle Chinese judge prompt format (综合得分)
- Update README with accurate install, dataset download, preparation, model config, and run instructions

## Verification

- python -m py_compile passed on all changed Python files
- llm_judgement score extraction tested for Overall Points, 综合得分, and 综合分数 patterns
- All YAML files parse with yaml.safe_load
- Dataset preparation tested against current GovAff Kaggle zip, produced 1600 objective rows and 1045 subjective rows
- No remaining /home, /root, hardcoded model paths, or TODO markers in target directory

Fixes #425

cc @MooreZheng

### DIFF
--- a/examples/government/singletask_learning_bench/objective/testalgorithms/gen/basemodel.py
+++ b/examples/government/singletask_learning_bench/objective/testalgorithms/gen/basemodel.py
@@ -14,22 +14,16 @@
 
 from __future__ import absolute_import, division
 
-import os
-import tempfile
-import time
-import zipfile
 import logging
-
-import numpy as np
+import os
 import random
+
+import torch
 from tqdm import tqdm
-from sedna.common.config import Context
 from sedna.common.class_factory import ClassType, ClassFactory
 from core.common.log import LOGGER
 
-
 from transformers import AutoModelForCausalLM, AutoTokenizer
-device = "cuda" # the device to load the model onto
 
 
 logging.disable(logging.WARNING)
@@ -43,12 +37,21 @@ os.environ['BACKEND_TYPE'] = 'TORCH'
 class BaseModel:
 
     def __init__(self, **kwargs):
-        self.model = AutoModelForCausalLM.from_pretrained(
-            "/home/icyfeather/models/Qwen2-0.5B-Instruct",
-            torch_dtype="auto",
-            device_map="auto"
+        self.model_name_or_path = os.getenv(
+            "GOVERNMENT_BENCH_MODEL",
+            "Qwen/Qwen2-0.5B-Instruct",
         )
-        self.tokenizer = AutoTokenizer.from_pretrained("/home/icyfeather/models/Qwen2-0.5B-Instruct")
+        self.device = "cuda" if torch.cuda.is_available() else "cpu"
+        model_kwargs = {"torch_dtype": "auto"}
+        if torch.cuda.is_available():
+            model_kwargs["device_map"] = "auto"
+        self.model = AutoModelForCausalLM.from_pretrained(
+            self.model_name_or_path,
+            **model_kwargs,
+        )
+        if not torch.cuda.is_available():
+            self.model.to(self.device)
+        self.tokenizer = AutoTokenizer.from_pretrained(self.model_name_or_path)
 
     def train(self, train_data, valid_data=None, **kwargs):
         LOGGER.info("BaseModel train")
@@ -59,19 +62,25 @@ class BaseModel:
 
     def predict(self, data, input_shape=None, **kwargs):
         LOGGER.info("BaseModel predict")
-        LOGGER.info(f"Dataset: {data.dataset_name}")
-        LOGGER.info(f"Description: {data.description}")
-        LOGGER.info(f"Data Level 1 Dim: {data.level_1_dim}")
-        LOGGER.info(f"Data Level 2 Dim: {data.level_2_dim}")
+        if hasattr(data, "dataset_name"):
+            LOGGER.info(f"Dataset: {data.dataset_name}")
+            LOGGER.info(f"Description: {data.description}")
+            LOGGER.info(f"Data Level 1 Dim: {data.level_1_dim}")
+            LOGGER.info(f"Data Level 2 Dim: {data.level_2_dim}")
         
+        questions = list(getattr(data, "x", data))
+        labels = list(getattr(data, "y", []))
         answer_list = []
-        for line in tqdm(data.x, desc="Processing", unit="question"):
-            # 3-shot
-            indices = random.sample([i for i, l in enumerate(data.x) if l != line], 3)
+        for line in tqdm(questions, desc="Processing", unit="question"):
+            candidates = [
+                i for i, question in enumerate(questions)
+                if question != line and i < len(labels)
+            ]
+            indices = random.sample(candidates, min(3, len(candidates)))
             history = []
             for idx in indices:
-                history.append({"role": "user", "content": data.x[idx]})
-                history.append({"role": "assistant", "content": data.y[idx]})
+                history.append({"role": "user", "content": questions[idx]})
+                history.append({"role": "assistant", "content": labels[idx]})
             history.append({"role": "user", "content": line})
             response = self._infer(history)
             answer_list.append(response)
@@ -89,7 +98,7 @@ class BaseModel:
             tokenize=False,
             add_generation_prompt=True
         )
-        model_inputs = self.tokenizer([text], return_tensors="pt").to(device)
+        model_inputs = self.tokenizer([text], return_tensors="pt").to(self.device)
         
         generated_ids = self.model.generate(
             model_inputs.input_ids,
--- a/examples/government/singletask_learning_bench/requirements.txt
+++ b/examples/government/singletask_learning_bench/requirements.txt
@@ -1,0 +1,18 @@
+# Install Ianvs itself from the repository root with:
+#   python -m pip install -e .
+#
+# Install the bundled Sedna wheel with:
+#   pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl
+
+accelerate==0.30.1
+fastapi==0.95.2
+numpy==1.24.4
+openai==1.30.5
+pandas==1.5.3
+prettytable==2.5.0
+pydantic==1.10.15
+PyYAML==6.0.2
+torch==2.2.2
+tqdm==4.66.5
+transformers==4.40.2
+uvicorn==0.22.0
--- a/examples/government/singletask_learning_bench/subjective/testalgorithms/gen/basemodel.py
+++ b/examples/government/singletask_learning_bench/subjective/testalgorithms/gen/basemodel.py
@@ -14,22 +14,16 @@
 
 from __future__ import absolute_import, division
 
-import os
-import tempfile
-import time
-import zipfile
 import logging
+import os
 
-import numpy as np
-import random
+import torch
 from tqdm import tqdm
-from sedna.common.config import Context
 from sedna.common.class_factory import ClassType, ClassFactory
 from core.common.log import LOGGER
 from openai import OpenAI
 
 from transformers import AutoModelForCausalLM, AutoTokenizer
-device = "cuda" # the device to load the model onto
 
 
 logging.disable(logging.WARNING)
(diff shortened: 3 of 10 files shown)


AswaniSahoo:
### Review: PR #426

**Recommendation: Merge with fixes.**

**Scope:** Example-only. All 10 modified files in `examples/government/singletask_learning_bench/`. Exposes a core-example contract mismatch: `core/testcasecontroller/.../singletask_learning.py:114` calls `job.preprocess()` which this example never implements.

Reviewed all 10 files and tested on Windows 11, Python 3.14.3, RTX 3050 (CUDA 12.6). gemini-code-assist reviewed this PR on May 12 and flagged two basemodel issues (line 71 getattr fallback, line 138 empty `response.choices` IndexError); Copilot bot flagged the README workspace path mismatch the same day. Both are valid. Findings below are separate from those.

**What works:**
- `basemodel.py` device detection via `torch.cuda.is_available()` is the right fix
- `prepare_dataset.py` automates Kaggle-to-Sedna data transformation
- `llm_judgement.py` regex now handles Chinese and English score formats
- YAML path fixes replace all `/home/icyfeather/...` hardcoded paths

**Issues found (not duplicates of gemini or Copilot review):**

1. **`requirements.txt` pins are too old for Python 3.13+.**
```
accelerate==0.30.1   # March 2024
torch==2.2.2         # March 2024
numpy==1.24.4        # fails on Python 3.12+ (pkgutil.ImpImporter removed)
```
Use `>=` instead of `==` so the example installs cleanly on current Python.

2. **PR strips 106 inline documentation comment lines from `benchmarkingjob.yaml` (both objective and subjective).** The original YAML files included contributor-facing help text explaining every config field — `selected_dataitem` options, `visualization` modes, `paradigms` values, `save_mode` semantics. This PR removes all of it. A new contributor landing on `benchmarkingjob.yaml` after a GitHub search no longer sees what `"selected_only"` means or which `paradigms` values are valid. The README rewrite covers the overall flow but does not replace this per-field inline reference. Recommend restoring the inline comments — the YAML file is the canonical config artifact contributors edit, and inline docs there are higher-leverage than equivalent README text.

3. **`prepare_dataset.py` doesn't remap field names.** GovAff uses `question` and `response`; Sedna expects `query` and `answer`. After running the prep script the pipeline still hits `KeyError: 'query'`. Either add the remap step in `prepare_dataset.py` or fix at the Sedna load layer.

4. **Missing `preprocess()` method (untracked in any open issue at time of review).** Core's `singletask_learning.py:114` calls `job.preprocess()`. This PR does not add it to subjective `basemodel.py`. The RAG version has `preprocess()` at line 157; the singletask version does not. Model loads (290 weights on GPU), then crashes:
```
AttributeError: 'BaseModel' object has no attribute 'preprocess'
```
Side-by-side code proof (objective basemodel does call it, subjective does not):

08-code-proof-preprocess-missing

Minimal fix:
```python
def preprocess(self, **kwargs):
    return None
```

**Verification:** Full reproduction screenshots (numpy install failure, preprocess crash, code proof, benchmark success after the local fix) in [issue #436](https://github.com/kubeedge/ianvs/issues/436). The preprocess crash and the post-fix benchmark success are images 7-8 and 13 there.

**Suggested fixes:**
1. Use `>=` version pins instead of `==`
2. Restore the 106 inline YAML doc comments (or move them into a `docs/config-reference.md` if README brevity was the goal)
3. Add field name remapping in `prepare_dataset.py`
4. Add `preprocess()` stub to singletask `basemodel.py`

Sahil-u07:
## Self-Review: Gaps Found During Further Testing
 
After submitting this PR I continued testing across different environments and found the following issues that this PR does not yet address. I am documenting them here for reviewer awareness and to be transparent about what a follow-up PR needs to fix.
 
---
 
### What This PR Successfully Fixes
 
- Bug 1: `requirements.txt` now exists - contributors no longer need to guess dependencies
- Bug 2: All five hardcoded `/home/icyfeather/` paths replaced with env-variable-driven relative paths  
- Bug 3: `device_map="auto"` added to `basemodel.py` - CPU fallback now works
- Bug 4: `prepare_dataset.py` added to normalize `"question"` to `"query"` field naming
- Bug 6 (partial): `llm_judgement.py` regex extended from 1 pattern to 3
---

### Problem 1 - torch==2.2.2 Strict Pin Forces Version Conflict
 
My `requirements.txt` pins `torch==2.2.2`. Testing showed that the current latest torch is `2.12.0` - 10 minor versions ahead of my pin. Installing my requirements on a system with newer torch forces a major downgrade. On Python 3.11 and 3.12 systems where the 2.2.2 wheel was never published, the install fails entirely.
 
```
ERROR: Could not find a version of torch that satisfies the requirement torch==2.2.2
ERROR: No matching distribution found for torch==2.2.2
```
 
SS3

 
Fix:
```
# Change from:
torch==2.2.2
transformers==4.40.2
 
# Change to:
torch>=2.2.2,<3.0.0
transformers>=4.35.0,<5.0.0
```
 
---
 
### Problem 2 - Missing preprocess() in Both basemodel.py Files
 
The `singletasklearning` paradigm calls `basemodel.preprocess(data)` before every inference run. I scanned both `basemodel.py` files in this example:
 
SS4

 
Both files - objective mode and subjective mode - are missing this method. At runtime this crashes as:
 
```
AttributeError: 'GovernmentModel' object has no attribute 'preprocess'
```
 
This happens after all path and dependency fixes are applied, meaning a contributor who successfully sets up the environment will still be completely blocked here.
 
Fix to add in both `objective/testalgorithms/gen/basemodel.py` and `subjective/testalgorithms/gen/basemodel.py`:
 
```python
def preprocess(self, data):
    """Required by the singletasklearning paradigm interface."""
    return data
```
 
---
 
### Problem 3 - llm_judgement.py Regex Misses English Output Format
 
I tested all three patterns in my PR against `"Overall Score: 7"` which is what Qwen2 returns in English mode:
 
SS5

 
All three return `None`. Every subjective evaluation where the judge responds in English silently records 0 with no error or warning. The benchmark appears to complete successfully while producing invalid scores.
 
Fix:
```python
match = re.search(
    r'(?:综合得分|综合分数|Overall Score|Overall Points|Score|Rating)[：:\s]+(\d+(?:\.\d+)?)',
    judge_response,
    re.IGNORECASE
)
```
 
---
 
### Problem 4 - fix_data.py Crashes on Subjective Rows
 
SS6

 
The script accesses `row['choices']` on every row. Subjective rows have no `choices` field. This aborts dataset preparation before any run begins.
 
Fix:
```python
for row in data:
    if 'choices' in row:
        # objective row
    else:
        # subjective row
```
 
---
 
### Problem 5 - Grid Search YAML Key Is Wrong and Systemic Across the Repo
 
I scanned all `algorithm.yaml` files in the repository:
 
SS7

 
The `hyperparameters` key was found in 6+ other examples across the repo including `federated-llm`, `llm-edge-benchmark-suite`, `RoboDK Palletizing`, and `yaoba`. Ianvs `TestCaseController` looks for `parameterGroup`. The wrong key causes grid search to be silently skipped in every affected example.
 
Fix:
```yaml
# Current (wrong):
hyperparameters:
  - temperature:
      values: [0.1, 0.7]
 
# Correct:
parameterGroup:
  - name: temperature
    values: [0.1, 0.7]
```
 
---
 
### Summary: Bugs Still Open After This PR
 
| Bug | Status |
|---|---|
| Bug 1 - No requirements.txt | Fixed |
| Bug 2 - Hardcoded paths | Fixed |
| Bug 3 - Forced CUDA | Fixed |
| Bug 4 - query/question field | Partial workaround |
| Bug 5 - Missing preprocess() | NOT fixed |
| Bug 6 - Regex | Partial |
| Bug 7 - fix_data.py KeyError | NOT fixed |
| Bug 8 - acc.py field mismatch | NOT fixed |
| Bug 9 - Grid Search YAML | NOT fixed |
 
I will address these in a follow-up commit.

examples/government/singletask_learning_bench/requirements.txt
```diff
@@ -0,0 +1,18 @@
...
+
+accelerate==0.30.1
+fastapi==0.95.2
+numpy==1.24.4
+openai==1.30.5
+pandas==1.5.3
+prettytable==2.5.0
+pydantic==1.10.15
+PyYAML==6.0.2
+torch==2.2.2
```
Sahil-u07, line 15: This pin is not available for Python 3.11 or 3.12 as a pre-built wheel.
Any contributor on Python 3.11+ gets a hard install error.
Suggest: torch>=2.2.2,<3.0.0

examples/government/singletask_learning_bench/objective/testalgorithms/gen/basemodel.py
```diff
@@ -43,12 +37,21 @@
 class BaseModel:
 
     def __init__(self, **kwargs):
+        self.model_name_or_path = os.getenv(
```
Sahil-u07, line 40: The singletasklearning paradigm calls basemodel.preprocess() once before 
predict() runs. Neither the objective nor the subjective basemodel.py in 
this PR implements that method.

What actually happens at runtime after all the path and dependency fixes 
in this PR are applied:

    AttributeError: 'BaseModel' object has no attribute 'preprocess'

So a contributor who applies this PR, sets up the dataset correctly, 
installs all dependencies, and runs the benchmark still hits a crash 
before any inference happens. That is a bad experience.

Minimum fix needed in both basemodel.py files:

    def preprocess(self, data, **kwargs):
        return data

examples/government/singletask_learning_bench/subjective/testalgorithms/gen/basemodel.py
```diff
@@ -43,12 +37,21 @@
 class BaseModel:
 
     def __init__(self, **kwargs):
+        self.model_name_or_path = os.getenv(
```
Sahil-u07, line 40: Same missing preprocess() as objective basemodel.py. The singletasklearning 
paradigm calls it before predict() - this class doesn't implement it, so 
both modes crash with AttributeError before any inference runs.

Fix needed:

    def preprocess(self, data, **kwargs):
        return data

examples/government/singletask_learning_bench/subjective/testenv/llm_judgement.py
```diff
@@ -13,28 +13,34 @@
...
 
+
 def extract_comprehensive_score(input_str):
-    # extract overall points
-    match = re.search(r"'Overall Points': (\d+)", input_str)
-    if match:
-        return int(match.group(1))
-    else:
-        return None
+    patterns = [
```
Sahil-u07, line 24: These patterns miss "Overall Score:" which is what Qwen2 returns in English mode.
Tested locally with all 3 patterns against "Overall Score: 7" - all return None.
Result: every subjective score silently recorded as 0 with no error or warning.
Suggested fix:

   ```
 match = re.search(
        r'(?:综合得分|综合分数|Overall Score|Overall Points|Score|Rating)[：:\s]+(\d+(?:\.\d+)?)',
        judge_response,
        re.IGNORECASE
    )
```

**Comment by Chidroopakanaparthy** [comment]
**Reproduction result - impact on smart_coding:**

Checked out pr-426. 10 files changed, all in `examples/government/singletask_learning_bench/`.

Ran:
```bash
ianvs -f examples/government/singletask_learning_bench/subjective/benchmarkingjob.yaml
Output:
ModuleNotFoundError: No module named 'sedna'
```

Fails at Sedna import - Sedna must be installed separately. The code changes themselves are sound. This is an environment prerequisite gap, not a bug in the PR.

The `os.getenv()` model path pattern and conditional `device_map` in this PR are what `smart_coding`'s basemodel.py needs for Bugs #2 and #3.

**Recommendation:** Merge but from the pov of smart_coding

Rudra-RNC:
Recommendation: major revision. 
The PR's compilation/YAML/regex/dataset-preparation checks support the portability portion, but they do not demonstrate the Issue #353 rich-schema invariant. Please add a tiny objective fixture proving `options` reaches the prompt and a subjective fixture proving judge/dimension metadata survives loading. Also report the exact end-to-end command, exit code, sample count, and metric output. Until those gates exist, this appears to fix startup symptoms without proving evaluation correctness. [Full analysis in Discussion: INSERT LINK](https://github.com/kubeedge/ianvs/discussions/981)

Rudra-RNC:
Restoration order should place the government example's rich-schema contract before claiming end-to-end success: setup fixes can expose the loader defect only after the process advances. Add a minimal fixture gate, then rerun it after prompt/evaluation changes. If the gate fails, revert the example integration without reverting the independent fixture or diagnostic tests.

PR #427, opened 2026-05-14
Add cloud-edge speculative decoding benchmark example

zjteee:
**What type of PR is this?**

Fixes #


examples/cloud-edge-speculative-decoding-benchmark/testalgorithms/speculative-decoding/algorithms/block/core.py
```diff
@@ -0,0 +1,422 @@
...
+                target_hidden=state["target_hidden"],
+                noise_embedding=noise_embedding,
+                position_ids=position_ids,
+                past_key_values=state["past_key_values_draft"],
+                use_cache=True,
+                is_causal=False,
+            )
+            hidden_slice = hidden_states[:, -block_size + 1 :, :]
+            draft_logits = self.lm_head(hidden_slice.to(self.lm_head_device))
+            state["past_key_values_draft"].crop(start)
```
Rishabh-git10: Calling `.crop()` directly on standard Hugging Face cache objects will trigger a runtime `AttributeError`. Refactor this block to utilize the safe tensor-slicing fallback helper implemented in `ar/core.py`.

core/testcasecontroller/algorithm/paradigm/joint_inference/joint_inference.py
```diff
@@ -179,9 +186,24 @@ def _inference(self, job):
...
+                    cloud_result = infer_res[3]
+            elif isinstance(infer_res, dict):
+                simulation = infer_res.get("simulation", {}) or {}
+                routed_to = simulation.get("routed_to")
+                if routed_to == "edge":
+                    edge_result = infer_res
+                elif routed_to == "cloud":
+                    cloud_result = infer_res
+
+            if edge_result:
```
Rishabh-git10, line 209: The parser loop lacks an explicit evaluation branch for the `"collaboration"` routing payload, leaving result pointers at `None` and freezing progress postfix metrics.

Rishabh-git10 (commented):
**Recommendation:** Request Changes / Do Not Merge.

The current implementation introduces significant monolithic code duplication by copy-pasting model execution blocks instead of extending existing core framework abstractions. Additionally, the benchmark configurations introduce hardcoded paths and environmental parameters that break cross-platform portability.

**Reproduce:** Cannot execute smoothly due to architectural and structural bugs. Specifically:
- Blindly invoking `.crop()` on standard `transformers.Cache` data fields triggers an immediate runtime `AttributeError`.
- The evaluation routing tree completely omits the `"collaboration"` state string payload, resulting in tracking metric lockups.

**Logs:**
```bash
  File ".../algorithms/block/core.py", line 152
    state["past_key_values_draft"].crop(start)
  AttributeError: 'DynamicCache' object has no attribute 'crop'
```

MooreZheng (commented):
1. discussion with nishant with the Sedna wheel. The KubeEdge SIG suggest to merge the wheel from this PR first to support the new joint inference scheme
2. tackle other review which have been fixed serveral days ago

core/testcasecontroller/algorithm/paradigm/joint_inference/joint_inference.py
zjteee: Thanks for checking these issues so carefully.

For the cache handling in `algorithms/block/core.py`, you are right. The current block path assumes the cache object can be cropped directly, which is not a safe assumption for standard Hugging Face cache objects. The AR path already has a safer cache-trimming fallback, and I will refactor the block path to reuse the same logic instead of calling `.crop()` directly.

For the `collaboration` routing branch in `joint_inference.py`, this point is also valid. The current counting logic only explicitly handles `"edge"` and `"cloud"`, and does not include a separate branch for `"collaboration"`, so that part is incomplete. In the current benchmark flow, the collaboration result is still carried in the main payload and parsed downstream, so this is not exactly the same as the main metric path failing, but the routing branch itself should definitely be made explicit. I will fix that as well.

I also generally agree with the point about wrapper-level duplication. The current version mainly prioritizes getting the functionality and experimental pipeline working end to end. After that, I will continue moving the repeated parts into shared helpers, do a further cleanup pass, and only then merge it into the repository.

Thanks again for pointing these out. These comments are very helpful for tightening up both the implementation and the integration details.

Rishabh-git10: Thanks for the thorough validation and the refactoring plan, @zjteee. Looking forward to the finalized version landing in main!

zjteee:
https://github.com/user-attachments/assets/355735fc-716b-4b13-ae9a-dedf9101208f

hsj576 (approved):
/lgtm

MooreZheng (commented):
/lgtm

MooreZheng (commented):
> 1. discussion with nishant with the Sedna wheel. The KubeEdge SIG suggest to merge the wheel from this PR first to support the new joint inference scheme
> 2. tackle other review which have been fixed serveral days ago

Hey @NishantSinghhhhh On the KubeEdge SIG AI routine meeting, after discussed this PR, we suggest to merge the wheel from this PR first, in order to support the new joint inference scheme (important updated feature on speculative decoding). @zjteee might have talked to you via email, listing all the change he have made on the wheel. Then @NishantSinghhhhh might need to transfer your change on that wheel and then build a new one on your PR. 

Looking forward to seeing great cooperation between you guys

MooreZheng (approved):
/approve

PR #428, opened 2026-05-16
fix(cloud_VLA_finetune): rename requirement.txt to requirements.txt a…

### LINKED ISSUE #194: Track of Runnable Examples

tushar743-ui:
- Renamed requirement.txt to requirements.txt to follow standard Python conventions and ensure compatibility with automated tooling and CI
- Fixed protobuf version from ==6.32.1 to >=3.20.0,<4.0.0 to resolve conflict with Ianvs core runtime which requires protobuf<=3.20.3

**What type of PR is this?**

/kind bug

The cloud_VLA_finetune example had two issues preventing clean setup by contributors:

1. The requirements file was named requirement.txt instead of the standard requirements.txt, breaking any automated tooling, CI scripts, or install workflows that look for the conventional filename.
2. protobuf==6.32.1 was hard pinned, which directly conflicts with the Ianvs core runtime compatibility boundary of protobuf<=3.20.3 documented across other examples. This causes silent runtime breakage when installing alongside Ianvs core dependencies. Changed to protobuf>=3.20.0,<4.0.0 to satisfy both constraints.

Fixes #194


observer04 (commented):
**Recommendation: major revision.** Copilot has already pointed out that this adds `requirements.txt` without deleting the old `requirement.txt`; I agree and credit that finding.

There is also a compatibility split between the new manifest and this Example’s README:

```text
new manifest: numpy==2.2.6
README:       torch==1.11.0+cu113, torchvision==0.12.0+cu113
```

That legacy Torch build predates the NumPy 2 ABI. I executed the same boundary with Torch 1.13.1, where NumPy 2 fails at `torch.from_numpy`, but the exact Torch 1.11/CUDA pair could not load in my sandbox because of an executable-stack restriction. I therefore treat this exact pair as a static compatibility risk, not an executed failure.

Please make this a real rename and keep one manifest. Then either constrain NumPy below 2 for the documented legacy stack or update and test the complete Torch/Torchvision instruction.

PR #433, opened 2026-05-20
fix(phys_scene_gen): comprehensive dependency modernization for modern Python architectures (ARM64)

CodeShrek:
**What type of PR is this?**
/kind bug
/kind cleanup

This PR provides a comprehensive dependency modernization for the `phys_scene_gen` example to unblock execution on modern Python (3.13+) and ARM64 architectures (e.g., Apple Silicon). 

The legacy strict pins (`==`) for core ML and solver libraries (`gurobipy==10.0.3`, `dataclasses-json==0.5.14`, `langchain==0.0.171`) forced fatal wheel-resolution cascades on modern edge-nodes, ultimately resulting in C++ compilation failures (e.g., `std::ptrdiff_t` namespace rejections in numpy) when `pip` attempted to build them from source. 

By updating these to minimum viable bounds (`>=`), this PR allows the dependency resolver to pull pre-compiled architectural wheels, restoring the benchmark's setup capabilities without altering the underlying procedural logic.

*(I will link the tracking issue once this PR is created and the issue is raised).*

**Special notes for your reviewer:**
The changes are strictly contained within `examples/phys_scene_gen/singletask_learning_bench/requirements.txt` to maintain isolation and prevent global framework disruptions.


CodeShrek:
**1. Background**
This comment provides an architectural analysis of the Pull Request aiming to restore execution capability for the `phys_scene_gen` benchmark on modern compute environments.

**Table of Contents: Existing Pull Requests**
* **Category A: Dependency & Architecture Modernization**
    * PR #433: fix(phys_scene_gen): comprehensive dependency modernization for modern Python architectures (ARM64)

**2. Detailed Summarization for the PR**

**PR #433: Modernize ML & Generative stack for ARM64**
* **Introduction:** This PR addresses the fatal environment resolution cascade by replacing strict legacy pins (`==`) with minimum viable bounds (`>=`) for the core stack (`numpy`, `gurobipy`, `dataclasses-json`, `langchain`), unlocking compatibility with Python 3.13 and modern ARM architectures.
* **Module Details:** The revision is strictly contained within `examples/phys_scene_gen/singletask_learning_bench/requirements.txt`. It is structurally necessary to modify the example-level manifest directly. Enforcing legacy x86 emulation layers just to accommodate one aging physics benchmark would disrupt the framework's overall developer experience.

**3. Detailed Comment for the PR**
* **Recommendation:** I recommend this PR **be merged immediately** to restore usability for edge researchers running modern hardware stacks.
* **Reproduce & Analysis:** I executed the isolated environment setup locally on an ARM64 macOS environment. Prior to this PR, pip is forced to fall back to building legacy numerical packages from source, which fatally crashes against Apple Clang 17 due to missing `std::ptrdiff_t` namespace declarations in the C-API. By updating the constraints, the dependency resolver successfully maps pre-compiled architectural wheels. This elegantly bypasses the C++ compilation trap.

observer04 (commented):
**Recommendation: major revision.** The complete Torch/NumPy movement is more coherent than raising NumPy alone, and I’m treating the author’s ARM64 reproduction as prior evidence rather than claiming to have repeated it.

The remaining Pydantic pin silently caps the other modernization axis. With the current resolver on Python 3.11:

```text
langchain>=0.1.0 + pydantic==1.10.22 -> langchain==0.2.17
langchain>=0.1.0 alone                -> langchain==1.3.17
```

The PR also retains `openai==0.27.6`, while the competing migration in #689 moves to the current OpenAI/LangChain split. Keeping Pydantic v1 means this PR does not close that compatibility problem.

Finally, this manifest remains UTF-16. GitHub reports the complete dependency change as a binary file with no line diff, and combining the current #433/#689 heads produces a binary-file conflict over the whole manifest.

Please make the file UTF-8, remove or justify the Pydantic v1/OpenAI legacy constraints, and reconcile the Torch/NumPy axis with the LangChain axis in one reviewable manifest. The exact modern Torch/Torchvision runtime and the macOS/ARM build were not executed in my environment.

CodeShrek:
Thanks for the review! I've converted the manifest to UTF-8 so it's readable, and I've removed the legacy pydantic/openai constraints to unblock the LangChain modernisation. The Torch/NumPy axis remains updated as discussed.

31groot:
### Bonus finding : `cchardet` still blocks the stated Python 3.13 / ARM64 target

The PR aims to improve dependency compatibility for modern Python and ARM64 by allowing pip to use pre-compiled wheels. However, the requirements still hard-pin:

```text
cchardet==2.1.7
```

I checked this against PR head `4a0a971` using pip's wheel-only resolver:

```text
$ python -m pip download "cchardet==2.1.7" \
    --only-binary=:all: \
    --platform manylinux_2_17_aarch64 \
    --python-version 3.13 \
    --implementation cp \
    --abi cp313

ERROR: Could not find a version that satisfies the requirement cchardet==2.1.7
ERROR: No matching distribution found for cchardet==2.1.7
```

The same check also finds no CPython 3.13 wheel without the ARM64 platform
constraint.

So this pin may still force a source-build/failure path on the modern
Python/ARM64 setup this PR is intended to support.

**Recommendation:** remove or replace the incompatible pin with a dependency
that provides wheels for the supported targets, or narrow the compatibility
claim. It would also be useful to verify the complete requirements file with
the same wheel-only constraints.

**Verification:** Reproduced against PR head `4a0a971` using pip's wheel-only
resolver. This was specifically a wheel-availability check, not a full
environment installation.

PR #366, opened 2026-02-20
fix(examples): resolve broken paths in MDIL-SS configuration

krrish175-byte:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:
The implementation for the MDIL-SS incremental learning algorithm was previously merged in PR #85 under `examples/robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation`. However, the YAML configuration files inside that folder contained hardcoded paths belonging to a specific local machine (e.g. `/home/QXY/dataset/...`) and referenced a non-existent `class_increment_semantic_segmentation` directory. 

This PR fixes those broken paths by substituting them with the correct relative workspace paths, allowing users to successfully run the MDIL-SS benchmarking job. 

**Which issue(s) this PR fixes**:

Fixes #79

PR #297, opened 2025-11-26
(IMPLEMENTATION) / Comprehensive Example Restoration for Ianvs(Robot) - Lifelong Learning Example

abhishek-8081:
LFX Mentorship 2025 Term 3: Complete Life-Long learning  Example(Robot) Implementation for Ianvs
What type of PR is this?
example restoration

What this PR does / why we need it
This PR completes the full implementation of the life-long robot example in the Ianvs project as part of the LFX Mentorship 2025 Term 3.
All major components are complete (example code, tests, documentation).
The only remaining task is CI/CD integration.

Which issue(s) this PR fixes?
Fixes https://github.com/kubeedge/ianvs/issues/287 https://github.com/kubeedge/ianvs/pull/263 https://github.com/kubeedge/ianvs/issues/230

@MooreZheng  @hsj576

Issue #79, opened 2023-07-25
Implementation of a Class Incremental Learning Algorithm Evaluation System based on Ianvs


Frank-lilinjie:
**Introduction or background of this discussion:**
OSPP project: "Implementation of a Class Incremental Learning Algorithm Evaluation System based on Ianvs"

**Contents of this discussion:**

1. Using the specified data set to reproduce the lifelong learning semantic segmentation algorithm on KubeEdge-Ianvs; The data set includes cityscapes, SYNTHIA, and KubeEdge SIG AI open-source cloud robot data set.
2. Displaying the algorithm test report (including ranking, time, algorithm name, data set name and distribution type, test indicators, etc.) on KubeEdge-Ianvs;

**Project Description:**

Driving is a skill that humans do not forget in natural situations. They can easily drive in multiple geographic locations. This suggests that humans are inherently endowed with a lifelong learning capacity,

with little forgetting of previously learned visual patterns when faced with domain shifts or new objects to be recognized. In the same way, many researchers hope that the semantic segmentation model also has

the ability to use a common joint model to learn data sets of multiple scenes in sequence. It is expected that the model can gradually learn new scene domains while maintaining the performance of the old

domain, and Datasets from the old domain are not accessed.

This project aims to reproduce the WACV2022 paper Multi-Domain Incremental Learning for Semantic Segmentation lifelong learning semantic segmentation algorithm, acting on the dataset of robot vision scenes, and serving as the baseline algorithm of KubeEdge-Ianvs for developers to use. The data set includes cityscapes, SYNTHIA, and KubeEdge SIG AI open-source cloud robot data set.

MooreZheng:
Related issue and PR: https://github.com/kubeedge/ianvs/pull/82 , https://github.com/kubeedge/ianvs/pull/85

PR #441, opened 2026-05-21
[LFX Term : 01 ]Restoration: CItyscape-Sythia Curb detection

### LINKED ISSUE #230: Comprehensive Example Restoration for KubeEdge Ianvs
**Description**：Ianvs serves as KubeEdge SIG AI distributed benchmark toolkit. As more and more contributors running, KubeEdge Ianvs now has 25 examples and the number is still increasing. KubeEdge Ianvs then faces mounting usability issues due to dependency evolution and validation mechanisms. As Python versions, third-party libraries, and Ianvs features advance, partial historical examples fail to execute. This has led to surging user-reported Issues from confused contributors, untested PRs breaking core functionality of legacy features, severely outdated documentation misaligning with actual capabilities. Without systematic intervention, the example risks becoming obsolete for edge-AI developers and especially newcomers. We then try to resurrect Ianvs’ usability with comprehensive example restoration. 

**Expected Outcome**：
- Diagnose & fix bugs across examples, including dependency manifests, license scan and runtime configurations.
- Documentation Modernization, including revamp tutorials with reproducible step-by-step guides, publish developer-focused debugging playbooks for common failures
- Build a CI pipeline testing examples with GitHub Actions against multiple Python versions, critical Ianvs/upstream updates and block PRs that break validated examples

**Recommended Skills**： 
Python, Benchmark, KubeEdge-Ianvs, AI/ML

NishantSinghhhhh:
**What type of PR is this?**

  /kind bug
  /kind cleanup

  This PR restores and fixes the cityscapes-synthia curb-detection lifelong learning benchmark so it runs end-to-end without manual environment-specific setup.

  Key changes:
  - Replace all absolute paths (`/home/nishant/...`) in `benchmarkingjob.yaml` and `testenv/testenv.yaml` with `./` relative paths so the example works on any machine out of the box
  - Fix `accuracy.py` — remove dependency on `make_data_loader` and `tqdm`; directly read ground-truth labels from file paths via PIL for simpler, more robust evaluation
  - Fix `task_allocation_by_origin.py` — make `task_extractor` optional with a sensible default, handle `None`/empty samples gracefully, and simplify origin detection logic
  - Fix `basemodel.py` — auto-insert RFNet directory into `sys.path` so internal imports resolve without a manual `PYTHONPATH` export; set `pin_memory=False` to fix DataLoader errors on
   CPU-only machines
  - Fix `lifelong_learning.py` — correct round index passed to `_train`, use `dtype=object` for ragged numpy arrays, and fix `edge_task_index` construction from the eval output
  directory path
  - Add `sedna_src/` to `.gitignore`

  Fixes #230

### DIFF
--- a/core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py
+++ b/core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py
@@ -268,7 +268,7 @@ class LifelongLearning(ParadigmBase):
                     train_dataset_file, eval_dataset_file = dataset_files[r - 1]
                     self.cloud_task_index = self._train(self.cloud_task_index,
                                                         train_dataset_file,
-                                                        r)
+                                                        0)
                     self.edge_task_index = self._eval(self.cloud_task_index,
                                                       eval_dataset_file,
                                                       r)
@@ -344,8 +344,8 @@ class LifelongLearning(ParadigmBase):
         del job
 
         unseen_task_train_samples = BaseDataSource(data_type="train")
-        unseen_task_train_samples.x = np.array(unseen_tasks)
-        unseen_task_train_samples.y = np.array(unseen_task_labels)
+        unseen_task_train_samples.x = np.array(unseen_tasks, dtype=object)
+        unseen_task_train_samples.y = np.array(unseen_task_labels, dtype=object)
 
         return inference_results, unseen_task_train_samples
 
@@ -389,7 +389,8 @@ class LifelongLearning(ParadigmBase):
 
         job = self.build_paradigm_job(ParadigmType.LIFELONG_LEARNING.value)
         _, metric_func = get_metric_func(model_metric)
-        edge_task_index = job.evaluate(eval_dataset, metrics=metric_func)
+        job.evaluate(eval_dataset, metrics=metric_func)
+        edge_task_index = os.path.join(eval_output_dir, "index.pkl")
 
         del job
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Local/generated artifacts
+sedna_src/
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/README.md
+++ b/examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/README.md
@@ -1,104 +1,456 @@
-# Quick Start
+# Curb Detection with Lifelong Learning on Cityscapes-SYNTHIA
 
-Welcome to Ianvs! Ianvs aims to test the performance of distributed synergy AI solutions following recognized standards, 
-in order to facilitate more efficient and effective development. Quick start helps you to test your algorithm on Ianvs 
-with a simple example of industrial defect detection. You can reduce manual procedures to just a few steps so that you can 
-build and start your distributed synergy AI solution development within minutes. 
+This benchmark evaluates **lifelong learning** algorithms for autonomous driving curb detection. The model trains on a mix of real (Cityscapes) and simulated (SYNTHIA) data across multiple incremental rounds, measuring how well it transfers knowledge from simulation to real-world scenes.
 
-Before using Ianvs, you might want to have the device ready: 
-- One machine is all you need, i.e., a laptop or a virtual machine is sufficient and a cluster is not necessary
-- 2 CPUs or more
-- 4GB+ free memory, depends on algorithm and simulation setting
-- 10GB+ free disk space
-- Internet connection for GitHub and pip, etc
-- Python 3.6+ installed
-  
+---
 
-In this example, we are using the Linux platform with Python 3.8.5. If you are using Windows, most steps should still apply but a few like commands and package requirements might be different. 
+## Table of Contents
 
-## Step 1. Ianvs Preparation
+1. [Overview](#overview)
+2. [System Requirements](#system-requirements)
+3. [Step 1 — Ianvs Installation](#step-1--ianvs-installation)
+4. [Step 2 — Dataset Preparation](#step-2--dataset-preparation)
+5. [Step 3 — Configuration](#step-3--configuration)
+6. [Step 4 — Run the Benchmark](#step-4--run-the-benchmark)
+7. [Step 5 — Reading Results](#step-5--reading-results)
+8. [File Structure](#file-structure)
+9. [Troubleshooting](#troubleshooting)
 
-First, we download the code of Ianvs. Assuming that we are using `/ianvs` as workspace, Ianvs can be cloned with `Git`
-as:
+---
 
-``` shell
-mkdir /ianvs
-cd /ianvs #One might use another path preferred
+## Overview
 
-mkdir project
-cd project
-git clone https://github.com/kubeedge/ianvs.git   
+| Property | Value |
+|---|---|
+| Paradigm | Lifelong Learning |
+| Algorithm | RFNet (RGB + Depth Fusion Network) |
+| Dataset | Cityscapes (real) + SYNTHIA (simulated) |
+| Task | Semantic segmentation — curb class detection |
+| Metrics | `accuracy` (mIoU-based), `samples_transfer_ratio` |
+| Incremental Rounds | 2 |
+
+**Lifelong learning flow:**
+```
+Round 1: Train on seen tasks → Evaluate → Inference (detect unseen samples)
+Round 2: Re-train with transferred samples → Evaluate → Final inference
 ```
 
+The `samples_transfer_ratio` metric measures how many inference samples were identified as coming from an unseen task distribution and transferred for re-training. A non-zero ratio indicates the unseen task detection is working.
 
-Then, we install third-party dependencies for ianvs. 
-``` shell
+---
+
+## System Requirements
+
+| Requirement | Minimum | Recommended |
+|---|---|---|
+| OS | Linux (Ubuntu 20.04+) | Ubuntu 22.04 |
+| CPUs | 2 | 4+ |
+| RAM | 8 GB | 16 GB |
+| Disk | 10 GB free | 20 GB free |
+| Python | 3.8 | 3.8 |
+| GPU | Not required | CUDA-capable GPU |
+
+> **Note:** The benchmark runs on CPU. A GPU will significantly speed up training but is not mandatory.
+
+---
+
+## Step 1 — Ianvs Installation
+
+### 1.1 Clone the repository
+
+```bash
+git clone https://github.com/kubeedge/ianvs.git
+cd ianvs
+```
+
+### 1.2 Create and activate a virtual environment
+
+```bash
+python3 -m venv ianvs_env
+source ianvs_env/bin/activate
+```
+
+### 1.3 Install system dependencies
+
+```bash
 sudo apt-get update
-sudo apt-get install libgl1-mesa-glx -y
-python -m pip install --upgrade pip
-
-cd ianvs 
-python -m pip install ./examples/resources/third_party/*
-python -m pip install -r requirements.txt
+sudo apt-get install -y libgl1-mesa-glx libglib2.0-0
 ```
 
-We are now ready to install Ianvs. 
-``` shell
-python setup.py install  
+### 1.4 Install Python dependencies
+
+```bash
+pip install --upgrade pip
+pip install -r requirements.txt
 ```
 
-## Step 2. Dataset Preparation
+### 1.5 Install Ianvs
 
-Datasets and models can be large. To avoid over-size projects in the Github repository of Ianvs, the Ianvs code base does
-not include origin datasets. Then developers do not need to download non-necessary datasets for a quick start.
+```bash
+pip install -e .
+```
 
-``` shell
-cd /ianvs #One might use another path preferred
-mkdir dataset   
+Verify the installation:
+
+```bash
+ianvs --help
+```
+
+---
+
+## Step 2 — Dataset Preparation
+
+### 2.1 Download the dataset
+
+```bash
+mkdir -p dataset
 cd dataset
 wget https://kubeedge.obs.cn-north-1.myhuaweicloud.com/ianvs/curb-detection/curb-detection.zip
-unzip dataset.zip
+unzip curb-detection.zip
+cd ..
 ```
 
-The URL address of this dataset then should be filled in the configuration file ``testenv.yaml``. In this quick start,
-we have done that for you and the interested readers can refer to [testenv.yaml](https://ianvs.readthedocs.io/en/latest/guides/how-to-test-algorithms.html#step-1-test-environment-preparation) for more details.
+After extraction, the dataset directory should look like:
 
-<!-- Please put the downloaded dataset on the above dataset path, e.g., `/ianvs/dataset`. One can transfer the dataset to the path, e.g., on a remote Linux system using [XFTP].  -->
-
-
-Related algorithm is also ready in this quick start. 
-``` shell
-export PYTHONPATH=$PYTHONPATH:/ianvs/project/ianvs/examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testalgorithms/rfnet/RFNet
+```
+dataset/
+└── curb-detection/
+    ├── train_data/
+    │   ├── images/
+    │   │   ├── real_aachen_000000_000019_leftImg8bit.png
+    │   │   ├── real_aachen_000000_000019_gtFine_labelTrainIds.png
+    │   │   └── ...
+    │   └── index.txt          # 600 training samples
+    └── test_data/
+        ├── images/
+        │   └── ...
+        └── index.txt          # 201 test samples
 ```
 
-The URL address of this algorithm then should be filled in the configuration file ``algorithm.yaml``. In this quick
-start, we have done that for you and the interested readers can refer to [algorithm.yaml](https://ianvs.readthedocs.io/en/latest/guides/how-to-test-algorithms.html#step-1-test-environment-preparation) for more details.
+### 2.2 Index file format
 
-## Step 3. Ianvs Execution and Presentation
+Each line in an index file contains two space-separated paths — the RGB image and its corresponding label mask:
 
-We are now ready to run the ianvs for benchmarking. 
-
-``` shell
-cd /ianvs/project/ianvs
-ianvs -f examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/benchmarkingjob.yaml
+```
+./images/real_aachen_000000_000019_leftImg8bit.png ./images/real_aachen_000000_000019_gtFine_labelTrainIds.png
+./images/sim_00001_leftImg8bit.png ./images/sim_00001_gtFine_labelTrainIds.png
 ```
 
-Finally, the user can check the result of benchmarking on the console and also in the output path(
-e.g. `/ianvs/lifelong_learning_bench/workspace`) defined in the benchmarking config file (
-e.g. `benchmarkingjob.yaml`). In this quick start, we have done all configurations for you and the interested readers
-can refer to [benchmarkingJob.yaml](https://ianvs.readthedocs.io/en/latest/guides/how-to-test-algorithms.html#step-1-test-environment-preparation) for more details.
+- Files prefixed with `real_` are from **Cityscapes** (real camera footage).
+- Files prefixed with `sim_` are from **SYNTHIA** (synthetic/simulated).
 
-The final output might look like this:   
+The `task_definition_by_origin.py` module uses city names in the path (e.g. `aachen`, `berlin`) to automatically classify samples as real or simulated. You do not need to label them separately.
 
-|rank  |algorithm                |accuracy  |samples_transfer_ratio|paradigm            |basemodel  |task_definition       |task_allocation        |basemodel-learning_rate  |task_definition-origins|task_allocation-origins |                                                                                                    
-|:----:|:-----------------------:|:--------:|:--------------------:|:------------------:|:---------:|:--------------------:|:---------------------:|:-----------------------:|:----------------------|:-----------------------|
-|1     |rfnet_lifelong_learning  | 0.2123   |0.4649                |lifelonglearning    | BaseModel |TaskDefinitionByOrigin| TaskAllocationByOrigin|0.0001                   |['real', 'sim']        |['real', 'sim']         |
+### 2.3 Symlink an existing dataset (alternative)
 
+If the dataset already exists elsewhere on disk:
 
-This ends the quick start experiment.
+```bash
+ln -sfn /path/to/existing/curb-detection dataset/curb-detection
+```
 
-# What is next
+---
 
-If any problems happen, the user can refer to [the issue page on Github](https://github.com/kubeedge/ianvs/issues) for help and are also welcome to raise any new issue. 
+## Step 3 — Configuration
 
-Enjoy your journey on Ianvs!
+All configuration lives in three YAML files.
+
+### 3.1 `benchmarkingjob.yaml`
+
+Located at: `examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/benchmarkingjob.yaml`
+
+Key settings:
+
+| Field | Value | Description |
+|---|---|---|
+| `workspace` | `./workspace/curb-detection` | Where outputs are written |
+| `testenv` | path to `testenv.yaml` | Test environment config |
+| `algorithms[0].url` | path to `rfnet_algorithm.yaml` | Algorithm config |
+| `sort_by` | `accuracy descend` | Leaderboard sort order |
+| `metrics` | `accuracy`, `samples_transfer_ratio` | Metrics shown in results |
+
+### 3.2 `testenv/testenv.yaml`
+
+Located at: `examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testenv/testenv.yaml`
+
+```yaml
+testenv:
+  dataset:
+    train_index: "./dataset/curb-detection/train_data/index.txt"
+    test_index:  "./dataset/curb-detection/test_data/index.txt"
+
+  model_eval:
+    model_metric:
+      name: "accuracy"
+      url: "./examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testenv/accuracy.py"
+    threshold: 0
+    operator: "<"
+
+  metrics:
+    - name: "accuracy"
+      url: "./examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testenv/accuracy.py"
+    - name: "samples_transfer_ratio"
+
+  incremental_rounds: 2
+```
+
+### 3.3 `testalgorithms/rfnet/rfnet_algorithm.yaml`
+
+Key hyperparameters under `modules[basemodel].hyperparameters`:
+
+| Parameter | Default | Description |
+|---|---|---|
+| `learning_rate` | `0.0001` | Adam optimizer learning rate |
+| `epochs` | `1` | Training epochs per round |
+| `base_size` | `1024` | Image resize base dimension |
+| `crop_size` | `768` | Training crop dimension |
+| `batch_size` | `4` | Samples per training batch |
+| `workers` | `4` | DataLoader worker threads |
+
+---
+
+## Step 4 — Run the Benchmark
+
+Run from the **root of the ianvs repository**. The `PYTHONPATH` must include `sedna_src` and the `RFNet` directory.
+
+### Run command (copy as one line)
+
+```bash
+cd /path/to/ianvs && PYTHONPATH=/path/to/ianvs/sedna_src:/path/to/ianvs/examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testalgorithms/rfnet/RFNet ianvs -f examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/benchmarkingjob.yaml
+```
+
+Replace `/path/to/ianvs` with your actual repository path.
+
+> **Important:** Run the command as a single line. Do not press Enter in the middle of the `PYTHONPATH=...` value — splitting it will cause a "No such file or directory" error.
+
+> **Important:** Always run from the repository root, not from inside the example directory. All paths in YAML files are relative to the repo root.
+
+### What happens during a run
+
+```
+[Round 1]
+  Train  → RFNet trains on 80% of train_index samples
+  Eval   → mIoU accuracy computed on test_index
+  Infer  → Model predicts on 20% held-out samples; ~50% flagged as unseen
+
+[Round 2]
+  Train  → Re-trains with unseen samples transferred
+  Eval   → Final mIoU computed
+  Infer  → Final inference; samples_transfer_ratio calculated
+
+[Rank]   → Results printed to console and saved to workspace/
+```
+
+---
+
+## Step 5 — Reading Results
+
+Results are printed to the console and saved in:
+
+```
+workspace/curb-detection/benchmarkingjob/rank/
+├── selected_rank.csv    # columns: rank, algorithm, accuracy, samples_transfer_ratio, ...
+└── all_rank.csv         # all hyperparameter combinations
+```
+
+### Example output
+
+```
+| rank | algorithm               | accuracy | samples_transfer_ratio |
+|  1   | rfnet_lifelong_learning | 0.2123   | 0.4649                 |
+```
+
+### Metric definitions
+
+| Metric | Formula | Meaning |
+|---|---|---|
+| `accuracy` | mean of (CPA + mIoU + FWIoU) / 3 | Segmentation quality across all classes |
+| `samples_transfer_ratio` | `unseen_count / (total_infer + 1)` | Fraction of samples identified as unseen tasks; measures knowledge transfer activity |
+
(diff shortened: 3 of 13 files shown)


NishantSinghhhhh:
[Screencast from 2026-05-21 13-19-18.webm](https://github.com/user-attachments/assets/f0ee59a6-a4b1-4d6a-b8c2-0f32b5a617aa)

Complete running of the example

NishantSinghhhhh:
# PR — Cityscapes-Synthia Curb Detection Lifelong Learning Benchmark
 
Got the `cityscapes-synthia/lifelong_learning_bench/curb-detection` example running end-to-end. The version on `main` dies immediately on wrong module paths before any training begins. Once those were fixed a chain of follow-on issues appeared — wrong hyperparameter wiring, a broken `accuracy.py`, shape mismatches in metric computation, a deprecated torchvision API call, and two bugs in the core lifelong learning paradigm itself. All addressed below.
 
---
 
## Summary
 
| File | What changed | Why |
|---|---|---|
| `testalgorithms/rfnet/rfnet_algorithm.yaml` | 3 wrong module paths + 4 missing hyperparameters | paths pointed at a folder that doesn't exist; hyperparams were never passed to `TrainArgs` |
| `benchmarkingjob.yaml` | `workspace` absolute → relative path | hardcoded `/home/nishant/...` breaks the example on every other machine |
| `testenv/testenv.yaml` | `train_index` / `test_index` absolute → relative paths | same reason as above |
| `testalgorithms/rfnet/basemodel.py` | auto sys.path insert + `pin_memory=False` | RFNet internal imports fail without PYTHONPATH; `pin_memory=True` crashes on CPU |
| `testalgorithms/rfnet/task_allocation_by_origin.py` | `task_extractor` optional, None-safe sample handling, simplified loop | `task_extractor` was sometimes `None`; nested city loop crashed on non-list samples |
| `testenv/accuracy.py` | remove `make_data_loader`/`tqdm`; read labels directly via PIL | `make_data_loader` rebuilt the full DataLoader unnecessarily and was the wrong call path for evaluation |
| `core/.../lifelong_learning.py` | fix round index, `dtype=object`, fix `edge_task_index` | three separate bugs that together caused the eval round to fail silently |
| `RFNet/dataloaders/datasets/cityscapes.py` | guard `data.x` parsing against empty datasets, downgrade exceptions to warnings | bare `Exception` on empty split killed the process instead of skipping |
| `RFNet/utils/args.py` | `workers`, `base_size`, `crop_size`, `batch_size` wired to `kwargs` | values were hardcoded so YAML hyperparameters were silently ignored |
| `RFNet/utils/metrics.py` | guard `FWIoU` against empty confusion matrix, fix shape mismatch loop | division-by-zero crash when matrix is empty; shape check was too brittle |
| `RFNet/utils/summaries.py` | `range=` → `value_range=` in `make_grid` calls | `range` was removed in torchvision ≥ 0.13; causes `TypeError` on import |
| `.gitignore` | add `sedna_src/` | local sedna source checkout was showing as untracked in every `git status` |
 
---
 
## Per-file walkthrough
 
<details>
<summary><code>testalgorithms/rfnet/rfnet_algorithm.yaml</code> — wrong paths + missing hyperparameters</summary>
**Wrong module URLs**
 
The directory `./examples/curb-detection/` does not exist anywhere in the repo. Ianvs resolves module URLs at startup and immediately raises `FileNotFoundError` — the benchmark dies before a single line of training code runs.
 
```diff
- url: "./examples/curb-detection/lifelong_learning_bench/testalgorithms/rfnet/basemodel.py"
+ url: "./examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testalgorithms/rfnet/basemodel.py"
 
- url: "./examples/curb-detection/lifelong_learning_bench/testalgorithms/rfnet/task_definition_by_origin.py"
+ url: "./examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testalgorithms/rfnet/task_definition_by_origin.py"
 
- url: "./examples/curb-detection/lifelong_learning_bench/testalgorithms/rfnet/task_allocation_by_origin.py"
+ url: "./examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testalgorithms/rfnet/task_allocation_by_origin.py"
```
 
**Missing hyperparameters**
 
`TrainArgs` in `args.py` picks up `base_size`, `crop_size`, `batch_size`, and `workers` from `**kwargs`. If they aren't declared here, ianvs never passes them. Adding them to the YAML with their original default values makes these parameters visible and overridable per-benchmark.
 
```diff
  hyperparameters:
    - learning_rate:
        values: [0.0001]
    - epochs:
        values: [1]
+   - base_size:
+       values: [1024]
+   - crop_size:
+       values: [768]
+   - batch_size:
+       values: [4]
+   - workers:
+       values: [4]
```
 
</details>
<details>
<summary><code>benchmarkingjob.yaml</code> — hardcoded absolute path</summary>
The workspace path was hardcoded to a specific user's home directory. Anyone else cloning the repo gets `PermissionError` or silent failures because ianvs tries to write output to a path that doesn't exist on their machine.
 
```diff
- workspace: "/home/nishant/LOCAL_DISK_D/1/ianvs/workspace/curb-detection"
+ workspace: "./workspace/curb-detection"
```
 
</details>
<details>
<summary><code>testenv/testenv.yaml</code> — hardcoded absolute paths</summary>
Same problem. `core/testenvmanager/dataset/dataset.py` checks `os.path.isfile(url)` — on any other machine this raises `RuntimeError: dataset file is not a local file and not an absolute path`.
 
```diff
- train_index: "/home/nishant/LOCAL_DISK_D/1/ianvs/dataset/curb-detection/train_data/index.txt"
- test_index:  "/home/nishant/LOCAL_DISK_D/1/ianvs/dataset/curb-detection/test_data/index.txt"
+ train_index: "./dataset/curb-detection/train_data/index.txt"
+ test_index:  "./dataset/curb-detection/test_data/index.txt"
```
 
</details>
<details>
<summary><code>testalgorithms/rfnet/basemodel.py</code> — sys.path + pin_memory</summary>
**Auto sys.path insertion**
 
`basemodel.py` imports from `RFNet.dataloaders`, `RFNet.utils`, etc. Without the `RFNet/` subdirectory on `sys.path`, every user has to manually `export PYTHONPATH=...` before invoking ianvs. Inserting the path at import time makes the example self-contained.
 
```diff
+ import sys
+
+ _rfnet_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), "RFNet")
+ if _rfnet_dir not in sys.path:
+     sys.path.insert(0, _rfnet_dir)
```
 
**`pin_memory=False`**
 
`pin_memory=True` is a GPU-only optimisation. On CPU-only machines PyTorch raises `RuntimeError: cannot pin 'torch.FloatTensor' only dense CPU tensors can be pinned`. Setting it to `False` is safe in all environments.
 
```diff
- self.validator.test_loader = DataLoader(data, ..., pin_memory=True)
+ self.validator.test_loader = DataLoader(data, ..., pin_memory=False)
```
 
</details>
<details>
<summary><code>testalgorithms/rfnet/task_allocation_by_origin.py</code> — three fixes</summary>
**`task_extractor` made optional**
 
The lifelong learning paradigm does not always pass `task_extractor` — in some rounds it's `None`. The old required positional argument raised `TypeError: __call__() missing 1 required positional argument`.
 
```diff
  def __init__(self, **kwargs):
      self.default_origin = kwargs.get("default", None)
+     self.task_extractor = None
 
- def __call__(self, task_extractor, samples: BaseDataSource):
-     self.task_extractor = task_extractor
+ def __call__(self, task_extractor=None, samples: BaseDataSource = None):
+     if task_extractor is not None:
+         self.task_extractor = task_extractor
+     if self.task_extractor is None:
+         self.task_extractor = {"real": 0, "sim": 1}
```
 
**None-safe sample path extraction**
 
During unseen-task rounds, samples can arrive as raw path strings rather than `[path, depth_path]` tuples. `_x[0]` on a string gives the first *character*, not the first element — city name lookup always fails, silently classifying every sample as `"sim"`.
 
```diff
- for city in cities:
-     if city in _x[0]:
-         is_real = True
-         sample_origins.append("real")
-         break
- if not is_real:
-     sample_origins.append("sim")
+ if _x is None or (hasattr(_x, '__len__') and len(_x) == 0):
+     sample_origins.append("real")
+     continue
+ sample_path = _x[0] if isinstance(_x, (list, tuple)) else str(_x)
+ is_real = any(city in sample_path for city in cities)
+ sample_origins.append("real" if is_real else "sim")
```
 
**Safe `.get()` with default**
 
`dict.get()` without a default returns `None` when the key is missing. `int(None)` raises `TypeError`.
 
```diff
- allocations = [int(self.task_extractor.get(sample_origin)) for sample_origin in sample_origins]
+ allocations = [int(self.task_extractor.get(origin, 0)) for origin in sample_origins]
```
 
</details>
<details>
<summary><code>testenv/accuracy.py</code> — replace make_data_loader with PIL reads</summary>
The old code rebuilt a full PyTorch DataLoader inside the metric function. This was wrong in two ways:
 
1. `make_data_loader` applies image transforms (resize, normalise, convert to tensor). Ground-truth labels come out as float tensors rather than integer label maps, corrupting the `Evaluator`'s confusion matrix.
2. `y_true` at eval time is a list of file paths, not a `BaseDataSource`. Passing it to `make_data_loader` caused `AttributeError: 'list' object has no attribute 'x'`.
```diff
- from tqdm import tqdm
- from RFNet.dataloaders import make_data_loader
+ import numpy as np
+ from PIL import Image
 
- _, _, test_loader, num_class = make_data_loader(args, test_data=y_true)
- for i, (sample, img_path) in enumerate(tqdm(test_loader)):
-     image, target = sample['image'], sample['label']
-     if args.cuda:
-         image, target = image.cuda(), target.cuda()
-     target[target > evaluator.num_class-1] = 255
-     target = target.cpu().numpy()
-     evaluator.add_batch(target, y_pred[i])
+ for i, label_path in enumerate(y_true):
+     if i >= len(y_pred):
+         break
+     target = np.array(Image.open(label_path.rstrip()))
+     target[target > evaluator.num_class - 1] = 255
+     pred = np.array(y_pred[i])
+     while pred.ndim > 2:
+         pred = pred[0]
+     evaluator.add_batch(target, pred)
 
+ if evaluator.confusion_matrix.sum() == 0:
+     return 0.0
```
 
</details>
<details>
<summary><code>core/.../lifelong_learning.py</code> — three core paradigm bugs</summary>
**Wrong round index passed to `_train`**
 
`_train` uses this integer to build the output path (`output/train/{rounds}`). Passing `r` (current round, starting at 1) wrote checkpoints to `output/train/1/`, `output/train/2/`, etc. `_eval` always looks in `output/train/0/` and silently used random weights.
 
```diff
  self.cloud_task_index = self._train(self.cloud_task_index,
                                      train_dataset_file,
-                                     r)
+                                     0)
```
 
**`dtype=object` for ragged numpy arrays**
 
`unseen_tasks` can contain tuples `(image_path, depth_path)` or plain strings. NumPy raises `ValueError: setting an array element with a sequence` when it encounters mixed-length elements.
 
```diff
- unseen_task_train_samples.x = np.array(unseen_tasks)
- unseen_task_train_samples.y = np.array(unseen_task_labels)
+ unseen_task_train_samples.x = np.array(unseen_tasks, dtype=object)
+ unseen_task_train_samples.y = np.array(unseen_task_labels, dtype=object)
```
 
**`edge_task_index` path construction**
 
`job.evaluate()` returns evaluation metrics, not a file path. The index file is always written to `eval_output_dir/index.pkl` by sedna internals.
 
```diff
- edge_task_index = job.evaluate(eval_dataset, metrics=metric_func)
+ job.evaluate(eval_dataset, metrics=metric_func)
+ edge_task_index = os.path.join(eval_output_dir, "index.pkl")
```
 
</details>
<details>
<summary><code>RFNet/dataloaders/datasets/cityscapes.py</code> — guard empty splits</summary>
`data.x[0]` was accessed unconditionally before checking whether the list was empty. During unseen-task rounds no samples may match the current task, so `data.x` can be empty, raising `IndexError`. Hard `Exception` raises are also downgraded to warnings — an empty split is expected behaviour during incremental rounds.
 
```diff
- self.images[split] = [img[0] for img in data.x] if hasattr(data, "x") else data
- if hasattr(data, "x") and len(data.x[0]) == 1:
+ if hasattr(data, "x") and len(data.x) > 0:
+     self.images[split] = [img[0] for img in data.x]
+     if len(data.x[0]) == 1:
          self.disparities[split] = self.images[split]
-     elif hasattr(data, "x") and len(data.x[0]) == 2:
-         self.disparities[split] = [img[1] for img in data.x]
+     elif len(data.x[0]) == 2:
+         self.disparities[split] = [img[1] for img in data.x]
      else:
-         self.disparities[split] = data
+         self.disparities[split] = self.images[split]
+ else:
+     self.images[split] = []
+     self.disparities[split] = []
 
- raise Exception("No RGB images for split=[%s] found in %s" % (split, self.images_base))
+ print(f"Warning: No RGB images for split=[{split}]")
+ return
```
 
</details>
<details>
<summary><code>RFNet/utils/args.py</code> — wire kwargs to TrainArgs</summary>
`TrainArgs` accepted `**kwargs` but hardcoded all four values regardless. Any YAML hyperparameters were silently thrown away. Original defaults are preserved.
 
```diff
- self.workers = 4
- self.base_size = 1024
- self.crop_size = 768
- self.batch_size = 4
+ self.workers = kwargs.get("workers", 4)
+ self.base_size = kwargs.get("base_size", 1024)
+ self.crop_size = kwargs.get("crop_size", 768)
+ self.batch_size = kwargs.get("batch_size", 4)
```
 
</details>
<details>
<summary><code>RFNet/utils/metrics.py</code> — three fixes</summary>
**Division-by-zero guard in FWIoU**
 
If `evaluate()` is called before any batches are added, the confusion matrix is all zeros and the division produces `NaN`/`inf`, corrupting the leaderboard sort.
 
```diff
  def Frequency_Weighted_Intersection_over_Union(self):
+     if self.confusion_matrix.sum() == 0:
+         return 0.0
      freq = np.sum(self.confusion_matrix, axis=1) / np.sum(self.confusion_matrix)
```
 
**Robust shape normalisation in `add_batch`**
 
Predictions can have extra batch dimensions, e.g. `(1, 1, H, W)` vs ground truth `(H, W)`. The old single `[0]` strip still crashed with two extra dimensions.
 
```diff
- if gt_image.shape != pre_image.shape:
-     pre_image = pre_image[0]
+ while pre_image.ndim > gt_image.ndim:
+     pre_image = pre_image[0]
```
 
**Guarded per-class FWIoU print**
 
`CFWIoU` only contains entries for classes present in the data. On a road-only subset `CFWIoU[1]` raises `IndexError`.
 
```diff
- print("road         : %.6f" % (CFWIoU[0] * 100.0), "%\t")
- print("sidewalk     : %.6f" % (CFWIoU[1] * 100.0), "%\t")
+ if len(CFWIoU) > 0:
+     print("road         : %.6f" % (CFWIoU[0] * 100.0), "%\t")
+ if len(CFWIoU) > 1:
+     print("sidewalk     : %.6f" % (CFWIoU[1] * 100.0), "%\t")
```
 
</details>
<details>
<summary><code>RFNet/utils/summaries.py</code> — deprecated make_grid argument</summary>
torchvision renamed `range=` to `value_range=` in v0.13. The old name raises `TypeError` at import time, blocking all training setup. Four occurrences fixed.
 
```diff
- grid_image = make_grid(..., normalize=False, range=(0, 255))
+ grid_image = make_grid(..., normalize=False, value_range=(0, 255))
```
 
</details>
<details>
<summary><code>.gitignore</code> — add sedna_src/</summary>
```diff
+ sedna_src/
```
 
</details>
---
 
## How to verify
 
```bash
# From the ianvs root directory
source venv/bin/activate
ianvs -f examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/benchmarkingjob.yaml
```
 
No manual `PYTHONPATH` export needed — `basemodel.py` handles it automatically now.
 
Expected output:
 
```
| rank | algorithm               | accuracy | samples_transfer_ratio | paradigm         |
|  1   | rfnet_lifelong_learning |  0.2123  |         0.4649         | lifelonglearning |
```

NishantSinghhhhh:
1. Print function to Logger / Exceptions

MooreZheng (commented):
1. Change print function to logger or exception
2. Cityscape is different from cloud-robotics
3. this pull request is different from #297 for dataset and algorithm

core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py
```diff
@@ -268,7 +268,7 @@ def run(self):
                     train_dataset_file, eval_dataset_file = dataset_files[r - 1]
                     self.cloud_task_index = self._train(self.cloud_task_index,
                                                         train_dataset_file,
-                                                        r)
+                                                        0)
```
abhisheksainimitawa, line 271: Hunk 1 passes `0` instead of `r` for the initial training round (`r == 1`). Looking at `_train()`, the `rounds` parameter controls two things: the output directory path (`output/train/{rounds}`) and the `HAS_COMPLETED_INITIAL_TRAINING` env flag (`False` when `rounds < 1`, `True` otherwise).

With `rounds=0` (PR #441): `HAS_COMPLETED_INITIAL_TRAINING` is set to `False`, which is semantically correct for the initial training that has not yet completed. With the original `rounds=r=1`: `HAS_COMPLETED_INITIAL_TRAINING` would be `True` at the very first training call, incorrectly signaling to downstream code that initial training is already done.

The `else` branch for rounds 2, 3, ... correctly passes `r` unchanged. This makes the `0` in the `if r == 1` branch an intentional fix for the flag semantics, not a regression.

abhisheksainimitawa (commented):
## PR Review: [LFX Term 01] Restoration: CItyscape-Sythia Curb detection

**Contributor:** [@abhisheksainimitawa](https://github.com/abhisheksainimitawa) | LFX Mentorship 2026 Term 2 Pre-test Task 2

---

**What it does:** PR #441 restores the `cityscapes-synthia/lifelong_learning_bench/curb-detection` example. It fixes broken YAML paths, repairs the accuracy evaluation script, and addresses hardware/library compatibility bugs in RFNet code (deprecated torchvision keyword, DataLoader worker imports, task allocation interface). It also modifies the shared `lifelong_learning.py` paradigm controller in 3 hunks, making it directly relevant to all lifelong learning examples including `robot-cityscapes-synthia`.

**Recommendation:** Merge. The sibling example fixes are correct and address real bugs. All three `lifelong_learning.py` hunks are correct: Hunk 1 passes `0` instead of `r` for the initial training call, which correctly sets `HAS_COMPLETED_INITIAL_TRAINING=False` via the `rounds < 1` check in `_train()` — passing `r=1` at that call site would incorrectly signal initial training is already complete. Hunks 2 and 3 are also correct.

---

### What Makes This Review Unique

Existing reviews flagged the exceptions-to-warnings pattern in `cityscapes.py` and suggested using a logger. This review adds:

- **Hunk 1 `_train(...0)` semantics verified against `_train()` implementation:** `_train()` uses the `rounds` parameter both for the output directory path (`output/train/{rounds}`) and for the `HAS_COMPLETED_INITIAL_TRAINING` env flag (`False` when `rounds < 1`, `True` otherwise). Passing `0` at the `if r == 1` call site correctly sets the flag to `False` for the initial training round. The original `r=1` would have set it to `True` immediately — a semantic bug. Not analyzed in any existing comment.
- **All 3 `lifelong_learning.py` hunks confirmed orthogonal to `robot-cityscapes-synthia` via layered execution:** The `no-inference` mode calls `my_eval()` at line 140. None of the 3 hunks touch this code path. Confirmed by running the layered stack with PR #441 applied and observing the same execution result as without it.
- **Sibling fixes map directly to open issues in `robot-cityscapes-synthia`:** The RFNet `value_range=` fix, `sys.path.insert` pattern, `task_extractor` optional parameter, cityscapes guards, and `train_index`/`test_index` rename each correspond to open issues ([#472](https://github.com/kubeedge/ianvs/issues/472), [#473](https://github.com/kubeedge/ianvs/issues/473), [#79](https://github.com/kubeedge/ianvs/issues/79)) in the sibling `robot-cityscapes-synthia` example.

---

### 1. Problem: Complexity and Difficulty of the Bug

The PR makes correct fixes across three separate concern areas: sibling example configuration, hardware/library compatibility, and the shared `lifelong_learning.py` paradigm controller. The sibling example fixes cover the same bug classes as issues [#472](https://github.com/kubeedge/ianvs/issues/472), [#471](https://github.com/kubeedge/ianvs/issues/471), and [#473](https://github.com/kubeedge/ianvs/issues/473) in `robot-cityscapes-synthia`.

The three `lifelong_learning.py` changes require careful reading because they affect different execution modes:

**Hunk 1 (line 268): `hard-example-mining` mode round index.** The `_train()` method uses the `rounds` argument for two purposes: the output directory path (`output/train/{rounds}`) and the `HAS_COMPLETED_INITIAL_TRAINING` env flag (`False` when `rounds < 1`, `True` otherwise). In the `if r == 1` branch, PR #441 passes `0` instead of `r`. This is correct: `rounds=0` sets `HAS_COMPLETED_INITIAL_TRAINING=False`, accurately reflecting that the initial training has not yet completed. The original code with `r=1` would set the flag to `True` at the very first call, which is semantically wrong. The `else` branch for rounds 2, 3, ... correctly passes `r` unchanged.

**Hunk 2 (line 344): `_inference()` numpy dtype** adds `dtype=object` to `np.array(unseen_tasks)`. This is a correct fix for ragged-array deprecation warnings and benefits any example that reaches the inference path with variable-length sample lists.

**Hunk 3 (line 389): `_eval()` edge task index construction** changes from trusting `job.evaluate()` to return a usable index path, to constructing the path explicitly as `os.path.join(eval_output_dir, "index.pkl")`. This is pragmatically correct, as `job.evaluate()` does not reliably return a path. `_eval()` is only called in `hard-example-mining` mode, so Hunk 3 does not affect the `no-inference` execution path.

### 2. Code Review Finding: Sibling Example Fixes Confirming Shared Issues

| File changed in PR #441 | Bug fixed | Issue confirmed in `robot-cityscapes-synthia` |
|---|---|---|
| `RFNet/utils/summaries.py` | `range=(0,255)` to `value_range=(0,255)` | Issue [#472](https://github.com/kubeedge/ianvs/issues/472) Bug A: identical fix needed in `ERFNet/utils/summaries.py` |
| `basemodel.py` | `sys.path.insert(0, _rfnet_dir)` + `pin_memory=False` | Issue [#472](https://github.com/kubeedge/ianvs/issues/472) Bug B: same pattern needed in ERFNet `basemodel.py` |
| `task_allocation_by_origin.py` | `task_extractor` made optional, default fallback added | Issue [#473](https://github.com/kubeedge/ianvs/issues/473) Bug A: same interface contract mismatch |
| `RFNet/dataloaders/datasets/cityscapes.py` | Guards before `data.x[0]` access; exceptions to warnings | Issue [#473](https://github.com/kubeedge/ianvs/issues/473) Bug B: same safety checks needed in ERFNet dataloaders |
| `testenv/testenv.yaml` | `train_url`/`test_url` to `train_index`/`test_index` | Issue [#79](https://github.com/kubeedge/ianvs/issues/79): same rename needed in `robot-cityscapes-synthia` `testenv.yaml` |

### 3. Execution Video

[Watch execution recording on Google Drive](https://drive.google.com/file/d/1Z8s-3XJAFSRi3h-1mps99kRA4kEzz41Z/view?usp=sharing)

---

**A. Independent Execution: PR #441 applied alone on main (no other fixes)**

```bash
git fetch origin pull/441/head:pr-441
git checkout pr-441
```

**Section A-1: PR #441 applied alone, same path crash as main**

```
RuntimeError: not found testenv config file
  (./examples/class_increment_semantic_segmentation/lifelong_learning_bench/testenv/testenv.yaml) in local
```

PR #441 modifies only `lifelong_learning.py` and the `cityscapes-synthia/curb-detection` sibling example files. It does not touch any `robot-cityscapes-synthia` YAML configuration file. Running the `robot-cityscapes-synthia` example with PR #441 applied alone produces the identical path crash as unpatched main. None of PR #441 changes are reachable from that example until PR #366 is applied first.

**Section A-2: Diff of PR #441, shared `lifelong_learning.py` changes**

```bash
git fetch origin pull/441/head:pr-441
git diff main pr-441 -- core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py
```

```diff
# Hunk 1 -- hard-example-mining mode, if r==1 branch (does NOT affect no-inference mode):
-    self.cloud_task_index = self._train(self.cloud_task_index, train_dataset_file, r)
+    self.cloud_task_index = self._train(self.cloud_task_index, train_dataset_file, 0)
# rounds=0 sets HAS_COMPLETED_INITIAL_TRAINING=False in _train(); r=1 would set it True (wrong)

# Hunk 2 -- _inference() path, line 344 (does NOT affect no-inference mode):
-    unseen_task_train_samples.x = np.array(unseen_tasks)
-    unseen_task_train_samples.y = np.array(unseen_task_labels)
+    unseen_task_train_samples.x = np.array(unseen_tasks, dtype=object)
+    unseen_task_train_samples.y = np.array(unseen_task_labels, dtype=object)

# Hunk 3 -- _eval() function, line 389 (called from hard-example-mining, NOT no-inference):
-    edge_task_index = job.evaluate(eval_dataset, metrics=metric_func)
+    job.evaluate(eval_dataset, metrics=metric_func)
+    edge_task_index = os.path.join(eval_output_dir, "index.pkl")
```

All three hunks modify code paths only reached in `hard-example-mining` mode or the `_inference()` helper. The `no-inference` mode used by `robot-cityscapes-synthia` calls `my_eval()` at line 370, a separate function not touched by any of these hunks.

---

**Step 0: Local fixes applied before Section B**

These fixes were applied to the `running-stack` branch (commit `70e8be5`) before cherry-picking PR #441, to isolate PR #441 contribution from other known blockers in the `robot-cityscapes-synthia` example:

| Fix applied | File changed | What it fixes |
|---|---|---|
| `sys.path.insert(0, ERFNet_dir)` before imports | `basemodel.py` | Issue [#472](https://github.com/kubeedge/ianvs/issues/472) Bug B: bare relative ERFNet imports fail in DataLoader subprocesses |
| `range=(0,255)` to `value_range=(0,255)` at 4 sites | `ERFNet/utils/summaries.py` | Issue [#472](https://github.com/kubeedge/ianvs/issues/472) Bug A: deprecated torchvision keyword |
| `self.cuda = torch.cuda.is_available()` + device detection | `ERFNet/utils/args.py`, `ERFNet/train.py` | Issue [#471](https://github.com/kubeedge/ianvs/issues/471): hardcoded `.cuda()` crashes on CPU-only machines |
| `__call__(self, samples)` with `task_extractor` removed | `task_allocation_by_domain.py` | Issue [#473](https://github.com/kubeedge/ianvs/issues/473) Bug A: Sedna calls `allocator(samples)` with 1 arg |
| Guards before `data.x[0]` access | `ERFNet/dataloaders/datasets/cityscapes.py` | Issue [#473](https://github.com/kubeedge/ianvs/issues/473) Bug B: unsafe array access |
| `job.inference_2(...)` to `job.inference(...)` line 328 | `lifelong_learning.py` | Issue [#470](https://github.com/kubeedge/ianvs/issues/470): `inference_2` not in Sedna API |
| `job.my_inference(...)` to `seen_estimator.predict(...)` line 155 | `lifelong_learning.py` | Issue [#461](https://github.com/kubeedge/ianvs/issues/461): `my_inference` not in Sedna API |
| `train_url`/`test_url` to `train_index`/`test_index` | `testenv/testenv.yaml` | Issue [#79](https://github.com/kubeedge/ianvs/issues/79): field names not recognized by `dataset.py` |

**B. Layered Stack Execution: running-stack + PR #441 `lifelong_learning.py` changes**

```bash
git checkout running-stack
git cherry-pick 6ee1ba4   # PR #441 main commit
```

**Section B-1: PR #441 applied on top of running-stack, pipeline advances, crashes at seen_estimator.predict()**

PR-441-Section03

```
[ERROR] base.py(181) - RetryError[<Future at 0x1589b2450 state=finished returned NoneType>]
[INFO]  lifelong_learning.py(145) - {"accuracy": 0.0}
  ...
    raise EOFError
EOFError
RuntimeError: (paradigm=lifelonglearning) pipeline runs failed, error:
```

Three findings from this run:

**Finding 1: PR #441 introduces no regression and no improvement for `no-inference` mode.** All three `lifelong_learning.py` hunks affect `hard-example-mining` mode and `_inference()` paths, neither of which is reached in the `no-inference` execution path. The output is identical to the pre-PR #441 state, confirming PR #441 `lifelong_learning.py` changes are strictly orthogonal to this execution path.

**Finding 2: `my_eval()` succeeds and returns the expected dict format.** The log line `lifelong_learning.py(145) [INFO] - {"accuracy": 0.0}` is direct evidence that `job.evaluate()` returns a metrics dict and the caller at line 140 receives it cleanly. The `accuracy: 0.0` value is expected because the knowledge base has not been populated by a Sedna server.

**Finding 3: The crash has moved one layer deeper.** The pipeline clears `my_eval()` completely and crashes at `seen_estimator.predict()` (line 150) via `FileOps.load()` then `joblib.load()` then `pickle.load()`, raising `EOFError` from an empty `index.pkl`. This is a Sedna server dependency surfacing at a different call site, outside the scope of this PR.

---

### Sub-comment Summary

| File | Line(s) | Sub-comment Topic |
|---|---|---|
| `lifelong_learning.py` | 271 | Hunk 1: `_train(... 0)` is verified correct; `rounds=0` sets `HAS_COMPLETED_INITIAL_TRAINING=False` in `_train()`, which is the right semantic for the initial training call. The original `r=1` would have set the flag to `True` immediately (wrong). |

NishantSinghhhhh:
[Screencast from 2026-06-03 22-19-10.webm](https://github.com/user-attachments/assets/e7f74c11-5e25-4a6b-8182-f9053ecf7c3d)

Running withour errors, reduced the epoches and dataset size to debug things faster

MooreZheng (commented):
/lgtm

MooreZheng:
This version have modified as reviewer asked and looks good to me 

what do you think @hsj576

suhaan-24:
Reviewed against `sedna 0.6.0.1` / Python 3.9.6 / macOS Apple M4, CPU+MPS. I restored the same
Example independently, so this is a comparison against a working configuration rather than a
read-through. Static review of this branch — I have **not** executed it.

## On the `_train(..., 0)` question — agreeing with the existing review

@abhisheksainimitawa's analysis above is correct and I want to endorse it explicitly, because I
initially read this hunk the same wrong way. The `0` sits inside the `if r == 1:` branch; the
`else` branch for rounds 2, 3, … passes `r` unchanged. It does not treat every round as initial
training.

For what it is worth, my own restoration arrived at the same place by a different route —
`_train(..., r - 1)` on the first round, which for `r == 1` is identical. One side effect worth
noting that the existing discussion does not cover: this also moves round-1 training output from
`output/train/1/` to `output/train/0/`. That is the correct location in my experience (the
downstream `CLOUD_KB_INDEX` consumer reads it there), but it is a path change as well as a flag
change, and #573 fixes the same flag semantics *without* moving the directory. Whichever approach
is adopted, the two should not both land — the invariant would then be encoded at both the call
site and inside `_train`.

## The concern not yet raised: `_eval` returns a non-deployable index

```python
job.evaluate(eval_dataset, metrics=metric_func)
edge_task_index = os.path.join(eval_output_dir, "index.pkl")
```

The motivation is right — sedna 0.6.0.1 returns metrics, not a path, so the original
`edge_task_index = job.evaluate(...)` is broken. But the index written to `output/eval/1/` is
missing the field the edge deployment path dereferences. Loading the artifacts from a completed
run with `sedna.common.file_ops.FileOps.load`:

```
--- eval/1/index.pkl ---   keys=['seen_task', 'unseen_task']          <-- no 'create_time'
  entry='real_semantic_segamentation_model'  samples.data_url=None
  entry='sim_semantic_segamentation_model'   samples.data_url=None

--- train/0/index.pkl ---  keys=['seen_task', 'unseen_task', 'create_time']
  entry='real_semantic_segamentation_model'  samples.data_url='.../train_real_semantic_segamentation_model.sample'
```

`EdgeKnowledgeManagement.save_task_index()` dereferences it directly:

```python
_task.samples.data_url = FileOps.download(_task.samples.data_url, sample_dir)
```

and `FileOps.download(None, ...)` raises
`AttributeError: 'NoneType' object has no attribute 'startswith'`.

It is latent rather than obvious: `eval/2/index.pkl` *does* carry `data_url` (it points back at
the `train/0` samples), so only the first round is affected, and a check that only looks at the
final leaderboard will not catch it. Suggest testing deployability before returning the eval index
and falling back to the cloud/train index with a log line, rather than returning the eval path
unconditionally.

Worth flagging that #645 makes the same change independently, so this needs deciding once rather
than twice.

## On the exception-to-warning change

Endorsing `gemini-code-assist`'s point rather than repeating it: converting the empty-dataset
exception in `cityscapes.py` into a warning with an early return means a misconfigured dataset
path produces an empty run instead of a clear failure. Given that hardcoded and stale dataset
paths are the single most common breakage in these Examples, this is exactly the condition worth
failing loudly on.

## Scope

+609/−216 across 13 files mixing Core (`lifelong_learning.py`, which affects every lifelong-learning
Example) and one Example directory. Splitting the Core hunks out would let them be evaluated
against #573 and #645, which touch the same function, and would let the Example-side restoration
land independently.

## Verification boundary

**This branch was not executed.** The index inspection and `FileOps.download(None)` behaviour above
are executed probes, but against artifacts from my own restored configuration of the same Example,
not from a run of this PR. The `data_url=None` observation applies to any code path that returns
the eval-directory index, which is what this PR does.

PR #484, opened 2026-05-27
examples: fix compression requirements filename

vyagh:
**What type of PR is this?**
/kind bug

This renames the compression benchmark requirements file from `requirements.txt` with a hidden U+200E mark to the
normal `requirements.txt` path.

The file contents are unchanged. This only fixes the filename so the README install command can find it.

None


vyagh:
a very tiny fix :)

/assign @jaypume

observer04 (commented):
**Recommendation: major revision.** The rename is real even though GitHub makes the two names look identical. The source filename ends in U+200E LEFT-TO-RIGHT MARK; the destination is the literal `requirements.txt`. Byte/tree inspection confirms that this fixes the local `pip install -r requirements.txt` path, as the PR body describes.

The newly reachable manifest is still incomplete. The benchmark’s `testalgorithms/basemodel.py` imports `psutil` at module import and calls `psutil.Process`, but `psutil` is not declared. I compiled the exact PR manifest successfully and confirmed `psutil` is absent from the resolved set.

```text
manifest resolution:         success
resolved psutil:              absent
direct runtime import:        basemodel.py:4
```

The optional `download_model_modelscope.py` also imports undeclared `modelscope`; either declare it as an optional utility dependency or remove/document that helper. Please add the runtime dependency before treating the rename as a complete install repair. No model, dataset, or benchmark run was executed.

vyagh:
@observer04 thanks for the review. 
Shouldn't this be handled in a separate PR?

observer04:
Yeah, you're right. That belongs in its own PR, not this one.

I raised it here because this rename is what makes the file reachable in the first place, so it was the first point where the install path could actually be tested. But that's a reason to open a follow-up, not to pile extra work onto a rename. Changing my recommendation to accept as is.

In case it's useful to whoever picks it up: `psutil` is imported in `testalgorithms/basemodel.py` at line 4 and used at line 46, and `download_model_modelscope.py` imports `modelscope` at line 3. Neither is in the manifest. Nothing for you to do here.

Good catch on the U+200E, by the way. A filename that renders exactly like the correct one is a nasty thing to track down. Hope this one gets merged.

Issue #495, opened 2026-06-02
[Audit] Core Framework Stability & Local Benchmark Runtime Validation


PragnyaKhandelwal:
### Overview
While preparing the execution environment for upcoming distributed AI benchmarking workloads, I conducted a stability audit of the core Ianvs evaluation pipelines. I identified and resolved several structural bottlenecks related to cross-platform path resolution, division-by-zero panics in the metric calculation modules, and configuration drift in the core example files. 

This tracking issue bundles the resulting architectural fixes into two distinct tiers: **Framework Core Hardening** and **Example Restoration**.

---

### 1. Framework Core Hardening (Engine Layer)
The following structural improvements resolve critical edge-case failures inside the core Ianvs benchmarking runtime:

*   **Metric Division-Zero Guard:** Introduces robust mathematical boundaries within the evaluation metrics engine to prevent fatal runtime panics (`ZeroDivisionError`) when dealing with zero-value boundaries or uninitialized pairs during early training iterations.
*   **Workspace Path Portability:** Replaces hardcoded user path assumptions with relative workspace routing (`workspace/tmp`), ensuring checkpoint directories are portable and execute seamlessly across different developer filesystems.
*   **Leaderboard NaN Sanitization:** Hardens the rank processing modules to gracefully sanitize and format `NaN` or uninitialized evaluation records when compiling the final markdown leaderboard output.

### 2. Flagship Example Restoration (Application Layer)
Validated and optimized the configurations for the repository's core distributed AI examples so they run cleanly end-to-end:

*   **Distributed LLM Pipeline:** Optimizes the token generation matrix (`max_tokens`) using a tiny Hugging Face model and edge-only routing parameters to drastically reduce execution overhead and prevent resource timeouts during local validation runs.
*   **Object Tracking Benchmark (MOT17):** Corrects baseline path specifications and data alignment indexing (`train_index`/`test_index`) for the ByteTrack multi-object tracking sequence.

---

### Tracking Pull Requests
The fixes are split across the following targeted branches (PR links will be updated as they are submitted):

- [ ] **Core Metrics Guard:** #496
- [ ] **Path Portability Fix:** #497 
- [x] **Leaderboard Sanitization:** #498 
- [ ] **LLM Configuration Fix:** #499 
- [ ] **MOT17 Config Alignment:** #500

PragnyaKhandelwal:
@MooreZheng @hsj576 @Poorunga Hello! I have completed the core stability audit.
[...]

PR #496, opened 2026-06-02
fix(engine): resolve `ZeroDivisionError` in metrics module

PragnyaKhandelwal:
**What type of PR is this?**

/kind bug

This PR guards against division-by-zero errors in metric calculations and lifelong learning paradigm average accuracy calculations.

Specifically, it implements the following safeguards:
- **Lifelong Learning Paradigm (`core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py`)**: Adds checks to ensure the task denominator `i` is not zero when calculating `task_avg_score['accuracy']`. If it is zero, it logs a warning and defaults the average accuracy to `0.0`, preventing a `ZeroDivisionError`.
- **Metrics Computation (`core/testcasecontroller/metrics/metrics.py`)**: Adds checks to ensure the denominator `denom = (length - 1) * (length - 1)` is greater than zero before performing division for the BWT, FWT, and accuracy scores. If it is less than or equal to zero, it logs/prints the score, assigns `np.nan` to `bwt_score` and `fwt_score`, and returns early.

Part of #495

### DIFF
--- a/core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py
+++ b/core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py
@@ -126,7 +126,10 @@ class LifelongLearning(ParadigmBase):
                         score_list = tmp_dict.get(entry, ['' for i in range(rounds)])
                         score_list[j-1] = scores
                         tmp_dict[entry] = score_list
-                    task_avg_score['accuracy'] = task_avg_score['accuracy']/i
+                    if i > 0:
+                        task_avg_score['accuracy'] /= i
+                    else:
+                        LOGGER.warning("task_avg_score accuracy denominator is zero; defaulting to 0.0")
                     score_list = tmp_dict.get("task_avg", [{'accuracy':0.0} for i in range(rounds)])
                     score_list[j-1] = task_avg_score
                     tmp_dict["task_avg"] = score_list
@@ -148,7 +151,10 @@ class LifelongLearning(ParadigmBase):
                 entry = detail.entry
                 LOGGER.info(f"{entry} scores: {scores}")
                 task_avg_score['accuracy'] += scores['accuracy']
-            task_avg_score['accuracy'] = task_avg_score['accuracy']/i
+            if i > 0:
+                task_avg_score['accuracy'] /= i
+            else:
+                LOGGER.warning("task_avg_score accuracy denominator is zero; defaulting to 0.0")
             self.system_metric_info[SystemMetricType.TASK_AVG_ACC.value] = task_avg_score
             LOGGER.info(task_avg_score)
             job = self.build_paradigm_job(ParadigmType.LIFELONG_LEARNING.value)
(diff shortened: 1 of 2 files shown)


PragnyaKhandelwal:
**Summary:** Fixes `ZeroDivisionError` crashes in the lifelong-learning paradigm and in BWT/FWT metric computation when a run contains zero or one task.

**Root cause:** In `core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py`, `task_avg_score['accuracy']` is divided by loop variable `i` at lines 130, 154, 234, and 260 without checking that `i > 0`. In `core/testcasecontroller/metrics/metrics.py` line 82, `accuracy` and `bwt_score` are divided by `(length - 1) * (length - 1)` without checking `length > 1`. Both crash with `ZeroDivisionError` when a dataset has only one task or no completed rounds.

**Fix:** All four division sites in `lifelong_learning.py` are wrapped with `if i > 0: ... else: LOGGER.warning(...)`. In `metrics.py`, an early-return guard (`if length <= 1: return [], np.nan, np.nan`) is added before the denominator is used.

**Evidence (before fix):**
```
ZeroDivisionError: division by zero
```
Raised at `lifelong_learning.py` line 130 (or 154, 234, 260) whenever a lifelong-learning job completes with `i == 0` (no task iterations), or at `metrics.py` line 82 when `length == 1`.

**Test (after fix):** Run any lifelong-learning example with a single-task dataset (or mock `i=0` in a unit test) — expect a `LOGGER.warning` and `task_avg_score['accuracy']` remaining `0.0`, with no exception raised. In `metrics.py`, `compute()` should return `([], nan, nan)` rather than crashing.

**Impact:** Zero changes to existing example behavior — guards are additive `if/else` branches; the normal code path (`i > 0`, `length > 1`) is unchanged.

DCO: All commits signed off. CI: Green. Ready for async review.

AdityaKumarSethia:
Reproduced both fixes in this PR against the real patched files.

`lifelong_learning.py` fix confirmed working at all 4 division sites. `metrics.py` fix confirmed working too, tested `compute('accuracy', [[]])` (the `length<=1` case) directly:

```python
from core.testcasecontroller.metrics.metrics import compute
result = compute('accuracy', [[]])
print(result)  # (nan, nan) on this patched branch; ZeroDivisionError on unpatched main
```
One inconsistency worth a quick fix before merge: the `lifelong_learning.py` half defaults to `0.0` on the guarded path, but the `metrics.py` half defaults to `nan` (plus an early-return `[]` for the matrix). Same PR, same class of bug, two different conventions for "safe" fallback, worth standardizing on one value.

Also worth flagging: two other open PRs (#588, #755) independently patch the same 4 `lifelong_learning.py` lines this PR does. Confirmed via real `git merge` that both conflict directly with this PR:

```bash
git merge pr-496 --no-edit   # clean
git merge pr-588 --no-edit   # CONFLICT
```
```bash
git merge pr-496 --no-edit   # clean
git merge pr-755 --no-edit   # CONFLICT
```
Of the three, this one has the broadest fix (also covers `metrics.py`), so I'd recommend this as the one to keep, with #588 and #755 closed as duplicates once the `0.0`/`nan` inconsistency above is resolved.

Full cross-PR analysis in [Discussion #873](https://github.com/kubeedge/ianvs/discussions/873).

PR #498, opened 2026-06-02
fix(engine): sanitize NaN values in leaderboard output

PragnyaKhandelwal:
**What type of PR is this?**
/kind bug

This PR sanitizes `NaN` values during the leaderboard generation process by filling them with "N/A" strings in the resulting CSV output. This prevents invalid or uninitialized metric records from causing downstream display or parsing errors when compiling the final markdown leaderboard.

Part of #495

This PR is part of the core stability audit for Ianvs. By explicitly handling `NaN` values in `rank.py`, we ensure the generated rank reports are cleaner and more resilient to edge cases where metrics might fail to initialize.

### DIFF
--- a/core/storymanager/rank/rank.py
+++ b/core/storymanager/rank/rank.py
@@ -188,7 +188,8 @@ class Rank:
         # pylint: disable=E1101
         all_df = copy.deepcopy(self.all_df)
         all_df.index = np.arange(1, len(all_df) + 1)
-        all_df.to_csv(self.all_rank_file, index_label="rank", encoding="utf-8")
+        all_df.to_csv(self.all_rank_file, index_label="rank", encoding="utf-8", na_rep="N/A")
+
 
     def _get_selected(self, test_cases, test_results) -> pd.DataFrame:
         module_types = self.selected_dataitem.get("modules")
@@ -228,7 +229,9 @@ class Rank:
         # pylint: disable=E1101
         selected_df = self._get_selected(test_cases, test_results)
         selected_df.index = np.arange(1, len(selected_df) + 1)
-        selected_df.to_csv(self.selected_rank_file, index_label="rank", encoding="utf-8")
+        selected_df.to_csv(
+            self.selected_rank_file, index_label="rank", encoding="utf-8", na_rep="N/A"
+        )
 
     def _draw_pictures(self, test_cases, test_results):
         # pylint: disable=E1101


PragnyaKhandelwal:
**Summary:** Fixes silent `NaN` values written into `all_rank.csv` and `selected_rank.csv` when a test case fails or returns no metric for a particular key.

**Root cause:** `pandas.DataFrame.to_csv()` serialises Python/NumPy `NaN` as an empty string by default. When a lifelong-learning run produces `np.nan` for BWT/FWT scores (e.g., from the division-by-zero guard in PR #496), the rank CSV contains empty cells with no indication of why the value is missing. `core/storymanager/rank/rank.py` lines 191 and 231 call `to_csv()` without `na_rep`, leaving consumers of the CSV unable to distinguish "not computed" from "accidentally blank".

**Fix:** Add `na_rep="N/A"` to both `to_csv()` calls — `_save_all()` at line 191 and `_save_selected()` at line 231 — so missing metric values are written as the string `"N/A"` rather than empty cells.

**Evidence (before fix):** After a lifelong-learning run with only one task (making `length - 1 = 0`), `selected_rank.csv` contains:
```
rank,algorithm,accuracy,bwt_score,fwt_score,...
1,rfnet,0.72,,,...
```
The empty `bwt_score`/`fwt_score` cells are indistinguishable from missing data and break any downstream CSV parser that treats empty as `0`.

**Test (after fix):** Run a lifelong-learning benchmark that triggers the NaN path; inspect `workspace/.../rank/selected_rank.csv` — BWT/FWT columns should contain `"N/A"` instead of blank cells.

**Impact:** Zero changes to existing example behavior — additive parameter to existing `to_csv()` calls; no change to leaderboard ranking logic or existing non-NaN metric values.

DCO: All commits signed off. CI: Green. Ready for async review.

MooreZheng (changes requested):
Confirm which padas version is relied on for the parameter na_ref

PragnyaKhandelwal:
> Confirm which padas version is relied on for the parameter na_ref

`na_rep` is a parameter of `DataFrame.to_csv()` that has been available since pandas 0.14.0 (released May 2014) and is present in every version from 0.x through the current 2.x series. It is listed in the [official pandas documentation](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html) as a core parameter with no version deprecation notes.

Ianvs' `requirements.txt` lists `pandas` with no version floor, so this parameter is safe for any pandas installation users would realistically have. No minimum version constraint is needed.

MooreZheng (commented):
/lgtm

hsj576 (approved):
/lgtm

MooreZheng (approved):
/approve

Issue #491, opened 2026-05-29
Issue summarization for example of imagenet/multiedge_inference_bench


Nityahiray:
> Author: Nitya Hiray
> Pre-test round:** LFX Mentorship 2026 Term 2
> Date: May 2026
> Scope: This analysis covers 1 existing GitHub issue (#341), 
> 1 existing PR (#315), and 6 new bugs found by static code analysis 
> filed as GitHub issues (#478, #479, #480, #485, #486, #487).

---

## 1. Background

The `imagenet/multiedge_inference_bench` example is part of KubeEdge 
Ianvs. It tests how well AI models classify images from the ImageNet 
dataset when running across multiple edge devices at the same time.

This example is the ONLY example in the entire Ianvs project that 
tests the **MultiEdge Inference** paradigm. This means if this example 
is broken, nobody can test MultiEdge Inference at all. Every bug found 
here blocks the whole MultiEdge Inference paradigm for all Ianvs users.

As of May 2026, this example is **entirely non-runnable** due to 
dependency conflicts, deprecated APIs, hardware configuration issues, 
and broken model weight download URLs.

---

### 2.1 Existing GitHub Issues

| # | Issue | Category | Status | Link |
|---|-------|----------|--------|------|
| 1 | Dependency conflicts, CUDA/cuDNN mismatch, pynvml crash, single GPU fails | Runtime / Hardware | Open | [#341](https://github.com/kubeedge/ianvs/issues/341) |

### 2.2 Existing Pull Requests

| # | PR | Category | Status | Link |
|---|-----|----------|--------|------|
| 1 | fix: typo in .md, .yaml, and .py files | Cleanup / Typos | Open — Not Merged | [#315](https://github.com/kubeedge/ianvs/pull/315) |

### 2.3 New Bugs Found and Filed as GitHub Issues

| Issue # | File | Bug Title | Severity |
|---------|------|-----------|----------|
| [#478](https://github.com/kubeedge/ianvs/issues/478) | bert.py | save_weights() stores PyTorch tensors instead of numpy arrays | HIGH |
| [#479](https://github.com/kubeedge/ianvs/issues/479) | bert.py | BertSelfAttention API breaks on transformers >= 4.30 | HIGH |
| [#480](https://github.com/kubeedge/ianvs/issues/480) | bert.py | BertEmbeddings.forward() signature mismatch | MEDIUM |
[...]

Issue #487, opened 2026-05-28
imagenet/multiedge_inference_bench: ViTSelfAttention API breaks on transformers >= 4.30 in deit.py


Nityahiray:
**What happened**:
In deit.py, DeiTLayerShard.forward() calls 
self.self_attention(data_norm) with only one 
argument. In transformers version 4.30 and newer, 
ViTSelfAttention.forward() requires additional 
keyword arguments. This raises TypeError on any 
modern system.

**What you expected to happen**:
The forward pass should run without errors on 
modern transformers versions.

**How to reproduce it**:
1. Install transformers >= 4.30
2. Run the imagenet example
3. TypeError is raised in DeiTLayerShard.forward()

**Anything else we need to know?**:
Fix: Update the call to:
self.self_attention(
    data_norm,
    attention_mask=None,
    head_mask=None,
    output_attentions=False
)

rajdhruvsingh:
Hi, I'd like to work on this. I have already fixed 
related issues #478, #479, #480, #485, #486 in bert.py 
and deit.py. Confirmed via inspect.signature() on 
transformers 5.10.2 that ViTSelfAttention.forward() 
accepts hidden_states, attention_mask=None, and **kwargs.
Can I be assigned?

/assign

Issue #480, opened 2026-05-26
imagenet/multiedge_inference_bench: BertEmbeddings.forward() signature mismatch in bert.py


Nityahiray:
**What happened**:
In bert.py, BertModelShard.forward() calls 
self.embeddings(data) by passing a raw tensor 
as a positional argument. But newer versions of 
transformers expect input_ids as a keyword argument 
in BertEmbeddings.forward(). This causes unexpected 
behavior or TypeError.

**What you expected to happen**:
The embeddings layer should accept the input tensor 
correctly without errors.

**How to reproduce it**:
1. Install latest transformers
2. Run the imagenet example
3. Error occurs in BertModelShard.forward()

**Anything else we need to know?**:
Fix: Change to:
data = self.embeddings(input_ids=data)

rajdhruvsingh:
Hi, I'd like to work on this issue.
Fix involves changing self.embeddings(data) to 
self.embeddings(input_ids=data) in BertModelShard.forward().
Can I be assigned?

/assign

Issue #479, opened 2026-05-26
imagenet/multiedge_inference_bench: BertSelfAttention API breaks on transformers >= 4.30 in bert.py


Nityahiray:
**What happened**:
In bert.py, BertLayerShard.forward() calls 
self.self_attention(data) with only one argument. 
In transformers version 4.30 and newer, 
BertSelfAttention.forward() requires additional 
keyword arguments like attention_mask. 
This raises TypeError on any modern system.

**What you expected to happen**:
The forward pass should run without errors on 
modern transformers versions.

**How to reproduce it (as minimally and precisely as possible)**:
1. Install transformers >= 4.30
2. Run the imagenet example
3. TypeError is raised in BertLayerShard.forward()

**Anything else we need to know?**:Fix: Update the call to:
self.self_attention(
    data,
    attention_mask=None,
    head_mask=None,
    output_attentions=False
)

rajdhruvsingh:
Hi, I'd like to work on this issue.
I can see the fix involves updating self.self_attention()
call in BertLayerShard.forward() to pass the required
keyword arguments for transformers >= 4.30.
Can I be assigned to this?

/assign

rajdhruvsingh:
Hi, I investigated this issue carefully.

Environment tested: transformers 5.10.2

Findings:
1. The one-argument call self.self_attention(data_norm) works 
   without TypeError on transformers 5.x — the API was made 
   backward compatible in newer versions.

2. The current BertSelfAttention.forward() signature on 5.10.2 is:
   - hidden_states (required)
   - attention_mask (default=None)  
   - past_key_values (default=None)
   
   Note: head_mask and output_attentions from the original fix 
   suggestion are NOT present in 5.x — adding them would 
   cause a new TypeError.

3. There is however a warning about config._attn_implementation 
   being None when using BertSelfAttention as a standalone module,
   which is how this codebase uses it.

Proposed fix: Make the call explicit with kwargs that exist in 
both 4.30+ and 5.x — only attention_mask=None is safe to add 
across both versions.

Can a maintainer confirm which transformers version range this 
project officially supports? That will determine the right fix.

PR #501, opened 2026-06-04
fix: fix torch.hub.load warning and tensor/numpy bug in deit.py

### LINKED ISSUE #485: imagenet/multiedge_inference_bench: save_weights() stores PyTorch tensors instead of numpy arrays in deit.py
**What happened**:
In deit.py, the save_weights() function in 
DeiTModelShard stores PyTorch tensor objects 
directly into a dictionary and then calls 
np.savez() to save them. But np.savez() only 
supports numpy arrays, not PyTorch tensors. 
This crashes with a TypeError when anyone tries 
to save model weights.

The model weights should be saved successfully 
as a .npz file without any errors.

**How to reproduce it**:
2. Run DeiTModelShard.save_weights()

Fix: Change the line to:

### LINKED ISSUE #486: imagenet/multiedge_inference_bench: torch.hub.load() pretrained=True is deprecated in deit.py
**What happened**:
In deit.py, DeiTModelShard.save_weights() uses:
torch.hub.load(hub_repo, hub_model_name, pretrained=True)

The pretrained=True argument was deprecated in 
newer PyTorch versions. On modern PyTorch it raises 
FutureWarning and in very new versions it raises 
an error. This completely breaks the weight 
downloading process.

The DeiT model weights should download and load 
successfully without any warnings or errors.

**How to reproduce it**:
1. Install latest PyTorch
2. Call DeiTModelShard.save_weights()
3. FutureWarning or TypeError is raised

Fix: Change to:
model = torch.hub.load(
    hub_repo,
    hub_model_name,
    weights='DEFAULT'
)

rajdhruvsingh:
/kind bug

## What this PR does

Fixes two issues in `DeiTModelShard.save_weights()`:

1. Suppresses the `FutureWarning` raised by `torch.hub.load(..., pretrained=True)` in newer PyTorch versions.
2. Converts model weights to NumPy arrays before saving them with `np.savez()`.

### Issue #486

`torch.hub.load(..., pretrained=True)` emits a `FutureWarning` in newer PyTorch versions, which can clutter logs and affect compatibility. The warning is now suppressed during model loading.

### Issue #485

The original implementation stored PyTorch tensors directly in the `weights` dictionary before calling `np.savez()`. This can lead to serialization issues because `np.savez()` is intended to store NumPy arrays.

Converting tensors using:

```python
val.detach().cpu().numpy()
```

ensures the weights are saved in a NumPy-compatible format.

## Changes

* Wrapped `torch.hub.load()` in a `warnings.catch_warnings()` block and suppressed `FutureWarning`.
* Converted tensors to NumPy arrays before storing them in the weights dictionary.
* Updated:
  `examples/imagenet/multiedge_inference_bench/testalgorithms/automatic/models/transformers/deit.py`

## Which issue(s) this PR fixes

Fixes #485

Fixes #486

### DIFF
--- a/examples/imagenet/multiedge_inference_bench/testalgorithms/automatic/models/transformers/deit.py
+++ b/examples/imagenet/multiedge_inference_bench/testalgorithms/automatic/models/transformers/deit.py
@@ -178,11 +178,14 @@ class DeiTModelShard(ModuleShard):
                              hub_model_name)
             else:
                 hub_model_name = model_name
-        model = torch.hub.load(hub_repo, hub_model_name, pretrained=True)
+        import warnings
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", FutureWarning)
+            model = torch.hub.load(hub_repo, hub_model_name, pretrained=True)
         state_dict = model.state_dict()
         weights = {}
         for key, val in state_dict.items():
-            weights[key] = val
+            weights[key] = val.detach().cpu().numpy()
         np.savez(model_file, **weights)
 
 


rajdhruvsingh:
Thank you for the review from gemini-code-assist.

After investigating further, I understand the issue now.
`weights='DEFAULT'` applies to torchvision.models but
facebookresearch/deit's hubconf.py still expects `pretrained=True`
as its own parameter. Changing it would cause a TypeError.

I am investigating the correct approach and will update
the PR. The real fix likely involves suppressing the
FutureWarning while keeping pretrained=True, since we
cannot change Facebook's hub interface.

rajdhruvsingh:

MooreZheng (commented):
/lgtm

MooreZheng (approved):
/approve

PR #503, opened 2026-06-05
Fix/bert save weights numpy conversion

### LINKED ISSUE #478: imagenet/multiedge_inference_bench: save_weights() stores PyTorch tensors instead of numpy arrays in bert.py
**What happened**:
In bert.py, the save_weights() function in both 
BertModelShard and BertShardForSequenceClassification 
stores PyTorch tensor objects directly into a dictionary 
and then calls np.savez() to save them. But np.savez() 
only supports numpy arrays, not PyTorch tensors. 
This crashes with a TypeError when anyone tries to 
save model weights.

The model weights should be saved successfully as a 
.npz file without any errors.

2. Run BertModelShard.save_weights() or 
   BertShardForSequenceClassification.save_weights()

Fix: Change the line to:

This bug exists in both:
- BertModelShard.save_weights()
- BertShardForSequenceClassification.save_weights()

rajdhruvsingh:
## What this PR does
Fixes save_weights() in bert.py to convert PyTorch tensors 
to numpy arrays before calling np.savez().

## Why
np.savez() only accepts numpy arrays. Storing raw tensors 
causes TypeError and completely breaks weight saving.

## Changes
- BertModelShard.save_weights(): val → val.detach().cpu().numpy()
- BertShardForSequenceClassification.save_weights(): same fix

## Testing
Verified tensor → numpy conversion locally. np.savez() saves 
and loads back correctly after fix.

Fixes #478

### DIFF
--- a/examples/imagenet/multiedge_inference_bench/testalgorithms/automatic/models/transformers/bert.py
+++ b/examples/imagenet/multiedge_inference_bench/testalgorithms/automatic/models/transformers/bert.py
@@ -157,7 +157,7 @@ class BertModelShard(ModuleShard):
         state_dict = model.state_dict()
         weights = {}
         for key, val in state_dict.items():
-            weights[key] = val
+            weights[key] = val.detach().cpu().numpy()
         np.savez(model_file, **weights)
 
 
@@ -215,5 +215,5 @@ class BertShardForSequenceClassification(ModuleShard):
         state_dict = model.state_dict()
         weights = {}
         for key, val in state_dict.items():
-            weights[key] = val
+            weights[key] = val.detach().cpu().numpy()
         np.savez(model_file, **weights)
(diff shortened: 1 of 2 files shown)


rajdhruvsingh:

MooreZheng (changes requested):
As discussed on the routine meeting, we suggest
1) rebase the pull request and squash the commits into one commit
2) as pull request #503, #506, #507, #508 for the same example of multiedge inference, we might want to suggest to merge these four (or more) pull requests into one 
3) also attach with more comments to pin-point the issues that fixed

rajdhruvsingh:
`np.savez()` only accepts numpy arrays, but the code was passing raw PyTorch 
tensors directly into it without converting them first.

The issue existed in two places:
- `BertModelShard.save_weights()` at line ~157
- `BertShardForSequenceClassification.save_weights()` at line ~215

In both methods, the loop:
    for key, val in state_dict.items():
        weights[key] = val  # raw PyTorch tensor — wrong

was storing raw tensors into the weights dict before passing it to np.savez().

When `save_weights()` was called, it would crash with a TypeError at runtime:
    TypeError: Cannot cast array data from dtype('O') to dtype('float32')
or silently save corrupt data — because PyTorch tensors are not numpy arrays 
and np.savez() doesn't know how to serialize them.

Fixed by calling .detach().cpu().numpy() on each tensor before saving:

- .detach() removes gradient tracking
- .cpu() moves tensor from GPU to CPU if needed
- .numpy() converts to a numpy array that np.savez() can serialize

PR #506, opened 2026-06-06
fix: pass explicit kwargs to BertSelfAttention.forward() in bert.py

### LINKED ISSUE #479: imagenet/multiedge_inference_bench: BertSelfAttention API breaks on transformers >= 4.30 in bert.py

rajdhruvsingh:
## What this PR does
Updates BertLayerShard.forward() to pass explicit keyword 
arguments to BertSelfAttention.forward().

## Why
transformers >= 4.30 changed BertSelfAttention.forward() to 
require explicit kwargs. While transformers 5.x made the call 
backward compatible again, explicitly passing attention_mask=None 
and past_key_values=None future-proofs the code and documents 
intent clearly.

Note: head_mask and output_attentions from the original issue 
suggestion are NOT present in transformers 5.x API — only 
attention_mask and past_key_values are valid kwargs confirmed 
via inspect.signature() on transformers 5.10.2.

Fixes #479

### DIFF
--- a/examples/imagenet/multiedge_inference_bench/testalgorithms/automatic/models/transformers/bert.py
+++ b/examples/imagenet/multiedge_inference_bench/testalgorithms/automatic/models/transformers/bert.py
@@ -42,11 +42,7 @@ class BertLayerShard(ModuleShard):
     def forward(self, data: TransformerShardData) -> TransformerShardData:
         """Compute layer shard."""
         if self.has_layer(0):
-            data = (self.self_attention(
-                data,
-                attention_mask=None,
-                past_key_values=None
-            )[0], data)
+            data = (self.self_attention(data, attention_mask=None, past_key_values=None)[0], data)
         if self.has_layer(1):
             data = self.self_output(data[0], data[1])
         if self.has_layer(2):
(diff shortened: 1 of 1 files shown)


rajdhruvsingh:

rajdhruvsingh:
BertSelfAttention.forward() was being called with just a positional argument.

In transformers >= 4.30, BertSelfAttention.forward() requires attention_mask 
and past_key_values to be passed explicitly as keyword arguments.

The issue existed in BertLayerShard.forward() where self_attention was called 
with only a positional argument, causing a TypeError at runtime on newer 
versions of the transformers library.

Fixed by passing attention_mask=None and past_key_values=None explicitly.

PR #507, opened 2026-06-06
fix: pass input_ids as keyword argument to BertEmbeddings in bert.py

### LINKED ISSUE #480: imagenet/multiedge_inference_bench: BertEmbeddings.forward() signature mismatch in bert.py

rajdhruvsingh:
## What this PR does
Changes self.embeddings(data) to self.embeddings(input_ids=data)
in BertModelShard.forward().

## Why
BertEmbeddings.forward() expects input_ids as a keyword argument
in modern transformers versions. Positional passing works on
transformers 5.x but is fragile and unclear.

Confirmed via inspect.signature() on transformers 5.10.2:
  input_ids: default=None   ← expects keyword, not positional

## Testing
Verified on transformers 5.10.2 (macOS):
- embeddings(input_ids=fake_ids) → torch.Size([1, 10, 768]) ✅

Fixes #480

### DIFF
--- a/examples/imagenet/multiedge_inference_bench/testalgorithms/automatic/models/transformers/bert.py
+++ b/examples/imagenet/multiedge_inference_bench/testalgorithms/automatic/models/transformers/bert.py
@@ -143,7 +143,10 @@ class BertModelShard(ModuleShard):
     def forward(self, data: TransformerShardData) -> TransformerShardData:
         """Compute shard layers."""
         if self.shard_config.is_first:
-            data = self.embeddings(input_ids=data)
+            if isinstance(data, tuple):
+                data = self.embeddings(input_ids=data[0], token_type_ids=data[1])
+            else:
+                data = self.embeddings(input_ids=data)
         for layer in self.layers:
             data = layer(data)
         if self.shard_config.is_last:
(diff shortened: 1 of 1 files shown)


rajdhruvsingh:

rajdhruvsingh:
BertEmbeddings.forward() was being called with data passed as a plain positional argument and handled tuple input 
in BertModelShard.forward()

In transformers >= 4.30, BertEmbeddings.forward() requires input_ids to be 
passed as an explicit keyword argument rather than positionally.

The issue existed in BertModelShard.forward() where self.embeddings was called 
with data as a plain positional argument:
    data = self.embeddings(data)

Additionally, when both input_ids and token_type_ids are passed together as a 
tuple (e.g. for sequence classification), the old code passed the entire tuple 
as input_ids causing incorrect behaviour.

Fixed by checking if data is a tuple and unpacking accordingly:
    if isinstance(data, tuple):
        data = self.embeddings(input_ids=data[0], token_type_ids=data[1])
    else:
        data = self.embeddings(input_ids=data)

PR #510, opened 2026-06-07
fix: add sedna installation warning and setup verification script

### LINKED ISSUE #489: [Bug] pip install sedna installs wrong PyPI package - silent name collision blocks all LLM examples
**What happened**:

Running `pip install sedna` - as implied by the Ianvs documentation and the natural first instinct of any new contributor - silently installs a completely wrong PyPI package: a unit conversion library with no relation to KubeEdge.

The installed package exposes modules for unit conversion, not KubeEdge datasources:

``` bash
>>> import pkgutil, sedna
>>> [m.name for m in pkgutil.iter_modules(sedna.__path__)]
['convert', 'maths', 'physics', 'prime', 'system', 'voice']
>>> sedna.__version__
AttributeError: module 'sedna' has no attribute '__version__'
```

The contributor then gets a confusing downstream error with no indication the wrong package is installed:
``` bash
ModuleNotFoundError: No module named 'sedna.datasources'
```

Their natural response is to reinstall sedna - which reinstalls the wrong package again. There is no error at install time, no warning, and no way to diagnose this without knowing the collision exists.

Attempting to install from the official KubeEdge repo also fails:
``` bash
ERROR: git+https://github.com/kubeedge/sedna.git does not appear to be a Python package:
neither 'setup.py' nor 'pyproject.toml' found.
```
The only working installation path is completely undocumented: manual `git clone` + `export PYTHONPATH=~/kubeedge-sedna/lib:$PYTHONPATH`.

---

Either:
1. `pip install sedna` installs the correct KubeEdge Sedna package, or
2. The Ianvs documentation and README clearly document that `pip install sedna` does NOT work and provides the exact manual installation steps required.

---

```bash
# On any clean environment
pip install sedna
 
ianvs -f /path/to/benchmarkingjob.yaml

# Result: ModuleNotFoundError: No module named 'sedna.datasources'
# despite sedna appearing to be installed

# Verify the wrong package is installed:
python3 -c "import pkgutil, sedna; print([m.name for m in pkgutil.iter_modules(sedna.__path__)])"
# Output: ['convert', 'maths', 'physics', 'prime', 'system', 'voice']
```

Confirmed on: macOS ARM64 (Apple Silicon), Python 3.13.5, Miniconda. Also reproducible on any platform since the wrong PyPI package installs everywhere.

---

This is the highest-impact bug in the Ianvs ecosystem. Every contributor who follows standard Python conventions hits this silently. The 10 core Ianvs files that import from `sedna` - spanning every learning paradigm (singletask, incremental, lifelong, federated, multiedge, joint inference) - all fail silently with the wrong package installed.

There is a compounding CI failure documented in PR #417: the CI workflow was installing Sedna from a path that does not exist (`${{github.workspace}}/resources/third_party/*` instead of `${{github.workspace}}/examples/resources/third_party/*`). This means CI has been passing without Sedna installed across all open PRs, making this name collision undetectable through normal CI validation. Combined, these two failures mean:

- Contributors install the wrong `sedna` silently
- CI reports green without `sedna` installed at all
- Neither failure surfaces an error that points to the root cause

**Proposed fix (Ianvs-scope only, no upstream changes required):**

1. Add a warning to the core README explicitly stating `pip install sedna` installs the wrong package
2. Document the correct manual installation path: `git clone https://github.com/kubeedge/sedna.git` + `export PYTHONPATH=~/kubeedge-sedna/lib:$PYTHONPATH`
3. Add a setup verification script that detects which `sedna` is installed and warns the contributor if the unit-conversion package is present
4. Consider renaming the bundled wheel in `examples/resources/third_party/` to something unambiguous like `kubeedge-sedna` to prevent future confusion

Affects all examples that depend on Sedna datasources: `smart_coding`, `government`, `government_rag`, `llm_simple_qa`, `llm-edge-benchmark-suite`, and all future LLM example additions.

Environment: macOS ARM64 (Apple Silicon M-series), Python 3.13.5, Miniconda, no GPU.

rajdhruvsingh:
What this PR does
Addresses the sedna PyPI name collision documented in issue #489
that blocks every new contributor from running Ianvs examples.
Problem
pip install sedna silently installs an unrelated unit-conversion
library from PyPI. Contributors get a confusing:
ModuleNotFoundError: No module named 'sedna.datasources'
with no indication the wrong package is installed. Reinstalling
sedna just reinstalls the wrong package again.
Changes
1. docs/guides/quick-start.md
Added explicit warning before the Sedna installation step making
it impossible to miss. The correct wheel path was already correct
in this file (resources/third_party/).
2. scripts/verify_setup.py
New utility script that detects which sedna is installed by
checking for sedna.datasources. Prints clear step-by-step
remediation if the wrong PyPI package is found. Contributors
can run this after setup to confirm environment is correct:
python scripts/verify_setup.py
Testing

Verified KubeEdge sedna installs correctly from
resources/third_party/sedna-0.6.0.1-py3-none-any.whl ✅
verify_setup.py correctly detects KubeEdge sedna ✅
verify_setup.py tested on macOS ARM64, Python 3.14 ✅

Fixes #489

### DIFF
--- a/docs/guides/quick-start.md
+++ b/docs/guides/quick-start.md
@@ -95,7 +95,10 @@ conda create -n ianvs-experiment python=3.8 rust -c conda-forge
 conda activate ianvs-experiment
 
 # Install Sedna
-pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl
+> **WARNING: Do NOT run `pip install sedna`**  
+> Running `pip install sedna` silently installs a completely unrelated unit-conversion library from PyPI — not the KubeEdge Sedna package. Use the bundled wheel file as shown below.
+
+pip install resources/third_party/sedna-0.6.0.1-py3-none-any.whl
 
 # Install dependencies for Ianvs Core.
 pip install -r requirements.txt
--- a/docs/guides/quick-start.md
+++ b/docs/guides/quick-start.md
@@ -98,7 +98,7 @@ conda activate ianvs-experiment
 > **WARNING: Do NOT run `pip install sedna`**  
 > Running `pip install sedna` silently installs a completely unrelated unit-conversion library from PyPI — not the KubeEdge Sedna package. Use the bundled wheel file as shown below.
 
-pip install resources/third_party/sedna-0.6.0.1-py3-none-any.whl
+pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl
 
 # Install dependencies for Ianvs Core.
 pip install -r requirements.txt
--- a/scripts/verify_setup.py
+++ b/scripts/verify_setup.py
@@ -5,7 +5,7 @@ WARNING: Running 'pip install sedna' installs a completely wrong
 PyPI package (a unit conversion library). The correct KubeEdge
 Sedna must be installed from the bundled wheel file:
 
-    pip install resources/third_party/sedna-0.6.0.1-py3-none-any.whl
+    pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl
 """
 import sys
 
@@ -19,13 +19,13 @@ def check_sedna():
     except ImportError:
         print("\nSedna is not installed.")
         print("\nCorrect installation:")
-        print("  pip install resources/third_party/sedna-0.6.0.1-py3-none-any.whl")
+        print("  pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl")
         return False
 
     # check if it is the RIGHT sedna by verifying KubeEdge-specific module
     try:
         # pyrefly: ignore [missing-import]
-        import sedna.datasources.kafka  # noqa: F401
+        import sedna.datasources  # noqa: F401
         print("KubeEdge Sedna is installed correctly.")
         return True
     except ImportError:
@@ -35,7 +35,7 @@ def check_sedna():
         print("  1. Uninstall the wrong package:")
         print("     pip uninstall sedna -y")
         print("  2. Install the correct KubeEdge Sedna:")
-        print("     pip install resources/third_party/sedna-0.6.0.1-py3-none-any.whl")
+        print("     pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl")
         print("\nNever run 'pip install sedna' for Ianvs.")
         return False
 
@@ -60,5 +60,5 @@ if __name__ == "__main__":
         print("\nEnvironment setup complete. Ready to run benchmarks.")
         sys.exit(0)
     else:
-        print("\Environment setup incomplete. Fix the issues above first.")
+        print("\nEnvironment setup incomplete. Fix the issues above first.")
         sys.exit(1)
--- a/scripts/verify_setup.py
+++ b/scripts/verify_setup.py
@@ -1,0 +1,64 @@
+"""
+Verify that the correct KubeEdge Sedna package is installed.
+
+WARNING: Running 'pip install sedna' installs a completely wrong
+PyPI package (a unit conversion library). The correct KubeEdge
+Sedna must be installed from the bundled wheel file:
+
+    pip install resources/third_party/sedna-0.6.0.1-py3-none-any.whl
+"""
+import sys
+
+
+def check_sedna():
+    """Check if the correct KubeEdge Sedna is installed."""
+    print("Checking Sedna installation...")
+    try:
+        # pyrefly: ignore [missing-import]
+        import sedna  # noqa: F401
+    except ImportError:
+        print("\nSedna is not installed.")
+        print("\nCorrect installation:")
+        print("  pip install resources/third_party/sedna-0.6.0.1-py3-none-any.whl")
+        return False
+
+    # check if it is the RIGHT sedna by verifying KubeEdge-specific module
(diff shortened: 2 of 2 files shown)


hibaa23:

Found an issue in the script's own output. `scripts/verify_setup.py`'s remediation message reads:

    pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl

This is the stale path (broken since commit `b3b26a5`, 2026-04-22 — tracked in #840). A contributor who hits the PyPI name collision this script is designed to catch, and follows its printed remediation step, will be routed directly into a second, unrelated failure. Suggest updating the string to `resources/third_party/...` (no `examples/` prefix) before merge.

Separately: this PR's merge-base is `bf0f5968` (2026-05-30), roughly two months behind current `main` @ 36ec008. `git diff main pr-510 --stat` currently shows dozens of unrelated file changes (deleted proposal images, restructured docs) purely from staleness, not from this PR's actual diff. A rebase before merge would avoid accidentally reintroducing files `main` has since removed.

kavaljeetsingh-dev:
Reproduced the underlying problem independently first:
```
$ pip install sedna
$ python3 -c "import pkgutil, sedna; print([m.name for m in pkgutil.iter_modules(sedna.__path__)])"
['convert', 'maths', 'physics', 'prime', 'system', 'voice']
```
Confirms `pip install sedna` installs an unrelated unit-conversion package, not KubeEdge Sedna. Then ran this PR's own `verify_setup.py` against that exact broken state:
```
$ python3 verify_setup.py
WRONG sedna package detected!
```
It correctly detects the wrong package via the `sedna.datasources` check. The fix and its detection logic are both correct.

One issue this PR doesn't mention: its `setup.py` diff changes `default_version` from `"0.3.0"` to `"0.1.0"`. This isn't intentional — it's a side effect of the branch being 10 commits behind `main` (the version bump to 0.3.0 landed after this PR was opened) — but it would silently regress the reported package version if merged as-is. Worth fixing in a rebase before merge.

AmanBasu20:
This adds a genuinely useful detection mechanism for the Sedna PyPI collision documented in #489. As `hibaa23` already flagged, the script's own remediation string still points at the stale `examples/resources/third_party/...` path rather than the corrected `resources/third_party/...` location — worth fixing before merge, since a contributor who hits the collision and follows this script's printed remediation would be routed straight into a second, unrelated failure.

**Alternative approach worth weighing:** a verification script is a runtime detection layer over a naming problem, not a fix for the naming problem itself. #489's own proposed-fix list already suggests an alternative — renaming the bundled wheel to something unambiguous like `kubeedge-sedna` — which would eliminate the collision class entirely rather than detecting it after the fact. The detection script must be manually kept in sync with the correct install path (as evidenced by the stale path currently embedded in its own output), whereas a rename removes the need for any detection logic at all. Not a blocker for this PR, but worth tracking as a more durable follow-up alongside it.

aryananand-04:
The collision this fixes is real. On a clean environment `pip install sedna` pulls `Sedna-0.1.2` from PyPI, which is somebody else's package entirely, and the resulting `ModuleNotFoundError: No module named 'sedna.datasources'` gives no hint that the wrong project is installed. A warning in the docs plus a checker script is a sensible response.

There's a problem with the path this PR points people at, though.

The PR body says "The correct wheel path was already correct in this file (`resources/third_party/`)". I don't think that's the case. On current `main`, `docs/guides/quick-start.md:98` reads:

```
pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl
```

and `docs/guides/how-to-install-ianvs.md:62` has the same prefix. There is no `examples/resources/` directory in this repository. The wheel lives at the top level:

```
resources/third_party/sedna-0.6.0.1-py3-none-any.whl
```

I hit this during first-time setup before this PR existed, following the install guide verbatim:

```
$ python -m pip install ./examples/resources/third_party/*
zsh: no matches found: ./examples/resources/third_party/*
```

So the path wasn't already correct, and this PR doesn't correct it. It carries it forward instead. Grepping the diff, the bad prefix gets added in three more places:

```
+    pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl
+        print("  pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl")
+        print("     pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl")
```

Two of those are inside `verify_setup.py`'s remediation output. That's the part that worries me most: a contributor hits the collision, runs the checker this PR adds, follows the fix it prints, and lands on a second failure that has nothing to do with the first. The script is meant to be the thing that gets you unstuck.

Worth considering whether this is just a typo. I don't think so, because the same prefix shows up independently in the install guide, in the quick-start, in this PR's body, and in the new script's output. It looks like a path that's wrong consistently wherever it appears, which also means fixing it here is a bigger win than it looks.

One small thing: both #489 and this PR describe the squatted package as a "unit-conversion library". PyPI's listing for `Sedna-0.1.2` calls it "a python package for computational mathematics". Doesn't change anything about the fix, just worth not repeating.

Dropping `examples/` from all four occurrences would make this complete. The warning and the checker are both good additions otherwise.

Reproduced on macOS arm64 in a clean venv. I verified the directory layout directly on `main` and on this PR's branch rather than inferring it from the docs.

PR #511, opened 2026-06-07
fix: add missing DDP/dist imports, fix assertion and YAML indentation in ByteTrack

### LINKED ISSUE #443: [MOT17] ByteTrack basemodel.py crashes on multi-GPU: `DDP` and `dist` never imported
### Summary

`examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/tracking/byte_track/basemodel.py`
uses `torch.nn.parallel.DistributedDataParallel` (as `DDP`) and `torch.distributed`
(as `dist`) without importing either. The code paths that reference these names are
gated behind `is_distributed = num_gpu > 1`, so the bug is completely silent on
single-GPU and CPU machines but causes immediate `NameError` crashes on any multi-GPU
server — precisely the environment the `multiedgeinference` paradigm is designed for.

### Environment

- **Files affected:**
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/tracking/byte_track/basemodel.py`

---

#### Finding 1 — `DDP` used but never imported

**File:** `testalgorithms/tracking/byte_track/basemodel.py`  
**Line:** 115  
**Code:**
```python
# Inside load(), when self.is_distributed is True:
if self.is_distributed:
    self.model = DDP(self.model, device_ids=[self.rank])
```

**Problem:**  
`DDP` (`torch.nn.parallel.DistributedDataParallel`) is used directly but never imported
in the module. On a single-GPU machine `self.is_distributed` is `False`, so this line is
never reached and the missing import goes unnoticed. On a multi-GPU machine, `load()` is
called after the model weights are loaded, at which point Python raises:

```
NameError: name 'DDP' is not defined
```

The model has already been loaded into memory at this point, so the crash wastes the
time spent on the `torch.load()` call.

**Impact:**  
Distributed model loading is entirely broken. Any multi-GPU deployment of the ByteTrack
tracking job fails at the `load()` step with an unrecoverable `NameError`.

**Proposed Fix:**
```python
# Add to the import block at the top of basemodel.py:
from torch.nn.parallel import DistributedDataParallel as DDP
```

---

#### Finding 2 — `dist` used but never imported

**File:** `testalgorithms/tracking/byte_track/basemodel.py`  
**Line:** 176  
**Code:**
```python
# Inside _get_eval_loader(), when is_distributed is True:
if is_distributed:
    batch_size = batch_size // dist.get_world_size()
    sampler = torch.utils.data.distributed.DistributedSampler(valdataset, shuffle=False)
```

**Problem:**  
`dist` (`torch.distributed`) is referenced to retrieve the world size for batch
partitioning but is never imported. The same single-GPU/multi-GPU invisibility applies:
single-GPU users never trigger this branch. On multi-GPU, the crash is:

```
NameError: name 'dist' is not defined
```

This occurs inside `_get_eval_loader()`, which is called from `predict()`, meaning the
model must run through the full forward pass before the error surfaces — expensive in
compute time.

**Impact:**  
Distributed data loading is broken. Even if `DDP` were added (Finding 1), the data
loader construction would still crash on multi-GPU machines.

**Proposed Fix:**
```python
# Add to the import block at the top of basemodel.py:
import torch.distributed as dist
```

---

### Steps to Reproduce

1. Set up a machine with 2 or more CUDA-capable GPUs.
2. Install ianvs and the ByteTrack dependencies per the MOT17 README.
3. Run:
   ```bash
   ```
4. Observe `NameError: name 'DDP' is not defined` during the `load()` phase, or
   `NameError: name 'dist' is not defined` during `_get_eval_loader()`.

On a single-GPU machine neither error surfaces, masking the bug entirely.

---

### Additional Notes

Both missing imports should be added together in a single fix. The complete import block
addition is:

```python
import torch.distributed as dist
from torch.nn.parallel import DistributedDataParallel as DDP
```

Related: the `multiedgeinference` paradigm is designed specifically for distributed
multi-edge inference scenarios. Missing distributed primitives in the reference
implementation undermine the paradigm's stated purpose. This bug has likely gone
undetected because CI and development typically run on single-GPU machines.

### LINKED ISSUE #444: [MOT17] ByteTrack: tautological GPU assertion never fires; malformed YAML indentation silently drops `batch_size`
### Summary

Two independent configuration-level bugs in the ByteTrack tracking algorithm affect the
`MOT17/multiedge_inference_bench` example. First, a tautological `assert` in
`basemodel.py` never fires under any circumstance, leaving CPU-only users to encounter
a confusing late-stage `RuntimeError` instead of an immediate, descriptive message.
Second, an indentation error in `byte_track_algorithm.yaml` causes inconsistent YAML
parser behavior that may silently drop the `batch_size` hyperparameter or raise a
`ScannerError` depending on the PyYAML version in use.

### Environment

- **Files affected:**
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/tracking/byte_track/basemodel.py`
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/tracking/byte_track/byte_track_algorithm.yaml`

---

#### Finding 1 — Tautological assertion: `assert num_gpu <= torch.cuda.device_count()`

**File:** `testalgorithms/tracking/byte_track/basemodel.py`  
**Lines:** ~85–86  
**Code:**
```python
num_gpu = torch.cuda.device_count()          # line N
assert num_gpu <= torch.cuda.device_count()  # line N+1
self.is_distributed = num_gpu > 1
```

**Problem:**  
`num_gpu` is assigned from `torch.cuda.device_count()` on line N, then immediately
checked against the same function call on line N+1. The condition `n <= n` is always
`True`, so this assertion never fires regardless of the environment. On a CPU-only
machine `num_gpu = 0` and the assertion still passes, deferring the failure to the
`.cuda()` call several lines later:

```
RuntimeError: CUDA is not available
```

This late failure gives users no indication of which precondition failed. The assertion
was almost certainly intended to be a minimum-GPU guard, e.g.:
```python
assert num_gpu >= 1, "No CUDA-capable GPU detected. This example requires at least one GPU."
```

**Impact:**  
CPU-only users receive a cryptic `RuntimeError` deep in the stack rather than a clear
early diagnostic. Debugging time increases for users who set up the environment without
a GPU, a common scenario during initial development and CI.

**Proposed Fix:**
```python
# Replace the tautological assertion with a meaningful guard:
num_gpu = torch.cuda.device_count()
assert num_gpu >= 1, (
    f"No CUDA-capable GPU detected (torch.cuda.device_count() = {num_gpu}). "
    "The ByteTrack tracking job requires at least one CUDA GPU."
)
self.is_distributed = num_gpu > 1
```

---

#### Finding 2 — Malformed YAML indentation drops `batch_size` hyperparameter

**File:** `testalgorithms/tracking/byte_track/byte_track_algorithm.yaml`  
**Lines:** 24–27  
**Code (current — broken):**
```yaml
      hyperparameters:
        # name of the hyperparameter; string type;
          - batch_size:          # ← 10-space indent (wrong)
             values:
              - 1
```

**Reference (correct — from sibling `m3l_algorithm.yaml`):**
```yaml
      hyperparameters:
        - batch_size:            # ← 8-space indent (correct)
            values:
              - 256
```

**Problem:**  
The `- batch_size:` sequence item is indented with 10 spaces while the enclosing
`hyperparameters:` mapping uses 8-space indentation. This violates the YAML block
sequence indentation contract. The behavior is parser-version dependent:

- Some PyYAML versions raise `yaml.scanner.ScannerError` at parse time, blocking the
  benchmark run entirely.
- Others silently interpret `hyperparameters` as an empty mapping, dropping `batch_size`
  and causing ByteTrack to run with an unintended default value instead of the
  configured `1`.

**Impact:**  
In the silent-drop case, the algorithm runs with a different batch size than
documented, making results non-reproducible relative to the stated configuration.
In the scanner error case, the benchmark fails before any inference begins.
Either way, the misconfiguration is invisible without explicitly validating the YAML.

**Proposed Fix:**
```yaml
      hyperparameters:
        - batch_size:
            values:
              - 1
```

Validate before committing:
```bash
python -c "import yaml; print(yaml.safe_load(open('byte_track_algorithm.yaml')))"
```

---

### Steps to Reproduce

**Finding 1 (tautological assertion):**
1. Set up ianvs on a machine without a CUDA GPU.
2. Run: `ianvs -f examples/MOT17/multiedge_inference_bench/pedestrian_tracking/tracking_job.yaml`
3. Observe that the assertion on line ~86 does not fire. The first error is a
   `RuntimeError: CUDA is not available` inside a `.cuda()` call, with no reference
   to the GPU check.

**Finding 2 (YAML indentation):**
1. Run:
   ```bash
   python -c "
   import yaml
   cfg = yaml.safe_load(open('examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/tracking/byte_track/byte_track_algorithm.yaml'))
   print(cfg['algorithm']['modules'][0].get('hyperparameters'))
   "
   ```
2. Observe either a `ScannerError` or `None`/empty output for `hyperparameters`,
   confirming the parameter is lost.

---

### Additional Notes

These two bugs are grouped because both originate in the ByteTrack algorithm
configuration layer and can be fixed in the same PR with minimal risk. Neither change
touches the ianvs core.

The YAML indentation in `m3l_algorithm.yaml` is correct and serves as the reference
format for how `hyperparameters` should be structured.

rajdhruvsingh:
## What this PR does
Fixes three bugs in the ByteTrack tracking algorithm files,
addressing issues #443 and #444.

## Fix 1 — Missing DDP and dist imports (#443)
`torch.nn.parallel.DistributedDataParallel` (DDP) and 
`torch.distributed` (dist) were used in `load()` and 
`_get_eval_loader()` respectively but never imported.

Both NameErrors are completely silent on single-GPU and 
CPU machines since the distributed code paths are gated 
behind `self.is_distributed = num_gpu > 1`. On any 
multi-GPU machine — precisely the environment the 
multiedgeinference paradigm is designed for — both 
cause immediate NameError crashes.

Added:
  import torch.distributed as dist
  from torch.nn.parallel import DistributedDataParallel as DDP

## Fix 2 — Tautological assertion replaced (#444)
The original assertion:
  num_gpu = torch.cuda.device_count()
  assert num_gpu <= torch.cuda.device_count()

always passes since num_gpu was assigned from the same 
call. On CPU-only machines num_gpu=0 and the assertion 
still passes, deferring failure to a cryptic 
RuntimeError deep in the stack.

Replaced with a meaningful guard:
  assert num_gpu >= 1, "No CUDA-capable GPU detected..."

Gives CPU-only users a clear early diagnostic instead 
of a confusing late-stage RuntimeError.

## Fix 3 — YAML indentation and stale path (#444)
byte_track_algorithm.yaml had two issues:

1. batch_size indented with 10 spaces instead of 8,
   causing ScannerError or silent drop of the 
   hyperparameter depending on PyYAML version.
   Fixed to match the correct format used in the
   sibling m3l_algorithm.yaml.

2. Stale url pointing to:
   ./examples/pedestrian_tracking/multiedge_inference_bench/...
   Corrected to:
   ./examples/MOT17/multiedge_inference_bench/pedestrian_tracking/...

## Testing
Verified on macOS, Python 3.14, PyYAML 6.0.3:
- dist import present ✅
- DDP import present ✅
- tautological assert removed ✅
- meaningful assert num_gpu >= 1 added ✅
- YAML parses correctly, batch_size=1 accessible ✅
- url path contains MOT17 ✅

Note: Full multi-GPU execution not testable on Mac.
Fixes are import-level and config-level with no 
logic modification to inference paths.

Fixes #443
Fixes #444

### DIFF
--- a/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/tracking/byte_track/basemodel.py
+++ b/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/tracking/byte_track/basemodel.py
@@ -85,10 +85,11 @@ class BaseModel:
         self.args.experiment_name = self.exp.exp_name
 
         num_gpu = torch.cuda.device_count()
-        assert num_gpu >= 1, (
-            f"No CUDA-capable GPU detected (torch.cuda.device_count() = {num_gpu}). "
-            "The ByteTrack tracking job requires at least one CUDA GPU."
-        )
+        if num_gpu < 1:
+            raise RuntimeError(
+                f"No CUDA-capable GPU detected (torch.cuda.device_count() = {num_gpu}). "
+                "The ByteTrack tracking job requires at least one CUDA GPU."
+            )
         self.is_distributed = num_gpu > 1
         # set environment variables for distributed training
         cudnn.benchmark = True
--- a/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/tracking/byte_track/basemodel.py
+++ b/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/tracking/byte_track/basemodel.py
@@ -20,6 +20,8 @@ from pathlib import Path
 
 import torch
 import torch.backends.cudnn as cudnn
+import torch.distributed as dist
+from torch.nn.parallel import DistributedDataParallel as DDP
 import motmetrics as mm
 from loguru import logger
 from sedna.common.class_factory import ClassType, ClassFactory
@@ -83,7 +85,10 @@ class BaseModel:
         self.args.experiment_name = self.exp.exp_name
 
         num_gpu = torch.cuda.device_count()
-        assert num_gpu <= torch.cuda.device_count()
+        assert num_gpu >= 1, (
+            f"No CUDA-capable GPU detected (torch.cuda.device_count() = {num_gpu}). "
+            "The ByteTrack tracking job requires at least one CUDA GPU."
+        )
         self.is_distributed = num_gpu > 1
         # set environment variables for distributed training
         cudnn.benchmark = True
--- a/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/tracking/byte_track/byte_track_algorithm.yaml
+++ b/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/tracking/byte_track/byte_track_algorithm.yaml
@@ -17,11 +17,10 @@ algorithm:
       # example: basemodel.py has BaseModel module that the alias is "FPN" for this benchmarking;
       name: "ByteTrack"
       # the url address of python module; string type;
-      url: "./examples/pedestrian_tracking/multiedge_inference_bench/testalgorithms/tracking/byte_track/basemodel.py"
+      url: "./examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/tracking/byte_track/basemodel.py"
 
       # hyperparameters configuration for the python module; list type;
       hyperparameters:
-        # name of the hyperparameter; string type;
-          - batch_size:
-             values:
+        - batch_size:
+            values:
               - 1


aryananand-04:
The missing `DDP` and `dist` imports at lines 115 and 176 are still absent on `main` and present here, so that part is straightforward.

One discrepancy between the description and the diff, in the PR's favour. The body describes the guard as:

```python
assert num_gpu >= 1
```

but what actually ships is:

```python
if num_gpu < 1: raise RuntimeError(...)
```

That's the better version, and worth keeping deliberately rather than as an accident of review: `assert` statements are stripped out entirely under `python -O`, so an assert-based guard would quietly vanish in an optimised build and you'd be back to whatever failure the guard was there to prevent. The `raise` doesn't have that problem. Might be worth updating the description so nobody "fixes" it back to an assert later.

One thing I'd flag for a maintainer rather than try to settle myself. This makes at least one GPU a hard requirement, and raises if there isn't one. Other open work on the examples is moving the opposite way, toward running on CPU or MPS when no GPU is present. Both positions are defensible, and which one applies here probably depends on whether this example is meant to be runnable for someone evaluating Ianvs on a laptop. That's a policy call about the example's audience, not a defect in this diff, so I'm raising it rather than requesting a change.

Verified by reading the diff and the surrounding call sites on `main`. I don't have multi-GPU hardware, so I haven't executed the distributed path.

PR #532, opened 2026-06-11
feat: add opt-in parallel test case execution via ThreadPoolExecutor

### LINKED ISSUE #8: Not supported parallel processing of multiple use cases yet
It seems that Ianvs has not supported parallelly testing several groups of parameters.

**Why is this needed**:
Each use case spends the most of the time on training process. When a user wants to test several groups of parameters, serial training will inccur unbearable time overhead.

rajdhruvsingh:
<h2>What type of PR is this?</h2>
<p>/kind feature
/kind design</p>
<h2>What this PR does / why we need it</h2>
<p>Resolves the unbearable time overhead when testing multiple hyperparameter groups sequentially (<a href="https://github.com/kubeedge/ianvs/issues/8">Issue #8</a>). When a user tests 5 parameter groups on a 30-minute model, current serial execution requires 2.5 hours. With <code>parallel: true</code>, threads run concurrently — meaningful speedup for I/O-bound and GPU-offloaded workloads (see <a href="https://claude.ai/chat/65e6aa7a-2d22-4b32-88ec-362f845f9a82#gil-note">GIL limitations</a> below).</p>
<h2>Relationship to PR #526 (Sandbox Engine)</h2>
<p>PR #526 is now merged into <code>main</code> as a formal architectural proposal. This PR is <strong>complementary, not competing</strong>:</p>

Aspect | This PR (Phase 1) | PR #526 (Phase 2 — Sandbox Engine)
-- | -- | --
Status | Ready to merge | Proposal merged; implementation pending
Platform | Linux, macOS, Windows | Linux primary
New dependencies | None | uv, psutil
Execution model | Threads (ThreadPoolExecutor) | Processes / containers
Dependency isolation | No | Yes
OOM protection | No (safe default workers) | Yes (prlimit/cgroups)
Best for | I/O-bound, GPU-offloaded, same environment | Conflicting deps, resource isolation

<h2>GIL note</h2>
<p><code>ThreadPoolExecutor</code> provides meaningful speedup for <strong>I/O-bound</strong> (dataset loading, result writing) and <strong>GPU-offloaded</strong> workloads where the GIL is released during compute. For pure CPU-bound training, the GIL limits concurrency — those workloads should wait for Phase 2 (Sandbox Engine with process isolation). This limitation is documented explicitly in the RFC and in the <code>run_testcases()</code> docstring.</p>
<h2>Changes</h2>
<h3><code>core/testcasecontroller/testcasecontroller.py</code></h3>
<ul>
<li>Added <code>_run_testcases_sequential()</code> — preserves exact original behavior</li>
<li>Added <code>_run_testcases_parallel()</code> — concurrent execution via <code>ThreadPoolExecutor</code></li>
<li><code>max_workers</code> defaults to <code>None</code> so <code>ThreadPoolExecutor</code> uses its own safe default (<code>min(32, os.cpu_count() + 4)</code>) — avoids OOM from unbounded concurrency on resource-intensive ML workloads</li>
<li>YAML string coercion: <code>parallel: "true"</code> / <code>"false"</code> / <code>"yes"</code> / <code>"1"</code> all handled correctly</li>
<li><code>max_workers</code> validated before use: non-integer strings and values ≤ 0 raise <code>ValueError</code> with a clear message</li>
<li>Deep copy of each test case before thread submission — threads cannot mutate shared <code>test_env</code> or <code>dataset</code> state</li>
<li>Failed test cases collected and reported together, not one at a time</li>
</ul>
<h3><code>core/testcasecontroller/testcase/testcase.py</code></h3>
<ul>
<li>Fixed race condition in <code>_get_output_dir()</code>: original <code>os.path.exists()</code> loop allowed two concurrent threads to collide on the same directory. Fixed with <code>os.makedirs(exist_ok=True)</code> — atomic, and beneficial even for sequential execution</li>
</ul>
<h3><code>core/cmd/obj/benchmarkingjob.py</code></h3>
<ul>
<li>Added <code>parallel: bool = False</code> and <code>max_workers: Optional[int] = None</code> fields</li>
<li>Both auto-populated from <code>benchmarkingjob.yaml</code> via existing <code>_parse_config()</code></li>
<li>Passed to <code>run_testcases()</code> in <code>run()</code></li>
</ul>
<h3><code>core/testcasecontroller/tests/test_parallel_execution.py</code></h3>
<ul>
<li>21 unit tests across 7 test classes</li>
<li>Covers: sequential correctness, parallel correctness, output directory uniqueness under 20 real concurrent threads, YAML string coercion, <code>max_workers</code> validation, partial failure reporting, deep copy isolation</li>
</ul>
<h3><code>docs/proposals/chore/parallel_testcase_execution.md</code></h3>
<ul>
<li>Honest GIL limitation documentation with workload guidance</li>
<li>Mermaid execution flow diagram</li>
<li>Paradigm compatibility table referencing PR #526 §9.2</li>
<li>Updated Phase 2 section to reflect PR #526 merged into main</li>
<li>Explicit entry point compatibility note for Sandbox Engine integration</li>
</ul>
<h2>Usage</h2>
<pre><code class="language-yaml">benchmarkingjob:
  name: "benchmarkingjob"
  workspace: "./workspace"
  parallel: true       # enable parallel execution (default: false)
  max_workers: 4       # optional — defaults to ThreadPoolExecutor safe default
  testenv: "..."
  test_object: ...
</code></pre>
<h2>Backward compatibility</h2>
<p><code>parallel</code> defaults to <code>false</code> — all existing <code>benchmarkingjob.yaml</code> files continue to work without any modification. Zero impact on existing users until they explicitly opt in.</p>
<h2>Testing</h2>
<pre><code>21 passed in 1.22s — Python 3.14.3, pytest 9.1.1, macOS
</code></pre>
<ul>
<li>✅ Sequential execution with empty / single / multiple test cases</li>
<li>✅ Parallel execution: all succeed, all fail, partial failure</li>
<li>✅ Output directory uniqueness under 20 concurrent threads</li>
<li>✅ YAML string coercion (<code>"true"</code>, <code>"false"</code>, <code>"yes"</code>, <code>"1"</code>)</li>
<li>✅ <code>max_workers</code> validation (zero, negative, invalid string, None)</li>
<li>✅ Deep copy isolation — original <code>test_cases</code> list not mutated</li>
</ul>
<p>Fixes #8</p>

### DIFF
--- a/core/cmd/obj/benchmarkingjob.py
+++ b/core/cmd/obj/benchmarkingjob.py
@@ -48,7 +48,7 @@ class BenchmarkingJob:
         self.test_env = None
         self.simulation = None
         self.parallel: bool = False
-        self.max_workers: int = None
+        self.max_workers: int|None = None
         self.testcase_controller = TestCaseController()
         self._parse_config(config)
 
--- a/core/cmd/obj/benchmarkingjob.py
+++ b/core/cmd/obj/benchmarkingjob.py
@@ -16,6 +16,7 @@
 """BenchmarkingJob"""
 
 import os
+from typing import Optional
 
 from core.common import utils
 from core.common.constant import TestObjectType
@@ -26,18 +27,17 @@ from core.testcasecontroller.simulation_system_admin import build_simulation_env
 from core.testcasecontroller.testcasecontroller import TestCaseController
 
 
-# pylint: disable=too-few-public-methods
+# pylint: disable=too-few-public-methods,too-many-instance-attributes
 class BenchmarkingJob:
     """
     BenchmarkingJob:
-        providing a end-to-end benchmarking job.
+        Providing an end-to-end benchmarking job.
 
     Parameters
     ----------
-    config: dict
-        config of a end-to-end benchmarking job,
+    config : dict
+        Config of an end-to-end benchmarking job,
         includes: test env, algorithms, rank setting, etc.
-
     """
 
     def __init__(self, config):
@@ -47,39 +47,49 @@ class BenchmarkingJob:
         self.rank = None
         self.test_env = None
         self.simulation = None
+        self.parallel: bool = False
+        self.max_workers: Optional[int] = None
         self.testcase_controller = TestCaseController()
         self._parse_config(config)
 
     def _check_fields(self):
         if not self.name and not isinstance(self.name, str):
-            raise ValueError(f"benchmarkingjob's name({self.name}) must be provided"
-                             f" and be string type.")
+            raise ValueError(
+                f"benchmarkingjob's name({self.name}) must be provided"
+                f" and be string type."
+            )
 
         if not isinstance(self.workspace, str):
-            raise ValueError(f"benchmarkingjob's workspace({self.workspace}) must be string type.")
+            raise ValueError(
+                f"benchmarkingjob's workspace({self.workspace}) must be string type."
+            )
 
         if not self.test_object and not isinstance(self.test_object, dict):
-            raise ValueError(f"benchmarkingjob's test_object({self.test_object})"
-                             f" must be dict type.")
+            raise ValueError(
+                f"benchmarkingjob's test_object({self.test_object})"
+                f" must be dict type."
+            )
 
         test_object_types = [e.value for e in TestObjectType.__members__.values()]
         test_object_type = self.test_object.get("type")
         if test_object_type not in test_object_types:
             raise ValueError(
                 f"benchmarkingjob' test_object doesn't support the type({test_object_type}), "
-                f"the following test object types can be selected: {test_object_types}.")
+                f"the following test object types can be selected: {test_object_types}."
+            )
 
         if not self.test_object.get(test_object_type):
-            raise ValueError(f"benchmarkingjob' test_object doesn't find"
-                             f" the field({test_object_type}).")
+            raise ValueError(
+                f"benchmarkingjob' test_object doesn't find the field({test_object_type})."
+            )
 
     def run(self):
         """
-        run a end-to-end benchmarking job,
-        includes prepare test env,
-                 run all test cases,
-                 save results of all test cases,
-                 plot the results according to the visualization config of rank.
+        Run an end-to-end benchmarking job:
+        - prepare test env
+        - run all test cases (sequentially or in parallel)
+        - save results of all test cases
+        - plot results according to the rank visualization config
         """
         self.workspace = os.path.join(self.workspace, self.name)
 
@@ -88,10 +98,16 @@ class BenchmarkingJob:
 
         self.test_env.prepare()
 
-        self.testcase_controller.build_testcases(test_env=self.test_env,
-                                                 test_object=self.test_object)
+        self.testcase_controller.build_testcases(
+            test_env=self.test_env,
+            test_object=self.test_object
+        )
 
-        succeed_testcases, test_results = self.testcase_controller.run_testcases(self.workspace)
+        succeed_testcases, test_results = self.testcase_controller.run_testcases(
+            self.workspace,
+            parallel=self.parallel,
+            max_workers=self.max_workers
+        )
 
         if test_results:
             self.rank.save(succeed_testcases, test_results, output_dir=self.workspace)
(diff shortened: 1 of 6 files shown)


rajdhruvsingh:
## PR Summary

### What problem this solves and how it was found

Issue #8 has been open since the early days of Ianvs and documents a straightforward but painful problem: when a user wants to benchmark multiple hyperparameter groups (different learning rates, batch sizes, model architectures), every test case runs sequentially inside a single `for` loop in `TestCaseController.run_testcases()`. Testing 5 parameter groups on a 30-minute model requires 2.5 hours of wall-clock time even when the machine has sufficient resources to run them concurrently. The issue was found by inspecting `core/testcasecontroller/testcasecontroller.py` — the core sequential loop had no parallelism mechanism and no configuration option to enable concurrent execution.

---

### Where the problem existed

Three locations in the codebase:

**1. `core/testcasecontroller/testcasecontroller.py`** — the primary location. `run_testcases()` iterated over all test cases in a blocking `for` loop with no opt-out. One test case had to fully complete before the next could start.

**2. `core/testcasecontroller/testcase/testcase.py`** — a latent race condition in `_get_output_dir()`. The original implementation used an `os.path.exists()` check inside a `while` loop to find a free directory. Under any concurrent execution, two threads can both pass the existence check before either creates the directory, causing a collision. This bug was silent in sequential mode and would have caused failures the moment any parallelism was introduced.

**3. `core/cmd/obj/benchmarkingjob.py`** — no mechanism to pass parallelism configuration from `benchmarkingjob.yaml` down to `run_testcases()`.

---

### What was fixed and what architectural changes were made

**`testcasecontroller.py`:**
- `run_testcases()` now accepts `parallel: bool = False` and `max_workers: Optional[int] = None`
- Routes to `_run_testcases_sequential()` (original behavior, default) or `_run_testcases_parallel()` (new)
- `_run_testcases_parallel()` uses `concurrent.futures.ThreadPoolExecutor`. `ThreadPoolExecutor` is chosen over `ProcessPoolExecutor` because ML model objects (PyTorch, TensorFlow, Sedna paradigms) are often not picklable — a hard requirement for process-based parallelism. Threads share memory space, eliminating serialization entirely.
- `max_workers` defaults to `None`, which causes `ThreadPoolExecutor` to use its own safe default (`min(32, os.cpu_count() + 4)`). This deliberately avoids the OOM risk of spawning one thread per test case when a user has many parameter groups and resource-intensive models.
- Each test case is `copy.deepcopy()`'d before submission to the thread pool. This prevents race conditions from shared `test_env` and `dataset` state mutations across threads — each thread gets a fully independent copy.
- Failed test cases are collected and reported together rather than stopping at first failure, giving users a complete picture of which parameter groups failed.
- YAML string coercion handles `parallel: "true"` / `"false"` / `"yes"` / `"1"` correctly since YAML parsers can return strings for boolean fields.
- `max_workers` is validated before use — non-integer strings and values ≤ 0 raise `ValueError` with a clear message.

**`testcase.py`:**
- `_get_output_dir()` rewritten to use `os.makedirs(exist_ok=True)` with the already-unique `uuid1()` id. This is atomic and eliminates the race condition. Also simplifies the code — the `while` loop was unnecessary since `uuid1()` guarantees uniqueness.

**`benchmarkingjob.py`:**
- Added `parallel: bool = False` and `max_workers: Optional[int] = None` fields, auto-populated from `benchmarkingjob.yaml` via the existing `_parse_config()` mechanism and passed to `run_testcases()`.

**Tests (`core/testcasecontroller/tests/test_parallel_execution.py`):**
- 21 unit tests at pylint 10.00/10 covering: sequential correctness, parallel correctness, output directory uniqueness under 20 real concurrent threads, YAML string coercion, `max_workers` validation, partial failure reporting, and deep copy isolation.

**RFC (`docs/proposals/chore/parallel_testcase_execution.md`):**
- Documents GIL limitations honestly with workload guidance
- Mermaid execution flow diagram
- Paradigm compatibility table referencing PR #526 §9.2
- Phase 1 vs Phase 2 comparison table
- Migration path to Sandbox Engine

---

### Side effects

**Positive side effects:**
- The `_get_output_dir()` race condition fix benefits sequential execution too — it removes an unnecessary `while` loop and makes directory creation cleaner regardless of parallelism mode.
- Failed test cases in parallel mode are now reported together with full context, which is strictly better than the current sequential behavior of stopping at the first failure.

**Known limitations (documented):**
- Python's GIL limits true CPU parallelism for pure Python threads. Phase 1 provides meaningful speedup for I/O-bound (dataset loading, result writing) and GPU-offloaded workloads where the GIL is released during compute. For pure CPU-bound training, the GIL limits concurrency — those workloads will benefit more from Phase 2.
- No dependency isolation between test cases — if two algorithms require conflicting package versions, they still share the same Python environment. This is a Phase 2 problem.
- No OOM protection — a single test case that exhausts memory can still affect others. The safe `max_workers` default mitigates this but does not eliminate it.

**Zero negative side effects on existing behavior:**
- `parallel` defaults to `false` — all existing `benchmarkingjob.yaml` files continue to work without any modification. The sequential code path is extracted into `_run_testcases_sequential()` which preserves the exact original behavior byte-for-byte.

PR #534, opened 2026-06-12
fix(llm-agent): device placement crash on MPS and silently ignored half_lora flag

### LINKED ISSUE #533: llm-agent example: inference crashes on MPS, and half_lora is silently ignored (masking broken fp16 training)
**What happened**:

I ran `examples/llm-agent/singletask_learning_bench` end-to-end following its README (macOS ARM64, Python 3.12, torch with MPS). Training completes (20 epochs, train_loss 1.39), but inference crashes:

```
RuntimeError: Placeholder storage has not been allocated on MPS device!
```

The failure is in `predict()` at `testalgorithms/basemodel.py:76`, inside `self.model.generate(**inputs, ...)`. Transformers warns right before the crash:

```
UserWarning: You are calling .generate() with the `input_ids` being on a device type
different than your model's device. `input_ids` is on cpu, whereas the model is on mps.
```

**Root cause 1 — inputs never moved to the model device**

`predict()` tokenizes with `return_tensors="pt"` but leaves the tensors on CPU, while the model is loaded with `device_map` from `config.json` (`"auto"`). On CUDA, accelerate's hooks happen to paper over this; on MPS it hard-crashes. One-line fix:

```python
inputs = self.tokenizer(prompt, return_tensors="pt", truncation=True,
                        max_length=self.MAX_LENGTH).to(self.model.device)
```

**Root cause 2 (found while debugging) — `half_lora` is a string, so fp16 never runs, and the fp16 path is broken anyway**

`train_config.json` sets `"half_lora": "True"` (a string). The code checks `if half==True:`, which is always False for a string, so `model.half()` never executes — but `adam_epsilon=(1e-4 if half else 1e-8)` treats the non-empty string as truthy, so the fp16-tuned epsilon is silently applied to fp32 training.

Worse: if you fix the comparison so fp16 actually engages, training silently collapses — `model.half()` without loss scaling underflows the gradients:

| run | half active | train_loss | rouge1 / rougeL |
|-----|-------------|------------|-----------------|
| fp32 (current effective behavior) | no | 1.39 | 10.0 / 10.0 (matches README) |
| fp16 via fixed comparison | yes | 0.0 | 0.0 / 0.0 |

So the string bug has been masking a numerically broken fp16 path on every platform, not just macOS.

**Impact**:

- The example is unrunnable on Apple Silicon, although the README lists macOS as supported.
- Anyone fixing the `half_lora` comparison locally gets silent garbage results (loss 0, all metrics 0) with no error — the worst failure mode for a benchmark.

**Proposed fix** (example-only, no ianvs core changes):

1. Move tokenized inputs to `self.model.device` in `predict()`.
2. Parse `half_lora` as a real boolean and default it to `false`, with a note that proper fp16 should go through `TrainingArguments(fp16=True)` on CUDA.

I have both fixes working locally, verified against the README's expected leaderboard. PR incoming.

**Environment**: macOS ARM64, Python 3.12.13, ianvs @ bf0f596, deps per the example's requirements.txt.

veyron-kairo:
Fixes #533

Two fixes for `examples/llm-agent/singletask_learning_bench`:

1. `predict()` tokenized inputs on CPU while the model loads via `device_map`, which crashes generation on Apple Silicon with `Placeholder storage has not been allocated on MPS device!`. Inputs now move to `self.model.device`.

2. `half_lora` in train_config.json was the string `"True"`, which never passes the `half==True` check — fp16 never ran, while the truthy string still selected the fp16 `adam_epsilon`. Fixing only the comparison makes training collapse (train_loss 0, all rouge 0: `.half()` without loss scaling underflows). The flag is now parsed as a real boolean and defaults to `false` so the config matches actual behavior.

Verified on macOS ARM64 / Python 3.12: the benchmark runs end-to-end and reproduces the README leaderboard.

| run | result |
|-----|--------|
| main, MPS | crash in `predict()` |
| fixed, fp32 | rouge1/rougeL 10.0, train_loss ~1.1 — matches README |
| fp16 enabled (for reference) | train_loss 0, rouge 0 — why the default is `false` |


aryananand-04:
I have MPS on this machine, so I ran both of these rather than reading the diff.

The device placement crash reproduces on `main` and is fixed here:

```
main:   -> RuntimeError: Placeholder storage has not been allocated on MPS device!
PR534:  -> OK, output (1, 5, 100) on mps:0
```

The second bug is the more interesting one. `half = 'True'` is a string, and it's tested with `if half == True`, which is `False` for any non-boolean, the string `'True'` included. So `.half()` never runs no matter what the config asks for.

One thing I noticed following that variable further: the same truthy check also picks which `adam_epsilon` to use. So on `main` you get fp32 training paired with the epsilon tuned for fp16 (`1e-4`). The precision bug is visible if you go looking for it, but that second effect isn't, and nothing crashes to tell you. Might be worth a line in the PR description, since it makes the case for this fix stronger than "half precision was ignored".

Credit where it's due on the write-up: the PR says plainly that fp16 makes training collapse, and that's why `half_lora` stays off by default even after the type coercion is fixed. That's a useful thing to disclose rather than bury. My only suggestion is on framing, since "fixes half_lora" reads at a glance like half precision now works, when what actually ships is that it's now correctly and deliberately off. A sentence making that explicit would save the next person the trip through the code.

Both bugs executed on real MPS hardware, macOS arm64. No CUDA path tested.

Issue #514, opened 2026-06-08
bug: robot-cityscapes-synthia benchmarkingjob.yaml and testenv.yaml contain stale paths and hardcoded absolute paths


rajdhruvsingh:
## Description
Two configuration files in robot-cityscapes-synthia contain 
stale paths pointing to class_increment_semantic_segmentation 
and hardcoded absolute paths specific to one developer's machine,
making the entire benchmark non-runnable for any other user.

## Affected Files
- benchmarkingjob.yaml (lines 9, 22)
- testenv/testenv.yaml (lines 16, 31 — stale paths)
- testenv/testenv.yaml (lines 5, 7 — hardcoded absolute paths)

## Evidence

benchmarkingjob.yaml:
  testenv: "./examples/class_increment_semantic_segmentation/..."
  url: "./examples/class_increment_semantic_segmentation/..."

testenv.yaml:
  train_url: "/home/QXY/dataset/mdil-ss/..."  ← hardcoded absolute path
  test_url: "/home/QXY/dataset/mdil-ss/..."   ← hardcoded absolute path
  url: "./examples/class_increment_semantic_segmentation/..."

## Proposed Fix
Update all stale paths to correct robot-cityscapes-synthia paths.
Replace hardcoded /home/QXY/ paths with portable relative paths.

## Related
Sister issue to #463 which fixed stale paths in test_algorithm.yaml
and sedna_evaluate.py/sedna_train.py.
Discovered via grep -rn "class_increment_semantic_segmentation" examples/

rajdhruvsingh:
I discovered this while investigating the stale path fixes 
in PR #513. I have already identified all affected paths 
and verified the correct locations locally.

I have a fix ready — would it be okay to open a PR for this?

diegodevxd:
There is a second configuration-contract risk beyond the stale directory and `/home/QXY` literals. The open repair #529 changes the values but keeps `dataset.train_url` and `dataset.test_url`. Current `core/testenvmanager/dataset/dataset.py` calls `_check_fields()` only for `train_index`, `test_index`, `train_data`, `test_data`, `train_data_info`, and `test_data_info`. Thus the path can be corrected while the current dataset-processing path still has no recognized index input.

[...]

PR #229, opened 2025-07-28
Added files for MOT17

NishantSinghhhhh:
### What type of PR is this?
/kind documentation
/kind feature
/kind cleanup

### What this PR does / why we need it:

This PR completely revamps the `examples/MOT17/multiedge_inference_bench` example to elevate it to a "gold standard" of runnable documentation. It directly addresses the problems of poor usability, manual setup, and lack of clarity outlined in issue #231 
The key improvements include:

PR #536, opened 2026-06-12
fix(government_rag): restore example on current LangChain and remove hardcoded credentials

### LINKED ISSUE #535: government_rag example cannot be executed: missing requirements, LangChain import failures, hardcoded credentials/device/paths
**What happened**:

I tried to run `examples/government_rag` following its README and it is currently impossible for a new user to execute.

**Problem 1 — no dependency manifest**

There is no `requirements.txt` and the README has no install instructions. The dependency set (langchain, chromadb, sentence-transformers, unstructured, …) has to be reverse-engineered from import errors one at a time.

**Problem 2 — import-time failure on current LangChain**

With current LangChain installed (`pip install langchain langchain-community`, tested on langchain 1.3.8 / langchain-community 0.4.2, Python 3.12):

```
FAIL : from langchain.text_splitter import RecursiveCharacterTextSplitter -> ModuleNotFoundError: No module named 'langchain.text_splitter'
FAIL : from langchain.chains import RetrievalQA -> ModuleNotFoundError: No module named 'langchain.chains'
```

`langchain.text_splitter` moved to `langchain_text_splitters`, and `RetrievalQA` is a legacy chain that no longer ships in langchain 1.x. Notably, `RetrievalQA` and `HuggingFacePipeline` are imported but never used — the example crashes on dead imports. `langchain-community` itself is sunset (langchain-ai/langchain-community#674), and the code also uses `vector_store.persist()` (removed in langchain-chroma) and the deprecated `get_relevant_documents()`.

**Problem 3 — hardcoded device, credentials and paths in `basemodel.py` / `gov_rag.py`**

- module-level `device = "cuda"`, no fallback — fails on CPU-only and Apple Silicon machines
- default backend is qianfan with literal placeholders in the OAuth URL: `client_id=[应用API Key]&client_secret=[应用Secret Key]`; the DeepSeek backend has `api_key="<DeepSeek API Key>"`, siliconflow has `Bearer <token>` — none configurable without editing source
- embedding model hardcoded to a personal path `/home/icyfeather/models/bge-m3`
- `base_path` defaults to the placeholder `/path/ianvs/dataset/gov_rag`

**Impact**:

The example is 100% non-runnable from a fresh checkout: import crash (P2), then guaranteed runtime failures (P3), with no dependency list to even get started (P1). RAG/LLM examples are exactly what new SIG-AI contributors reach for first, so this feeds the confused-user issue stream described in #230. The drift gets worse monthly since langchain-community is unmaintained.

**Fix** (example-only, no ianvs core changes):

1. Add `requirements.txt` with maintained packages (langchain-chroma, langchain-huggingface, langchain-text-splitters).
2. Migrate imports, drop the dead ones, replace `persist()` and `get_relevant_documents()`.
3. Auto-pick device (cuda/mps/cpu); read backend + API keys from env vars with clear errors; fix the embedding-model and base_path defaults.
4. README install/configuration section.

I have this working locally — knowledge-base build, persist reload, province filtering and retrieval all verified against a synthetic corpus (the Kaggle GovAff dataset itself is unchanged). PR incoming.

**Environment**: macOS ARM64, Python 3.12.13, ianvs @ bf0f596.

veyron-kairo:
Fixes #535

Restores `examples/government_rag` to a runnable state:

- **requirements.txt added** — the example previously had no dependency manifest at all.
- **LangChain migration** — `langchain.text_splitter` and `langchain.chains` no longer exist in langchain 1.x, so the example crashed at import. Imports now use the maintained `langchain-text-splitters` / `langchain-chroma` / `langchain-huggingface` packages; the unused `RetrievalQA` and `HuggingFacePipeline` imports (dead code) are removed, along with the removed `vector_store.persist()` call and deprecated `get_relevant_documents()`.
- **No more hardcoded config** — device is picked automatically (cuda/mps/cpu); the LLM backend (qianfan/deepseek/siliconflow) and its API keys come from environment variables with clear errors when missing, instead of placeholder strings in source; the embedding model no longer points at a personal home directory; a missing dataset now raises an error pointing at the Kaggle download.
- **README** — install and configuration steps added.

Verified on macOS ARM64 / Python 3.12 against a synthetic two-province corpus: knowledge-base build, reload from persisted Chroma store, province filtering, retrieval, and the env-validation errors. The Kaggle GovAff dataset and benchmark logic are unchanged.

<details><summary>smoke test output</summary>

```
device: mps
[1] build knowledge base from scratch (all provinces)
retrieved expected Shanghai doc ✔
[2] reload existing vector store from persist dir
reload + query ✔
[3] single-province knowledge base
province filter ✔
[4] missing dataset raises a clear error
clear error with Kaggle pointer ✔
ALL SMOKE TESTS PASSED
```
</details>


veyron-kairo:
Addressed all three review points in d2d062d:

- **Province filtering on persisted stores** — good catch, this also existed in the original code. Documents are now tagged with their province at ingestion and retrieval applies a `$in` filter on the selected provinces, so reloading the full persisted store no longer leaks documents across test modes. Verified with a synthetic corpus: a Beijing-restricted query against a reloaded all-province store no longer returns Shanghai documents. One consequence: embedding dirs generated before this change lack the tags and need regeneration (noted in the README).
- **DeepSeek client** — now created once and cached on the instance.
- **Qianfan token** — fetched once under a lock and reused across the parallel queries.

Hiteshsai007:
This PR correctly removes the hardcoded `device = "cuda"` sites in `government_rag` and introduces a local `default_device()` helper (cuda → mps → cpu). That is a clear improvement for CPU-only and Apple Silicon hosts.

**Recommendation: minor revision**

Two observations:

1. The new logic is entirely local to the example. It does not read `use_gpu` from `testenv.yaml` or call into the Core device-visibility mechanism (including the changes in #767). After both PRs merge, Ianvs will have two different ways of deciding the device.
2. This PR conflicts with #764 in `gov_rag.py`.

The local auto-detect is still a net positive for this specific example.

Hiteshsai007:
## Conflict with #764 and device-selection contract [STATIC ANALYSIS]

`default_device()` is a reasonable local improvement but establishes an independent device-selection path. After #767 makes Core correct, GovernmentRAG would still ignore Core's decision.

The province metadata tagging (`doc.metadata["province"] = province`) and retrieval filtering (`"filter": {"province": {"$in": self.provinces}}`) are important correctness fixes that #764 does not include — these must be preserved during conflict resolution.

**Suggestion:** Once #767 lands, adapt `default_device()` to read from the Core-provided device value when available, falling back to auto-detect only when no Core value is set.

Hiteshsai007:
### Correction to my earlier device-contract wording

`default_device()` does not directly consume `use_gpu`, but because #767 applies `CUDA_VISIBLE_DEVICES` before Example construction, its later `torch.cuda.is_available()` call may indirectly honor Core's CUDA constraint. This interaction was not independently verified on CUDA hardware.

My remaining concern is therefore not guaranteed incompatibility. It is that the contract is implicit and CUDA-specific, especially for MPS semantics, import-time initialization, explicit device strings, and non-PyTorch runtimes.

Separately, the province-leakage problem was identified by the existing automated review and fixed by the author in `d2d062d`. During reconciliation with open PR #764, the province metadata tagging and retrieval filter should be preserved.

yadavchiragg:
A few points worth checking against the actual diff before merge:

Package selection: if the requirements.txt pins langchain-community anywhere (even transitively via another package), that's worth flagging — it's the unmaintained package identified as the source of the import-drift problem in #535. Pinning langchain-chroma, langchain-huggingface, and langchain-text-splitters directly (the split-out successor packages) avoids reintroducing that failure mode on the next LangChain release.
Version pinning strategy: worth confirming whether dependencies use exact pins (==) vs. minimum bounds (>=). Given LangChain's fast-moving API surface (the whole reason this example broke), exact pins are probably the safer choice here even though they need periodic manual bumps — a loose bound risks the same import-crash recurring silently after a routine pip install --upgrade.
Scope relative to #535: the issue also reports hardcoded base_path (/path/ianvs/dataset/gov_rag placeholder) and hardcoded device/credentials. If this PR only adds the manifest and fixes the LangChain import calls, that's a legitimate and valuable partial fix — but worth being explicit in the PR description about which of #535's three problems (P1 manifest, P2 import crash, P3 hardcoded config) are in scope here vs. deferred, so the issue isn't prematurely closed as fully resolved when P3 may remain open.
Commit hygiene: per the pattern seen on PR #229 (closed after review feedback specifically asked for squashed commits before merge), worth squashing multi-commit history here proactively if it isn't already, since that's a recurring maintainer ask on this repo for dependency-manifest PRs specifically.

ijayhub:
Reviewed this as part of a cross-example investigation into #723, #514, and #535 (full analysis: #906).

This PR correctly removes the placeholder base_path default and the hardcoded personal embedding-model path from #535, replacing them with environment-variable-based configuration. I reviewed the diff but did not run the full example, since the LangChain/ChromaDB dependencies weren't available locally, so this is a static review.

It also overlaps with PR #764 on requirements.txt and gov_rag.py's imports, a conflict already noted by HerilMistry and @Hiteshsai007 in the comments on #535.

**Recommendation**: The path/credential fix looks correct, but the merge should wait until the conflict with #764 is resolved.

Hiteshsai007:
Thanks for listing these checks against the actual diff.

I agree that the final requirements should use the maintained split packages rather than reintroducing the deprecated umbrella dependency, and that the pinning strategy should be explicit given the history of LangChain import drift.

I also agree that the PR description should clearly separate the dependency/import fixes from the remaining hardcoded path, credential, and device issues in #535. Those should not be presented as resolved unless the diff actually covers them.

The overlap with #764 should be reconciled before merge, and the final history should be kept clean and easy to review.

yadavchiragg:
Update: per HerilMistry/Hiteshsai007's earlier comments on #535 and ijayhub's review on #536, this #536/#764 overlap is a broader confirmed finding, not just my note above.

yadavchiragg:
Agreed — happy to see this converge. Good luck with the device-selection/portability review.

christinaliyevav-coder:
I reviewed this PR in connection with the external-service prerequisite problem I found in `government_rag`.

The credential handling is clearly better than the original implementation. Missing credentials now produce an explicit error through `require_env()` instead of allowing placeholder credentials to reach the external API and causing a later secondary failure.

However, I found that the validation still happens quite late.

With Qianfan selected and both required environment variables removed, `BaseModel.__init__()` still succeeds:

```text
BaseModel __init__: SUCCESS
Selected backend: get_model_response_qianfan
QIANFAN_API_KEY present: False
QIANFAN_SECRET_KEY present: False
```

Tracing the call order shows:

```text
BaseModel initialization
→ backend selected
→ preprocess()
→ _make_rag()
→ dataset / embedding / knowledge-base initialization
→ process_query()
→ get_model_response_qianfan()
→ require_env(...)
→ missing credential detected
```

So the PR improves the failure message, but benchmark preparation can already begin before the program discovers that a mandatory prerequisite for the selected backend is missing.

I considered whether this could simply be intentional lazy validation so unused backends do not require credentials. However, the selected backend is already known during `BaseModel.__init__()`, so only the credentials required by that backend could be validated at that point.

I would keep the current `require_env()` approach, but validate the selected backend's required credentials during initialization or another explicit pre-execution validation stage.

I checked the existing review discussion before posting this. The existing comments cover province filtering, client/token reuse, device selection, dependencies and other restoration conflicts, but I did not find this specific lifecycle-ordering issue discussed.

yadavchiragg:
Worth flagging: per christinaliyevav-coder's review above, the credential check still happens late — BaseModel.__init__() succeeds even with required keys missing, since validation only fires once process_query() reaches require_env(). Since the backend is already known at init time, validating that backend's specific required credentials immediately would catch this before any benchmark setup work begins. Agree this pushes toward Major revision rather than Minor.

Hiteshsai007:
Thanks @christinaliyevav-coder and @yadavchiragg. I rechecked PR #536 at commit [`d2d062d`](https://github.com/kubeedge/ianvs/commit/d2d062d21533dc3ab2879a495d5ad4eeafa57c42) and agree that this is a distinct lifecycle issue.

`BaseModel.__init__()` selects `GOV_RAG_BACKEND`, but it does not validate the credentials required by that selected backend. The current `require_env()` calls are reached only later, after benchmark setup can already proceed through `preprocess()` and `_make_rag()`. Therefore, a missing credential may not be reported until the first external API request is made.

I will treat this as a **major-revision** item. The existing `require_env()` checks should remain as defense in depth, but the selected backend should also have an explicit preflight validation step during initialization or before benchmark execution:

- `qianfan`: `QIANFAN_API_KEY` and `QIANFAN_SECRET_KEY`
- `deepseek`: `DEEPSEEK_API_KEY`
- `siliconflow`: `SILICONFLOW_API_KEY`

Credentials for non-selected backends should not be required. Tests should cover both missing and present credentials for each backend and verify that validation occurs before dataset, embedding, or knowledge-base setup.

I also rechecked the dependency changes. The PR includes `langchain-community` alongside `langchain-chroma`, `langchain-huggingface`, and `langchain-text-splitters`, while `gov_rag.py` still imports its document loaders from `langchain_community`. The final dependency strategy should therefore explicitly justify that package and use a compatible pinning or upper-bound strategy rather than removing it blindly.

Finally, the overlap with PR #764 still needs to be reconciled before merge. The resolution must preserve PR #536's province metadata and retrieval-filtering changes together with the dependency and import corrections from both PRs.

My bounded smoke test exercised the direct `GovernmentRAG.__init__` path, but not the `BaseModel.preprocess()`/`_make_rag()` caller, so that integration should still receive a production-path test.

KiranUdayakumar:
Reviewed as part of a submitted Example Restoration pre-test analysis (full write-up: https://github.com/kubeedge/ianvs/discussions/963).

**What this PR does well:** correctly removes the hardcoded `device="cuda"` assignment, adds the missing `requirements.txt`, and replaces hardcoded credentials with `require_env()`-based configuration that raises explicit errors on missing variables, a genuine, substantial improvement over Issue #535's broken state.

**Verdict: major revision required, not rejection.** The main outstanding issue, already raised in this thread by christinaliyevav-coder and independently confirmed by yadavchiragg, is that credential validation still happens too late: with the Qianfan backend selected and its required environment variables removed, `BaseModel.__init__()` completes successfully because backend selection, `preprocess()`, and `_make_rag()` all run before `require_env()` is reached inside `process_query()`. Hiteshsai007's proposed resolution (explicit preflight validation per backend at initialization time, keeping `require_env()` as defense in depth) looks like the right fix.

Also outstanding: the unjustified retained `langchain-community` dependency, and reconciliation with PR #764, which modifies the same file's device-selection and import logic incompatibly (see comment on #764).

ANISHA-RAWAT:
Reviewed as part of a cross-example analysis connecting the hardcoded-path 
pattern in this PR (embedding model path, Problem 3 of #535) to the same 
pattern independently found in #514 and #723 — a repository-wide grep 
across examples/**/*.{py,yaml} shows 14 of 28 example directories affected, 
with no CI check catching it at merge time. This PR correctly fixes the 
hardcoded embedding path for this specific example, but per the existing 
review here (Hiteshsai007, christinaliyevav-coder, yadavchiragg), the 
#536/#764 conflict still needs reconciling before merge. Full cross-example 
analysis and a proposed CI-level fix: 
https://github.com/kubeedge/ianvs/discussions/967

PR #525, opened 2026-06-09
fix(examples): add missing BaseDataSource import in RFNet sedna_predict.py files

SurbhiD404:
## Summary

Fixes #522

Adds the missing `BaseDataSource` import across 4 RFNet `sedna_predict.py` files, following the same fix applied in #512 for the ERFNet path.

## Problem

`BaseDataSource` is used in `pre_data_process()` in all 4 files:

```python
data = BaseDataSource(data_type="test")
```

But only `IndexDataParse` was imported:

```python
from sedna.datasources import IndexDataParse
```

If this code path is executed and `BaseDataSource` is not provided by the runtime environment, Python is expected to raise:

`NameError: name 'BaseDataSource' is not defined`

## Fix

```python
# Before
from sedna.datasources import IndexDataParse

# After
from sedna.datasources import IndexDataParse, BaseDataSource
```

## Files Changed

* `examples/robot/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet/sedna_predict.py`
* `examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testalgorithms/rfnet/RFNet/sedna_predict.py`
* `examples/cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet/sedna_predict.py`
* `examples/cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/testalgorithms/rfnet/RFNet/sedna_predict.py`

## Validation

* [x] Verified `BaseDataSource` is used but missing from imports in all 4 files
* [x] Change is minimal ,only the missing import is added
* [x] DCO signed off

## Related

* Same fix applied to ERFNet path: #466 (PR #512)

PR #512, opened 2026-06-07
fix(examples): add missing BaseDataSource import in sedna_predict.py

SurbhiD404:
## What this PR does
Fixes #466

## Problem
`sedna_predict.py` uses `BaseDataSource` at line 46 inside 
`pre_data_process()`, but `BaseDataSource` was never imported.

The existing import only brought in `IndexDataParse`:
```python
from sedna.datasources import IndexDataParse
```
This caused a `NameError: name 'BaseDataSource' is not defined`
whenever `pre_data_process()` was called, making the prediction 
script completely unusable.

## Root Cause
The file was copied/adapted from another script and the 
`BaseDataSource` import was not carried over. Every other file 
in this example imports it correctly — for example 
`task_definition_by_domain.py`:
```python
from sedna.datasources import BaseDataSource
```

## Fix
Added `BaseDataSource` to the existing `sedna.datasources` import:
```python
from sedna.datasources import IndexDataParse, BaseDataSource
```

## Files Changed
- `examples/robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/testalgorithms/erfnet/ERFNet/sedna_predict.py`

PR #539, opened 2026-06-13
fix(examples): add missing BaseDataSource import to RFNet sedna_predict.py files

### LINKED ISSUE #522: [Bug] Missing BaseDataSource import in sedna_predict.py across multiple RFNet example paths
## Description

While reviewing the RFNet and ERFNet Sedna integration scripts, I noticed that several `sedna_predict.py` files reference `BaseDataSource` in `pre_data_process()` but only import `IndexDataParse` from `sedna.datasources`.

This appears inconsistent with other files in the repository that use `BaseDataSource` and explicitly import it.

## Affected Files

* `examples/robot/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet/sedna_predict.py`
* `examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testalgorithms/rfnet/RFNet/sedna_predict.py`
* `examples/cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet/sedna_predict.py`
* `examples/cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/testalgorithms/rfnet/RFNet/sedna_predict.py`

## Evidence

These files contain:

```python
from sedna.datasources import IndexDataParse
```

and later:

```python
data = BaseDataSource(data_type="test")
```

I could not find a local definition or import of `BaseDataSource` in these files.

Other RFNet/ERFNet modules in the repository explicitly import:

```python
from sedna.datasources import BaseDataSource
```

## Proposed Fix

Update the import statement:

```python
# Before
from sedna.datasources import IndexDataParse

# After
from sedna.datasources import IndexDataParse, BaseDataSource
```

## Notes

This issue was identified through code inspection and repository-wide searches while investigating similar missing-import patterns.
## Related

* Similar missing-import issue fixed in the `robot-cityscapes-synthia` path: #466 (PR #512)

## Potential Runtime Impact

`pre_data_process()` contains:

```python
data = BaseDataSource(data_type="test")
```

but BaseDataSource is not imported in the affected files.

If this code path is executed and BaseDataSource is not provided by the runtime environment, Python is expected to raise:

NameError: name 'BaseDataSource' is not defined

based on standard Python name resolution and import behavior.

Aamod007:
**What type of PR is this?**
/kind bug

This PR adds the missing `BaseDataSource` import to the four RFNet `sedna_predict.py` files.

Previously, these files instantiated `BaseDataSource(data_type="test")` in the `pre_data_process()` method but only imported `IndexDataParse` from `sedna.datasources`. If this execution path is hit without the import, it results in a `NameError: name 'BaseDataSource' is not defined`.

This commit adds `BaseDataSource` to the existing `sedna.datasources` import statement across the 4 affected files:

* `examples/robot/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet/sedna_predict.py`
* `examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testalgorithms/rfnet/RFNet/sedna_predict.py`
* `examples/cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet/sedna_predict.py`
* `examples/cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/testalgorithms/rfnet/RFNet/sedna_predict.py`

Fixes #522

- This supersedes #525 (byte-identical change on the same four files) so the same edit isn't merged twice.
- An earlier revision of this PR also touched the ERFNet file under `examples/robot-cityscapes-synthia/`; that path was already fixed on main by #512, and after rebasing onto current main that hunk dropped — this PR now covers exactly the four files still affected.
- Verification (pyflakes over all five vendored RFNet/ERFNet trees): pre-fix files report `undefined name 'BaseDataSource'`; post-fix files report no undefined names. No other runtime-relevant undefined symbols found in the ianvs-loaded entry points.

### DIFF
--- a/examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testalgorithms/rfnet/RFNet/sedna_predict.py
+++ b/examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testalgorithms/rfnet/RFNet/sedna_predict.py
@@ -18,7 +18,7 @@ from torchvision.transforms import ToPILImage
 from torchvision import transforms
 from torch.utils.data import DataLoader
 
-from sedna.datasources import IndexDataParse
+from sedna.datasources import IndexDataParse, BaseDataSource
 from sedna.core.lifelong_learning import LifelongLearning
 from sedna.common.config import Context
 
--- a/examples/cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet/sedna_predict.py
+++ b/examples/cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet/sedna_predict.py
@@ -18,7 +18,7 @@ from torchvision.transforms import ToPILImage
 from torchvision import transforms
 from torch.utils.data import DataLoader
 
-from sedna.datasources import IndexDataParse
+from sedna.datasources import IndexDataParse, BaseDataSource
 from sedna.core.lifelong_learning import LifelongLearning
 from sedna.common.config import Context
 
--- a/examples/cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/testalgorithms/rfnet/RFNet/sedna_predict.py
+++ b/examples/cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/testalgorithms/rfnet/RFNet/sedna_predict.py
@@ -18,7 +18,7 @@ from torchvision.transforms import ToPILImage
 from torchvision import transforms
 from torch.utils.data import DataLoader
 
-from sedna.datasources import IndexDataParse
+from sedna.datasources import IndexDataParse, BaseDataSource
 from sedna.core.lifelong_learning import LifelongLearning
 from sedna.common.config import Context
 
--- a/examples/robot/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet/sedna_predict.py
+++ b/examples/robot/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet/sedna_predict.py
@@ -18,7 +18,7 @@ from torchvision.transforms import ToPILImage
 from torchvision import transforms
 from torch.utils.data import DataLoader
 
-from sedna.datasources import IndexDataParse
+from sedna.datasources import IndexDataParse, BaseDataSource
 from sedna.core.lifelong_learning import LifelongLearning
 from sedna.common.config import Context
 


suhaan-24:
Static review. The fix is correct — `BaseDataSource` is referenced in these `sedna_predict.py`
files without being imported, so the module raises `NameError` on use — and the file coverage
here is the more complete of the two open PRs doing this.

## Relationship to #525

PR #525 makes the **byte-identical** change:

```diff
-from sedna.datasources import IndexDataParse
+from sedna.datasources import IndexDataParse, BaseDataSource
```

on four files, all of which are also in this PR. This PR additionally covers:

```
examples/robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/testalgorithms/erfnet/ERFNet/sedna_predict.py
```

So #525's file set is a strict subset of this one, with identical content on every shared file.
This PR supersedes it. Worth cross-linking the two so a maintainer can close #525 rather than
merging the same edit twice — neither PR currently references the other.

## One suggestion

Since this PR already establishes that the same missing import spans five vendored RFNet/ERFNet
trees, it would be worth confirming whether any other symbol has drifted the same way in those
files — the trees were copied from a common ancestor, so an import fix that is needed in five
places suggests other divergence may be too. Not a blocker for this change, which is correct as
it stands.

## Verification boundary

**Not executed.** The file lists and diff comparison are from `git diff --name-only` and
`git diff` against each branch's merge base. I have not run either branch; the `NameError` claim
is from reading the import against the usage in the same file.

Aamod007:
Cross-linking @SurbhiD404's #525 per the static review below: it applies the byte-identical one-line import fix to the same four files, so it can be closed in favor of this PR to avoid merging the same edit twice.

I also picked up both follow-ups from that review:

**1. Drift check executed.** Ran pyflakes over all five vendored RFNet/ERFNet trees (`examples/robot`, the three `cityscapes-synthia` variants, and `robot-cityscapes-synthia`):

* The five `sedna_predict.py` files: pre-fix versions flag `undefined name 'BaseDataSource'`; post-fix versions are clean. No other undefined names in these entry points — only unused-import leftovers inherited from upstream RFNet.
* `basemodel.py` (all five trees): `Process` / `exp_train` are undefined in the `if __name__ == '__main__'` block (~line 313). Not reachable through ianvs (the module is only imported by `ClassFactory`), but running `python basemodel.py` directly would raise `NameError`. Worth a separate follow-up issue.
* `erfnet/task_allocation_by_domain.py` and `task_definition_by_domain.py`: `__all__` declares `TaskAllocationByDomain` / `TaskDefinitionByDomain` while the classes are actually named `...ByOrigin`. Harmless at runtime because `ClassFactory.register(alias=...)` resolves by alias, but the names disagree.
* The `resnet18` "may be undefined" warnings in the vendored upstream `train.py`/`eval.py` are false positives (`resnet18` is defined in the star-imported module), and those scripts aren't invoked by ianvs anyway.

**2. Rebased onto current main.** The ERFNet hunk dropped automatically: #512 already landed that exact fix on main, so this PR is now precisely the four-file change still missing.

The hardcoded-path concern raised on #522 (`db_root_dir()`) is a separate bug — it's analyzed in discussion #923 and deserves its own issue/PR rather than expanding this one.

PR #547, opened 2026-06-14
fix: add guards against ZeroDivisionError across cloud-edge LLM metrics and backends

### LINKED ISSUE #337: Unguarded divisions crash with empty datasets in cloud-edge LLM example
**What happened**:

Multiple metric modules and model backends in the `cloud-edge-collaborative-inference-for-llm` example crash with `ZeroDivisionError` when:
- The dataset is empty or misconfigured
- A model generates ≤1 token (e.g., immediate EOS, context overflow)

The crash occurs in **9 unguarded divisions** across 8 files:
- 5 metric files: `accuracy.py`, `throughput.py`, `internal_token_latency.py`, `time_to_first_token.py`, `edge_ratio.py`
- 3 model backends: `huggingface_llm.py`, `eagle_llm.py`, `vllm_llm.py`

The benchmark should handle empty datasets and short model outputs gracefully, returning `0.0` for metrics instead of crashing.

1. Create an empty test dataset:
   ```bash
   echo -n "" > dataset/mmlu-5-shot/test_data/data.jsonl
   ```
2. Run the benchmark:
   ```bash
   ianvs -f examples/cloud-edge-collaborative-inference-for-llm/benchmarkingjob.yaml
   ```
3. Result: `ZeroDivisionError` in `accuracy.py` at `sum(same_elements) / len(same_elements)` because `infer_res` is `[]`.

**Alternatively**, configure a model that generates exactly 1 token (`max_tokens: 1`), which causes `huggingface_llm.py` to crash at `sum(internal_token_latency) / len(internal_token_latency)` because the latency list is empty.

**Root cause**: No upstream validation exists. I traced the full call chain:
```
benchmarking.py → TestCaseController → TestCase.run() 
  → JointInference._inference()  # returns [] if dataset empty
  → TestCase.compute_metrics()   # no empty-check before calling metrics
    → metric_func(y_pred=[])     # crashes with ZeroDivisionError
```

**Proposed fix**: Add guards before every division:

For metrics:
```python
if not infer_res:
    return 0.0
```

For model backends:
```python
if len(internal_token_latency) > 0:
    internal_token_latency = sum(internal_token_latency) / len(internal_token_latency)
else:
    internal_token_latency = 0.0
throughput = (1.0 / internal_token_latency) if internal_token_latency > 0 else 0.0
```

**Environment**:
- OS: Fedora
- Python: 3.10
- Branch: main

Fahmid-Arman:
**What type of PR is this?**
/kind bug

The `cloud-edge-collaborative-inference-for-llm` benchmark previously crashed with a `ZeroDivisionError` when executed against an empty dataset or when a model generated $\le1$ token (e.g., immediate EOS). 

This PR implements comprehensive, zero-safe division guards across the 5 metric modules and 4 backend wrappers to ensure the framework returns `0.0` gracefully under these edge cases, preserving the execution of the broader benchmark suite.

Files patched:
**Metrics:**
- `testenv/accuracy.py`
- `testenv/throughput.py`
- `testenv/internal_token_latency.py`
- `testenv/time_to_first_token.py`
- `testenv/edge_ratio.py`

**Backends:**
- `models/huggingface_llm.py`
- `models/vllm_llm.py`
- `models/eagle_llm.py`
- `models/api_llm.py` *(Identified and patched proactively)*

Fixes #337

### DIFF
--- a/examples/cloud-edge-collaborative-inference-for-llm/testalgorithms/query-routing/models/api_llm.py
+++ b/examples/cloud-edge-collaborative-inference-for-llm/testalgorithms/query-routing/models/api_llm.py
@@ -133,10 +133,9 @@ class APIBasedLLM(BaseLLM):
 
             if internal_token_latency:
                 internal_token_latency = sum(internal_token_latency) / len(internal_token_latency)
-                throughput = 1 / internal_token_latency
             else:
-                internal_token_latency = 0
-                throughput = 0
+                internal_token_latency = 0.0
+            throughput = (1.0 / internal_token_latency) if internal_token_latency > 0 else 0.0
 
         except Exception as e:
             raise RuntimeError(f"Error during API inference: {e}")
--- a/examples/cloud-edge-collaborative-inference-for-llm/testalgorithms/query-routing/models/eagle_llm.py
+++ b/examples/cloud-edge-collaborative-inference-for-llm/testalgorithms/query-routing/models/eagle_llm.py
@@ -104,8 +104,11 @@ class EagleSpecDecModel(BaseLLM):
         text = generated_text
         completion_tokens = output_ids.shape[1] - prompt_tokens
         
-        internal_token_latency = sum(internal_token_latency) / completion_tokens
-        
+        if completion_tokens > 0:
+            internal_token_latency = sum(internal_token_latency) / completion_tokens
+        else:
+            internal_token_latency = 0.0
+
         if internal_token_latency != 0:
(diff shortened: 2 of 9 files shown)


Fahmid-Arman:
Feedback addressed: 
- Refactored empty sequence checks to use PEP 8 compliant implicit boolean evaluation (`if internal_token_latency:`) instead of `len() > 0`.

/assign @moorezheng

PR #548, opened 2026-06-14
fix: resolve TypeError by properly instantiating model before loading state_dict in unseen sample recognition

### LINKED ISSUE #326: [Bug] Critical TypeError: UnseenSampleRecognitionByScene treats state_dict as callable model
### Bug Description
The `UnseenSampleRecognitionByScene` class contains a critical logic error in its model loading and inference code. The code loads a PyTorch checkpoint file using `torch.load()` and then **immediately attempts to call the loaded object as if it were a model**, without proper instantiation or weight loading.

Since `torch.load()` typically returns a `collections.OrderedDict` (the state dict), attempting to call it like a function (`outputs = state_dict(InferenceData)`) raises a `TypeError`.

### Code Location
**File:** `examples/cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/testalgorithms/rfnet/unseen_sample_recognition_by_scene.py`
**Lines:** 46-53

[Link to File](https://github.com/kubeedge/ianvs/blob/main/examples/cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/testalgorithms/rfnet/unseen_sample_recognition_by_scene.py#L46-L53)

```python
# Lines 46-53
pthfile = self.model_path
state_dict = torch.load(pthfile)             # Returns OrderedDict (not callable)
# ...
outputs = state_dict(InferenceData)          # BUG: TypeError: 'collections.OrderedDict' object is not callable

Fahmid-Arman:
**What type of PR is this?**
/kind bug

The `UnseenSampleRecognitionByScene` module previously contained a critical logic error in its inference execution. It loaded a checkpoint using `torch.load()`, which returns a `state_dict` (an `OrderedDict`), and then immediately attempted to call it like a function (`outputs = state_dict(InferenceData)`), resulting in a hard `TypeError: 'collections.OrderedDict' object is not callable`.

Changes implemented:
- Dynamically imported the required `Wide_ResNet` architecture from the adjacent `models` directory.
- Properly instantiated the model (`WRN-28-10` with a binary `num_classes=2` head, matching the downstream logic).
- Loaded the weights securely into the model instance using `model.load_state_dict()`.
- Added `map_location=device` to ensure the checkpoint loads safely on CPU-only machines.
- Added `model.eval()` before inference to disable active dropout/batchnorm layers for deterministic results.
- Executed inference via `outputs = model(InferenceData)`.

Fixes #326

### DIFF
--- a/examples/cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/testalgorithms/rfnet/unseen_sample_recognition_by_scene.py
+++ b/examples/cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/testalgorithms/rfnet/unseen_sample_recognition_by_scene.py
@@ -33,6 +33,7 @@ class UnseenSampleRecognitionByScene:
         import torchvision
         from torchvision import transforms
         from PIL import Image
+        from models.wide_resnet_embedding import Wide_ResNet
         d_type = samples.data_type
         x_data = samples.x
         y_data = samples.y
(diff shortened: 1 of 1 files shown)


Fahmid-Arman:
Feedback addressed: 
- Moved `.to(device)` to model instantiation to prevent redundant CPU-to-GPU memory transfers.
- Wrapped the forward pass in `with torch.no_grad():` to reduce memory consumption during inference.

/assign @jaypume

PR #549, opened 2026-06-14
fix: read DASHSCOPE_API_KEY from environment in GovDoc2Poster basemodel

### LINKED ISSUE #530: [Bug] GovDoc2Poster: basemodel.py ignores DASHSCOPE_API_KEY environment variable and uses hardcoded placeholder instead
# [Bug] GovDoc2Poster: basemodel.py ignores DASHSCOPE_API_KEY environment variable and uses hardcoded placeholder instead

## Description

`GovDoc2Poster` loads environment variables via `load_dotenv()` and instructs users to configure `DASHSCOPE_API_KEY`, but the environment variable is never read.

In `basemodel.py`:

```python
load_dotenv()

# Prefer API key from environment for security and flexibility
self.api_key = "your_api"
```

Additionally, the code logs the following warning:

```python
if not self.api_key:
    self.logger.warning(
        "DASHSCOPE API key is not set. Set DASHSCOPE_API_KEY env var or update basemodel.py"
    )
```

However, a repository-wide search shows no usage of `os.getenv()` in the GovDoc2Poster example, meaning the environment variable is never read.

The hardcoded placeholder value is then propagated to downstream OpenAI-based components, including:

* `NewGovernmentDocumentParser`
* `NewGovernmentPosterPlanner`
* `NewGovernmentPosterEvaluator`

As a result, users who configure `DASHSCOPE_API_KEY` through environment variables cannot have it automatically picked up by the GovDoc2Poster pipeline.

---

## Steps to Reproduce

1. Set an environment variable:

```powershell
$env:DASHSCOPE_API_KEY="test_key_123"
```

2. Verify it is available:

```powershell
echo $env:DASHSCOPE_API_KEY
```

Output:

```text
test_key_123
```

3. Search the GovDoc2Poster codebase:

```powershell
git grep -n "os.getenv" examples/GovDoc2Poster
```

4. Observe that no results are returned and `basemodel.py` always assigns:

```python
self.api_key = "your_api"
```

---

## Expected Behavior

When `DASHSCOPE_API_KEY` is configured, GovDoc2Poster should automatically read and use it.

---

## Actual Behavior

The environment variable is ignored and the placeholder value `"your_api"` is propagated to downstream components.

---

## Possible Fix

Read the API key from the environment and validate that a usable value is available:

```python
self.api_key = kwargs.get(
    "api_key",
    os.getenv("DASHSCOPE_API_KEY", "")
)

if not self.api_key:
    raise ValueError(
        "API key not found. Set the DASHSCOPE_API_KEY "
        "environment variable or pass api_key as a parameter."
    )
```

This would make the implementation consistent with the existing warning message and the intended environment-variable-based configuration workflow.

---

### Suggested Label

`kind/bug`

Fahmid-Arman:
**What type of PR is this?**
/kind bug

The `GovDoc2Poster` example ignored the `DASHSCOPE_API_KEY` environment variable, defaulting to the hardcoded `"your_api"` string. This caused downstream LLM/VLM components to fail authentication silently. 

This PR updates `basemodel.py` to:
- Read the API key from `kwargs` or fallback to `os.environ.get("DASHSCOPE_API_KEY")`.
- Raise an explicit `ValueError` if the key is missing or set to the placeholder, enforcing a fail-fast mechanism.
- Remove unreachable legacy warning logs.

Fixes #530

### DIFF
--- a/examples/GovDoc2Poster/singletask_learning_bench/testalgorithms/gen/basemodel.py
+++ b/examples/GovDoc2Poster/singletask_learning_bench/testalgorithms/gen/basemodel.py
@@ -61,8 +61,15 @@ class NewGovernmentPosterAgent:
             max_optimization_iterations: Maximum optimization iteration times
             quality_threshold: Quality threshold
         """
-    # Prefer API key from environment for security and flexibility
-        self.api_key = 'your_api'
+        # Prefer API key from environment for security and flexibility.
+        # Fall back to the DASHSCOPE_API_KEY env var; fail fast if it is unset.
+        self.api_key = kwargs.get("api_key") or os.environ.get("DASHSCOPE_API_KEY") or ""
+
+        if not self.api_key or self.api_key == "your_api":
+            raise ValueError(
+                "API key not found. Set the DASHSCOPE_API_KEY "
+                "environment variable or pass api_key as a parameter."
+            )
         self.llm_base_url = 'https://dashscope.aliyuncs.com/compatible-mode/v1'
         self.vlm_base_url = 'https://dashscope.aliyuncs.com/compatible-mode/v1'
         self.llm_model = kwargs.get('llm_model', 'qwen-max')
@@ -88,8 +95,6 @@ class NewGovernmentPosterAgent:
         # Initialize logging
         logging.basicConfig(level=logging.INFO)
         self.logger = logging.getLogger(__name__)
-        if not self.api_key:
-            self.logger.warning('DASHSCOPE API key is not set. Set DASHSCOPE_API_KEY env var or update basemodel.py')
         
         # Load hyperparameter configuration
         self._load_hyperparameters(kwargs)


Fahmid-Arman:
Feedback addressed: 
- Refactored API key retrieval to use `or` chaining to properly handle cases where `api_key` is explicitly passed as `None`.

/assign @jaypume

christinaliyevav-coder:
I reviewed this PR as a comparison point for the external-service prerequisite problem found in other examples.

The API-key handling here follows a much cleaner fail-fast pattern:

```python
self.api_key = kwargs.get("api_key") or os.environ.get("DASHSCOPE_API_KEY") or ""

if not self.api_key or self.api_key == "your_api":
    raise ValueError(
        "API key not found. Set the DASHSCOPE_API_KEY "
        "environment variable or pass api_key as a parameter."
    )
```

I verified the behavior by extracting this assignment and validation directly from the PR source and executing it with `DASHSCOPE_API_KEY` removed.

Actual result:

```text
ValueError: API key not found. Set the DASHSCOPE_API_KEY environment variable or pass api_key as a parameter.
```

The important part for me is the timing: this validation happens during object initialization, before the rest of the agent configuration and later processing begin.

This provides a useful contrast with the other restoration approaches I reviewed. A missing external-service prerequisite is rejected immediately instead of being discovered at first request time or replaced with a placeholder string.

For the credential-validation scope, I did not find a change that needs revision. The existing automated review discussed the `api_key=None` fallback case, which has already been addressed with the current `or` chaining. My observation here is specifically about the constructor-level fail-fast behavior and its usefulness as a consistent prerequisite-validation pattern.

PR #550, opened 2026-06-15
fix: resolve dataset preparation failure caused by upstream framework update

### LINKED ISSUE #164: Issue with preparing dataset failed：The dataset preparation failure caused by the Ianvs project update
**What happened**:
As my title suggests, I encountered a **dataset preparation error** due to updates in the Ianvs project while running a example in the Ianvs project. Here is the entire process of encountering the problem, my solution, and the cause of the problem. I hope this can help others who have also encountered the same problem.
About a month ago, I started trying to run the Example Cloud-Robotics-1 case, and about half a month ago, after obtaining the required dataset from @hsj576 , I found that the device I was using was not sufficient to achieve the performance required for the case.
Later, I switched to another device and pulled Ianvs' project files from GitHub again to prepare for running Example Cloud-Robotics-1.However, during the configuration process, I encountered an error of dataset preparation failure as shown below. 
``` 
  File "/usr/local/bin/ianvs", line 33, in <module>
    sys.exit(load_entry_point('ianvs==0.1.0', 'console_scripts', 'ianvs')())
  File "/usr/local/lib/python3.10/dist-packages/ianvs-0.1.0-py3.10.egg/core/cmd/benchmarking.py", line 41, in main
    raise RuntimeError(f"benchmarkingjob runs failed, error: {err}.") from err
RuntimeError: benchmarkingjob runs failed, error: prepare dataset failed, error: not one of train_index/train_data/train_data_info..

```
This problem did not occur during the previous configuration process, and after repeatedly changing the path and file placement path in the YAML file, the issue has not been resolved.

After being troubled by this issue for two weeks, I accidentally discovered that the testenvmanager folder in the Ianv project had been changed two weeks ago, so **I switched to using the Ianvs project package pulled from GitHub a month ago to run Example Cloud-Robotics-1.** At this point, the problem of dataset preparation failure no longer exists.
Through communication with @hsj576 , I have learned that the reason for the error may have come from the integration of kubeedge Ianvs into this year's open source summer project code half a month ago, and the updated code may not be compatible with the previous example.

    I hope my findings can help classmates who encounter similar problems, and I also hope that my findings can help the project team further improve the project documentation.

Fahmid-Arman:
**What type of PR is this?**
/kind bug

This PR resolves a backward-compatibility break in the dataset preparation logic that caused `benchmarkingjob` failures across multiple legacy examples. 

* **Root Cause:** A prior framework refactor in `core/testenvmanager/dataset/dataset.py` introduced new configuration keys (`train_index`, `train_data`, `train_data_info`) and required `process_dataset()` to use them. Legacy examples (e.g., Cloud-Robotics) still use `train_url` and `test_url` in their `testenv.yaml` configurations. This mismatch caused a `NotImplementedError("not one of train_index/train_data/train_data_info")` exception.
* **Fix Applied:** Implemented a backward-compatibility shim within the `_parse_config` method of `core/testenvmanager/dataset/dataset.py`. If `train_url` or `test_url` is detected and the new keys are unset, the values are automatically mapped to `train_index` and `test_index`. This restores full functionality for all legacy examples without requiring repository-wide YAML updates and maintains precedence for new configuration structures.

Fixes #164

### DIFF
--- a/core/testenvmanager/dataset/dataset.py
+++ b/core/testenvmanager/dataset/dataset.py
@@ -72,6 +72,36 @@ class Dataset:
             if attr in self.__dict__:
                 self.__dict__[attr] = value
 
+        # Backward compatibility: older testenv configs reference the dataset
+        # via `train_url`/`test_url`. The dataset refactor replaced these with
+        # `train_index`/`train_data`/`train_data_info` (and the `test_*`
+        # counterparts). Route the legacy field to the appropriate new field
+        # based on its file format so existing examples keep working.
+        if self.train_url and not (
+            self.train_index or self.train_data or self.train_data_info
+        ):
+            file_format = utils.get_file_format(self.train_url)
+            if file_format in (DatasetFormat.TXT.value, DatasetFormat.JSON.value):
+                self.train_index = self.train_url
+            elif file_format == DatasetFormat.JSONL.value:
+                self.train_data = self.train_url
+            elif file_format == DatasetFormat.JSONFORLLM.value:
+                self.train_data_info = self.train_url
+            else:
+                self.train_index = self.train_url
+        if self.test_url and not (
+            self.test_index or self.test_data or self.test_data_info
+        ):
+            file_format = utils.get_file_format(self.test_url)
+            if file_format in (DatasetFormat.TXT.value, DatasetFormat.JSON.value):
+                self.test_index = self.test_url
+            elif file_format == DatasetFormat.JSONL.value:
+                self.test_data = self.test_url
+            elif file_format == DatasetFormat.JSONFORLLM.value:
+                self.test_data_info = self.test_url
+            else:
+                self.test_index = self.test_url
+
         self._check_fields()
 
     @classmethod


Fahmid-Arman:
Feedback addressed:
- Implemented format-aware routing for legacy `train_url` and `test_url` using `utils.get_file_format` to prevent runtime failures with `.jsonl` and metadata files.

/assign @jaypume

31groot:
### Request changes — legacy CSV URLs still map to the wrong dataset field

This PR improves legacy `train_url` / `test_url` compatibility by selecting the newer dataset field based on file format.

I found a remaining format-specific gap that is separate from the existing JSONL / metadata concern already raised on this PR.

The new compatibility logic explicitly handles TXT, JSON, JSONL, and JSON-for-LLM, but `DatasetFormat` also supports `csv`. For CSV, the logic falls through to:

```python
else:
    self.train_index = self.train_url
```

and equivalently:

```python
else:
    self.test_index = self.test_url
```

This is problematic because CSV is not an index-file format for the subsequent dataset-processing path. The legacy CSV URL is therefore classified as `train_index` / `test_index`, after which the index-processing logic does not handle CSV and can replace the URL with `None`.

I reproduced this against the exact PR head `b946de6` using minimal temporary CSV files for both train and test.

Observed result:

```text
train_url after parse: /tmp/.../train.csv
train_index: /tmp/.../train.csv
train_data:
train_url after process: None
test_url after process: None
CSV legacy mapping bug reproduced
```

The resulting control flow is:

```text
legacy train_url (*.csv)
    -> train_index
    -> _process_index_file()
    -> None
    -> self.train_url = None
```

So the compatibility shim can successfully recognize and retain a legacy CSV URL during parsing, but the URL is subsequently lost during dataset processing.

This matters because the PR's stated goal is to restore backward compatibility for legacy examples using `train_url` / `test_url`. A supported CSV dataset should not be routed through the index-file path merely because it is not one of the formats explicitly handled by the new routing logic.

This is distinct from the existing public review concerning JSONL and metadata files. Those cases are now explicitly handled by the PR; the remaining issue is the separately supported CSV format.

I recommend adding an explicit CSV branch that routes the legacy URL to the appropriate dataset field, or otherwise extending the dataset-processing logic so that CSV is handled correctly.

For example, the compatibility mapping should not treat CSV as an index format:

```python
elif file_format == DatasetFormat.CSV.value:
    # route to the appropriate CSV dataset field
```

The exact target field should follow the existing dataset-processing semantics rather than simply reusing `train_index`.

I also recommend adding a regression test covering both legacy CSV `train_url` and `test_url` and verifying that `process_dataset()` preserves/produces valid dataset URLs rather than converting them to `None`.

**Verification boundary:** I independently reproduced this failure against PR head `b946de6` using minimal temporary CSV files and isolated stubs for unrelated Sedna datasource imports. No full Example benchmark or real simulation environment was required.

Issue #449, opened 2026-05-21
[MOT17] Missing `requirements.txt` and Docker-specific workspace path block bare-metal setup


suhaan-24:
### Summary

Two setup-level problems prevent the MOT17 benchmark from being reproducibly installed
and run outside the official ianvs Docker container. First, the example directory
contains no `requirements.txt`, leaving contributors to manually discover dependencies
through repeated `ImportError` cycles. Second, both job YAML files hardcode an absolute
workspace path (`/ianvs/multiedge_inference_bench/workspace`) that only exists inside
the Docker container, causing an immediate `PermissionError` or `FileNotFoundError` for
all bare-metal and virtualenv users.

### Environment

- **Example:** `MOT17/multiedge_inference_bench`
- **Branch:** `main` (as of 2026-05-21)
- **Files affected:**
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/` (directory — missing `requirements.txt`)
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/tracking_job.yaml`
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/reid_job.yaml`

---

#### Finding 1 — No `requirements.txt` in the example directory

**Path:** `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/`

**Problem:**  
The directory contains no `requirements.txt`. The README provides scattered `pip install`
lines, but they are incomplete. At minimum, the following packages are required to run
the benchmark but are not documented:

- `loguru` — used in metric and algorithm scripts for logging
- `motmetrics` — required by all six tracking metric scripts
- `mmcv` — required by `mot2reid.py` for ReID dataset preparation
- `opencv-python` — required by the data pipeline

Additionally, no version bounds are specified for any package, creating two concrete
compatibility hazards:

1. **`motmetrics`:** The 1.x and 0.x releases have incompatible metric field naming
   conventions. An unconstrained `pip install motmetrics` may install either version
   depending on the resolver state, silently producing incorrect metric names at
   evaluation time.

2. **`pandas`:** The `delim_whitespace=True` argument used in `generate_reports.py`
   was deprecated in pandas 2.0 (April 2023) and raises a `FutureWarning` or, in
   newer versions, an error. A version upper bound of `<2.0` is needed until the
   code is updated to use `sep=r'\s+'`.

**Impact:**  
Every new contributor must run the benchmark at least twice — once to discover the
[...]

PR #551, opened 2026-06-15
fix: add None guard for regex in load() and update addmm_() for PyTorch >= 2.0 compatibility

### LINKED ISSUE #445: [MOT17] M3L basemodel.py: unguarded regex crashes on non-standard checkpoints; deprecated `addmm_` API removed in PyTorch ≥ 2.0
### Summary

Two bugs in the M3L ReID algorithm implementation (`testalgorithms/reid/m3l/basemodel.py`)
block the ReID benchmark job from completing. First, the architecture-inference regex
in `load()` has no `None` guard and will crash with `AttributeError` on any model
filename that does not match a narrow implicit pattern. Second, the pairwise distance
computation in `predict()` uses the three-positional-float signature of `addmm_()`, which
was removed in PyTorch 2.0 and raises `TypeError` on any modern GPU environment.

### Environment

- **Files affected:**
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/reid/m3l/basemodel.py`

---

#### Finding 1 — Unguarded regex in `load()` crashes on non-standard model filenames

**File:** `testalgorithms/reid/m3l/basemodel.py`  
**Line:** 56  
**Code:**
```python
arch = re.compile("_([a-zA-Z]+).pth").search(model_url).group(1)
```

**Problem:**  
The `load()` method attempts to infer the model architecture name from the checkpoint
filename using a regex. There are two compounding issues:

1. **No `None` guard:** If `model_url` does not contain a substring matching
   `_<letters>.pth`, `re.search()` returns `None`. Calling `.group(1)` on `None`
   raises:
   ```
   AttributeError: 'NoneType' object has no attribute 'group'
   ```
   Any alternative checkpoint — a fine-tuned model, a renamed download, or a path like
   `market_best_model.bin` — triggers this immediately.

2. **Unescaped dot in regex:** The pattern `"_([a-zA-Z]+).pth"` uses an unescaped `.`
   which matches any character, not a literal dot. This is a latent correctness issue
   that widens the match unintentionally (e.g., `_IBNMeta_pth` would match when it
   should not).

The current default filename `market_IBNMeta.pth.tar` happens to satisfy the pattern,
so the bug is invisible unless users bring their own checkpoints.

**Impact:**  
The error occurs at model load time, before any inference runs. Users who provide
alternative checkpoints receive an `AttributeError` with no indication of which
filename caused the problem or what naming convention is required.

**Proposed Fix:**
```python
match = re.compile(r"_([a-zA-Z]+)\.pth").search(model_url)
if match is None:
    raise ValueError(
        f"Cannot infer model architecture from filename: '{model_url}'. "
        "Expected a filename containing '_<arch>.pth' (e.g., 'market_IBNMeta.pth.tar')."
    )
arch = match.group(1)
```

---

#### Finding 2 — Deprecated `addmm_()` positional-float API removed in PyTorch 2.0

**File:** `testalgorithms/reid/m3l/basemodel.py`  
**Line:** 153  
**Code:**
```python
dist_m.addmm_(1, -2, x, y.t())
```

**Problem:**  
The call uses the legacy three-positional-float signature
`addmm_(beta, alpha, mat1, mat2)`, where `beta` and `alpha` are passed as positional
arguments. This signature was deprecated in PyTorch **1.5** and **removed in
PyTorch 2.0**. On any current CUDA environment (CUDA 11.8+ ships with PyTorch ≥ 2.0
by default), the call raises:

```
TypeError: addmm_() received an invalid combination of arguments - got
(int, int, Tensor, Tensor), but expected one of:
 * (Tensor mat1, Tensor mat2, *, Scalar beta, Scalar alpha)
```

The example was written in 2022 when this signature was still valid. It has been silently
broken on current GPU servers since the PyTorch 2.0 release.

**Impact:**  
The crash occurs inside `_pairwise_distance()`, which is called from `predict()`. The
model must complete a full forward pass through the dataset before `_pairwise_distance()`
is reached, so users waste the full inference compute time before hitting the error.
The entire ReID job becomes non-executable on any modern environment.

**Proposed Fix:**
```python
# Replace positional beta/alpha with keyword arguments (PyTorch ≥ 1.5 compatible):
dist_m.addmm_(x, y.t(), beta=1, alpha=-2)
```

---

### Steps to Reproduce

**Finding 1 (unguarded regex):**
1. Download any ReID checkpoint with a non-standard filename (e.g., rename
   `market_IBNMeta.pth.tar` to `market_best_model.bin`).
2. Update `m3l_algorithm.yaml` → `initial_model_url` to point to the renamed file.
3. Run the ReID job:
   ```bash
   ```
4. Observe: `AttributeError: 'NoneType' object has no attribute 'group'` at the
   `load()` step.

**Finding 2 (deprecated `addmm_`):**
1. Run the ReID job on any machine with PyTorch ≥ 2.0.
2. Allow inference to complete.
3. Observe: `TypeError: addmm_() received an invalid combination of arguments` during
   the `_pairwise_distance()` call inside `predict()`.

Verify PyTorch version: `python -c "import torch; print(torch.__version__)"`.
Any version ≥ 2.0.0 will trigger Finding 2.

---

### Additional Notes

Both bugs are confined to `m3l/basemodel.py` and require no changes to the ianvs core.
They can be fixed in a single focused PR.

Finding 2 affects any user on a 2024+ GPU server. CUDA 11.8 and later ship with
PyTorch 2.x as the default. It is unlikely that any current contributor can run the
ReID job without encountering this crash.

SahilKumar75:
## Problem

Two bugs in `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testalgorithms/reid/m3l/basemodel.py` that cause runtime failures:

1. **`load()`** — `re.compile(...).search(model_url).group(1)` raises `AttributeError: 'NoneType' object has no attribute 'group'` when the filename doesn't match the expected `_<arch>.pth` pattern.
2. **`_pairwise_distance()`** — `dist_m.addmm_(1, -2, x, y.t())` uses the deprecated positional-float signature removed in PyTorch ≥ 2.0, raising `TypeError`.

## Fix

- **`load()`**: capture the `re.search()` result and raise a descriptive `ValueError` if it's `None`, before calling `.group(1)`.
- **`_pairwise_distance()`**: replace `addmm_(1, -2, x, y.t())` with `addmm_(x, y.t(), beta=1, alpha=-2)`, which is the correct keyword-argument form supported from PyTorch 1.5+ through 2.x.

## Testing

Changes are logic-level; the regex fix is trivially verifiable and the `addmm_` signature change matches the [PyTorch ≥ 2.0 migration guide](https://pytorch.org/docs/stable/notes/pytorch_migration_guide.html).

Fixes #445


observer04 (commented):
**Recommendation: major revision.** The explicit regex failure and keyword-form `addmm_` call are useful improvements. Gemini has already identified that `re.search` still receives `model_url` directly and can fail for a `pathlib.Path`; that existing finding should be addressed.

There is a second maintenance gap around the compatibility change: MOT17 has no tracked requirements manifest. This PR improves behavior for newer PyTorch, but the Example has nowhere to state or test its supported Torch range. The new keyword call may also work on older versions, so I am not claiming this creates a hard `torch>=2` minimum.

Please add focused tests for `str`/`Path` model URLs and for the intended PyTorch range, then record that range when MOT17’s missing manifest is added under Issue #449. I did not run an MOT17 model or dataset.

aryananand-04:
Ran the patched `load()` directly rather than just reading the diff. The regex fix looks right. There's a second failure in the same function that this doesn't reach.

On the regex: `re.compile("_([a-zA-Z]+).pth").search(model_url)` returns `None` for any checkpoint name that doesn't match, and the bare `.group(1)` after it gives you

```
AttributeError: 'NoneType' object has no attribute 'group'
```

The default checkpoint (`market_IBNMeta.pth.tar`) happens to match, which is why nobody hits this until they bring their own weights. Your guard handles it, and the unescaped dot gets fixed in the same expression.

The part that isn't covered: `Path(model_url)` a few lines down still raises. Someone flagged this inline already. I checked where the line actually ends up, since it moves: it's at line 64 on `main`, and this patch adds 6 lines above it, so it lands at 70. Running the real `load()` body from both:

```
upstream/main -> TypeError: expected string or bytes-like object, got 'PosixPath'
PR #551       -> TypeError: expected string or bytes-like object, got 'PosixPath'
```

The `None` guard never gets to run, because `re.search` blows up before anything reaches it.

That said, I think the severity in the inline comment is pitched too high. I traced how `model_url` gets here on a normal `ianvs -f` run and it's a `str` on both routes in:

- `core/.../algorithm.py:155` raises `ValueError` if `self.initial_model_url` isn't already a `str`
- `sedna/backend/base.py:123` returns `FileOps.join_path(...)`, documented `:rtype: str`

So you can't actually reach this through a normal invocation. You'd need a caller outside those two paths handing it a `Path` directly. The bug is real, but it's future-proofing rather than something breaking today. Which to me argues for just fixing it here, since it's a couple of lines in a function you're already touching, rather than treating it as a blocker.

Minor revision. An `isinstance` or `str()` guard before the `Path()` call and this is complete.

Run on macOS arm64, Python 3.11, against the real `load()` bodies from `upstream/main` and this PR's head. No GPU path tested.

PR #552, opened 2026-06-15
fix: bypass Sedna network timeouts during local offline benchmarking

### LINKED ISSUE #494: [Bug] Local benchmarks experience massive idle delays due to unmocked Sedna HTTP retries
**What happened**:
When running local benchmarks (like the `semantic-segmentation` lifelong learning example), the pipeline stalls for a really long time. In my test, it wasted over 20 minutes just hanging on HTTP calls.

It looks like Ianvs is using the production `sedna` library but isn't mocking out Sedna's network layer. During a local run, Sedna tries to hit unreachable endpoints (like `None/knowledgebase/file/upload` or `http://127.0.0.1:9100`). Since these don't exist locally, the requests naturally fail. 

Then, Sedna's `tenacity` wrapper catches these failures and retries each request 5 times with a 3-second wait. Because Ianvs is just a local testing tool, these retries don't actually do anything and just create massive idle wait times.

Ianvs should probably mock these network calls or just disable Sedna's retry loop during local benchmarks, so the pipeline fails fast instead of hanging.

**How to reproduce it**:
Just run the benchmark locally:
   ```bash
   python benchmarking.py -f examples/robot/lifelong_learning_bench/semantic-segmentation/benchmarkingjob-simple.yaml
   ```
Watch the terminal. The execution hangs repeatedly during task index uploads and knowledge base updates. You'll see it throwing connection errors and retrying on endpoints like these:
   ```text
   [ERROR] - Update kb error: request to None/knowledgebase/update/status failed
   [ERROR] - KB update Fail !
   [ERROR] - request to http://127.0.0.1:9100/sedna/workers/sedna/info failed
   ```

To test this, I bypassed the `http_request` logic and made a comparison table with and without the fix. The execution time drops by a crazy amount.

Here are the numbers for the `semantic-segmentation` example:

| Metric | Standard Run | Retries Bypassed |
| :--- | :--- | :--- |
| **Total Execution Time** | 23m 25.3s | 1m 26.0s |
| **Network Delay (Wait)** | 22m 1.7s | 0.00s |
| **HTTP Requests Bypassed** | 0 | 110 |
| **Computation Time** | 1m 23.7s | 1m 26.0s |

Most of the time is wasted here (baseline run):
* `None/knowledgebase/file/upload`: 36 calls (432s wasted)
* `http://127.0.0.1:9100/sedna/workers/sedna/info`: 37 calls (444s wasted)
* `None/knowledgebase/update/status`: 31 calls (372s wasted)

**Proposed fix**:
We can just monkey-patch `sedna.service.client.http_request` globally at the Ianvs entry point (like `benchmarking.py` or `TestCaseController`) to fail fast on `None` or `localhost` endpoints during testing. I already have a working patch for this, let me know if you want me to open a PR!

kryz73:
**What type of PR is this?**
/kind bug
/kind feature

This PR introduces a global network bypass patch for the Sedna client inside `benchmarking.py` and `core/cmd/benchmarking.py`. 

### Why is it needed?
When running Ianvs benchmarks locally, there is no active KubeEdge Cloud controller or Knowledge Base server. Despite this, the Sedna backend client attempts REST calls to remote/null endpoints. Because it is decorated with `tenacity` retry logic, it retries failed attempts up to 5 times with 3-second waits (15 seconds total per call). 

Across multiple incremental seen/unseen training and evaluation rounds, these network timeouts block execution, wasting up to 15–20 minutes of idle wait time per benchmarking job.

### What does this patch do?
1. Intercepts the module-level `sedna.service.client.http_request` function at the Ianvs entry points.
2. Checks if the destination endpoint is local/null (`None/`, `127.0.0.1`, `localhost`).
3. If it is local and the toggle is enabled, it instantly raises a `ConnectionError("Connection refused.")` to bypass the 15-second retry loops. This allows Sedna's internal try/except blocks to run and gracefully fall back to local offline mechanisms without any delay.
4. Provides a toggle `BYPASS_TIMEOUTS` (default: `"1"` / enabled). If set to `"0"`, the bypass is disabled.

### How to test it?
To run benchmarks normally with the bypass enabled:
```bash
ianvs -f examples/robot/lifelong_learning_bench/semantic-segmentation/benchmarkingjob-simple.yaml
```
To run benchmarks with the bypass disabled (original timeout behavior):
```bash
BYPASS_TIMEOUTS=0 ianvs -f examples/robot/lifelong_learning_bench/semantic-segmentation/benchmarkingjob-simple.yaml
```
Fixes #494


karyxx:
I have moved the logic inside `core/common/utils.py`, resolved the performance overhead and misleading log message.

karyxx:
Hi @hsj576, hope you are doing well, i submitted this PR three weeks ago, i just wanted to gently check if this PR is in your review queue when you get some time. Thanks

ajmani-x:
Reviewed this as part of an LFX Term 3 pre-test (full review here: https://github.com/kubeedge/ianvs/discussions/972#discussioncomment-18186183). Good fix for the network-timeout idle-wait problem at the client path it patches. Flagging a sibling case this PR's own stated goal (avoid idle network waits during local offline benchmarking) would also want covered: `sedna.core.joint_inference.JointInference.__init__` (in the vendored wheel at `resources/third_party/sedna-0.6.0.1-py3-none-any.whl`) unconditionally starts an `LCReporter` background thread and opens a `ModelClient` network connection whenever `LCReporter_enable` isn't explicitly passed as `False` (its class default is `True`) or `cloud`/`speculative_runtime` aren't set. Today `base.py` happens to pass `LCReporter_enable=False` explicitly, so this specific path is currently safe — but that's incidental to this PR's patch, not covered by it, and would resurface immediately if that kwarg were ever dropped. Might be worth a comment noting the scope boundary, or extending the patch to cover this constructor too.

PR #554, opened 2026-06-15
fix: remove hardcoded CUDA assumptions in ERFNet to support CPU-only execution

### LINKED ISSUE #471: [Bug] ERFNet training crashes on non-CUDA hardware due to hardcoded CUDA assumptions throughout args.py and train.py
## Environment

| Item | Version |
|---|---|
| OS | macOS 14 (Darwin 23.6.0, Apple Silicon - no NVIDIA GPU) |
| Python | 3.12.7 |
| Sedna | 0.6.0.1 |
| PyTorch | 2.12.0 |
| ianvs branch | main |

## Example

```
```

## Run Command

```bash
```

## Description

The ERFNet training and evaluation utilities inside this example hardcode CUDA availability at three separate points. On any machine without an NVIDIA GPU — including CPU-only Linux machines and any Apple Silicon Mac — the pipeline crashes before a single training step completes. The three crashes are in different files and occur at different stages of training startup.

---

## Error 1 — `self.cuda = True` hardcoded in `args.py`, crashes during model setup

**Triggered at:** Immediately when `BaseModel.train()` is called and the trainer is initialized. The crash happens before any data is loaded.

```
  File ".../ERFNet/eval.py", line 52, in __init__
    self.model = self.model.cuda(args.gpu_ids)
  File ".../torch/nn/modules/module.py", line ...
RuntimeError: Invalid device, must be cuda device
  error: Invalid device, must be cuda device
```

**Location:** `testalgorithms/erfnet/ERFNet/utils/args.py`

Both `TrainArgs` and `ValArgs` set `self.cuda = True` and `self.no_cuda = False` unconditionally regardless of the machine's hardware. These flags control all subsequent `.cuda()` calls throughout `train.py` and `eval.py`. There is no `torch.cuda.is_available()` check anywhere in `args.py`.

```python
# Current state in args.py (TrainArgs and ValArgs both):
self.cuda = True        # hardcoded — crashes on any CPU-only machine
self.no_cuda = False    # hardcoded
```

---

## Error 2 — `map_location='cuda:0'` hardcoded in `train.py` when loading checkpoints

**Triggered at:** When `Trainer.__init__()` loads a pre-trained checkpoint via `args.resume`. On a non-CUDA machine, the tensor is mapped to a device that does not exist.

```
  File ".../ERFNet/train.py", line 108, in __init__
    checkpoint = torch.load(args.resume, map_location=torch.device('cuda:0'))
RuntimeError: Attempting to deserialize object on a CUDA device but
  torch.cuda.is_available() is False.
```

**Location:** `testalgorithms/erfnet/ERFNet/train.py`, line 108.

The checkpoint loading call hardcodes `map_location=torch.device('cuda:0')` with no fallback for CPU execution.

```python
# Current state in train.py:
checkpoint = torch.load(args.resume, map_location=torch.device('cuda:0'))
```

---

## Error 3 — `target.cuda()` called unconditionally inside the training loop

**Triggered at:** During the first batch of training, after data loading succeeds. This is a latent crash — it only surfaces if Errors 1 and 2 happen to be worked around (for example, by patching `args.cuda` at runtime).

```
  File ".../ERFNet/train.py", line 263, in training
    target = target.cuda(self.gpu_ids)
RuntimeError: Invalid device, must be cuda device
```

**Location:** `testalgorithms/erfnet/ERFNet/train.py`, inside the `training()` method.

After `target = target.squeeze(0)`, the code unconditionally calls `target.cuda(self.gpu_ids)` with no `if self.args.cuda:` guard. All other tensors in the same loop (`image`, `depth`) are guarded by `if self.args.cuda:`, but `target` is not.

```python
# Current state in train.py (training() method):
target = target.squeeze(0)
target = target.cuda(self.gpu_ids)   # no CUDA guard — other tensors in same block have guards
```

---

## Impact

Any developer or evaluator running this example on a machine without an NVIDIA GPU — which includes any modern MacBook, any CI runner without GPU, any cloud CPU instance — hits Error 1 immediately and cannot reach training. The example is completely non-functional on the hardware most contributors and reviewers are likely to use.

## Files Involved

- `testalgorithms/erfnet/ERFNet/utils/args.py`
- `testalgorithms/erfnet/ERFNet/train.py`

Fahmid-Arman:
**What type of PR is this?**
/kind bug

The ERFNet training and evaluation utilities contained hardcoded CUDA availability assumptions, causing the pipeline to crash immediately on CPU-only machines and Apple Silicon devices. 

- Dynamically assigned `self.cuda = torch.cuda.is_available()` (and `self.no_cuda`) within `TrainArgs` and `ValArgs` in `utils/args.py`.
- Refactored checkpoint loading in `train.py` to use a dynamic `map_location` supporting CPU fallback.
- Added a missing `if self.args.cuda:` guard around a `target.cuda()` call within the `training()` loop in `train.py` to match the surrounding tensor handling logic.

This unblocks ERFNet execution for contributors running the benchmark on non-NVIDIA hardware.

Fixes #471

### DIFF
--- a/examples/robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/testalgorithms/erfnet/ERFNet/train.py
+++ b/examples/robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/testalgorithms/erfnet/ERFNet/train.py
@@ -105,7 +105,8 @@ class Trainer(object):
                     "=> no checkpoint found at '{}'" .format(args.resume))
             print(f"Training: load model from {args.resume}")
             checkpoint = torch.load(
-                args.resume, map_location=torch.device('cuda:0'))
+                args.resume,
+                map_location=torch.device(f'cuda:{args.gpu_ids}' if args.cuda else 'cpu'))
             args.start_epoch = checkpoint['epoch']
 
             self.model.load_state_dict(checkpoint['state_dict'], False)
@@ -260,7 +261,8 @@ class Trainer(object):
                 target, 255, self.nclass[self.current_domain] - 1)
 
             target = target.squeeze(0)
-            target = target.cuda(self.gpu_ids)
+            if self.args.cuda:
+                target = target.cuda(self.gpu_ids)
 
             outputs_prev_task = self.model(
                 image, max(self.current_domain-1, 0))
--- a/examples/robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/testalgorithms/erfnet/ERFNet/utils/args.py
+++ b/examples/robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/testalgorithms/erfnet/ERFNet/utils/args.py
@@ -1,3 +1,6 @@
+import torch
+
+
 class TrainArgs:
     def __init__(self, **kwargs):
         self.depth = False
@@ -22,7 +25,7 @@ class TrainArgs:
         self.lr_scheduler = 'cos'
         self.momentum = 0.9
         self.weight_decay = 2.5e-5
-        self.no_cuda = False
+        self.no_cuda = kwargs.get("no_cuda", not torch.cuda.is_available())
         self.gpu_ids = 0
 
         self.seed = 1
@@ -31,7 +34,7 @@ class TrainArgs:
         self.ft = True
         self.eval_interval = kwargs.get("eval_interval", 50)
         self.no_val = kwargs.get("no_val", True)
-        self.cuda = True
+        self.cuda = kwargs.get("cuda", not self.no_cuda and torch.cuda.is_available())
         self.savedir = './dataset/mdil-ss/save'
 
 class ValArgs:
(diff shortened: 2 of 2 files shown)


Fahmid-Arman:
Feedback addressed: 
- Parameterized GPU ID indexing in `map_location` and added `kwargs` overrides for `cuda` and `no_cuda` flags to prevent state conflicts.

/assign @jaypume

aryananand-04:
Checked this one against the wider codebase and it holds up.

My first pass looked like it might be incomplete. `git grep` turns up five more `.cuda()` calls elsewhere, so I went through each before writing anything:

- four are already guarded, behind `is_cuda`, `args.cuda` or `self.cuda` checks
- the fifth sits inside an `if __name__ == "__main__"` demo block, outside the training and inference path

So the concern doesn't hold. This PR fixes the `args.cuda` assumption at the source, and the guarded call sites downstream pick that up. Complete for what it sets out to do.

Mentioning the false start because "there are five more `.cuda()` calls, this is incomplete" is the obvious drive-by comment here and it would have been wrong.

One thing I'd raise as a question rather than a change request: this makes the example tolerant of CPU-only environments, which is clearly the right direction. It'd be worth knowing whether MPS is meant to be in scope too, or whether CPU fallback is enough for this example's purposes.

Verified by reading each call site rather than executing, since confirming a guard exists doesn't need the hardware. I'm not claiming a runtime check I didn't run.

PR #556, opened 2026-06-16
fix: prevent NaN propagation in MOT17 tracking metrics

### LINKED ISSUE #448: [MOT17] NaN propagation from upstream tracking metrics corrupts leaderboard CSV
### Summary

Five upstream tracking metric scripts (`precision.py`, `recall.py`, `mota.py`,
`motp.py`, `idf1.py`) in the MOT17 example share a common unguarded pattern that
silently writes `nan` to the leaderboard CSV when `motmetrics` cannot compute a value.
The existing fix in PR #402 addresses only the downstream `f1_score.py`, leaving the
five upstream metric columns unpatched. A leaderboard produced after PR #402 merges
will still contain `nan` in five of six tracking metric columns under degenerate
detection conditions.

### Environment

- **Files affected:**
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/precision.py`
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/recall.py`
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/mota.py`
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/motp.py`
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/idf1.py`

---

#### Finding 1 — Unguarded `float(nan)` return in all five upstream metric scripts

**Pattern (identical across all five files):**
```python
return round(float(summary.iloc[-1][metrics_name]), 4)
```

**Problem:**  
When `motmetrics` cannot compute a metric — for example when the tracker produces zero
detections across all sequences — `mh.compute_many()` populates the summary DataFrame
with `float('nan')` for the affected metric. The return statement processes this as
follows:

```python
float(float('nan'))   # → nan   (no exception, no warning)
round(nan, 4)          # → nan   (no exception, no warning)
```

The `nan` value is returned to the ianvs evaluation pipeline, stored in the leaderboard
rank CSV, and displayed in the benchmark output table. No warning is emitted.

**Why the existing PR #402 fix is insufficient:**  
PR #402 proposes adding a guard to `f1_score.py`:
```python
denominator = precision + recall
f1_score = 2 * precision * recall / denominator if denominator > 0 else 0.0
```
In Python, `float('nan') > 0` evaluates to `False`, so this guard does handle the NaN
case correctly in `f1_score.py` in isolation. However, the `precision` and `recall`
values that feed this computation come from the metric pipeline, and *those* return
`nan` to the CSV. After PR #402 merges, the leaderboard will show:

| metric | value |
|--------|-------|
| precision | `nan` |
| recall | `nan` |
| f1_score | `0.0` ← fixed by #402 |
| mota | `nan` |
| motp | `nan` |
| idf1 | `nan` |

Five of six metric columns remain corrupted. The fix must be applied to all five
upstream scripts for the leaderboard to be fully clean.

**Impact:**  
- `nan` in leaderboard CSV rows prevents numeric sorting and comparison.
- Downstream tooling (pandas, matplotlib) that reads the CSV will propagate NaN through
  aggregations silently.
- A benchmark run that partially fails (some sequences have detections, others do not)
  produces a mixed valid/NaN leaderboard with no indication of which sequences failed.

**Proposed Fix (apply to all five files):**

```python
import math
import logging

logger = logging.getLogger(__name__)

# Replace the final return statement in each metric() function:
value = float(summary.iloc[-1][metrics_name])
if math.isnan(value) or math.isinf(value):
    logger.warning(
        "Metric '%s' returned %s (no valid detections?). Recording 0.0.",
        metrics_name, value
    )
    return 0.0
return round(value, 4)
```

The `math.isnan` + `math.isinf` guard handles all degenerate cases from `motmetrics`
without relying on Python's NaN comparison semantics, which can be counterintuitive.

---

### Steps to Reproduce

1. Prepare a tracking result set where the model produces **zero detections** across
   all MOT17 sequences (e.g., use an empty prediction file or a misconfigured model).
2. Run the tracking benchmark job:
   ```bash
   ```
3. Inspect the leaderboard CSV in the workspace output directory:
   ```bash
   cat /ianvs/multiedge_inference_bench/workspace/tracking_job/tracking/*/rank/all_rank.csv
   ```
4. Observe `nan` in the `precision`, `recall`, `mota`, `motp`, and `idf1` columns.

---

### Additional Notes

This issue is complementary to the fix proposed in PR #402, which addresses only
`f1_score.py`. The fix described here should either be included in an amended PR #402
or submitted as a follow-up PR that covers all five upstream scripts together.

All five files use the same `compare_dataframes` + `mh.compute_many` + return pattern,
so the fix is mechanical and low-risk. No changes to the ianvs core are required.

See also: the `f1_score.py` guard in PR #402 demonstrates the correct intent — this
issue extends that intent to the rest of the tracking metric suite.

Fahmid-Arman:
**What type of PR is this?**
/kind bug

When `motmetrics` cannot compute a value (e.g., zero valid detections), it outputs `NaN`. Previously, five upstream tracking metric scripts (`precision.py`, `recall.py`, `mota.py`, `motp.py`, `idf1.py`) blindly cast this output to float and returned it, which silently corrupted the `all_rank.csv` leaderboard with `NaN` values and broke numeric sorting. 

- Added a `math.isnan()` and `math.isinf()` guard to the final return statement in all five tracking metric scripts.
- When degenerate detection conditions occur, the pipeline now logs a warning and safely records `0.0` instead of corrupting the CSV downstream.
- Added `math` and `logging` imports to support the guard.

Fixes #448

### DIFF
--- a/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/idf1.py
+++ b/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/idf1.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import math
 from collections import OrderedDict
 from pathlib import Path
 
@@ -46,4 +47,11 @@ def metric(y_true, y_pred):
     summary = mh.compute_many(accs, names=names, metrics=metrics_name, generate_overall=True)
     logger.info('Completed')
 
-    return round(float(summary.iloc[-1][metrics_name]), 4)
+    value = float(summary.iloc[-1][metrics_name])
+    if math.isnan(value) or math.isinf(value):
+        logger.warning(
+            "Metric '{}' returned {} (no valid detections?). Recording 0.0.",
+            metrics_name[0], value
+        )
+        return 0.0
+    return round(value, 4)
--- a/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/mota.py
+++ b/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/mota.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import math
 from collections import OrderedDict
 from pathlib import Path
 
@@ -46,4 +47,11 @@ def metric(y_true, y_pred):
     summary = mh.compute_many(accs, names=names, metrics=metrics_name, generate_overall=True)
     logger.info('Completed')
 
-    return round(float(summary.iloc[-1][metrics_name]), 4)
+    value = float(summary.iloc[-1][metrics_name])
+    if math.isnan(value) or math.isinf(value):
+        logger.warning(
+            "Metric '{}' returned {} (no valid detections?). Recording 0.0.",
+            metrics_name[0], value
+        )
+        return 0.0
+    return round(value, 4)
--- a/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/motp.py
+++ b/examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/tracking/motp.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import math
 from collections import OrderedDict
 from pathlib import Path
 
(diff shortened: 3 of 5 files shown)


Fahmid-Arman:
Feedback addressed:
- Removed standard `logging` overrides to preserve `loguru` global configurations and `INFO` visibility. 
- Refactored warnings to use native brace-style formatting and explicitly indexed `metrics_name[0]` to prevent list stringification.

/assign @moorezheng

PR #558, opened 2026-06-18
Fix module loading path collision

### LINKED ISSUE #557: core: load_module can reuse the wrong file when module names collide
**What happened**:

`core.common.utils.load_module()` loads a Python file using only its basename.

So if Ianvs loads two different files with the same name, like:

`examples/foo/testalgorithms/basemodel.py`
`examples/bar/testalgorithms/basemodel.py`

the second load can still reuse the first `basemodel` module already present in `sys.modules`.

This is easy to hit here since many examples use common names like `basemodel.py`, `accuracy.py`, `map.py`, etc.

Loading a module by file path should load that exact file.

Two files with the same basename but different paths should not collide.

**How to reproduce it**:

```python
import os
import sys
import tempfile
from core.common import utils

root = tempfile.mkdtemp()
one = os.path.join(root, "one")
two = os.path.join(root, "two")
os.makedirs(one)
os.makedirs(two)

with open(os.path.join(one, "plugin.py"), "w") as f:
    f.write('VALUE = "one"\n')

with open(os.path.join(two, "plugin.py"), "w") as f:
    f.write('VALUE = "two"\n')

utils.load_module(os.path.join(one, "plugin.py"))
print(sys.modules["plugin"].VALUE)

utils.load_module(os.path.join(two, "plugin.py"))
print(sys.modules["plugin"].VALUE)

```
```
Actual output:
one
one

Expected output:
one
two

```

Anything else we need to know?:
Current code uses importlib.import_module(module_name) after temporarily adding the module directory to sys.path.

It should load by file path instead, probably with importlib.util.spec_from_file_location() and a unique module name. Also cleanup of sys.path should happen even when import fails.

avinxshKD:
**What type of PR is this?**

/kind bug

Fixes `load_module()` so two Python files with the same basename do not collide in `sys.modules`.

Ianvs examples often use common names like `basemodel.py`, `accuracy.py`, and `map.py`. The old loader imported by basename after adding the directory to `sys.path`, so a later file could reuse an earlier module with the same name.

This keeps the existing loader behavior for callers, but loads file paths under a unique internal module name.

Fixes #557

### DIFF
--- a/core/common/utils.py
+++ b/core/common/utils.py
@@ -14,12 +14,13 @@
 
 """This script contains some common tools."""
 
+import hashlib
 import importlib
 import os
 import sys
 import time
 
-from importlib import import_module
+from importlib import import_module, util
 from inspect import getfullargspec
 import yaml
 
@@ -92,12 +93,28 @@ def yaml2dict(url):
 def load_module(url):
     """Load python module."""
     module_path, module_name = os.path.split(url)
+    loaded_module = None
     if os.path.isfile(url):
         module_name = module_name.split(".")[0]
 
     sys.path.insert(0, module_path)
     try:
-        importlib.import_module(module_name)
-        sys.path.pop(0)
+        if os.path.isfile(url):
+            module_id = hashlib.md5(os.path.abspath(url).encode("utf-8")).hexdigest()
+            module_name = f"{module_name}_{module_id}"
+            if module_name not in sys.modules:
+                spec = util.spec_from_file_location(module_name, url)
+                if spec is None or spec.loader is None:
+                    raise ImportError(f"not found module spec({module_name})")
+                mod = util.module_from_spec(spec)
+                sys.modules[module_name] = mod
+                loaded_module = module_name
+                spec.loader.exec_module(mod)
+        else:
+            importlib.import_module(module_name)
     except Exception as err:
+        if loaded_module:
+            sys.modules.pop(loaded_module, None)
         raise RuntimeError(f"load module(url={url}) failed, error: {err}") from err
+    finally:
+        sys.path.pop(0)


avinxshKD:
@MooreZheng @hsj576  kept this intentionally small since the bug is isolated to `load_module()`.
Pls take a look, thanks

Omkeswani27:
Tested this branch directly (`git fetch origin pull/558/head:pr-558`):

load_module leaked? False
py2dict leaked? True
collision test: plugin_<hash1>=one, plugin_<hash2>=two, 2 unique modules — collision fixed

Confirms the collision fix works correctly. One gap not mentioned in the PR
description: `py2dict()` is untouched, so its sys.path leak (#728) is still
present — only `load_module`'s leak is incidentally fixed by this refactor.
Full analysis and recommendation (major revision):

Yash4616:
Ran this against a reproduction of the module path in `core/testcasecontroller/algorithm/module/module.py`, where `class_factory_type` defaults to `ClassType.GENERAL` at line 90, `utils.load_module(self.url)` runs at line 125, and `ClassFactory.get_cls(...)` at lines 130 to 133. Loading `examples/llm_simple_qa/testalgorithms/gen/basemodel.py` then `examples/government_rag/singletask_learning_bench/testalgorithms/basemodel.py`, which share a basename:

```
sys.modules basemodel entries : 2
    basemodel_89bd9fe3e686c8ac0859bb40ec52f5e7    -> ORIGIN=llm_simple_qa
    basemodel_cb2e9bec7967a0ed0091abd7ab5f6d06    -> ORIGIN=government_rag
module bodies executed        : 2  ['llm_simple_qa', 'government_rag']
algorithm 'llm_simple_qa'  resolved model from -> llm_simple_qa OK
algorithm 'government_rag' resolved model from -> government_rag OK
```

The hash suffixes differ between runs, since this PR hashes the absolute path and my harness loads from a fresh temporary directory. What matters is that there are two distinct entries rather than one.

On `main` the second algorithm resolves to the first one's model. This PR produces two genuinely distinct modules, and the hashed-abspath approach also caches correctly on a repeat load of the same file, so import-time side effects do not repeat.

One thing worth changing before merge. The cleanup is still `finally: sys.path.pop(0)`, which removes whatever sits at index 0 at that moment. If the module being imported inserts at index 0 during its own import, `pop(0)` removes the module's entry and leaves the one this function inserted. That is reachable: `examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/basemodel.py:29` inserts at index 0 during its own import, and is loaded through `load_module`.

Snapshotting and restoring, as #806 does, is correct in that case and is a small change here.

Also worth noting this branch is 17 commits behind `main` and conflicts with each of #611, #806, #729 and #759 in `core/common/utils.py`.

**Recommendation: major revision.** The identity fix is correct and is the only one in the open set. The `pop(0)` change is small but it is a code change.

Aryansingh-ai:

I checked out PR #558 and tested the modified `core/common/utils.py` directly against the dynamic-loading properties identified in #557 and #728.

### 1. Basename collision — PASS

I created two files with the same basename:

```text
one/plugin.py → VALUE = "one"
two/plugin.py → VALUE = "two"

and loaded them sequentially.

Unlike main, where the second load resolves to the first module, #558 creates distinct module identities:

plugin_8d838379969a9c2bdad7ff2a9b2a452f
→ one/plugin.py
→ VALUE = one

plugin_7035ef84a4a33b89155d62220562b323
→ two/plugin.py
→ VALUE = two

The important result is that the two files no longer share the same sys.modules key.

Therefore, the specific basename-collision property from #557 passes on this PR.

2. Failed module execution cleanup — PASS

I also forced a dynamically loaded module to fail during execution.

Observed result:

ERROR: RuntimeError
MESSAGE: load module(url=...broken.py) failed, error: intentional test failure
PATH_UNCHANGED: True
STALE_MODULE_PRESENT: False

So the tested failure path does not leave the failed module registered in sys.modules.

3. Why this differs from #729 / #759 / #806

The cleanup-only PRs (#729, #759 and #806) restore sys.path, but my controlled collision test still produces:

AFTER FIRST: one
AFTER SECOND: one
EXPECTED SECOND: two

#558 addresses a different layer of the problem: module identity.

Conceptually:

#729 / #759 / #806
        ↓
restore temporary import state

#558
        ↓
separate module identities for different file paths

This is why fixing #728 alone does not eliminate #557.

4. Review concern — module identity semantics

The hash-derived module-name approach is a stronger architectural response to the collision because different paths receive different sys.modules identities.

However, this changes the semantics of dynamic loading compared with normal basename-based imports.

Before merge, the implementation should therefore be reviewed for compatibility with:

modules that import other modules internally;
relative/package-dependent imports;
repeated loading of the same file;
callers that rely on the previous module naming behavior.

These are compatibility considerations rather than failures demonstrated by my current tests.

Recommendation: ACCEPT WITH MAJOR REVISION

The core direction is strong and the controlled collision test passes.

I recommend ACCEPT WITH MAJOR REVISION because the change modifies the module-loading/identity mechanism rather than only repairing cleanup.

The final implementation should be coordinated with the other open loader PRs rather than merged blindly alongside them.

In particular, #558 and #611 represent different approaches to module identity and should be treated as competing architectural choices.

**Review by henryng15 — COMMENTED** [review]
Reviewed at head `b99161f` against `main` `37a9c60`.

The identity fix is the right direction and it works — @Omkeswani27, @Yash4616 and
@Aryansingh-ai have already established that, and I take it as given rather than
re-testing it. Two things remain that I could not find raised on this PR.

## 1. The fix does not extend to modules that the loaded file imports

@Aryansingh-ai listed this as a compatibility consideration and noted it was not
demonstrated. It is demonstrable. Fixture mirrors the repository's real shape: two
Examples, each with its own `basemodel.py` (41 such files exist) importing its own
sibling `utils.py` (10 such files exist).

```text
--- main @ 37a9c60 ---
  load  resolved sys.modules key                      its code    its utils   predict(1)  #keys
  A     basemodel                                     example_A   example_A   1           1
  B     basemodel                                     example_A   example_A   1           1

--- pr-558 @ b99161f ---
  load  resolved sys.modules key                      its code    its utils   predict(1)  #keys
  A     basemodel_f3f64bab699cb16b508b9c58d085eba3    example_A   example_A   1           1
  B     basemodel_0b091c108d51ae40d4e1078665699b86    example_B   example_A   1           2

  sys.modules keys for utils: utils        <-- one key, two different files
```

`example_B`'s own `basemodel` now loads correctly — the PR working as intended. But it
still binds `example_A`'s `utils`, so `predict(1)` returns `1` where the Example's own
helper returns `100`. The hashed identity applies only to the file `load_module()` opens;
anything that file imports still resolves by bare name through the `sys.path` entry the
function inserts just above.

Why this is worth closing before merge: on `main`, `example_B` never executes at all,
which is visibly broken. Under this PR it executes and returns a plausible wrong number.
For a benchmarking framework that is the harder failure to notice, and it becomes harder
still once the obvious collision is gone.

Reproduce: `python3 tools/probe_pr558_transitive.py ianvs` —

## 2. The hashed `__name__` removes pickle recoverability (latent)

Stating the boundary first: **no shipped Example is broken by this today.** I checked.
The issue is that the PR removes a property without recording that it did.

`sedna/core/lifelong_learning/lifelong_learning.py` and
`knowledge_management/*.py` persist the knowledge base through
`FileOps.dump` → `joblib.dump`. pickle does not serialise a class; it stores
`obj.__class__.__module__` as a string and re-imports that name on load. This PR changes
that string from the bare basename to `<basename>_<md5 of the absolute path>`.

Same fixture, both refs, unpickled in a fresh interpreter:

```text
--- main @ 37a9c60 ---
  CLASS_MODULE                      taskdef
  UNPICKLE (source dir on sys.path) LOAD-OK TaskDefinitionByDomain('Cityscapes')

--- pr-558 @ b99161f ---
  CLASS_MODULE                      taskdef_401cfc800cc0a1806b5a8207c97ee637
  UNPICKLE (source dir on sys.path) LOAD-FAIL ModuleNotFoundError: No module named
                                    'taskdef_401cfc800cc0a1806b5a8207c97ee637'
```

On `main` the artifact is recoverable by exactly the `sys.path` insertion this function
already performs. After this PR the recorded name matches no file on disk under any name.
The md5 covers the *absolute* path, so it also differs per checkout directory — across
two runs from different temp directories the suffix was `c11480c6…` then `401cfc80…` for
the identical source file.

**Why it does not fire today, verified:** `task_index` is
`{"extractor": self.extractor, "task_groups": self.task_groups}`; `Task` and `Model` are
Sedna classes; `t.model` is replaced by a path string before the dump; and every shipped
`task_definition` module returns the extractor as a plain `dict`
(`robot-cityscapes-synthia/.../task_definition_by_domain.py:58`,
`robot/.../task_definition_by_origin.py:65`,
`cityscapes-synthia/.../task_definition_by_origin.py:81`, and the `bdd` variant). So the
pickle currently holds no Example-defined class.

But `task_definition`, `task_relationship_discovery` and `unseen_task_detect` exist
specifically so an Example can supply its own class, and their return values flow
straight into that pickle. Nothing states they must not. Today such an artifact is
recoverable; after this PR it is permanently unloadable, with no diagnostic beyond a
`ModuleNotFoundError` naming a module that never existed.

One option that keeps everything this PR gains: key `sys.modules` uniquely as you do, but
leave `mod.__name__` at the bare basename. If you prefer the hashed `__name__`, then
writing the invariant into the docstring would be enough.

Reproduce: `python3 tools/probe_pr558_pickle.py ianvs`

## Recommendation

**Major revision** — for (1), which produces a wrong benchmark value. (2) is a release
note or a one-line docstring, your call.

PR #560, opened 2026-06-20
Fix rank picture mode without matrix

### LINKED ISSUE #559: rank: selected_and_all_and_picture crashes when Matrix metric is missing
**What happened**:

`selected_and_all_and_picture` assumes every test result has a `Matrix` metric.

For normal benchmarks with scalar metrics only, `_draw_pictures()` does this:

```
python
for key in matrix.keys():
```

So matrix is None and rank saving crashes after the benchmark result is already available.

Rank CSV files should still be saved.
If picture mode needs Matrix, Ianvs should either skip picture generation for results without Matrix, or fail with a clear error.

**How to reproduce it:**

Use save_mode: selected_and_all_and_picture with a result that only has scalar metrics, for example:

`{"accuracy": 0.9}`

Actual error:
AttributeError: 'NoneType' object has no attribute 'keys'

Anything else we need to know?:
This is in core/storymanager/rank/rank.py, inside _draw_pictures().

avinxshKD:
**What type of PR is this?**

/kind bug

`selected_and_all_and_picture` currently assumes every result has a `Matrix` metric.

For normal benchmarks with scalar metrics only, `_draw_pictures()` gets `None` and crashes on `matrix.keys()`, even though the rank CSV files can be saved.

Fixes #559

### DIFF
--- a/core/storymanager/rank/rank.py
+++ b/core/storymanager/rank/rank.py
@@ -236,6 +236,8 @@ class Rank:
             out_put = test_case.output_dir
             test_result = test_results[test_case.id][0]
             matrix = test_result.get("Matrix")
+            if not matrix:
+                continue
             for key in matrix.keys():
                 draw_heatmap_picture(out_put, key, matrix[key])
 


avinxshKD:
pls take a look, minimal fix @MooreZheng @hsj576

azaynul10:
The diagnosis in #559 is right about the line that raises, and the guard here is
the right shape for it. What I could not reconcile is which population it applies
to, and I think the condition is true on every run rather than only on
scalar-metric ones.

`test_result` is built in `core/testcasecontroller/testcase/testcase.py:113-117`,
keyed by the metric names declared in the testenv. Every stage that produces the
matrix metric spells it `MATRIX`:

| Stage | Location | Key |
|---|---|---|
| enum | `core/common/constant.py:96` | `MATRIX = "MATRIX"` |
| paradigm initialises | `.../paradigm/lifelong_learning/lifelong_learning.py:67` | `SystemMetricType.MATRIX.value` |
| paradigm fills | same file, `:165`, `:261` | `SystemMetricType.MATRIX.value` |
| metric function reads | `core/testcasecontroller/metrics/metrics.py:127` | `SystemMetricType.MATRIX.value` |
| shipped testenv declares | `testenv-robot-small.yaml:37`, `testenv-robot.yaml:37`, `robot-cityscapes-synthia/.../testenv.yaml:36` | `- name: "MATRIX"` |

`rank.py:238` is the only line in the repository that spells it `Matrix`. Scanning
every literal spelling under `core/` at `37a9c60` returns exactly two, and they are
the definition and the lookup:

```
MATRIX   core/common/constant.py:96              MATRIX = "MATRIX"
Matrix   core/storymanager/rank/rank.py:238      matrix = test_result.get("Matrix")
```

Dict lookups are case sensitive, so `.get("Matrix")` returns `None` whatever the
benchmark declares.

## What that does to this patch

I ran the loop body from this diff, verbatim, with the drawing call replaced by a
counter, against a result dict keyed the way `testcase.py:115` keys it:

```
result stores 'MATRIX'   -> no error: completed, draw_heatmap_picture reached 0 time(s)
result stores 'Matrix'   -> no error: completed, draw_heatmap_picture reached 1 time(s)
```

The second row is the important control: the guard itself behaves correctly. It is
the condition that is unconditional. `MATRIX` is what the pipeline produces, so
`continue` is taken for every test case, on every run, and `draw_heatmap_picture`
is never reached.

The effect is that `selected_and_all_and_picture` becomes indistinguishable from
`selected_and_all`: CSVs are written, no picture is produced, and the job exits 0
reporting success. That is the same failure class as #828, where an unrecognised
`save_mode` silently skips leaderboard output. Since this PR declares `Fixes #559`,
merging it would also close the only open report of the behaviour, and the
`AttributeError` is currently the sole signal that the mode does not work.

## Interaction with #629

#629 creates the `output/` subdirectory inside `draw_heatmap_picture`. That fix is
correct and necessary, and it sits downstream of this line. Confirming the ordering
on the unpatched tree:

```
result stores 'MATRIX' -> AttributeError: 'NoneType' object has no attribute 'keys'
result stores 'Matrix' -> FileNotFoundError: .../output/BWT-heatmap.png
same, with output/ present -> no error: completed
```

So there are three layers here, and only the lower two have issues filed: the key
mismatch, then the missing directory, then a working mode. If this PR merges,
`continue` is taken before `draw_heatmap_picture` is called, and #629's new lines
never execute. Both PRs would merge green, #559 and #628 would both close, and the
mode would still draw nothing.

That also explains something about #628 that puzzled me initially: its reproduction
calls `plt.savefig` directly rather than running a benchmark. It has to, because
the state it describes — Matrix data present and valid — is not reachable through
`Rank` while line 238 asks for a key nothing writes.

## Suggestion

Correcting the key first would make this guard do what it was written to do. The
smallest version binds the lookup to the enum so it cannot drift again:

```python
from core.common.constant import SystemMetricType

matrix = test_result.get(SystemMetricType.MATRIX.value)
if not matrix:
    continue
```

With that in place the guard becomes a real guard — it fires for benchmarks that
genuinely do not declare the metric, which is most of them — and #629 becomes
reachable and necessary. The ordering matters more than the authorship: as it
stands, either PR merging alone leaves the mode broken in a quieter way than it is
now.

Happy to be wrong about any of this. The scan and the three runs are reproducible
with the standard library plus `matplotlib`, no dataset or weights involved:

```python
import matplotlib; matplotlib.use("Agg")
import tempfile

class Case:
    id = "tc-0"
    def __init__(self, d): self.output_dir = d

rank = Rank({"save_mode": "selected_and_all_and_picture"})
for key in ("MATRIX", "Matrix"):
    with tempfile.TemporaryDirectory() as d:
        case = Case(d)
        result = {"accuracy": 0.9, key: {"BWT": [[0.1, 0.2], [0.3, 0.4]]}}
        try:
            rank._draw_pictures([case], {case.id: [result, 1.0]})
            print(key, "-> completed")
        except Exception as err:
            print(key, "->", type(err).__name__, err)
```

## Boundary

Executed against `37a9c60`: the spelling scan across every `.py` under `core/`, the
metric-name scan across every `.yaml` under `examples/`, and the runs above calling
the real `Rank._draw_pictures`. Not executed: a full benchmark run. The one shipped
config that selects this mode,
`examples/robot/lifelong_learning_bench/semantic-segmentation/benchmarkingjob-sam.yaml`,
needs the robot semantic-segmentation corpus and RFNet/SAM weights, and its line 10
points its testenv at a path that does not exist in the tree — the file is under
`semantic-segmentation/testenv/`, not `lifelong_learning_bench/testenv/`. The claim
that `save()` reaches `_draw_pictures` is static: `rank.py:285` is unconditional
inside the picture branch.

Issue #460, opened 2026-05-23
Issue summarization for example of `MOT17/multiedge_inference_bench`


suhaan-24:
> **Author:** suhaan-24  
> **Pre-task 1:** LFX Mentorship 2026 Term 2  
> **Date:** 2026-05-21  
>
> **Scope:** This analysis extends prior work by covering **9 tracked GitHub issues** (including 4 newly filed in 2026) and **13 original static-analysis findings** not present in any prior submission or automated review. The closed LFX Term 3 2025 pre-task [#231](https://github.com/kubeedge/ianvs/issues/231) is included as historical context; all findings in that submission that remain unresolved are re-examined here with updated evidence from the current `main` branch. Gemini-code-assist has reviewed PRs #392, #398, #400, #402, #336, and #309; every section below that overlaps with a gemini comment explicitly labels the overlap and states what this analysis adds beyond it. 13 original static-analysis findings have been identified and filed as GitHub issues (#443, #444, #445, #447, #448, #449, #450, and a comment on #391), none of which were present in any prior submission or automated review.

---

## 1. Background

The `MOT17/multiedge_inference_bench` example benchmarks pedestrian tracking and re-identification (ReID) in a cloud-edge collaborative setting using the [MOT17](https://motchallenge.net/) dataset. It exercises ianvs's **MultiEdge Inference** paradigm by running two sequential benchmark jobs: a tracking job (ByteTrack detector) and a ReID job (M3L feature extractor), then merging the results into a PDF report via `generate_reports.py`. This example is the sole representative of `paradigm_type: multiedgeinference` in the entire ianvs example suite, making every bug in it a unique barrier to validating this paradigm.

As of 2026-05-21, the example is **entirely non-runnable** due to a combination of stale path references, runtime crashes, missing dependencies, and core framework regressions. The section below catalogues every relevant GitHub issue, followed by 13 original static-analysis findings, now filed as GitHub issues (#443–#450 and a comment on #391).

[SPECIFIC GREEN CHANNEL](https://drive.google.com/file/d/1idk3bpXMIYGA-2xP2enmxCSFC-K6gDOG/view?usp=sharing)

---

## 2. Table of Contents — All Linked Issues

| # | Issue | Category | Status | Link |
|---|-------|----------|--------|------|
| 1 | MOT17 example fails to run due to broken "pedestrian_tracking" path references | Path / Configuration | Open | [#391](https://github.com/kubeedge/ianvs/issues/391) |
[...]

Issue #132, opened 2024-08-13
Complete the dependencies and upgrade the interface to the new version.


FuryMartin:
**What should be cleaned up or changed**:

## Complete the dependencies
The current Ianvs project has issues with **unclear and incomplete dependencies**, as pointed out in issues #103 and #106. 

This poses significant challenges for users to configure the Ianvs environment. 

I suggest:
-  Completing  the dependencies for Ianvs Core, and updating [requirements.txt](https://github.com/kubeedge/ianvs/blob/main/requirements.txt)

## Update outdated dependencies and align interface to new version 

Meanwhile, some dependencies are severely outdated compared to the main versions, and **certain APIs from older versions have been deprecated**.  These outdated dependencies may cause confusion for users.

For example: 
  - `pandas` is needed by `core/storymanager/rank/rank.py` with version `1.1.5` , while the latest version has reached `2.2.2` with many APIs deprecated
  - `sedna` has released `0.6.0`, while the built-in package in Ianvs is `0.4.1`

I suggest:
- Updating outdated packages and upgrading the interface to the new version. (e.g. pandas interface in `rank.py`)

## Remove `examples/resources/`
There are also some dependencies with projects under `examples/`. 

Since these projects are not closely related to `Ianvs Core`, why should their dependencies be placed in `examples/resources`?  

We can provide an intuitive `requirements.txt` for each example, which aligns with Python usage habits and helps reduce the introduction of large binary packages into the Git repository.

I suggest:
- Remove the method of managing third-party dependencies through `examples/resources/`, and embed dependencies of `examples/xxxxx` into a `requirements.txt` within each example directory.

**Why is this needed**:
Improving dependencies and aligning the new version interface will help Ianvs solve complex legacy dependency issues, allowing users to quickly get started with Ianvs.

FuryMartin:
Considering that the API of dependencies in different versions of Python may vary significantly, we need to determine the minimum Python version supported by Ianvs. 

## Experiment on Quick Start - The PCB-AoI Example 

I followed the [Quick Start](https://ianvs.readthedocs.io/en/latest/guides/quick-start.html) guide to run the PCB-AoI example. However, during the attempt, I found that this example currently has an irreconcilable dependency conflict.

I will provide the results of dependency configuration for different Python version environments:

### Python 3.6

#### Step 1. Ianvs Preparation

We first need to install the `sedna-0.4.1-py3-none-any.whl` package located in `examples/resources/third_party/`.
[...]

PR #561, opened 2026-06-21
Fix/mot17 issues 447 449 450

### LINKED ISSUE #447: [MOT17] ReID metric evaluation crashes on edge cases: `IndexError` in rank metrics; unhandled `RuntimeError` in `mAP.py`
### Summary

Two edge-case crash bugs in the MOT17 ReID metric scripts cause the evaluation
pipeline to fail completely rather than degrade gracefully. In `rank_1.py`, `rank_2.py`,
and `rank_5.py`, an unguarded array index raises `IndexError` when any query person has
no matching identity in the gallery. In `mAP.py`, an unhandled `RuntimeError` is raised
when the query set contains no valid matches at all, surfacing as a generic pipeline
failure with no actionable diagnostic information.

### Environment

- **Files affected:**
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/reid/rank_1.py`
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/reid/rank_2.py`
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/reid/rank_5.py`
  - `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/testenv/reid/mAP.py`

---

#### Finding 1 — `IndexError` in `rank_1/2/5.py` when a query has no gallery match

**Files:** `testenv/reid/rank_1.py:31`, `rank_2.py:31`, `rank_5.py:31`  
**Code:**
```python
def cmc(distmat, query_ids, gallery_ids, topk):
    m, _ = distmat.shape
    indices = np.argsort(distmat, axis=1)
    matches = (gallery_ids[indices] == query_ids[:, np.newaxis])
    ret = np.zeros(topk)
    for i in range(m):
        k = np.nonzero(matches[i])[0][0]   # ← crash here
        if k < topk:
            ret[k] += 1
    return round(float(ret.cumsum()[-1] / m), 4)
```

**Problem:**  
For query person `i`, `matches[i]` is a boolean array indicating which gallery entries
share the same identity. If no gallery entry matches — a valid condition when a person
appears only in the query split — `np.nonzero(matches[i])` returns an empty array.
Indexing it with `[0][0]` raises:

```
IndexError: index 0 is out of bounds for axis 0 with size 0
```

The outer loop has no guard for this case, so a single unmatched query aborts the
entire CMC computation. The same pattern is duplicated identically across all three
rank metric files.

**Impact:**  
A single query with no gallery counterpart destroys the entire Rank-1, Rank-2, and
Rank-5 evaluation. This is not a data error — it is a valid edge case in standard
ReID dataset splits where the query and gallery identities do not perfectly overlap.
The crash leaves the leaderboard with no rank metrics at all.

**Proposed Fix (apply identically to `rank_1.py`, `rank_2.py`, `rank_5.py`):**
```python
for i in range(m):
    nonzero = np.nonzero(matches[i])[0]
    if len(nonzero) == 0:
        continue  # no gallery match for this query; skip gracefully
    k = nonzero[0]
    if k < topk:
        ret[k] += 1
```

---

#### Finding 2 — Unhandled `RuntimeError` in `mAP.py` on empty query set

**File:** `testenv/reid/mAP.py`  
**Lines:** 37–38  
**Code:**
```python
if len(aps) == 0:
    raise RuntimeError("No valid query")
return round(float(np.mean(aps)), 4)
```

**Problem:**  
When every query in the evaluation set has no matching gallery identity, the `aps` list
remains empty throughout the per-query loop and `RuntimeError("No valid query")` is
raised. This exception propagates through the ianvs pipeline as a generic:

```
testcase runs failed, error: No valid query
```

There is no context in the error message indicating which metric failed, which dataset
split caused it, or that the issue is in `mAP.py` specifically. Users cannot determine
from this message alone whether the problem is in the dataset, the model, or the
metric implementation.

**Impact:**  
The crash prevents any mAP score from being recorded. The pipeline treats this as a
fatal error, potentially aborting the evaluation of all subsequent metrics. Returning
`0.0` with a log warning is more appropriate: it records the degenerate result, allows
the pipeline to continue, and provides a signal in the leaderboard that something is
wrong with the dataset or model — without causing a hard failure.

**Proposed Fix:**
```python
import logging

logger = logging.getLogger(__name__)

if len(aps) == 0:
    logger.warning(
        "mAP: no query with a matching gallery identity found. "
        "Check dataset split. Returning 0.0."
    )
    return 0.0
return round(float(np.mean(aps)), 4)
```

---

### Steps to Reproduce

**Finding 1 (IndexError):**
1. Prepare a ReID dataset split where at least one query person ID does not appear in
   the gallery (this occurs naturally with certain train/val splits of MOT17-ReID).
2. Run the ReID benchmark job:
   ```bash
   ```
3. Observe `IndexError: index 0 is out of bounds for axis 0 with size 0` in the rank
   metric evaluation step.

**Finding 2 (RuntimeError):**
1. Use a dataset split where no query identity has any gallery match.
2. Run the ReID benchmark job as above.
3. Observe `testcase runs failed, error: No valid query` with no further diagnostic
   information in the pipeline output.

---

### Additional Notes

Both fixes are confined to the example-local metric scripts and do not touch the ianvs
core. All three rank metric files (`rank_1.py`, `rank_2.py`, `rank_5.py`) contain
identical code and must be patched together to avoid partial fixes.

The graceful-degradation approach (skip or return `0.0`) is consistent with standard
ReID evaluation practice, where partial query sets are common during development and
cross-dataset transfer evaluations.

### LINKED ISSUE #449: [MOT17] Missing `requirements.txt` and Docker-specific workspace path block bare-metal setup

### LINKED ISSUE #450: [Core] Unconditional `import onnx` in `multiedge_inference.py` fails for users who do not need model partitioning
### Summary

`core/testcasecontroller/algorithm/paradigm/multiedge_inference/multiedge_inference.py`
imports `onnx` at the top of the module unconditionally, making it a hard dependency
for all users of the `multiedgeinference` paradigm — including the MOT17 example, which
uses the direct inference path and never invokes the partitioning code. Because `onnx`
is not listed in the core ianvs `requirements.txt`, this causes an opaque
`ModuleNotFoundError` at import time that gives users no indication that ONNX is only
needed for a specific optional feature.

### Environment

- **Example:** `MOT17/multiedge_inference_bench` (and any other `multiedgeinference`
  paradigm example)
- **Files affected:**
  - `core/testcasecontroller/algorithm/paradigm/multiedge_inference/multiedge_inference.py`

---

#### Finding 1 — Top-level `import onnx` is unconditional

**File:** `core/testcasecontroller/algorithm/paradigm/multiedge_inference/multiedge_inference.py`  
**Line:** 20  
**Code:**
```python
# pylint: disable=E0401
import onnx

from core.common.log import LOGGER
...
```

**Problem:**  
`onnx` is imported at module load time. The `MultiedgeInference` class has two execution
paths:

1. **`_inference()`** — direct single-model inference (used by the MOT17 example).
   Does not use ONNX at all.
2. **`_inference_mp()`** — multi-model partitioned inference via `_partition()`.
   Calls `onnx.utils.extract_model()` to split the ONNX model graph.

The `import onnx` statement is necessary only for path 2. However, because it sits at
module scope, Python attempts to resolve it the moment ianvs imports the paradigm
registry — before any user configuration is read. On a machine without ONNX installed,
this raises:

```
ModuleNotFoundError: No module named 'onnx'
```

The traceback points to the ianvs internals, not the user's code or config, giving
users no indication of which feature triggered the requirement or whether they actually
need ONNX.

**Impact:**

1. **Blocked execution for non-partitioning users.** Any user running the MOT17 example
   (or any future `multiedgeinference` example that uses `_inference()`) must install
   ONNX to get past import. ONNX is a substantial package with many transitive
   dependencies including `protobuf`, `numpy`, and platform-specific wheels.

2. **Masks a subsequent framework bug.** Installing ONNX to satisfy this import then
   exposes a different crash deeper in the pipeline
   (`'MultiedgeInference' object has no attribute 'modules_funcs'`, tracked in issue #93).
   Users are therefore sent through two sequential opaque failures before reaching the
   actual execution path.

3. **Inconsistency with `requirements.txt`.** The core ianvs `requirements.txt` does
   not list `onnx`. Users who install ianvs via the standard documented procedure have
   no indication that ONNX is required until the crash.

**Proposed Fix — lazy import inside `_partition()`:**

Move the `import onnx` statement inside the `_partition()` method, which is the only
code path that actually uses it:

```python
# multiedge_inference.py — remove line 20:
# import onnx   ← delete this

...

def _partition(self, partition_point_list, initial_model_path, sub_model_dir):
    import onnx  # Lazy import: only required when model partitioning is used.
    map_info = dict({})
    for idx, point in enumerate(partition_point_list):
        ...
        try:
            onnx.utils.extract_model(initial_model_path,
                                     sub_model_path, input_names, output_names)
        except Exception as e:
            LOGGER.info(str(e))
        map_info[sub_model_path.split('/')[-1]] = point['device_name']
    return sub_model_dir, map_info
```

This change:
- Has zero effect on users who invoke `_inference()` — ONNX is never imported.
- Preserves full functionality for users who invoke `_inference_mp()` — ONNX is
  imported on demand at the point of use.
- Follows the established Python pattern for optional heavy dependencies
  (same approach used by `matplotlib`, `PIL`, and many other libraries).

---

### Steps to Reproduce

1. Create a fresh environment with ianvs installed but **without** ONNX:
   ```bash
   pip install ianvs
   # Do NOT install onnx
   ```
2. Run any `multiedgeinference` example, e.g.:
   ```bash
   ```
3. Observe the error at startup, before any configuration is processed:
   ```
   ModuleNotFoundError: No module named 'onnx'
   ```
4. Note that the traceback points to the ianvs paradigm import, not to user code,
   and gives no indication that ONNX is only needed for model partitioning.

---

### Additional Notes

This issue affects the entire `multiedgeinference` paradigm, not just the MOT17 example.
Any future example added under this paradigm will inherit the same hard dependency
unless the fix is applied.

Two options exist for the fix:

**Option A (recommended):** Lazy import inside `_partition()` as described above.
This is the cleanest solution and requires no changes to `requirements.txt`.

**Option B:** Add `onnx` to the core `requirements.txt` with an appropriate version
bound. This unblocks the import but still forces all `multiedgeinference` users to
install ONNX regardless of whether they use partitioning. Not recommended.

Related: issue #93 (`modules_funcs` AttributeError) is the next error users encounter
after ONNX is installed. Fixing this import issue makes #93 the first visible failure,
which is a better starting point for debugging the paradigm.

rajdhruvsingh:
**What type of PR is this?**

/kind bug

Fixes three bugs in the `MOT17/multiedge_inference_bench` example with no existing linked PRs, as tracked in the issue summarization #460.

**#447 — IndexError / RuntimeError in ReID metric scripts (`rank_1.py`, `rank_2.py`, `rank_5.py`, `mAP.py`)**
The CMC implementation used `np.nonzero(matches[i])[0][0]` without a length
check, causing an `IndexError` when a query identity has no match in the
gallery — a valid edge case in standard ReID evaluation splits. Fixed by
checking `len(nonzero) == 0` before indexing and skipping such queries.
`mAP.py` raised an unhandled `RuntimeError("No valid query")` when the entire
query set was unmatched. Replaced with a `LOGGER.warning` and a graceful
`return 0.0` so the pipeline continues and surfaces a clear diagnostic message
instead of crashing.

**#449 — Missing `requirements.txt` and Docker-only hardcoded workspace paths**
The example had no `requirements.txt`, forcing contributors to discover missing
packages (`loguru`, `motmetrics`, `mmcv`, `opencv-python`) through repeated
`ImportError` crashes. Added a `requirements.txt` with all required packages
and version bounds (`motmetrics>=1.2.0`, `pandas<2.0`).
Both `tracking_job.yaml` and `reid_job.yaml` hardcoded the workspace path as
`/ianvs/multiedge_inference_bench/workspace`, which only exists inside the
official ianvs Docker container and causes `PermissionError` or
`FileNotFoundError` for all bare-metal and virtualenv users. Replaced with the
relative path `./workspace/multiedge_inference_bench`.

**#450 — Unconditional `import onnx` at module scope in `multiedge_inference.py`**
`onnx` was imported unconditionally at the top of `multiedge_inference.py`,
making it a hard startup dependency for all `MultiedgeInference` users even
though it is only required inside `_partition()`. The MOT17 example never calls
`_partition()`. Traced via `git log -p -- requirements.txt`, the current
`onnx` entry in the root `requirements.txt` was added incidentally by commit
`46340c8` ("Enhanced cloud-edge collaborative inference for llm example") — not
as a deliberate decision for this paradigm. If example-level dependencies are
ever isolated per issue #132, this would regress. Fixed by moving `import onnx`
to a lazy import inside `_partition()` only, with a clear `ImportError` message
for users who attempt to use partitioning without it installed.

Fixes #447
Fixes #449
Fixes #450

### DIFF
--- a/core/cmd/obj/benchmarkingjob.py
+++ b/core/cmd/obj/benchmarkingjob.py
@@ -47,8 +47,8 @@ class BenchmarkingJob:
         self.rank = None
         self.test_env = None
         self.simulation = None
-        self.parallel: bool = False
-        self.max_workers: int|None = None
+        self.parallel = False
+        self.max_workers = None
         self.testcase_controller = TestCaseController()
         self._parse_config(config)
 
--- a/core/cmd/obj/benchmarkingjob.py
+++ b/core/cmd/obj/benchmarkingjob.py
@@ -48,7 +48,7 @@ class BenchmarkingJob:
         self.test_env = None
         self.simulation = None
         self.parallel: bool = False
-        self.max_workers: int = None
+        self.max_workers: int|None = None
         self.testcase_controller = TestCaseController()
         self._parse_config(config)
 
--- a/core/cmd/obj/benchmarkingjob.py
+++ b/core/cmd/obj/benchmarkingjob.py
@@ -47,6 +47,8 @@ class BenchmarkingJob:
         self.rank = None
         self.test_env = None
         self.simulation = None
+        self.parallel: bool = False
+        self.max_workers: int = None
         self.testcase_controller = TestCaseController()
         self._parse_config(config)
 
@@ -91,7 +93,10 @@ class BenchmarkingJob:
         self.testcase_controller.build_testcases(test_env=self.test_env,
                                                  test_object=self.test_object)
 
-        succeed_testcases, test_results = self.testcase_controller.run_testcases(self.workspace)
+        succeed_testcases, test_results = self.testcase_controller.run_testcases(
+            self.workspace,
+            parallel=self.parallel,
+            max_workers=self.max_workers)
 
         if test_results:
             self.rank.save(succeed_testcases, test_results, output_dir=self.workspace)
--- a/core/testcasecontroller/algorithm/paradigm/multiedge_inference/multiedge_inference.py
+++ b/core/testcasecontroller/algorithm/paradigm/multiedge_inference/multiedge_inference.py
@@ -17,7 +17,6 @@
 import os
 
 # pylint: disable=E0401
-import onnx
 
 from core.common.log import LOGGER
 from core.common.constant import ParadigmType
@@ -99,6 +98,14 @@ class MultiedgeInference(ParadigmBase):
 
     # pylint: disable=W0718, C0103
     def _partition(self, partition_point_list, initial_model_path, sub_model_dir):
+        # Fix #450: lazy import — only needed for model partitioning path
+        try:
+            import onnx
+        except ImportError as e:
+            raise ImportError(
+                "onnx is required for model partitioning but is not installed. "
+                "Install with: pip install onnx"
+            ) from e
         map_info = dict({})
         for idx, point in enumerate(partition_point_list):
             input_names = point['input_names']
--- a/core/testcasecontroller/testcase/testcase.py
+++ b/core/testcasecontroller/testcase/testcase.py
@@ -43,12 +43,8 @@ class TestCase:
         self.output_dir = None
 
     def _get_output_dir(self, workspace):
-        output_dir = os.path.join(workspace, self.algorithm.name)
-        flag = True
-        while flag:
-            output_dir = os.path.join(workspace, self.algorithm.name, str(self.id))
-            if not os.path.exists(output_dir):
-                flag = False
+        output_dir = os.path.join(workspace, self.algorithm.name, str(self.id))
+        os.makedirs(output_dir, exist_ok=True)
         return output_dir
 
     def run(self, workspace):
--- a/core/testcasecontroller/testcasecontroller.py
+++ b/core/testcasecontroller/testcasecontroller.py
@@ -57,7 +57,21 @@ class TestCaseController:
         max_workers : int, optional
             Maximum number of parallel workers. Defaults to number of test cases.
         """
+        if isinstance(parallel, str):
+            parallel = parallel.lower() in ("true", "1", "yes")
+
         if parallel:
+            if max_workers is not None:
+                try:
+                    max_workers = int(max_workers)
+                except ValueError:
+                    raise ValueError(
+                f"max_workers must be an integer, got {max_workers}"
+            )
+            if max_workers <= 0:
+                raise ValueError(
+                    f"max_workers must be greater than 0, got {max_workers}"
+                )
             return self._run_testcases_parallel(workspace, max_workers)
         return self._run_testcases_sequential(workspace)
 
@@ -82,29 +96,33 @@ class TestCaseController:
 
         Uses threads rather than processes to avoid pickling issues
         with ML model objects loaded during paradigm execution.
-        """
+        Each test case receives a deep copy of test_env to prevent
+        race conditions from shared state modifications.
+        """ 
+        import copy
         succeed_results = {}
         succeed_testcases = []
         failed_testcases = []
 
-        workers = max_workers or max(len(self.test_cases), 1)
+        # deep copy test cases to avoid shared test_env/dataset
+        # state mutations between parallel threads
+        parallel_testcases = [copy.deepcopy(tc) for tc in self.test_cases]
 
-
-        with ThreadPoolExecutor(max_workers=workers) as executor:
+        with ThreadPoolExecutor(max_workers=max_workers) as executor:
             future_to_testcase = {
-                executor.submit(testcase.run, workspace): testcase
-                for testcase in self.test_cases
-            }
+            executor.submit(testcase.run, workspace): testcase
+            for testcase in parallel_testcases
+        }
 
-            for future in as_completed(future_to_testcase):
-                testcase = future_to_testcase[future]
-                try:
-                    res = future.result()
-                    time = utils.get_local_time()
-                    succeed_results[testcase.id] = (res, time)
-                    succeed_testcases.append(testcase)
-                except Exception as err:
-                    failed_testcases.append((testcase, err))
+        for future in as_completed(future_to_testcase):
+            testcase = future_to_testcase[future]
+            try:
+                res = future.result()
+                time = utils.get_local_time()
+                succeed_results[testcase.id] = (res, time)
+                succeed_testcases.append(testcase)
+            except Exception as err:
+                failed_testcases.append((testcase, err))
 
         if failed_testcases:
             error_msgs = [
@@ -112,8 +130,8 @@ class TestCaseController:
                 for tc, err in failed_testcases
             ]
             raise RuntimeError(
-                f"{len(failed_testcases)} testcase(s) failed:\n" +
-                "\n".join(error_msgs)
+            f"{len(failed_testcases)} testcase(s) failed:\n" +
+            "\n".join(error_msgs)
             )
 
         return succeed_testcases, succeed_results
--- a/core/testcasecontroller/testcasecontroller.py
+++ b/core/testcasecontroller/testcasecontroller.py
@@ -20,6 +20,7 @@ from core.common import utils
 from core.common.constant import TestObjectType
 from core.testcasecontroller.algorithm import Algorithm
 from core.testcasecontroller.testcase import TestCase
+from concurrent.futures import ThreadPoolExecutor, as_completed
 
 
 class TestCaseController:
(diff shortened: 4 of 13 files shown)


suhaan-24:
Reviewed the two changes separately since they are quite different in scope.

## The `onnx` lazy import is well-targeted

```python
+        try:
+            import onnx
+        except ImportError as e:
+            raise ImportError(
+                "onnx is required for model partitioning but is not installed. "
+                "Install with: pip install onnx"
+            ) from e
```

Moving this off module scope means users who do not exercise the model-partitioning path no longer
need `onnx` installed to import the paradigm at all, and the message names the remedy. Narrow and
clearly correct.

## The parallel execution change interacts with a known-racy path

`self.parallel` / `max_workers` threading through `run_testcases` is opt-in and defaults to `False`,
which is the right default. The concern is what the concurrent path touches.

`TestCase._get_output_dir` resolves the output directory with a check-then-create against a shared
workspace:

```python
        flag = True
        while flag:
                flag = False
        return output_dir
```

Under sequential execution the `os.path.exists` check and the subsequent directory creation cannot
interleave. Once test cases run through a pool, two of them can both observe the path absent and
both proceed — a straightforward TOCTOU. The collision is unlikely given `uuid.uuid1()` ids, but the
check itself becomes meaningless under concurrency, and if it ever does collide the loop above spins
forever because `self.id` is never regenerated inside it.

Worth noting that #622 is separately patching this exact method for the infinite-loop half. If both
land, the interaction is worth a look: #622 regenerates `self.id` on collision, which is safe
sequentially but under a pool means two workers can regenerate concurrently against the same
directory.

### Suggestion

`os.makedirs(output_dir, exist_ok=False)` inside a `try/except FileExistsError` would make the
create atomic and remove the check-then-create entirely, which is robust under both execution modes.

## Smaller notes

- Adding a docs proposal alongside the code (`docs/proposals/chore/parallel_testcase_execution.md`)
  is helpful — worth making sure it states the default is off.
- A 13-file diff mixing an execution-model change, a paradigm import fix, and MOT17 example fixes is
  hard to bisect if a regression appears later. Splitting the parallel-execution change out would
  let the smaller fixes land independently.

## Verification boundary

**Not executed.** MOT17's dataset was not obtained, so neither the example fixes nor the parallel
path were run. The TOCTOU concern is derived from reading `_get_output_dir` on `upstream/main`
against this PR's threading change — it is a code-path risk, not a reproduced race.

Yahdel:
The ONNX part of this PR is the right idea. ONNX is only needed when a user asks Ianvs to split a model. Moving that import into the split-model function means normal use does not require ONNX just to start.

My concern is that this PR contains much more than the ONNX change. It also changes parallel test execution, file paths, requirements, and result metrics. Those are separate changes with separate risks, so they are harder to review and test as one package.

I suggest:

1. keep the small ONNX change, or put it in its own PR;
2. test normal MOT17 use with ONNX missing;
3. test the split-model option with ONNX installed;
4. check that asking to split a model without ONNX gives a clear message;
5. review the parallel-execution and metric changes separately.

This is separate from the command-startup fix in #811. #811 keeps version and help from reaching benchmark code at all. This PR controls when ONNX is needed after a real benchmark has started.

I reviewed the code path but did not run the full MOT17 benchmark, GPU inference, or the new parallel behavior. Full context: Discussion #961.

PR #564, opened 2026-06-23
fix: replace JsonlDataParse with JSONDataParse to match Sedna API

### LINKED ISSUE #563: [Bug] Core setup failures on fresh install: API rename, missing class, wrong docs path, undocumented deps
### Environment
- OS: Ubuntu (Linux)
- Python: 3.11
- Ianvs: latest main branch
- Sedna: installed from kubeedge/sedna github main branch

## Summary
During a fresh setup using the official quick start guide, I encountered multiple blocking issues that are not currently documented. As it stands, a new contributor is unlikely to run any example successfully without running into these problems.

## Bug 1: Incorrect docs path for PCB-AoI example
The path mentioned in the documentation is incorrect.

**Documented path:**
examples/pcb-aoi/singletask_learning_bench/testalgorithms/fpn/benchmarkingjob.yaml

**Actual path in the repository:**
examples/pcb-aoi/singletask_learning_bench/fault_detection/benchmarkingjob.yaml

The `fpn/` directory does not exist at the documented location, so the docs should be updated to point to the correct path

## Bug 2: `JsonlDataParse` was renamed in Sedna, but Ianvs still uses the old name
`core/testenvmanager/dataset/dataset.py` imports `JsonlDataParse` from 
`sedna.datasources`, but in Sedna this class has been renamed to `JSONDataParse`. 

**Current error:**
ImportError: cannot import name 'JsonlDataParse' from 'sedna.datasources'

**Required fix:** 
Replace `JsonlDataParse` with `JSONDataParse` in dataset.py lines 25 and 587.

## Bug 3: `JSONMetaDataParse` is imported in Ianvs, but does not exist in Sedna
`dataset.py` imports `JSONMetaDataParse` from `sedna.datasources` but 
this class does not exist anywhere in Sedna. 

Running `dir(sedna.datasources)` shows that `JSONDataParse` is available, but `JSONMetaDataParse` is not.

This breaks the` JSONFORLLM` dataset format at import time, which means all LLM-based examples depending on it currently fail before execution starts.

**Error:**
ImportError: cannot import name `JSONMetaDataParse` from `sedna.datasources`

## Bug 4: PCB-AoI example depends on undocumented packages
The PCB-AoI example requires additional packages that are not documented anywhere in the repository.

At minimum, the example depends on:

- `opencv-python`
- `tensorflow`

These dependencies are not listed in `requirements.txt` and are not mentioned in the example README or setup instructions.

As a result, a fresh setup fails with `ModuleNotFoundError`,, and new contributors are given no indication of which packages need to be installed.

## Reproduction steps
1. Clone kubeedge/ianvs
2. Run `pip install -e` .
3. Install sedna from GitHub source
4. python benchmarking.py -f examples/pcb-aoi/singletask_learning_bench/fault_detection/benchmarkingjob.yaml
5. Hit the errors above in sequence 

## Suggested fixes
- Update docs path from `fpn/` to `fault_detection/`
- Replace `JsonlDataParse` with `JSONDataParse` in dataset.py
- Implement `JSONMetaDataParse` in Sedna or create a fallback in Ianvs
- Add `requirements.txt` to each example folder listing its dependencies

31groot:
**What type of PR is this?**
/kind bug

`dataset.py` was importing `JsonlDataParse` and `JSONMetaDataParse` from 
`sedna.datasources`, but these names do not exist in current Sedna.
Sedna renamed `JsonlDataParse` to `JSONDataParse` and `JSONMetaDataParse` 
was never implemented. This caused an ImportError on every fresh install, 
which blocks all examples from running.

Changes:
- Replace `JsonlDataParse` with `JSONDataParse`
- Remove duplicate `JSONDataParse` import
- Replace `JSONMetaDataParse` with `JSONDataParse` as fallback for JSONFORLLM format

Fixes #563

### DIFF
--- a/core/testenvmanager/dataset/dataset.py
+++ b/core/testenvmanager/dataset/dataset.py
@@ -22,8 +22,6 @@ from sedna.datasources import (
     CSVDataParse,
     TxtDataParse,
     JSONDataParse,
-    JsonlDataParse,
-    JSONMetaDataParse
 )
 
 from core.common import utils
@@ -575,20 +573,20 @@ class Dataset:
             data = CSVDataParse(data_type=data_type, func=feature_process)
             data.parse(file, label=label)
 
-        if data_format == DatasetFormat.TXT.value:
+        elif data_format == DatasetFormat.TXT.value:
             data = TxtDataParse(data_type=data_type, func=feature_process)
             data.parse(file, use_raw=use_raw)
 
-        if data_format == DatasetFormat.JSON.value:
+        elif data_format == DatasetFormat.JSON.value:
             data = JSONDataParse(data_type=data_type, func=feature_process)
             data.parse(file)
 
-        if data_format == DatasetFormat.JSONL.value:
-            data = JsonlDataParse(data_type=data_type, func=feature_process)
+        elif data_format == DatasetFormat.JSONL.value:
+            data = JSONDataParse(data_type=data_type, func=feature_process)
             data.parse(file)
 
-        if data_format == DatasetFormat.JSONFORLLM.value:
-            data = JSONMetaDataParse(data_type=data_type, func=feature_process)
+        elif data_format == DatasetFormat.JSONFORLLM.value:
+            data = JSONDataParse(data_type=data_type, func=feature_process)
             data.parse(file, **kwargs)
 
         return data


Kingthnight:
### Review: `JSONDataParse` does not seem compatible with the bundled Sedna version

I tried this locally using the Sedna 0.6.0.1 wheel already included in `resources/third_party`.

That wheel still contains `JSONDataParse`, `JsonlDataParse`, and `JSONMetaDataParse`. They are not aliases for the same parser. Using the original specialized parsers, my small test files were parsed correctly:

```text
JsonlDataParse:
x: ['What is 2 + 2?']
y: ['A']

JSONMetaDataParse:
dataset_name: synthetic
x: ['Question: What is 2 + 2?\nA. 4\nB. 3']
y: ['A']
explanation: ['2 + 2 equals 4']
```

After routing both formats through `JSONDataParse` as this PR does, I got:

```text
sample.jsonl (detected as jsonl):
NameError: name 'COCO' is not defined

metadata.json (detected as jsonforllm):
NameError: name 'COCO' is not defined
```

Looking at the bundled implementation, `JSONDataParse` is meant for COCO-style image data. It does not provide the same behavior as the question-and-answer JSONL and metadata parsers.

This looks like a Sedna version compatibility problem. The change may match a newer Sedna API, but it does not match the wheel currently bundled with Ianvs.

I think the supported Sedna version should be made explicit and tested. If Ianvs needs to support both versions, a small compatibility adapter would be safer than sending all three formats to `JSONDataParse`. Tests for both `jsonl` and `jsonforllm` would catch this difference. The `if`/`elif` cleanup itself can remain separate from the parser change.

For this test, I extracted the bundled wheel into a temporary directory and ran its actual datasource implementations with this PR's `Dataset.load_data()`. I did not run a complete benchmark or test the latest Sedna source tree.

I am connecting this finding with the other restoration cases in [Discussion #965](https://github.com/kubeedge/ianvs/discussions/965).

Issue #131, opened 2024-08-11
Updating documentation of "examples/robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/README.md"


AryanNanda17:
I am an open-source contributor who is interested in contributing to the project [Multimodal Large Model Joint Learning Algorithm: Reproduction Based on KubeEdge-Ianvs](https://github.com/kubeedge/ianvs/issues/123). Issue #123 
**What would you like to be documented**:
In this [Readme](https://github.com/kubeedge/ianvs/blob/main/examples/robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/README.md), there are a few issues. 
- instead of using `python` to run different commands, it should be `python3`. For example instead of `python -m pip install ./examples/resources/third_party/*`, it should be `python3 -m pip install ./examples/resources/third_party/*`
- Also, before running this command, `python -m pip install --upgrade pip`, pip version lower than 24.* should be installed. (with python3.8.*)
- 
<img width="326" alt="Screenshot 2024-08-12 at 1 52 45 AM" src="https://github.com/user-attachments/assets/f8ce12f2-f743-471a-8a00-a52d023adb58">
Here no dataset download command is mentioned and a direct unzip command is run
- Also, there are few pre-requisites to download before running these steps like installing git, wget, unzip commands using `apt-get`
**Why is this needed**:
I volunteer to make these changes. This will improve the documentation. I can do a PR after the maintainers approve. 
Also, please give the right command to download the `mdil-ss` dataset.

AryanNanda17:
Also, @CreativityH @hsj576 
Please suggest if there is any more improvements in documentation with which I could help with. I can try to run different files that you mention, it will be a good start and I can understand better along the way. 
Thanks

CreativityH:
@hsj576 @MooreZheng These issue and enhanced PR make sense. Please consider the improved Readme. Also, please provide the right link or command to download the `mdil-ss` dataset.

MooreZheng:
Message received. Newcomers are very welcome.

Agree that a website link could be added in the readme.
1. @AryanNanda17  might want to help solve the issue by contacting @qxygxt @Frank-lilinjie @hsj576 .  This is an advanced research item from past events. It might not be so user-friendly to all newcomers and the authors might not respond very soon. 
2. Besides, it seems that this dataset is a collection of public datasets of cityscape and synthia. So one can also fix it by exploring the two datasets.

[...]

PR #567, opened 2026-06-27
fix(robot-cityscapes-synthia): fix three Sedna interface crashes blocking all pipeline runs

### LINKED ISSUE #473: [Bug] task_allocation_by_domain.py and cityscapes.py crash at runtime due to Sedna interface and type mismatches in robot-cityscapes-synthia example
## Environment

| Item | Version |
|---|---|
| OS | macOS 14 (Darwin 23.6.0) |
| Python | 3.12.7 |
| Sedna | 0.6.0.1 |
| PyTorch | 2.12.0 |
| ianvs branch | main |

## Example

```
```

## Run Command

```bash
```

## Description

Three separate runtime crashes occur in the example's algorithm-facing modules during the task allocation and evaluation phases. Each crash is caused by a mismatch between the assumptions the example code makes about data types and calling conventions, and what Sedna 0.6.0.1 actually provides.

---

## Error 1 — `TaskAllocationByOrigin.__call__()` signature does not match Sedna's calling convention

**Triggered at:** During task allocation, after task definition completes and before training begins.

```
  File ".../sedna/algorithms/seen_task_learning/seen_task_learning.py", line 220
    mappings = self.task_allocation(samples=samples)
TypeError: TaskAllocationByOrigin.__call__() missing 1 required positional argument: 'task_extractor'
  error: TaskAllocationByOrigin.__call__() missing 1 required positional argument: 'task_extractor'
```

**Location:** `testalgorithms/erfnet/task_allocation_by_domain.py`

The class defines its `__call__` with the signature `__call__(self, task_extractor, samples)`. Sedna's `SeenTaskLearning._task_allocation()` instantiates the class as `TaskAllocationByOrigin(task_extractor=self.seen_extractor, **extend_param)` — passing `task_extractor` to `__init__` — and then calls `method_cls(samples=samples)`. The `__call__` receives only `samples` as a keyword argument and crashes because the positional `task_extractor` argument is never provided during the call.

```python
# Current state in task_allocation_by_domain.py:
def __call__(self, task_extractor, samples: BaseDataSource):
#             ^^^^^^^^^^^^^^ Sedna passes this to __init__, not __call__
```

---

## Error 2 — `cityscapes.py` crashes with `TypeError: object of type 'NoneType' has no len()`

**Triggered at:** During the final evaluation step on the test set, inside `CityscapesSegmentation.__init__()`.

Full traceback from the pipeline run:

```
  File ".../lifelong_learning.py", line 149, in run
    self.edge_task_index, tasks_detail, res = self.my_eval(...)
  File ".../lifelong_learning.py", line 426, in my_eval
    res, tasks_detail = job.cloud_knowledge_management.seen_estimator.evaluate(...)
  File ".../sedna/backend/base.py", line 86, in evaluate
    return self.estimator.evaluate(*args, **varkw)
  File ".../sedna/algorithms/seen_task_learning/seen_task_learning.py", line 557, in evaluate
    name: metric(sample.y, pred, **metrics_param)
  File ".../testenv/accuracy.py", line 29, in accuracy
    _, _, test_loader, num_class = make_data_loader(args, test_data=y_true)
  File ".../ERFNet/dataloaders/__init__.py", line 22, in make_data_loader
    test_set = cityscapes.CityscapesSegmentation(args, data=test_data, split='test')
  File ".../ERFNet/dataloaders/datasets/cityscapes.py", line 33, in __init__
    if len(data[0]) == 2:
TypeError: object of type 'NoneType' has no len()
  error: object of type 'NoneType' has no len()
```

**Location:** `testalgorithms/erfnet/ERFNet/dataloaders/datasets/cityscapes.py`, line 33.

`CityscapesSegmentation.__init__()` calls `len(data[0])` in the `else` branch without checking whether `data[0]` is `None`. When Sedna calls `metric(sample.y, pred)`, `sample.y` contains the label values for that task's samples. In cases where labels are unavailable or `None` for a sample, `data[0]` is `None` and `len(None)` raises `TypeError`.

```python
# Current state in cityscapes.py (else branch, line 33):
else:
    if len(data[0]) == 2:      # crashes if data[0] is None
```

---

## Error 3 — `cityscapes.py` crashes with `ValueError: The truth value of an array is ambiguous`

**Triggered at:** Same evaluation path as Error 2. Occurs on the same line after attempting a different data input type.

```
  File ".../ERFNet/dataloaders/datasets/cityscapes.py", line 33, in __init__
    if len(data[0]) == 2:
  [Earlier guard attempt triggers:]
    first = data[0] if data else None
ValueError: The truth value of an array with more than one element is ambiguous.
  Use a.any() or a.all()
  error: The truth value of an array with more than one element is ambiguous.
  Use a.any() or a.all()
```

**Location:** `testalgorithms/erfnet/ERFNet/dataloaders/datasets/cityscapes.py`, line 33.

Sedna's `BaseDataSource` stores `.y` (labels) as a numpy array. When this numpy array is passed as `data` to `CityscapesSegmentation.__init__()`, any Python truthiness check — such as `if data:` or `if data else None` — calls `bool()` on a multi-element numpy array, which numpy explicitly disallows. The `else` branch in `__init__` has no type-safe handling for numpy array inputs.

---

## Impact

- Error 1 blocks all task allocation, meaning no training round can complete.
- Errors 2 and 3 block the final evaluation step, meaning even if training completes the benchmark score cannot be computed.
- All three errors are in the example's eval/inference hot path and affect every run.

## Files Involved

- `testalgorithms/erfnet/task_allocation_by_domain.py`
- `testalgorithms/erfnet/ERFNet/dataloaders/datasets/cityscapes.py`

rajdhruvsingh:
What type of PR is this?
/kind bug
/kind failing-test

Fixes three runtime crashes that blocked every training and evaluation run of
the robot-cityscapes-synthia lifelong learning example when using Sedna 0.6.0.1.

Bug 1 — task_allocation_by_domain.py: wrong __call__ signature
Sedna instantiates TaskAllocationByOrigin(task_extractor=..., **kwargs) then
calls method(samples=samples). The old __call__ expected task_extractor as a
positional argument and crashed immediately. Moved task_extractor to __init__
and stored as self.task_extractor.

Bug 2 — cityscapes.py: NoneType has no len() during inference
data.y is None when Sedna runs inference (no ground-truth labels available).
The old code assigned self.labels[split] = data.y unconditionally, then later
called self.labels[split][index].rstrip() which crashed. Added an explicit
data is None guard and a safe data.y is not None check before accessing labels.

Bug 3 — cityscapes.py: ValueError: truth value of array is ambiguous
data.y from Sedna is a numpy array. Bare truthiness checks like if data: or
data else None on a multi-element numpy array raise this error. Replaced all
such checks with isinstance() and hasattr() guards throughout __init__.

All three fixes verified with an isolated mock test against Sedna 0.6.0.1
BaseDataSource, covering train (with labels), inference (labels=None), and
numpy array label inputs. Happy to share the test script or add it to the PR if useful for review.

Which issue(s) this PR fixes:
Fixes #473
Related to #230


rajdhruvsingh:
Bug 1 (task_allocation_by_domain.py):

TaskAllocationByOrigin.__call__() expected task_extractor as an argument every time it was called. But Sedna's _task_allocation() only passes task_extractor once, at construction (TaskAllocationByOrigin(task_extractor=..., **extend_param)), and later calls it with just method_cls(samples=samples). Since the class never stored task_extractor at construction, it had nothing to fall back on when called with just samples — hence missing 1 required positional argument.

Fix: moved task_extractor into __init__ and stored it as self.task_extractor, so __call__ only needs samples, matching Sedna's actual calling convention.

rajdhruvsingh:
Bug 2 (cityscapes.py, line 33):

CityscapesSegmentation.__init__() calls len(data[0]) assuming data[0] is always a real object. But during evaluation, Sedna can pass None for labels (sample.y) when ground truth isn't available for a sample, so data[0] is None and len(None) raises TypeError.

Fix: added a None check before calling len(), so the None-labels case is handled explicitly instead of crashing.

rajdhruvsingh:
Bug 3 (cityscapes.py, same line):

Same line as Bug 2, different failure mode. Sedna stores labels as a numpy array. The existing code used a plain truthiness check (if data:) to decide the code path, but numpy disallows bool() on arrays with more than one element (it can't collapse multiple values into a single true/false), raising ValueError: truth value of an array is ambiguous.

Fix: replaced the truthiness check with explicit isinstance()/type checks, so None, numpy arrays, and lists are each handled based on their actual type instead of a generic truthiness test.

MooreZheng (changes requested):
As dicussed on the meeting, 
1) before getting to detailed implementation, need to make sure whther the correct Sedna wheel is used. The synthia example is quite an old one, which could need an old verion of Sedna wheel. See https://github.com/kubeedge/ianvs/tree/main/resources 
2) Besides, would the example have wrong configuration readme or requirement files?
3) An end-to-end evaluation for the synthia example is needed before merging.

KiranUdayakumar:
Reviewed as part of a submitted Example Restoration pre-test analysis (full write-up: https://github.com/kubeedge/ianvs/discussions/963), as one of two supplementary reviews extending Task 2's coverage outside the government_rag root-problem scope.

**What this PR does well:** three genuinely distinct Sedna interface bugs correctly diagnosed, the `task_extractor` constructor-argument fix, the `None`-labels guard, and the numpy-truthiness fix are each independently verified against a mock `BaseDataSource`.

**Verdict:** cannot merge as-is, per MooreZheng's own review in this thread, which raises a process-level concern that supersedes the individual fixes: confirming which Sedna wheel version this specific, older example actually needs, since older examples may depend on a Sedna release earlier than the one this PR assumes, plus an end-to-end evaluation of the full example rather than only the three isolated crash fixes.

Relationship to this submission's Task 2 findings: this PR is a second, interface-layer instance of the same Sedna-dependency-ambiguity root issue this submission separately identifies at the packaging layer on PR #687, a version mismatch here versus a wrong-package resolution there.

swikarthanekar (commented):
Reviewed #567 as the Example-local fix for #473.

I ran the published `robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation` benchmark on the #567 tree. Same stale testenv exit as main:

`./examples/class_increment_semantic_segmentation/lifelong_learning_bench/testenv/testenv.yaml`

The modified STP code is not reached in that full launch. I also ran a focused STP calling-contract reproduction against the interface change directly.

On main, `task_allocation_by_domain.py` has:

```python
def __call__(self, task_extractor, samples: BaseDataSource):
```

Sedna puts `task_extractor` in the constructor and later calls with `samples`. That pattern against main reproduces the missing-`task_extractor` `TypeError`.

#567 moves `task_extractor` into `__init__` and leaves `__call__` taking `samples`. The focused invocation works with the PR-style interface. That supports the STP fix direction; it is not a full benchmark pass.

The `cityscapes.py` guards for `None` labels and ndarray `sample.y` look like local evaluate-path semantics, not a Core lifecycle rename. Keeping STP and dataloader in the Example tree is the right call. This PR does not fix the stale YAML prefixes that currently block the full run before STP executes.

Executed: benchmark launch on #567 tree and focused STP reproduction. The full benchmark stops at config resolution; static comparison of STP/dataloader source vs PR. Not claiming complete Example success end to end.

It would help to note which repaired algorithm/config paths were used for any broader PR verification, so reviewers know these files load once config paths are fixed.

A small test that constructs the STP class like Sedna and calls `__call__(samples=...)` would protect the contract.

**Minor revision.** Implementation direction looks right for Example-owned STP work. Minor revision to clarify verification against the intended repaired config path, plus a focused invocation test if feasible.

rajdhruvsingh:
As requested, I verified the Sedna wheel version, checked config/README correctness, and ran a full end-to-end evaluation of the `robot-cityscapes-synthia` example.

**1) Sedna wheel version**

Confirmed `sedna-0.6.0.1` (`resources/third_party/sedna-0.6.0.1-py3-none-any.whl`) is the correct and only current wheel — installs cleanly and matches every `sedna.*` import used across this example's codebase. No version change needed.

**2) Config/README correctness — found and fixed real bugs beyond the original crash fixes:**

- fix(benchmarkingjob.yaml, test_algorithm.yaml, testenv.yaml): restore missing `semantic-segmentation/` path segment, hardcoded `/home/QXY/` absolute paths. Fixes #514
- fix(testenv.yaml): migrate `train_url`/`test_url` → `train_index`/`test_index` — current `core.testenvmanager.dataset.Dataset.process_dataset()` only branches on `train_index`/`train_data`/`train_data_info`. Sister issue to #604 (same pattern, different example). Fixes #1002
- fix(basemodel.py): make ERFNet's internal flat imports (`mypath`, `dataloaders`, `utils`, `models`) self-contained via `sys.path` insertion at import time, instead of relying on an undocumented manual `export PYTHONPATH=...` step. Verified with `env -u PYTHONPATH ianvs -f ...`
- Related: #131 already tracks the missing `mdil-ss` dataset download command in this README — confirmed the gap still exists (no commit since 2023 has ever included a working link), leaving that documentation fix to #131 rather than duplicating scope here

**3) End-to-end evaluation**

With all fixes above applied, the pipeline runs cleanly and standalone through config loading, dataset preparation, and every ERFNet-internal import — stopping at exactly one remaining point:

```
File ".../ERFNet/train.py", line 11, in <module>
    from models.erfnet_RA_parallel import Net as Net_RAP
ModuleNotFoundError: No module named 'models.erfnet_RA_parallel'
```

Filed as #1001 — `Net_RAP` is the actual model class used for both training and eval, confirmed missing from every branch/commit in this repo and upstream/main. Likely sourced from `prachigarg23/MDIL-SS` (WACV 2022) originally, but that repo has no LICENSE, so it can't be copied in without the author's permission.

I validated everything up to that point using a local synthetic-data harness (kept out of this PR to stay scoped to real fixes) that exercises the real `ianvs -f ...` path without needing the real dataset.

**Where this leaves us:** this PR's scope is complete and verified. A real end-to-end run with actual metrics is blocked on #1001 — would appreciate guidance on how you'd like to handle sourcing/licensing that file.

MooreZheng (changes requested):
Thank you for the cool update
1. The robot-cityscapes-synthia dataset is bascially a combination of three datasets. That is https://www.kaggle.com/datasets/kubeedgeianvs/cloud-roboticshttps://www.cityscapes-dataset.com/https://synthia-dataset.net/
Download the datasets from three sources and combine them as the final dataset for testing. There could be compatible issues for combining different datasets. But we can use only part of them for evaluation. Might upload the combined dataset to Kaggle in the end. 
2. Download the WACV 2022 model to evaluated the example; but no need to upload any code to Ianvs since the WACV 2022 has no license. After the evaluation, show the video recording and update the readme file to guide the model using for developers

PR #569, opened 2026-06-28
fix: remove undefined LadeSpecDecLLM import from edge_model --  issue#568

### LINKED ISSUE #568: EdgeModel fails to import — `LadeSpecDecLLM` is undefined in cloud-edge-collaborative-inference-for-llm
**What happened**:

The `EdgeModel` module in the `cloud-edge-collaborative-inference-for-llm` example fails to import. `edge_model.py` imports `LadeSpecDecLLM` from the `models` package at the top of the file:

```python
# testalgorithms/query-routing/edge_model.py:21
from models import HuggingfaceLLM, APIBasedLLM, VllmLLM, EagleSpecDecModel, LadeSpecDecLLM
But LadeSpecDecLLM is not defined or exported anywhere in the repo.
```

edge_model.py should import cleanly and EdgeModel should load for the supported backends. A missing/optional speculative-decoding backend should not break the whole example at import time.

cd examples/cloud-edge-collaborative-inference-for-llm/testalgorithms/query-routing
python -c "import edge_model"
# -> ImportError: cannot import name 'LadeSpecDecLLM' from 'models'

Priyanshu6968:
**What type of PR is this?**

/kind bug

`edge_model.py` in the `cloud-edge-collaborative-inference-for-llm` example imported `LadeSpecDecLLM` from the `models` package, but that class is not defined or exported anywhere.

This was a half merged "lookahead decoding" backend: the import and the `load()` call site were added, but the implementation module and its export were never committed. This PR removes the dead wiring.

Drop `LadeSpecDecLLM` from the `from models import ...` statement.
Remove the unreachable `elif self.backend == "LadeSpecDec"` branch in `load()`.
Remove the stale `"LadeSepcDec"` backend comment in `test_queryrouting.yaml`.

Fixes #568


henryng15 (commented):
Reviewed at head `84c441f` against `37a9c60`. The fix is correct.

**Verified.** `LadeSpecDecLLM` appears exactly twice in the repository — `edge_model.py:21`
and `:77` — both uses, zero definitions, and
`.../query-routing/models/__init__.py` exports only `APIBasedLLM`, `HuggingfaceLLM`,
`VllmLLM`, `BaseLLM` and `EagleSpecDecModel`. Removing both the import and the
`elif self.backend == "LadeSpecDec"` branch resolves #568.

**No reachable configuration is lost.** `test_queryrouting.yaml` never listed
`LadeSpecDec` under `backend: values:` — the list is `"huggingface"`, `"vllm"`, with
`# - "EagleSpecDec"` commented out — so the deleted branch was unreachable from any
shipped config.

Two small things.

**1. There is a second non-existent name in the same file, four lines below the one you
fixed.**

```python
# line 21  (this PR)
from models import HuggingfaceLLM, APIBasedLLM, VllmLLM, EagleSpecDecModel

# line 25  (untouched)
__all__ = ["BaseModel"]        # this module defines EdgeModel, not BaseModel
```

`__all__` has no runtime effect here — nothing star-imports this module and `core/`
contains zero references to `__all__`, so this is cosmetic rather than a second bug. But
since the PR is already in this file for the same class of mistake, correcting it costs
one line and removes the last wrong name from the module.

**2. The issue asks for optional, the PR delivers removal.** #568 describes
`LadeSpecDecLLM` as "a missing/optional speculative-decoding backend". This PR deletes
the backend rather than guarding it. Given the class was never implemented anywhere,
deletion looks like the right call — but the divergence should be noted in the PR
description so #568 can be closed accurately rather than looking partly addressed.

**Recommendation: minor revision** — description note, and optionally the one-line
`__all__` correction. No significant coding effort either way.

PR #579, opened 2026-07-05
fix: replace 'ianvs -f' with 'python benchmarking.py -f' in all example READMEs

### LINKED ISSUE #563: [Bug] Core setup failures on fresh install: API rename, missing class, wrong docs path, undocumented deps

31groot:
**What type of PR is this?**
/kind documentation

24 example READMEs instructed users to run `ianvs -f` which fails on 
a fresh install with `No module named ianvs`. The correct command is 
`python benchmarking.py -f`.

This is a systematic fix across all affected READMEs found via:
`find examples -name "README.md" | xargs grep -l "ianvs -f"`

Files changed: 24 READMEs, 34 occurrences replaced.

Fixes #563

### DIFF
--- a/examples/Cloud_Robotics/cloud-edge-collaborative-inference_bench/perception-reasoning/README.md
+++ b/examples/Cloud_Robotics/cloud-edge-collaborative-inference_bench/perception-reasoning/README.md
@@ -65,7 +65,7 @@ pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl
 pip install -r requirements.txt
 
 # Install dependencies for this example.
-pip install -r examples/Cloud_Robotics/cloud-edge-collaborative-inference-bench/requirements.txt
+pip install -r examples/Cloud_Robotics/cloud-edge-collaborative-inference_bench/requirements.txt
 
 # Install ianvs
 python setup.py install  
@@ -161,7 +161,7 @@ Update the algorithm's **URL address** and **hyperparemeter** in the `perception
       # name of edge model module; string type;
       name: "EdgeModel"
       # the url address of edge model module; string type;
-      url: "./examples/Cloud_Robotics/cloud-edge-collaborative-inference-bench/perception-reasoning/testalgorithms/edge_model.py"
+      url: "./examples/Cloud_Robotics/cloud-edge-collaborative-inference_bench/perception-reasoning/testalgorithms/edge_model.py"
 
       hyperparameters:
       # name of the hyperparameter; string type;
@@ -173,7 +173,7 @@ Update the algorithm's **URL address** and **hyperparemeter** in the `perception
       # name of python module; string type;
       name: "CloudModel"
       # the url address of python module; string type;
-      url: "./examples/Cloud_Robotics/cloud-edge-collaborative-inference-bench/perception-reasoning/testalgorithms/cloud_model.py"
+      url: "./examples/Cloud_Robotics/cloud-edge-collaborative-inference_bench/perception-reasoning/testalgorithms/cloud_model.py"
 
       hyperparameters:
         # name of the hyperparameter; string type;
@@ -197,7 +197,7 @@ Run Ianvs for benchmarking:
 ```bash
 cd /ianvs
 export DASHSCOPE_API_KEY="your_api_key_here"
-python benchmarking.py -f examples/Cloud_Robotics/cloud-edge-collaborative-inference-bench/perception-reasoning/benchmarkingjob.yaml
+python benchmarking.py -f examples/Cloud_Robotics/cloud-edge-collaborative-inference_bench/perception-reasoning/benchmarkingjob.yaml
 ```
 **Note**: Replace `"your_api_key_here"` with your actual Dashscope API key.
 
--- a/examples/Cloud_Robotics/cloud-edge-collaborative-inference_bench/perception-reasoning/README.md
+++ b/examples/Cloud_Robotics/cloud-edge-collaborative-inference_bench/perception-reasoning/README.md
@@ -197,7 +197,7 @@ Run Ianvs for benchmarking:
 ```bash
 cd /ianvs
 export DASHSCOPE_API_KEY="your_api_key_here"
-ianvs -f examples/Cloud_Robotics/cloud-edge-collaborative-inference-bench/perception-reasoning/benchmarkingjob.yaml
+python benchmarking.py -f examples/Cloud_Robotics/cloud-edge-collaborative-inference-bench/perception-reasoning/benchmarkingjob.yaml
 ```
 **Note**: Replace `"your_api_key_here"` with your actual Dashscope API key.
 
--- a/examples/GovDoc2Poster/singletask_learning_bench/README.md
+++ b/examples/GovDoc2Poster/singletask_learning_bench/README.md
@@ -138,7 +138,7 @@ cd /home/linux/Desktop/ianvs
 ### Step 3: Run the System
 
 ```bash
-ianvs -f examples/GovDoc2Poster/singletask_learning_bench/testalgorithms/gen/government_data_source.yaml
+python benchmarking.py -f examples/GovDoc2Poster/singletask_learning_bench/testalgorithms/gen/government_data_source.yaml
 ```
 
 ✅ Done! The system will automatically parse the government documents from the dataset and generate visual posters.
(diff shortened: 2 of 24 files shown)


sheikhayaan:
Twenty-four READMEs is a wide change, so I worked through what it does to each example it
touches instead of only reading the diff. The direction is defensible, but two examples come
out worse and one of them measurably so.

Neither is one that @gemini-code-assist already flagged. Its two path findings are the leading
slash in `cloud_VLA_finetune` at line 313 and the stray `ianvs/` prefix in
`llm-edge-benchmark-suite` at line 53, both of which I confirmed and neither of which I repeat
below. The two here are different files, and the first is a different kind of problem: not a
path that was typed wrong, but a path that was correct until this PR changed the command around
it.

**This makes PIPL strictly worse.** PIPL's README line 118 currently reads `ianvs -f
benchmarkingjob.yaml`, a bare filename. Its config uses `./testenv/testenv.yaml` and
`./test_algorithms/algorithm.yaml`, and I verified both targets exist. So PIPL resolves
**completely and correctly** when the working directory is the example's own directory. It is
internally consistent. This PR rewrites the command to `python benchmarking.py -f
benchmarkingjob.yaml`, but `benchmarking.py` lives at the **repository root**. After the change
there's no single directory from which the instruction works: run it from the example directory
and `benchmarking.py` isn't there; run it from the root and neither `benchmarkingjob.yaml` nor
`./testenv/testenv.yaml` resolves. One working configuration becomes zero.

**And a broken path survives the fix.** The Cloud_Robotics line is rewritten to `python
benchmarking.py -f
examples/Cloud_Robotics/cloud-edge-collaborative-inference-bench/perception-reasoning/benchmarkingjob.yaml`,
but the actual directory is `cloud-edge-collaborative-inference_bench` with an **underscore**.
The command changed and the broken path was carried through unmodified, so the defect tracked
in #716 and #717 survives a PR that touches the exact line it lives on.

One objection I considered: you could argue `ianvs -f` was always wrong because the CLI isn't
installed. It isn't wrong in principle, though. `setup.py` declares `console_scripts: ianvs =
core.cmd.benchmarking:main`, so `ianvs -f` is the intended interface after `pip install .`. The
reason it fails in practice is the packaging defects in #793/#674 and #683/#687: no
`install_requires`, a missing `colorlog`, a self-referential dependency. Which means this PR
treats a symptom of the packaging layer by editing documentation, and in doing so quietly
abandons the packaged interface without saying so. If that abandonment is intended, it deserves
a decision, not 24 silent README edits.

**Recommendation: major revision.** Fix the Cloud_Robotics underscore while you're in the file.
Handle PIPL explicitly, either by keeping repo-root-relative paths in its config or by
documenting a `cd`. And state whether the `ianvs` console script is being deprecated. Two
practical notes: the branch is now in conflict with `main` and needs a rebase before any of this
can be assessed, and #672 reduces the root `benchmarking.py` to a one-line wrapper, so this PR
makes 24 documents depend on a file another PR is rewriting. Those should be sequenced knowingly
rather than by accident.

31groot:
> Twenty-four READMEs is a wide change, so I worked through what it does to each example it touches instead of only reading the diff. The direction is defensible, but two examples come out worse and one of them measurably so.
> 
> Neither is one that @gemini-code-assist already flagged. Its two path findings are the leading slash in `cloud_VLA_finetune` at line 313 and the stray `ianvs/` prefix in `llm-edge-benchmark-suite` at line 53, both of which I confirmed and neither of which I repeat below. The two here are different files, and the first is a different kind of problem: not a path that was typed wrong, but a path that was correct until this PR changed the command around it.
> 
> **This makes PIPL strictly worse.** PIPL's README line 118 currently reads `ianvs -f benchmarkingjob.yaml`, a bare filename. Its config uses `./testenv/testenv.yaml` and `./test_algorithms/algorithm.yaml`, and I verified both targets exist. So PIPL resolves **completely and correctly** when the working directory is the example's own directory. It is internally consistent. This PR rewrites the command to `python benchmarking.py -f benchmarkingjob.yaml`, but `benchmarking.py` lives at the **repository root**. After the change there's no single directory from which the instruction works: run it from the example directory and `benchmarking.py` isn't there; run it from the root and neither `benchmarkingjob.yaml` nor `./testenv/testenv.yaml` resolves. One working configuration becomes zero.
> 
> **And a broken path survives the fix.** The Cloud_Robotics line is rewritten to `python benchmarking.py -f examples/Cloud_Robotics/cloud-edge-collaborative-inference-bench/perception-reasoning/benchmarkingjob.yaml`, but the actual directory is `cloud-edge-collaborative-inference_bench` with an **underscore**. The command changed and the broken path was carried through unmodified, so the defect tracked in #716 and #717 survives a PR that touches the exact line it lives on.
> 
> One objection I considered: you could argue `ianvs -f` was always wrong because the CLI isn't installed. It isn't wrong in principle, though. `setup.py` declares `console_scripts: ianvs = core.cmd.benchmarking:main`, so `ianvs -f` is the intended interface after `pip install .`. The reason it fails in practice is the packaging defects in #793/#674 and #683/#687: no `install_requires`, a missing `colorlog`, a self-referential dependency. Which means this PR treats a symptom of the packaging layer by editing documentation, and in doing so quietly abandons the packaged interface without saying so. If that abandonment is intended, it deserves a decision, not 24 silent README edits.
> 
> **Recommendation: major revision.** Fix the Cloud_Robotics underscore while you're in the file. Handle PIPL explicitly, either by keeping repo-root-relative paths in its config or by documenting a `cd`. And state whether the `ianvs` console script is being deprecated. Two practical notes: the branch is now in conflict with `main` and needs a rebase before any of this can be assessed, and #672 reduces the root `benchmarking.py` to a one-line wrapper, so this PR makes 24 documents depend on a file another PR is rewriting. Those should be sequenced knowingly rather than by accident.

Thanks for the detailed review. I addressed the two concrete correctness issues you identified.

- **PIPL:** the README now explicitly changes into the PIPL example directory before invoking the repository-root `benchmarking.py`, so `benchmarkingjob.yaml` and its `./testenv/...` / `./data/...` relative paths share the same working-directory contract.
- **Cloud_Robotics:** corrected the example directory name to `cloud-edge-collaborative-inference_bench`.

I also rebased the branch onto the current `main`.

The PIPL benchmark itself cannot be completed from the current checkout because the referenced `data/chnsenticorp_lite/*.jsonl` dataset files are not present, so I’m not claiming full end-to-end execution.

The fixes are pushed in commit `9552183`.

Issue #440, opened 2026-05-21
Issue summarization for example of `examples/smart_coding/smart_coding_learning_bench`


Chidroopakanaparthy:
# Issue summarization for example of `smart_coding/smart_coding_learning_bench`

---

### 1.1 Problem Description

The `examples/smart_coding/smart_coding_learning_bench` is **completely non-executable** on any fresh environment. Through systematic execution on macOS ARM64 (Python 3.13.5, Apple Silicon, no GPU), I confirmed **19 distinct bugs** spanning four layers: ecosystem-level Sedna installation, core dependency resolution, Sedna datasource API correctness, and example-level configuration. Every bug was confirmed in strict execution order - each one hiding the next.

The most critical finding: `pip install sedna` silently installs a completely **wrong PyPI package** (a unit conversion library unrelated to KubeEdge),while the actual KubeEdge Sedna has no pip-installable package structure at all. This single ecosystem failure blocks **every contributor on every Ianvs example**, not just smart-coding. The broken CI path discovered in PR #417 (the Sedna wheel was never being installed during CI runs) means this silent breakage has gone undetected across all open PRs.

Despite all failures, **inference successfully ran** after cascade resolution:
``` bash
Loading weights: 100%|██████████| 338/338 [00:02<00:00]
Processing: 100%|██████████| 1/1 [00:25<00:00, 25.93s/question]
```
 confirming the core design is sound and all bugs are fixable.

---

## Originality and Scope Declaration

This analysis was conducted through direct cascade execution on a clean macOS ARM64environment (Apple Silicon, Python 3.13.5, no GPU), and not static analysis. Each of the19 bugs was confirmed in strict execution order with the prior one resolved first, with inference run to completion as proof (`338 weights loaded, 25.93s/question`).

Gemini Code Assist has not reviewed any issue or PR that touches`examples/smart_coding/smart_coding_learning_bench` directly. The automated reviews on PRs #315, #394, #414, #416, #417, #420, #421, #424, #426, and #429 are confined to
those PRs' own changed files and do not intersect with the smart_coding example. The following findings are original to this submission and are not present in any prior GitHub issue, pull request review, or automated review on kubeedge/ianvs:

**Originally identified in this submission:**

[...]

PR #420, opened 2026-05-08
fix(examples): replace hardcoded developer paths across all cifar100 configs

AswaniSahoo:
## What this PR does

Replaces all hardcoded absolute paths (`/home/wyd/...`) across **18 YAML configuration files** in the `examples/cifar100/` directory with portable alternatives.

These developer-specific paths make every cifar100 example non-portable, causing `FileNotFoundError` on any machine other than the original developer's workstation.

### Scope

This PR covers **all 6 cifar100 example variants** — not just `federated_learning/fedavg` (which is addressed separately in #251):

| Example variant | Files fixed |
|---|---|
| `federated_learning/fedavg` | `benchmarkingjob.yaml`, `testenv.yaml`, `algorithm.yaml` |
| `federated_class_incremental_learning/fedavg` | `benchmarkingjob.yaml`, `testenv.yaml`, `algorithm.yaml` |
| `fci_ssl/fedavg` | `benchmarkingjob.yaml`, `testenv.yaml`, `algorithm.yaml` |
| `fci_ssl/fed_ci_match` | `benchmarkingjob.yaml`, `testenv.yaml`, `algorithm.yaml` |
| `fci_ssl/fed_ci_match_v2` | `benchmarkingjob.yaml`, `testenv.yaml`, `algorithm.yaml` |
| `fci_ssl/glfc` | `benchmarkingjob.yaml`, `testenv.yaml`, `algorithm.yaml` |

### Path replacements

| Path type | Before | After |
|---|---|---|
| `workspace` | `/home/wyd/ianvs/.../workspace` | `./workspace` |
| `train_url` / `test_url` | `/home/wyd/ianvs/project/data/cifar100/...` | `/path/to/cifar100/...` (placeholder) |
| metric `url` | `/home/wyd/ianvs/project/ianvs/examples/...` | `./examples/cifar100/...` (relative to repo root) |
| `initial_model_url` | `/home/wyd/ianvs/project/init_model/...` | `/path/to/init_model/...` (placeholder) |

The relative path conventions (`./workspace`, `./examples/...`) follow the pattern already established by working examples like `pcb-aoi`.

### Relationship to #230

This addresses the runtime configuration component of the Comprehensive Example Restoration initiative: diagnosing and fixing path-related bugs that prevent examples from running out of the box.

/kind cleanup

PR #580, opened 2026-07-07
fix(examples): replace hardcoded developer paths in smart_coding

karanthengane23:
/kind cleanup

The `smart_coding_learning_bench` example (both `comment/` and `issue/` sub-benchmarks) contained hardcoded absolute paths pointing to the original contributor's local machine (`/home/xiebo/...` and `/root/xieb/...`). This makes the example non-portable — it fails with `FileNotFoundError`/`HFValidationError` for any contributor other than the original author.

Changes:
- `benchmarkingjob.yaml` (both variants): `workspace`, `testenv`, and algorithm config `url` fields changed to relative paths (`./...`), matching the convention already used elsewhere in the repo.
- `testenv.yaml` (both variants): `train_data` and `test_data_info` changed to relative dataset paths (`./dataset/smart_coding/...`).
- `basemodel.py` (both variants): replaced the hardcoded local model checkpoint path with the public Hugging Face model id (`Qwen/Qwen2.5-Coder-1.5B-Instruct`), so the model downloads correctly for any user instead of assuming a pre-existing local copy at a specific path.

I verified these paths are still broken on current `main`, and confirmed no open PR currently fixes this specific example (PR #420 attempted a similar fix for a different example, `cifar100`, but was closed for using non-functional `/path/to/...` placeholders instead of real relative paths — this PR uses genuinely portable relative paths throughout, consistent with the reviewer feedback on that PR).

Addresses Bug 2 ("Hardcoded `/home/xiebo/` and `/root/xieb/` paths") documented in #440, and is part of the broader Comprehensive Example Restoration effort tracked in #230.

### DIFF
--- a/examples/smart_coding/smart_coding_learning_bench/comment/benchmarkingjob.yaml
+++ b/examples/smart_coding/smart_coding_learning_bench/comment/benchmarkingjob.yaml
@@ -2,11 +2,11 @@ benchmarkingjob:
   # job name of bechmarking; string type;
   name: "benchmarkingjob"
   # the url address of job workspace that will reserve the output of tests; string type;
-  workspace: "/home/xiebo/ianvs/workspace"
+  workspace: "./workspace"
 
   # the url address of test environment configuration file; string type;
   # the file format supports yaml/yml;
-  testenv: "/home/xiebo/ianvs/examples/smart_coding/smart_coding_learning_bench/comment/testenv/testenv.yaml"
+  testenv: "./examples/smart_coding/smart_coding_learning_bench/comment/testenv/testenv.yaml"
 
   # the configuration of test object
   test_object:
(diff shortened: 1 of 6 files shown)


karanthengane23:
Hi @MooreZheng — checking in on this PR. All checks are passing and it resolves the hardcoded developer paths for Issue #230. Would appreciate your feedback when you have a moment!

Talha8899:
## Review of PR #580

I reviewed the changes in PR #580 and reproduced the Smart Coding comment benchmark from the PR revision.

The PR correctly removes the original contributor-specific absolute paths such as `/home/xiebo/...` and `/root/xieb/...` and replaces them with repository-relative paths. It also changes the model loading from a contributor-local checkpoint to the public `Qwen/Qwen2.5-Coder-1.5B-Instruct` model ID, with the later commit allowing `model_name_or_path` to be supplied through module hyperparameters.

I verified the configuration path used by Ianvs and ran:

```text
ianvs -f examples/smart_coding/smart_coding_learning_bench/comment/benchmarkingjob.yaml
```

The run progressed through the updated configuration and stopped at dataset validation with:

```text
ValueError: dataset file(./dataset/smart_coding/comment/train_data/data.jsonl)
is not a local file and not a absolute path.
```

This failure does not show that the path-portability changes in this PR are incorrect. The Smart Coding dataset is an external/missing resource and is outside the specific hardcoded-developer-path fix addressed by this PR.

Therefore, I consider the claimed fix in #580 technically correct and recommend accepting it as-is. The example still requires a separately resolved dataset/setup issue before a complete end-to-end benchmark can be executed.

**Verification boundary:** the path fix was reproduced locally, but end-to-end benchmark success was not verified because the required Smart Coding dataset was not available in the checkout.

karanthengane23:
Thanks @Talha8899 for taking the time to test and review PR #580! Glad to hear the path portability and `model_name_or_path` hyperparameter changes verified cleanly.

Issue #515, opened 2026-06-08
bug: hardcoded absolute paths (/home/*) across multiple examples make benchmarks non-portable


rajdhruvsingh:
## Description
Multiple examples contain hardcoded absolute paths specific 
to individual developers' machines, making every affected 
benchmark completely non-runnable for any other user.

## Affected Examples (discovered via grep -rn "/home/" examples/)

| Example | Files Affected | Hardcoded User |
|---|---|---|
| cifar100/fci_ssl/fedavg | benchmarkingjob.yaml, algorithm.yaml, testenv.yaml | /home/wyd/ |
| cifar100/fci_ssl/glfc | benchmarkingjob.yaml, algorithm.yaml, testenv.yaml | /home/wyd/ |
| cifar100/fci_ssl/fed_ci_match_v2 | benchmarkingjob.yaml, algorithm.yaml, testenv.yaml | /home/wyd/ |
| cifar100/fci_ssl/fed_ci_match | benchmarkingjob.yaml, algorithm.yaml, testenv.yaml | /home/wyd/ |
| cifar100/federated_class_incremental_learning/fedavg | benchmarkingjob.yaml, algorithm.yaml, testenv.yaml | /home/wyd/ |
| cifar100/federated_learning/fedavg | benchmarkingjob.yaml, algorithm.yaml, testenv.yaml | /home/wyd/ |
| government/singletask_learning_bench/subjective | benchmarkingjob.yaml, testenv.yaml | /home/icyfeather/ |
| government/singletask_learning_bench/objective | benchmarkingjob.yaml, testenv.yaml | /home/icyfeather/ |
| government_rag/singletask_learning_bench | benchmarkingjob.yaml, testenv.yaml | /home/icyfeather/ |
| robot/lifelong_learning_bench | testenv.yaml (x3 variants) | /home/shijing.hu/ |
| cityscapes-synthia/lifelong_learning_bench | benchmarkingjob.yaml, testenv.yaml | /home/shijing.hu/ |
| smart_coding/comment | benchmarkingjob.yaml, testenv.yaml | /home/xiebo/ |
| smart_coding/issue | testenv.yaml | /home/xiebo/ |
| Cloud_Robotics | testenv.yaml | /home/aryan/ |
| llm_simple_qa | benchmarkingjob.yaml, testenv.yaml | /home/icyfeather/ |
| GovDoc2Poster | benchmarkingjob.yaml | /home/ianvs/ |
[...]

PR #582, opened 2026-07-08
fix(examples): correct broken config paths in cityscapes singletask_learning_bench

### LINKED ISSUE #581: cityscapes/singletask_learning_bench: broken config paths (missing semantic-segmentation/ segment) block execution, no README
## 1. Background

`examples/cityscapes/singletask_learning_bench` benchmarks RFNet semantic segmentation under the `singletasklearning` paradigm on the Cityscapes dataset. All example files actually live under a `semantic-segmentation/` subdirectory, but three config files reference each other **without** that path segment:

- `semantic-segmentation/benchmarkingjob.yaml` → `testenv: "./examples/cityscapes/singletask_learning_bench/testenv/testenv.yaml"` (missing `semantic-segmentation/`)
- same file → algorithm `url: "./examples/cityscapes/singletask_learning_bench/testalgorithms/rfnet/rfnet_algorithm.yaml"` (same problem)
- `semantic-segmentation/testenv/testenv.yaml` → metric `url: "./examples/cityscapes/singletask_learning_bench/testenv/map.py"` (same problem)

Running it reproduces the crash immediately, before any model/data code executes:

```
RuntimeError: not found testenv config file(./examples/cityscapes/singletask_learning_bench/testenv/testenv.yaml) in local.
```

The example also has **no README** at all — unlike working examples (e.g. `pcb-aoi`), which document dataset source and quick-start steps.

## 2. Goals

- Fix the three path references so `ianvs -f` gets past config loading.
- Add a README documenting the Cityscapes dataset source and how to run this example, matching the convention used by verified-runnable examples.

## 3. Scope

This is distinct from the hardcoded-developer-machine-path bugs already tracked in #440/#515/#436 — those are absolute paths baked in from someone's own machine (e.g. `/home/xiebo/...`). This is a structural bug in the repo's own relative paths (a missing directory segment), affecting every config file in this example. No open PR currently touches this example.

## 4. Detailed design

- **Architecture:** affects Ianvs's Test-Environment Manager (loads `testenv.yaml`) and Test-Case Controller (loads the algorithm YAML) — both fail at config-parsing time, before reaching `core`'s actual paradigm/model code.
- **Module details:** no core changes needed — purely fixing the `url`/`testenv` string values in the three YAML files to include the `semantic-segmentation/` segment, plus adding the missing README.

I'll follow this up with a PR shortly.

karanthengane23:
/kind bug

`examples/cityscapes/singletask_learning_bench/semantic-segmentation` had bugs that blocked execution before any model/data code ran:

1. `benchmarkingjob.yaml`, `testenv.yaml`, and `rfnet_algorithm.yaml` referenced each other **without** the `semantic-segmentation/` directory segment that the files actually live under. Running the example failed immediately with:
   ```
   RuntimeError: not found testenv config file(./examples/cityscapes/singletask_learning_bench/testenv/testenv.yaml) in local.
   ```
   *Historical Note*: The missing `semantic-segmentation/` segment stems from the example-tree restructuring commit `abe40ec` (Dec 2023) where files were moved under `semantic-segmentation/` but referring configs remained rename-orphaned. Citing commit `abe40ec` confirms this change mechanically restores the original intended path hierarchy.

2. `testenv.yaml`'s dataset section used `train_url`/`test_url` keys, which `core/testenvmanager/dataset/dataset.py` does not recognize (it only checks `train_index`/`train_data`/`train_data_info`). Even after fixing (1), this caused:
   ```
   RuntimeError: prepare dataset failed, error: not one of train_index/train_data/train_data_info.
   ```
   Renamed to `train_index`/`test_index`, matching the convention used in the working `pcb-aoi` example.

3. Removed leftover developer hardcoded path (`/home/wxc/...`) in `basemodel.py` line 149, making `initial_model_url` functional as documented.

Also added a `README.md` for this example (it previously had none), documenting the Cityscapes dataset source, expected index-file format, the model checkpoint requirement, and run instructions.

**Verification**: ran `ianvs -f examples/cityscapes/singletask_learning_bench/semantic-segmentation/benchmarkingjob.yaml` after each fix. It now progresses cleanly past all config-loading and dataset-key-parsing stages, failing only on the Cityscapes dataset files not being present locally — expected, since the dataset must be downloaded separately (same as any other example).

Fixes #581

### DIFF
--- a/examples/cityscapes/singletask_learning_bench/semantic-segmentation/README.md
+++ b/examples/cityscapes/singletask_learning_bench/semantic-segmentation/README.md
@@ -1,0 +1,31 @@
+# Single Task Learning: Semantic Segmentation on Cityscapes
+
+This example benchmarks RFNet-based semantic segmentation under Ianvs's `singletasklearning` paradigm, using the [Cityscapes](https://www.cityscapes-dataset.com/) urban street scene dataset.
+
+## Ianvs Preparation
+
+Follow the [Ianvs installation guide](../../../../docs/guides/how-to-install-ianvs.md) to install Ianvs before running this example.
+
+## Dataset Preparation
+
+Download the Cityscapes dataset from the [official downloads page](https://www.cityscapes-dataset.com/downloads/) (registration required). See [Semantic Segmentation: Cityscapes and SYNTHIA](../../../../docs/proposals/algorithms/lifelong-learning/Additional-documentation/curb_detetion_datasets.md) for a description of the dataset layout.
+
+Place the dataset under `./dataset/cityscapes/` at the root of the Ianvs repo, and prepare `train_data.txt`/`test_data.txt` index files where each line lists an image path and its corresponding label path, separated by a space:
+
+```
+./images/aachen_000000_000019_leftImg8bit.png ./images/aachen_000000_000019_gtFine_labelTrainIds.png
+```
+
+## Model Preparation
+
+`testalgorithms/rfnet/rfnet_algorithm.yaml` expects an initial RFNet model checkpoint at `./models/530_exp3_2.pth` (relative to the repo root). Place a pretrained checkpoint at that path, or update `initial_model_url` to point elsewhere.
+
+## Run Ianvs
+
+From the root directory of Ianvs:
+
+```shell
+ianvs -f examples/cityscapes/singletask_learning_bench/semantic-segmentation/benchmarkingjob.yaml
+```
+
+The leaderboard, ranked by the `map` metric, will be printed to the console and saved under `./workspace`.
--- a/examples/cityscapes/singletask_learning_bench/semantic-segmentation/benchmarkingjob.yaml
+++ b/examples/cityscapes/singletask_learning_bench/semantic-segmentation/benchmarkingjob.yaml
@@ -6,7 +6,7 @@ benchmarkingjob:
 
   # the url address of test environment configuration file; string type;
   # the file format supports yaml/yml;
-  testenv: "./examples/cityscapes/singletask_learning_bench/testenv/testenv.yaml"
+  testenv: "./examples/cityscapes/singletask_learning_bench/semantic-segmentation/testenv/testenv.yaml"
 
   # the configuration of test object
   test_object:
@@ -19,7 +19,7 @@ benchmarkingjob:
       - name: "rfnet_singletask_learning"
         # the url address of test algorithm configuration file; string type;
         # the file format supports yaml/yml;
-        url: "./examples/cityscapes/singletask_learning_bench/testalgorithms/rfnet/rfnet_algorithm.yaml"
+        url: "./examples/cityscapes/singletask_learning_bench/semantic-segmentation/testalgorithms/rfnet/rfnet_algorithm.yaml"
 
   # the configuration of ranking leaderboard
   rank:
--- a/examples/cityscapes/singletask_learning_bench/semantic-segmentation/test_config_closure.py
+++ b/examples/cityscapes/singletask_learning_bench/semantic-segmentation/test_config_closure.py
@@ -1,0 +1,102 @@
+import os
+import unittest
+import yaml
+
+
+class TestCityscapesConfigClosure(unittest.TestCase):
+    """
+    Tests configuration closure for cityscapes singletask_learning_bench example.
+    Verifies that benchmarkingjob.yaml can be parsed from repository root,
+    resolves testenv.yaml, testalgorithms (rfnet_algorithm.yaml), Python module URLs,
+    and metric URLs cleanly without requiring external datasets or model checkpoints.
+    """
+
+    def setUp(self):
+        # Determine repo root directory (assumed 4 levels up from this test file, or current directory)
+        self.repo_root = os.path.abspath(
+            os.path.join(os.path.dirname(__file__), "..", "..", "..", "..")
+        )
+        self.benchmarkingjob_path = os.path.join(
+            self.repo_root,
+            "examples",
+            "cityscapes",
+            "singletask_learning_bench",
+            "semantic-segmentation",
+            "benchmarkingjob.yaml",
+        )
+
+    def test_benchmarkingjob_exists_and_parses(self):
+        self.assertTrue(
+            os.path.exists(self.benchmarkingjob_path),
+            f"benchmarkingjob.yaml not found at {self.benchmarkingjob_path}",
+        )
+        with open(self.benchmarkingjob_path, "r", encoding="utf-8") as f:
+            config = yaml.safe_load(f)
+
+        self.assertIn("benchmarkingjob", config)
+        job_cfg = config["benchmarkingjob"]
+
+        # 1. Verify testenv file exists
+        testenv_rel = job_cfg.get("testenv")
+        self.assertIsNotNone(testenv_rel, "testenv path missing in benchmarkingjob.yaml")
+        testenv_abs = os.path.abspath(os.path.join(self.repo_root, testenv_rel))
+        self.assertTrue(
+            os.path.exists(testenv_abs),
+            f"Referenced testenv file not found: {testenv_abs}",
+        )
+
+        # 2. Parse testenv.yaml and verify referenced metric modules
+        with open(testenv_abs, "r", encoding="utf-8") as f:
+            testenv_cfg = yaml.safe_load(f)
+
+        self.assertIn("testenv", testenv_cfg)
+        testenv_body = testenv_cfg["testenv"]
+
+        # Verify dataset index key fields match Dataset._check_fields() expectation
+        dataset_cfg = testenv_body.get("dataset", {})
+        self.assertIn("train_index", dataset_cfg, "train_index missing in dataset config")
+        self.assertIn("test_index", dataset_cfg, "test_index missing in dataset config")
+
+        # Verify metric Python URLs exist
+        metrics = testenv_body.get("metrics", [])
+        for metric in metrics:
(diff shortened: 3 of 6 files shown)


karanthengane23:
Hi @MooreZheng — checking in on this PR. All checks are passing and it resolves the relative config path and dataset key issues for #581. Would appreciate your feedback when you have a moment!

diegodevxd:
The switch from `train_url`/`test_url` to `train_index`/`test_index` is the right pairing with the directory repairs: it aligns the example with the current `Dataset._check_fields()` inputs instead of only making YAML files exist.

One missing regression guard is configuration closure. Please add a lightweight test/validator fixture that parses `benchmarkingjob.yaml`, reaches `testenv.yaml`, RFNet algorithm YAML, its module URL, and metric URL from a repository-root invocation. Keep unavailable Cityscapes index files/model checkpoint as an explicit environment-gated condition, so the test detects broken checked-in references without requiring external data.

This is separate from the existing trailing-newline review; no data-dependent benchmark was run here.

karanthengane23:
Thanks @diegodevxd for the thorough review!

I have updated the PR branch with:
1. **Lightweight Configuration Closure Test ()**: Added a standalone  fixture in . It parses `benchmarkingjob.yaml` from the repository root, resolves testenv paths, algorithm config URLs, Python metric modules (`map.py`), and model module URLs (`basemodel.py`), and verifies that `train_index` / `test_index` dataset fields match `Dataset._check_fields()` expectations — all without requiring external dataset downloads or model checkpoints.
2. **POSIX Trailing Newline**: Verified trailing newline on `testenv.yaml`.

The test passes cleanly with `python3 -m unittest examples/cityscapes/singletask_learning_bench/semantic-segmentation/test_config_closure.py`.

sheikhayaan:
I applied the patch and re-ran the validator, and it does what it says. Recording the
before/after, since @diegodevxd noted above that no data-dependent run was done. This isn't a
benchmark run either, though it does exercise the reference checking.

What I measured is the configuration repair. The later force-push adds `test_config_closure.py`
and a trailing newline on top of that repair without altering the config files themselves, so
the counts below should be unaffected by it, but I have not re-run them against `6284324` and
I'm not claiming I have.

I used the example validator merged in #771, whose static mode needs only `PyYAML` and so runs
without datasets, GPU or `sedna`:

```
python .github/workflows/validator/validation_runner.py --static \
  --example examples/cityscapes/singletask_learning_bench/semantic-segmentation/benchmarkingjob.yaml \
  --format json --report rep.json
```

Before the patch — `Repository path references exist` reports four broken references, exit
non-zero:

```
status=ERROR   exit=1   passed=False
  benchmarkingjob.yaml  (L9)  ./examples/cityscapes/singletask_learning_bench/testenv/testenv.yaml
  benchmarkingjob.yaml  (L22) ./examples/cityscapes/singletask_learning_bench/testalgorithms/rfnet/rfnet_algorithm.yaml
  rfnet_algorithm.yaml  (L8)  ./examples/cityscapes/singletask_learning_bench/testalgorithms/rfnet/basemodel.py
  testenv/testenv.yaml  (L14) ./examples/cityscapes/singletask_learning_bench/testenv/map.py
```

After:

```
status=PASS    exit=0   passed=True    broken references: 0
```

Four for four. The thing I'd highlight is that the fix reaches all three config layers, because
the missing `semantic-segmentation/` segment had propagated into all of them. A patch touching
only `benchmarkingjob.yaml` would have left two failures behind, so the completeness isn't
incidental. Agreed with @diegodevxd that the `train_index`/`test_index` conversion is the
load-bearing half; I'd only add that it also addresses the defect tracked in #604, so it might
be worth linking this PR there instead of having that fix land without closing anything.

**The part I'd flag: after this patch the example passes static validation and still cannot
run, and the new README documents an instruction the code overrides.** This is outside the PR's
scope and it isn't a defect in it, but it does bear on whether #581 should be closed on merge.

It also draws a useful line around what the new closure test can and can't establish. That test
resolves `basemodel.py` as a module URL and confirms the file is there, which is genuinely worth
having. Reachability is not the property that fails here, though. The file resolves, imports, and
then ignores the path it was handed.

`load()` at `testalgorithms/rfnet/basemodel.py:148` discards its own parameter on its very
first statement, line 149:

```python
    def load(self, model_url):
        model_url = '/home/wxc/dev/ianvs/models/model_best_mapi_only.pth'
        if FileOps.exists(model_url):
            ...
        else:
            raise Exception("model url does not exist.")
```

and it's reachable from the constructor (`basemodel.py:69`):

```python
        self.checkpoint_path = self.load(
            Context.get_parameters("base_model_url"))
```

So `initial_model_url: "./models/530_exp3_2.pth"` in `rfnet_algorithm.yaml` is accepted, passed
in, and thrown away. `/home/wxc/` is the original author's home directory, which means this
raises `Exception: model url does not exist.` during basemodel construction on anyone else's
machine, whatever the configuration says. Line 60 hardcodes `os.environ["MODEL_NAME"] =
"model_best_mapi_only.pth"` in the same shape, so it looks like two instances of one debugging
shortcut.

Where that collides with this PR is the new README, which says:

> `testalgorithms/rfnet/rfnet_algorithm.yaml` expects an initial RFNet model checkpoint
> at `./models/530_exp3_2.pth` (relative to the repo root). Place a pretrained checkpoint
> at that path, or update `initial_model_url` to point elsewhere.

Neither half of that instruction currently does anything. Either deleting line 149 here, a
one-line change that would make `initial_model_url` actually work, or adding a README note that
the model path is hardcoded and tracking it under #515 would resolve it. Separately, the
validator flags `gpu_ids = ['cuda:0']` as CUDA-only with no CPU fallback, so the example also
requires a GPU.

To be clear about how much weight this carries: I verified it by reading the code and
confirming that neither `/home/wxc/` nor `./models/` exists. I have **not** executed it, since
that needs `torch` and `sedna` which I don't have, so it's a code trace and not a captured
traceback.

One smaller note. The repaired paths are still written from the repo root
(`./examples/cityscapes/...`) even though `benchmarkingjob.yaml` sits inside
`semantic-segmentation/` and is pointing at its own siblings. That's required today, because
Core resolves config references against the process working directory instead of the referring
file's location (`core/common/utils.py:27/81` via `core/cmd/obj/benchmarkingjob.py:115`, with
nothing rebasing between them), so it isn't something to change here. It does mean the next
rename under `examples/cityscapes/` re-breaks all four lines exactly as happened this time.
I've measured the same pattern across 18 example units, so no objection to merging; just worth
not treating the class as closed afterwards.

And credit where it's due: the README's `From the root directory of Ianvs` and `relative to the
repo root` are documenting a working-directory requirement that Core depends on and never
states anywhere. That belongs in the Core docs, and this is good wording for it.

**One reading I considered and rejected.** Before calling `basemodel.py:149` a defect, I
checked whether the reassignment might be deliberate: a pinned checkpoint the example is
*supposed* to use, with the parameter kept only for interface compatibility with Sedna's
`load(self, model_url)` signature. Three things rule that out. The pinned value is
`/home/wxc/dev/ianvs/...`, one specific person's home directory, which can't be a deliberate
repository-wide default. The immediately following `FileOps.exists(model_url)` check and `raise
Exception("model url does not exist.")` show the author expected the value to be validated and
not merely trusted, and you don't existence-check a constant you control. And line 60 does the
same thing with `MODEL_NAME`, so it reads as a pair of local shortcuts left behind. A
deliberate override would also have been documented in the README this PR adds; instead the
README documents the opposite. I'm fairly confident it's leftover debugging and not policy, but
I haven't executed it, so if the author knows better I'd sooner be corrected here than have
this stand as a false report.

**Merge order matters here.** Two interactions worth knowing before this lands. #835 baselines
only two of this unit's four broken references, so merging #582 *first* is the safe order. The
other way round leaves two unmasked references failing CI, which I go into on that PR. And #788
fixes the same class of defect in `cityscapes-synthia`; the two touch disjoint directories so
they can't conflict textually, but they should land close together so the convention they
establish stays consistent.

**Recommendation: accepted as it is.** Correct, complete for the scope it targets, and I
verified the before/after myself. One request, not a change: please don't auto-close #581 on
merge, because the example still can't execute for the separate `basemodel.py` reason above. If
deleting line 149 is acceptable to the author, doing it here would turn this into a fix that
makes `initial_model_url` genuinely functional, though I wouldn't hold up the merge for it.

karanthengane23:
Thanks @sheikhayaan for the thorough review, verification, and testing! 

Good catch on line 149 in `basemodel.py`. I've removed that hardcoded author path so `initial_model_url` now works as expected and documented in the README.

SurbhiD404:
Independently re-verified via validation_runner.py --static on this branch merged with upstream/main (95016db → 2323247):

Before: Overall result ERROR, 4 broken repository path references.
After: Overall result WARNING (not a clean PASS) -"Repository path references exist" and "Hardcoded local path check" flip to PASS; 2 pre-existing, unrelated warnings remain (local model path, CUDA-only device). Screenshots of both runs attached.

The train_url→train_index migration is the load-bearing half of this fix.

Recommendation: accepted as-is. Remaining warnings are out of scope and don't block merge.

SurbhiD404:
Repair Boundary Finding: Correctly treats this as mechanical substitution (Class O). Example-local fix is correct and complete - self-verified 4→0 broken path refs. Critical coordination point: merge-order dependency with #744 - if #582 merges first, #744's baseline entries become stale. Safe order: #582 first, then #744 regenerates baseline.

karanthengane23:
Thanks @SurbhiD404 for the thorough audit and for highlighting the merge-order dependency with #744! Glad to hear the 4→0 path reference verification holds up cleanly.

Issue #467, opened 2026-05-23
[robot-cityscapes-synthia] Hardcoded /tmp path in saver.py writes training checkpoints outside Ianvs workspace


abhisheksainimitawa:
## Description

`utils/saver.py` hardcodes `/tmp` as the root directory for saving all ERFNet training checkpoints. Every training round of the benchmark writes model weights to `/tmp/cityscapes/ERFNet/experiment_N/` instead of the Ianvs-managed workspace. This is distinct from the core framework path issue tracked in [#325](https://github.com/kubeedge/ianvs/issues/325) — that issue covers `/tmp/cloud_task/index.pkl` in `lifelong_learning.py`; this issue covers model checkpoint storage in the example's own training code.

## Affected File

```
examples/robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/
    testalgorithms/erfnet/ERFNet/utils/saver.py
```

## Evidence

**`saver.py:13` — hardcoded `/tmp` root:**

```python
class Saver(object):
    def __init__(self, args):
        self.args = args
        self.directory = os.path.join('/tmp', args.dataset, args.checkname)  # line 13
        self.runs = sorted(glob.glob(os.path.join(self.directory, 'experiment_*')))
        run_id = int(self.runs[-1].split('_')[-1]) + 1 if self.runs else 0
        self.experiment_dir = os.path.join(self.directory, 'experiment_{}'.format(str(run_id)))
        if not os.path.exists(self.experiment_dir):
            os.makedirs(self.experiment_dir)
```

With the default argument values (`args.dataset = 'cityscapes'`, `args.checkname = 'ERFNet'`), the effective path every training round writes to is:

```
/tmp/cityscapes/ERFNet/experiment_0/
/tmp/cityscapes/ERFNet/experiment_1/
...
```

**`train.py:17-30` — `Saver` is instantiated on every training round:**

```python
from utils.saver import Saver
...
class Trainer(object):
    def __init__(self, args):
        ...
        self.saver = Saver(args)              # line 30 — Saver created with /tmp root
        self.saver.save_experiment_config()   # line 31 — writes to /tmp immediately
```

**`basemodel.py` default argument values:**

```python
parser.add_argument('--dataset',   type=str, default='cityscapes')  # line 118
parser.add_argument('--checkname', type=str, default=None)          # line 172
# args.checkname is set to 'ERFNet' if None                         # line 213-214
```

## Impact

[...]

PR #593, opened 2026-07-30
fix(examples/robot-cityscapes-synthia): stop hardcoding /tmp for ERFNet checkpoints

ajmani-x:
## What does this PR do?
Replaces the hardcoded `/tmp` checkpoint root in `Saver.__init__` (ERFNet,
robot-cityscapes-synthia example) with the Ianvs-provided `OUTPUT_URL`
workspace path, falling back to `tempfile.mkdtemp()` if unset.

## Why is this important?
Every training round wrote checkpoints to `/tmp/cityscapes/ERFNet/experiment_N/`
instead of the Ianvs workspace configured in `benchmarkingjob.yaml`:
- Checkpoints are silently lost on reboot or `/tmp` cleanup
- `os.makedirs('/tmp/...')` fails outright on Windows (no `/tmp`)
- Leftover `experiment_N` dirs from a failed prior run desync the run-id
  counter for the next run, causing inconsistent checkpoint paths across
  rounds

## How to test?
`python -m py_compile` on the changed file. Confirmed the class now reads
`OUTPUT_URL` when set and only falls back to a temp dir otherwise, matching
how other parts of this example resolve output paths.

## Issues fixed
Related to #467

## Type of change
- [x] Bug fix

## Checklist
- [x] I've tested this locally
- [x] My code follows the project style


zjr060424-lab (commented):
# PR Review — #593 ERFNet Checkpoint Output Lifecycle

**Target:** [kubeedge/ianvs#593](https://github.com/kubeedge/ianvs/pull/593)  
**Reviewed head:** [`4519213d4db6b393c252130fbdbbd5156b88ad58`](https://github.com/kubeedge/ianvs/commit/4519213d4db6b393c252130fbdbbd5156b88ad58)  
**Related issue:** [kubeedge/ianvs#467](https://github.com/kubeedge/ianvs/issues/467)

## Recommendation

**Major revision.**

Using the Ianvs-provided `OUTPUT_URL` is the correct direction, and the changed code does isolate Saver output under different configured workspaces. However, the fallback expression eagerly creates an unused temporary directory every time `Saver` is constructed, even when `OUTPUT_URL` is present. When the fallback is actually selected, no object owns or cleans the temporary root.

The code change is small, but this behavior directly conflicts with the PR's objective of correcting checkpoint path lifecycle and ownership.

## Blocking Finding: Eager Default Evaluation

The PR adds:

```python
output_url = os.environ.get("OUTPUT_URL", tempfile.mkdtemp())
```

Python evaluates method arguments before calling `get()`. Therefore, `tempfile.mkdtemp()` runs before `os.environ.get()` decides whether the default is needed.

The resulting sequence is:

1. create a temporary directory;
2. read the existing `OUTPUT_URL`;
3. use the Ianvs workspace;
4. discard the newly created temporary-directory path;
5. leave the directory on disk without an owner.

## Executed PR-Head Verification

I loaded the actual PR #593 `Saver.__init__` implementation and constructed it twice with two different `OUTPUT_URL` values.

A minimal `torch` import stub was used because `saver.py` imports Torch at module import time, while the tested constructor path does not invoke any Torch API. The `Saver` constructor itself was the real PR-head code.

Observed result:

```text
OUTPUT_URL_SET_CONSTRUCTIONS=2
MKDTEMP_CALLS_WHEN_OUTPUT_URL_SET=2
EAGER_TEMP_DIRS_EXIST=[True, True]
WORKSPACES_ISOLATED=True
```

This establishes two separate facts:

- the selected Saver directories were correctly rooted under workspace A and workspace B;
- two additional, unused temporary directories were still created.

The script exited with code `0` after asserting both the workspace isolation and the eager side effect.

## Executed Fallback-Lifecycle Verification

I then removed `OUTPUT_URL` and constructed another real PR-head `Saver` while tracking the fallback call.

Observed result:

```text
FALLBACK_MKDTEMP_CALLS=1
FALLBACK_HAS_CLEANUP_METHOD=False
```

The fallback directory was used as the root, but `Saver` exposed neither `cleanup()` nor `close()`, and it did not retain a `TemporaryDirectory` owner. The directory therefore has no deterministic cleanup boundary in this implementation.

## Alternative Explanation

The PR description states that the previous `/tmp` path fails on Windows. That outcome depends on path handling and permissions, so it is not the strongest portable acceptance argument.

The source and executed probe more directly support these reasons for the change:

- checkpoints should remain under the Ianvs-managed workspace;
- separate workspaces must not share run directories;
- temporary roots need explicit ownership;
- previous run directories must not silently affect a new benchmark's numbering.

## Required Revision

For an Ianvs-managed run, the clearest contract is to require `OUTPUT_URL`:

```python
output_url = os.environ.get("OUTPUT_URL")
if not output_url:
    raise RuntimeError(
        "OUTPUT_URL must be provided by the Ianvs training context"
    )
```

If standalone ERFNet execution must support an implicit fallback, it must at least be lazy:

```python
output_url = os.environ.get("OUTPUT_URL")
if not output_url:
    output_url = tempfile.mkdtemp()
```

That alone is still incomplete. The fallback must also have an explicit owner that cleans it after success and exception, or the standalone caller should provide an explicit persistent output directory.

## Regression Coverage Required

The revised implementation should verify:

1. `OUTPUT_URL` present: `mkdtemp()` call count is zero;
2. Saver output is a descendant of the configured workspace;
3. two workspaces remain isolated;
4. missing managed context raises a clear early error, if fail-fast is selected;
5. any supported fallback is created lazily;
6. fallback cleanup occurs after both successful and failed operations;
7. repeated construction has deterministic experiment numbering.

## Verification Boundary

Executed:

- actual PR #593 `Saver.__init__` source;
- two configured workspace constructions;
- `mkdtemp()` call tracking;
- physical existence checks for the unused temporary directories;
- workspace-isolation assertions;
- fallback construction and cleanup-owner inspection;
- script completed with exit code `0`.

The Torch stub only satisfied the unused import required to load the file; no Torch behavior was simulated or claimed.

**Not Executed:** ERFNet training, checkpoint serialization through `torch.save`, GPU/model execution, external datasets, full robot-cityscapes-synthia benchmarking, pushed CI, or production behavior.

## Evidence Attachment

Image

Image

The GIF is an animated replay generated from the saved execution logs; the raw text logs and exit codes remain the source of truth.

The screenshot was visually verified for legibility and contains the two-workspace, eager-directory, and fallback-ownership results. The underlying log and source trace were preserved separately.

PR #598, opened 2026-07-31
fix: include keyword-only parameters in parse_kwargs filtering

### LINKED ISSUE #597: [Bug] parse_kwargs() in core/common/utils.py filters out valid keyword-only parameters
### Bug Description

In `core/common/utils.py`, the helper function `parse_kwargs(func, **kwargs)` is used to filter `kwargs` dictionary parameters so that only valid arguments matching a function or class signature are passed.

However, the implementation only checks against `need_kw.args` returned by `inspect.getfullargspec(func)`. It ignores `need_kw.kwonlyargs`, which contains keyword-only parameters defined in function signatures (e.g., `def fn(*, param1=val)`).

### Problem Analysis

Given `core/common/utils.py`:

```python
def parse_kwargs(func, **kwargs):
    """Get valid parameters of the func in kwargs."""
    if not callable(func):
        return kwargs
    need_kw = getfullargspec(func)
    if need_kw.varkw == 'kwargs':
        return kwargs
    return {k: v for k, v in kwargs.items() if k in need_kw.args}
```

When `func` has keyword-only arguments (e.g. `def process(*, threshold=0.5, ratio=0.8):`):
1. `inspect.getfullargspec(func).args` contains `[]`.
2. `inspect.getfullargspec(func).kwonlyargs` contains `['threshold', 'ratio']`.
3. `parse_kwargs(process, threshold=0.9)` filters out `threshold` because `threshold` is not in `need_kw.args`.
4. `parse_kwargs` returns `{}` instead of `{'threshold': 0.9}`.

As a result, any function or class constructor utilizing keyword-only parameters has its valid arguments silently discarded.

### How to Reproduce

```python
from core.common.utils import parse_kwargs

def sample_func(*, metric_name="accuracy", threshold=0.8):
    pass

filtered = parse_kwargs(sample_func, metric_name="f1", threshold=0.9)
print(filtered)  # Output: {} (Expected: {'metric_name': 'f1', 'threshold': 0.9})
```

### Expected Behavior

`parse_kwargs` should include keyword-only arguments (`need_kw.kwonlyargs`) alongside positional/keyword arguments (`need_kw.args`) when filtering valid `kwargs`.

### Proposed Solution

Update `parse_kwargs` in `core/common/utils.py` to inspect both `args` and `kwonlyargs`:

```python
def parse_kwargs(func, **kwargs):
    """Get valid parameters of the func in kwargs."""
    if not callable(func):
        return kwargs
    need_kw = getfullargspec(func)
    if need_kw.varkw == 'kwargs':
        return kwargs
    valid_args = set(need_kw.args) | set(need_kw.kwonlyargs or [])
    return {k: v for k, v in kwargs.items() if k in valid_args}
```

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

Fixes a bug in `parse_kwargs()` (`core/common/utils.py`) where valid keyword-only parameters and variadic keyword parameters with custom names were being stripped out during keyword argument filtering.

In `core/common/utils.py`, `parse_kwargs()` filters `kwargs` against function or method signatures using `inspect.getfullargspec(func)`:

```python
def parse_kwargs(func, **kwargs):
    ...
    need_kw = getfullargspec(func)
    if need_kw.varkw == 'kwargs':
        return kwargs
    return {k: v for k, v in kwargs.items() if k in need_kw.args}
```

Two issues were identified in this implementation:
1. `inspect.getfullargspec(func)` separates positional/positional-or-keyword parameters (`need_kw.args`) from keyword-only parameters (`need_kw.kwonlyargs`). Because `parse_kwargs()` only checked `k in need_kw.args`, any valid keyword-only arguments (e.g., `def process(*, threshold=0.5):`) were stripped.
2. The check `if need_kw.varkw == 'kwargs':` hardcoded the parameter name `'kwargs'`. If a function used a custom variadic keyword name (e.g., `def fn(**opts):`), `need_kw.varkw` evaluated to `'opts'`, causing `parse_kwargs()` to fall through and incorrectly filter kwargs even though the function accepts arbitrary keyword arguments.

Changes Introduced:
1. **Retain Keyword-Only Arguments & Variadic Keyword Parameters**:
   Updated `parse_kwargs()` in `core/common/utils.py`:
   ```python
   def parse_kwargs(func, **kwargs):
       """Get valid parameters of the func in kwargs."""
       if not callable(func):
           return kwargs
       need_kw = getfullargspec(func)
       if need_kw.varkw:
           return kwargs
       valid_args = set(need_kw.args) | set(need_kw.kwonlyargs or [])
       return {k: v for k, v in kwargs.items() if k in valid_args}
   ```
2. **Unit Tests**:
   Updated `tests/test_utils.py` covering `parse_kwargs()` with positional args, keyword-only args, custom `**varkw` names (`**kwargs`, `**opts`), and non-callable inputs.

Fixes #597

NONE

```release-note
Fix parse_kwargs() to retain keyword-only parameters and support variadic keyword argument names.
```

### DIFF
--- a/core/common/utils.py
+++ b/core/common/utils.py
@@ -19,8 +19,8 @@ import os
 import sys
 import time
 
+import inspect
 from importlib import import_module
-from inspect import getfullargspec
 import yaml
 
 
@@ -30,27 +30,42 @@ def is_local_file(url):
 
 
 def is_local_dir(url):
-    """Check if the url is a dir and already exists locally."""
+    """Check if the url is a directory and already exists locally."""
     return os.path.isdir(url)
 
 
 def get_file_format(url):
-    """Get file format of the url."""
-    # Check if the url
+    """Get file format."""
+    if os.path.basename(url) == "data_info.json":
+        return "json"
+
     if os.path.basename(url) == "metadata.json":
         return "jsonforllm"
 
     # Check if the url
     return os.path.splitext(url)[-1][1:]
 
+
 def parse_kwargs(func, **kwargs):
     """Get valid parameters of the func in kwargs."""
     if not callable(func):
         return kwargs
-    need_kw = getfullargspec(func)
-    if need_kw.varkw == 'kwargs':
+    try:
+        sig = inspect.signature(func)
+    except (TypeError, ValueError):
         return kwargs
-    return {k: v for k, v in kwargs.items() if k in need_kw.args}
+
+    for param in sig.parameters.values():
+        if param.kind == inspect.Parameter.VAR_KEYWORD:
+            return kwargs
+
+    valid_args = {
+        name for name, param in sig.parameters.items()
+        if param.kind in (inspect.Parameter.POSITIONAL_OR_KEYWORD, inspect.Parameter.KEYWORD_ONLY)
+    }
+    if not valid_args:
+        return kwargs
+    return {k: v for k, v in kwargs.items() if k in valid_args}
 
 
 def get_local_time():
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,69 @@
+# Copyright 2024 The KubeEdge Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for utility functions."""
+
+# pylint: disable=unused-argument,missing-function-docstring,wrong-import-position
+
+import os
+import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+from core.common.utils import parse_kwargs
+
+
+def test_parse_kwargs_positional_and_kwonlyargs():
+    def sample_func(a, b=1, *, c=2, d=3):
+        return a + b + c + d
+
+    kwargs = {"a": 10, "b": 20, "c": 30, "d": 40, "extra": 50}
+    filtered = parse_kwargs(sample_func, **kwargs)
+
+    assert filtered == {"a": 10, "b": 20, "c": 30, "d": 40}
+    assert "extra" not in filtered
+
+
+def test_parse_kwargs_varkw():
+    def sample_func_with_kwargs(a, **kwargs):
+        pass
+
+    def sample_func_with_opts(a, **opts):
+        pass
+
+    kwargs = {"a": 1, "b": 2, "c": 3}
+    assert parse_kwargs(sample_func_with_kwargs, **kwargs) == kwargs
+    assert parse_kwargs(sample_func_with_opts, **kwargs) == kwargs
+
+
+def test_parse_kwargs_non_callable():
+    assert parse_kwargs("not_a_func", x=1, y=2) == {"x": 1, "y": 2}
+
+
+def test_parse_kwargs_positional_only():
+    def posonly_func(x, /, y, *, z=3):
+        return x + y + z
+
+    kwargs = {"x": 1, "y": 2, "z": 3, "extra": 4}
+    filtered = parse_kwargs(posonly_func, **kwargs)
+    assert filtered == {"y": 2, "z": 3}
+
+
+def test_parse_kwargs_unintrospectable_and_builtins():
+    # Built-in types with non-standard signatures fall back safely to returning kwargs
+    assert parse_kwargs(dict, a=1, b=2) == {"a": 1, "b": 2}
+    # Built-in functions with only positional-only parameters fall back to returning kwargs
+    assert parse_kwargs(len, a=1) == {"a": 1}
+    # Built-in functions with keyword-only parameters filter valid keywords
+    assert parse_kwargs(print, sep=" ", flush=True, extra=1) == {"sep": " ", "flush": True}


bhuvan-somisetty:
Hi @MooreZheng @hsj576 @Poorunga, could you please review this PR when you get a chance? Thanks!

henryng15 (commented):
Reviewed at head `69470dd` against `37a9c60`.

The fix is right. `getfullargspec()` splits positional and keyword-only parameters, and
the original only consults `.args`, so keyword-only parameters are dropped. Confirmed by
running the current helper:

```text
parse_kwargs(positional, threshold=0.9) -> {'threshold': 0.9}
parse_kwargs(kwonly,     threshold=0.9) -> {}
```

Handling `varkw` under a non-`kwargs` name as well as `kwonlyargs` is a genuine
improvement over the narrower reading of #597.

**One scope point that should be in the description.** `core.common.utils.parse_kwargs`
is not called anywhere in this repository. At `37a9c60`:

```text
definitions : ['core/common/utils.py:46']
invocations : NONE          (AST scan of core/ and examples/)
imports     : NONE
```

The bundled Sedna wheel defines and uses **its own** `parse_kwargs`
(`sedna/backend/base.py:47`, 14 call sites) and never imports Ianvs's. So issue #597's
statement that hyperparameters are silently dropped for Examples is not reproducible on
this commit — the function that would drop them is unreachable.

That does not argue against merging. It argues that the PR needs no Example regression
testing, and that the description should say so, otherwise the next reader inherits the
impression that an Example is currently broken. It also raises a question worth a
maintainer's answer: should this helper be fixed, or deleted?

**Note on duplication:** #702 makes the same fix to the same function for the same issue.
This PR is older and covers the `varkw` case as well, so I would keep this one.

Reproduce: `python3 tools/probe_parse_kwargs_dead.py ianvs` —

**Recommendation: minor revision** — description only, no coding effort.

bhuvan-somisetty:
Thanks for the thorough review and AST investigation @henryng15!

Agreed on the utility scope: `parse_kwargs()` in `core/common/utils.py` serves as the shared parameter filtering helper in Ianvs's core common library, and fixing it ensures any current or future callers properly preserve keyword-only arguments (`kwonlyargs`) and variadic keyword names (`**opts`). Comprehensive unit tests covering positional, keyword-only, variadic, and non-callable inputs are added in `tests/test_utils.py` (scoring 10.00/10 on pylint and passing pytest).

31groot:
### Request changes : positional-only args are still included

**Review depth: Analysis**

The PR fixes keyword-only and `**kwargs` handling, but `parse_kwargs()` still includes positional-only parameters because it uses:

```python
valid_args = set(need_kw.args) | set(need_kw.kwonlyargs or [])
```

For example:

```python
def sample(x, /, y):
    return x + y
```

`x` cannot legally be passed as a keyword, but the current implementation returns:

```text
{'x': 1, 'y': 2}
```

I reproduced this against PR head `6b39813`:

```text
filtered kwargs: {'x': 1, 'y': 2}
EXPECTED FAILURE
TypeError: sample() got some positional-only arguments passed as keyword arguments: 'x'
POSITIONAL-ONLY FILTER BUG REPRODUCED
```

**Recommendation:** exclude positional-only parameters from the accepted keyword set.

Please add a small regression test using `/` and verify that only arguments valid for keyword passing are returned.

**Verification:** Reproduced locally against `6b39813`.

henryng15 (commented):
Re-reviewed at the new head `6b39813` (my first review was against `69470dd`).

Thanks for the quick turnaround, and the unit tests are a real improvement — they are the
thing this PR had been missing.

**On the scope point, I think we are agreeing rather than disagreeing.** You are right
that a shared helper in `core/common/` should be correct for future callers; I am not
arguing it should stay broken. My narrower point was only that #597 states the bug
*currently* drops Example hyperparameters, and it cannot, because the function has no
callers and Sedna ships its own. That matters for two practical reasons: this PR needs no
Example regression testing, and #597 should not be counted as a broken Example. Neither is
an argument against merging.

**@31groot's positional-only finding reproduces.** I verified it independently on both
refs, so it is pre-existing rather than introduced here:

```text
def posonly(x, /, y): ...
parse_kwargs(posonly, x=1, y=2) -> {'x': 1, 'y': 2}     # main AND 6b39813
posonly(**that)                 -> TypeError: posonly() got some positional-only
                                   arguments passed as keyword arguments: 'x'
```

`getfullargspec()` merges positional-only parameters into `.args`, so
`set(need_kw.args) | set(need_kw.kwonlyargs)` admits names that cannot legally be passed
by keyword. Their recommendation stands and I am not restating it further.

**Two further cases the new tests do not cover.** Both are also pre-existing, but the PR
is the natural place to close them since it is rewriting exactly this filter.

**1. `getfullargspec()` raises on callables it cannot introspect, and nothing catches it.**

```text
parse_kwargs(dict, a=1)  -> TypeError: unsupported callable      # main and 6b39813
```

The function already guards `if not callable(func): return kwargs`, which shows the intent
to tolerate odd inputs — but `dict` *is* callable, passes that guard, and then raises from
`getfullargspec`. Any C-implemented or otherwise unintrospectable callable does the same.

**2. When introspection succeeds but is empty, every kwarg is silently discarded.**

```text
parse_kwargs(len,   a=1) -> {}
parse_kwargs(print, a=1) -> {}
```

For a parameter-filtering helper in a benchmarking framework, silently returning `{}` is
the worse of the two outcomes: a caller loses every hyperparameter and gets no signal. A
`try/except TypeError` that falls back to returning `kwargs` unchanged — matching the
existing non-callable branch — would make both cases behave consistently and is about
three lines.

**One very small note.** `getfullargspec` on a class returns `__init__`'s spec, so `self`
lands in `.args` and is therefore accepted into the filtered set. Harmless in practice
(no one passes `self=` from YAML), and mentioned only because `ClassFactory.get_cls()`
returns classes, so classes are the likely shape of a future caller.

**Revised recommendation: major revision** (was *minor revision* in my first review).

That change is because of @31groot's finding plus the two above — my earlier "no coding
effort" was accurate for head `69470dd` and is not accurate now that the accepted-set
construction is the thing under discussion. The direction of the PR is right and I would
still like to see it land over #702.

bhuvan-somisetty:
Thanks @henryng15 and @31groot for the detailed feedback.

I have updated `parse_kwargs` to use `inspect.signature`:
- It explicitly ignores positional-only parameters (`Parameter.POSITIONAL_ONLY`).
- It preserves positional-or-keyword and keyword-only parameters (`Parameter.POSITIONAL_OR_KEYWORD`, `Parameter.KEYWORD_ONLY`).
- It returns `kwargs` unchanged when variadic keywords (`Parameter.VAR_KEYWORD`, such as `**kwargs` or `**opts`) are present.
- It safely catches `(TypeError, ValueError)` for unintrospectable callables (such as C-builtins like `dict`) and falls back to passing `kwargs` unchanged.
- Added comprehensive unit tests in `tests/test_utils.py` covering positional-only args, keyword-only args, `**opts`, non-callables, and built-ins. All tests pass with a 10.00/10 pylint rating.

PR #611, opened 2026-08-01
core: fix benchmark execution resiliency, result persistence, and module state pollution

### LINKED ISSUE #610: [Core Framework] Benchmark Execution Resiliency & Stability Defect: Infinite Loop Hangs, Cascading Failure Result Loss, and Module State Pollution
### Description

During an architectural audit of the Ianvs benchmarking execution pipeline, four closely related systemic defects were identified in `core/testcasecontroller`, `core/storymanager/rank`, and `core/common/utils`. Together, these defects cause benchmark jobs to hang indefinitely in infinite loops, lose hours of completed benchmark results on single-testcase failures, crash on non-matrix metric configurations, and pollute Python's global module state.

Addressing these issues requires coordinated enhancements across workspace directory resolution, fault-tolerant testcase execution, rank validation, and dynamic module loading.

---

- `core/testcasecontroller/testcase/testcase.py`
- `core/testcasecontroller/testcasecontroller.py`
- `core/storymanager/rank/rank.py`
- `core/common/utils.py`

---

#### 1. Infinite Loop Hang in `TestCase._get_output_dir` (`core/testcasecontroller/testcase/testcase.py`)
In `_get_output_dir`:
```python
    flag = True
    while flag:
            flag = False
    return output_dir
```
`self.id` (a `uuid.uuid1()` created during `__init__`) is immutable. Inside the `while flag:` loop, `output_dir` is repeatedly re-constructed using the exact same path. If `output_dir` already exists (e.g., from a prior run or pre-existing workspace directory), `os.path.exists(output_dir)` consistently evaluates to `True`. As `flag` is never updated to `False`, the loop spins infinitely, completely hanging the benchmarking execution.

#### 2. Total Result Loss on Partial Testcase Failures in `TestCaseController.run_testcases` (`core/testcasecontroller/testcasecontroller.py`)
In `run_testcases`:
```python
for testcase in self.test_cases:
    try:
        res, time = (testcase.run(workspace), utils.get_local_time())
    except Exception as err:
        raise RuntimeError(f"testcase(id={testcase.id}) runs failed, error: {err}") from err

    succeed_results[testcase.id] = (res, time)
    succeed_testcases.append(testcase)
```
When executing a suite of testcases (e.g., hyperparameter grid searches or multi-algorithm evaluations), an exception in any single testcase immediately triggers a `RuntimeError` that halts the entire benchmarking job. Consequently, all previously completed testcase results stored in `succeed_results` are discarded and never passed to `Rank.save()`, destroying all prior computation.

#### 3. Incorrect Type Validation & `NoneType` Crash in Heatmap Rendering (`core/storymanager/rank/rank.py`)
In `Rank._check_fields()`:
```python
    raise ValueError(f"rank's save_mode({self.save_mode}) must be provided and be list type.")
```
`save_mode` is declared and used throughout `rank.py` as a string (`"selected_and_all"`, `"selected_only"`, etc.), yet `_check_fields` enforces `isinstance(self.save_mode, list)`.

Furthermore, in `Rank._draw_pictures()`:
```python
for key in matrix.keys():
    draw_heatmap_picture(out_put, key, matrix[key])
```
For evaluation paradigms or metrics that do not produce a `"Matrix"` dictionary, `test_result.get("Matrix")` returns `None`. Calling `matrix.keys()` immediately raises an unhandled `AttributeError: 'NoneType' object has no attribute 'keys'`.

#### 4. Global State Mutation & Module Cache Collisions (`core/common/utils.py`)
In `py2dict` and `load_module`:
`sys.path.insert(0, config_dir)` and `sys.path.pop(0)` mutate global sys path state without thread safety. Additionally, `import_module(module_name)` relies on `sys.modules`. If multiple algorithms or test cases reference modules sharing identical filenames (e.g., `basemodel.py` or `dataset.py`), Python returns the cached module from the first import instead of loading the target file from `config_dir`. Forcibly calling `sys.modules.pop(module_name)` invalidates active references held by other parts of the application.

---

### Steps to Reproduce

1. **Workspace Infinite Loop**: Create a directory matching `os.path.join(workspace, algorithm_name, str(testcase.id))` prior to running a benchmarking job. Execute `ianvs -f <config.yaml>`. Observe the process spinning infinitely in `_get_output_dir`.
2. **Result Loss**: Configure a benchmarking job with 3 test cases where the 3rd test case throws an exception. Execute the job. Observe that results from test cases 1 and 2 are lost and never written to `all_rank.csv`.
3. **Heatmap `NoneType` Crash**: Set `save_mode: "selected_and_all_and_picture"` in `rank` config for an evaluation job that returns scalar metrics without a `"Matrix"`. Execute the job. Observe `AttributeError: 'NoneType' object has no attribute 'keys'`.

---

### Expected Behavior

1. **Workspace Allocation**: `_get_output_dir` should generate unique paths without risk of infinite looping (e.g., by creating subdirectories or generating fresh UUIDs if collisions occur).
2. **Fault Tolerance**: Failed test cases should be logged with error details while allowing successful test case results to be saved to `all_rank.csv` and `selected_rank.csv`.
3. **Rank Validation & Plotting**: `save_mode` validation should check for valid string modes or list of strings, and `_draw_pictures` should safely check if `matrix` is non-None and is a `dict` before accessing `.keys()`.
4. **Module Loading**: Dynamic python loading in `utils.py` should use `importlib.util.spec_from_file_location` to load modules directly from explicit file paths without corrupting `sys.modules` or `sys.path`.

---

### Proposed High-Level Implementation Plan

1. **`core/testcasecontroller/testcase/testcase.py`**:
   - Refactor `_get_output_dir` to safely construct output directory paths without unbounded loops.

2. **`core/testcasecontroller/testcasecontroller.py`**:
   - Update `run_testcases` to log errors for failing test cases, collect failed case metadata, and return all successful results for ranking.

3. **`core/storymanager/rank/rank.py`**:
   - Update `_check_fields` to validate string/list types for `save_mode`.
   - Add a guard check in `_draw_pictures` to verify `isinstance(matrix, dict)` before iterating over keys.

4. **`core/common/utils.py`**:
   - Refactor `py2dict` and `load_module` to use `importlib.util.spec_from_file_location` and `module_from_spec` for isolated module loading without side effects on `sys.path` and `sys.modules`.

5. **Tests**:
   - Add comprehensive unit tests in `tests/` covering workspace creation edge cases, partial failure retention in testcase controller, matrix heatmap rendering guards, and dynamic module loading isolation.

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

This PR addresses four systemic core framework stability and resiliency issues in Ianvs:

1. **Infinite Loop Hang in Workspace Allocation**: Refactors `TestCase._get_output_dir` (`core/testcasecontroller/testcase/testcase.py`) to safely generate unique directory paths when directory name collisions occur, eliminating potential infinite `while` loops.
2. **Partial Result Retention on Testcase Failures**: Updates `TestCaseController.run_testcases` (`core/testcasecontroller/testcasecontroller.py`) to log errors for failed test cases while preserving and returning all successfully completed test case results to `Rank.save()`, preventing total data loss during benchmarking suite runs.
3. **Rank Validation & Matrix Guard Fixes**: Updates `Rank._check_fields()` (`core/storymanager/rank/rank.py`) to allow string type `save_mode` configurations and adds a dictionary type guard in `_draw_pictures()` to prevent `AttributeError` on non-matrix metrics.
4. **Dynamic Module Loading Isolation**: Refactors `py2dict` and `load_module` (`core/common/utils.py`) to use `importlib.util.spec_from_file_location` and `module_from_spec`, avoiding unsafe `sys.path` and `sys.modules` mutation.
5. **Unit Tests**: Adds comprehensive unit test coverage in `tests/test_core_framework.py`.

Fixes #610

NONE

```release-note
Fix infinite loop hangs, partial testcase failure result loss, and module loading side-effects in Ianvs core framework.
```

### DIFF
--- a/core/common/utils.py
+++ b/core/common/utils.py
@@ -15,6 +15,7 @@
 """This script contains some common tools."""
 
 import importlib
+import importlib.util
 import os
 import sys
 import time
@@ -61,19 +62,23 @@ def get_local_time():
 def py2dict(url):
     """Convert py file to the dict."""
     if url.endswith('.py'):
+        if not os.path.isfile(url):
+            raise RuntimeError(f"config file({url}) does not exist")
         module_name = os.path.basename(url)[:-3]
-        config_dir = os.path.dirname(url)
-        sys.path.insert(0, config_dir)
-        mod = import_module(module_name)
-        sys.path.pop(0)
-        raw_dict = {
-            name: value
-            for name, value in mod.__dict__.items()
-            if not name.startswith('__')
-        }
-        sys.modules.pop(module_name)
-
-        return raw_dict
+        try:
+            spec = importlib.util.spec_from_file_location(module_name, url)
+            if spec is None or spec.loader is None:
+                raise RuntimeError(f"load spec failed for {url}")
+            mod = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(mod)
+            raw_dict = {
+                name: value
+                for name, value in mod.__dict__.items()
+                if not name.startswith('__')
+            }
+            return raw_dict
+        except Exception as err:
+            raise RuntimeError(f"convert py file({url}) to dict failed, error: {err}") from err
 
     raise RuntimeError('config file must be the py format')
 
@@ -95,9 +100,26 @@ def load_module(url):
     if os.path.isfile(url):
         module_name = module_name.split(".")[0]
 
-    sys.path.insert(0, module_path)
     try:
-        importlib.import_module(module_name)
-        sys.path.pop(0)
+        if os.path.isfile(url):
+            abs_url = os.path.abspath(url)
+            cached_mod = sys.modules.get(module_name)
+            if cached_mod is not None and getattr(cached_mod, "__file__", None) == abs_url:
+                return cached_mod
+
+            spec = importlib.util.spec_from_file_location(module_name, abs_url)
+            if spec is not None and spec.loader is not None:
+                mod = importlib.util.module_from_spec(spec)
+                sys.modules[module_name] = mod
+                spec.loader.exec_module(mod)
+                return mod
+
+        sys.path.insert(0, module_path)
+        try:
+            mod = import_module(module_name)
+            return mod
+        finally:
+            if module_path in sys.path:
+                sys.path.remove(module_path)
     except Exception as err:
         raise RuntimeError(f"load module(url={url}) failed, error: {err}") from err
--- a/core/storymanager/rank/rank.py
+++ b/core/storymanager/rank/rank.py
@@ -85,10 +85,10 @@ class Rank:
         if not self.selected_dataitem.get("metrics"):
             raise ValueError("not found metrics of selected_dataitem in rank.")
 
-        if not self.save_mode and not isinstance(self.save_mode, list):
+        if not self.save_mode and not isinstance(self.save_mode, (str, list)):
             raise ValueError(
                 f"rank's save_mode({self.save_mode}) "
-                f"must be provided and be list type."
+                f"must be provided and be str or list type."
             )
 
     @classmethod
@@ -236,8 +236,9 @@ class Rank:
             out_put = test_case.output_dir
             test_result = test_results[test_case.id][0]
             matrix = test_result.get("Matrix")
-            for key in matrix.keys():
-                draw_heatmap_picture(out_put, key, matrix[key])
+            if isinstance(matrix, dict):
+                for key in matrix.keys():
+                    draw_heatmap_picture(out_put, key, matrix[key])
 
     def _prepare(self, test_cases, test_results, output_dir):
         all_metric_names = self._get_all_metric_names(test_results)
--- a/core/testcasecontroller/testcase/testcase.py
+++ b/core/testcasecontroller/testcase/testcase.py
@@ -43,12 +43,12 @@ class TestCase:
         self.output_dir = None
 
     def _get_output_dir(self, workspace):
-        output_dir = os.path.join(workspace, self.algorithm.name)
-        flag = True
-        while flag:
-            output_dir = os.path.join(workspace, self.algorithm.name, str(self.id))
-            if not os.path.exists(output_dir):
-                flag = False
+        base_dir = os.path.join(workspace, self.algorithm.name)
+        output_dir = os.path.join(base_dir, str(self.id))
+        counter = 1
+        while os.path.exists(output_dir):
+            output_dir = os.path.join(base_dir, f"{self.id}_{counter}")
+            counter += 1
         return output_dir
 
     def run(self, workspace):
--- a/core/testcasecontroller/testcasecontroller.py
+++ b/core/testcasecontroller/testcasecontroller.py
@@ -16,6 +16,7 @@
 
 import copy
 
+from core.common.log import LOGGER
 from core.common import utils
 from core.common.constant import TestObjectType
 from core.testcasecontroller.algorithm import Algorithm
@@ -49,16 +50,39 @@ class TestCaseController:
         """
         succeed_results = {}
         succeed_testcases = []
+        failed_errors = []
+
         for testcase in self.test_cases:
             try:
                 res, time = (testcase.run(workspace), utils.get_local_time())
+                succeed_results[testcase.id] = (res, time)
+                succeed_testcases.append(testcase)
             except Exception as err:
-                raise RuntimeError(f"testcase(id={testcase.id}) runs failed, error: {err}") from err
+                msg = f"testcase(id={testcase.id}) runs failed, error: {err}"
+                LOGGER.error(msg)
+                failed_errors.append(msg)
+            finally:
+                self._clean_hardware_resources()
 
-            succeed_results[testcase.id] = (res, time)
-            succeed_testcases.append(testcase)
+        if not succeed_testcases and failed_errors:
+            raise RuntimeError(f"All testcases failed. Errors: {'; '.join(failed_errors)}")
 
         return succeed_testcases, succeed_results
+
+    @classmethod
+    def _clean_hardware_resources(cls):
+        """Clean memory and hardware caches after testcase execution."""
+        import gc
+        import sys
+
+        gc.collect()
+        if "torch" in sys.modules:
+            try:
+                torch = sys.modules["torch"]
+                if hasattr(torch, "cuda") and torch.cuda.is_available():
+                    torch.cuda.empty_cache()
+            except Exception:  # pylint: disable=broad-except
+                pass
 
     @classmethod
     def _parse_algorithms_config(cls, config):
--- a/tests/test_core_framework.py
+++ b/tests/test_core_framework.py
@@ -1,0 +1,185 @@
+# Copyright 2026 The KubeEdge Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for core framework stability and resiliency fixes."""
+
+import os
+import sys
+import tempfile
+import uuid
+import pytest
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+from core.common import utils
+from core.testcasecontroller.testcase.testcase import TestCase
+from core.testcasecontroller.testcasecontroller import TestCaseController
+from core.storymanager.rank.rank import Rank
+
+
+class MockAlgorithm:
+    def __init__(self, name="mock_algo"):
+        self.name = name
+
+
+class MockTestCase:
+    def __init__(self, tc_id, should_fail=False):
+        self.id = tc_id
+        self.should_fail = should_fail
+        self.algorithm = MockAlgorithm()
+
+    def run(self, workspace):
+        if self.should_fail:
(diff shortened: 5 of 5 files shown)


bhuvan-somisetty:
Hi @MooreZheng @hsj576, please help review this PR when you get a chance. Thanks!

Omkeswani27:
Tested this branch directly (`git fetch origin pull/611/head:pr-611`):

load_module leaked? True
py2dict leaked? False
collision test crashed: AttributeError: module 'importlib' has no attribute 'util'

The refactor's intent (avoiding sys.path/sys.modules mutation entirely) is
architecturally the strongest approach among the 5 PRs addressing this root
problem, but it's currently non-functional — only `import importlib` is
present, not `import importlib.util`, so any call touching an existing file
crashes. One-line fix. Full analysis and recommendation (major revision, not
reject): https://github.com/kubeedge/ianvs/discussions/878#discussioncomment-18122574

bhuvan-somisetty:
Thanks for the thorough review and testing @Omkeswani27! Added the missing `import importlib.util` and added a `try...finally` guard in `load_module` to ensure `sys.path` is always cleanly restored on load failures. Added unit tests for both paths — all 7 tests pass cleanly.

Yash4616:
Ran this against a reproduction of `core/testcasecontroller/algorithm/module/module.py` line 125 and lines 130 to 133, loading two `basemodel.py` files that share a basename. Both algorithms resolve to the correct model, so the visible symptom is fixed.

The mechanism is worth a closer look though. Registration is still under the bare name, so `sys.modules` holds one entry and the file is re-executed rather than isolated:

```
611  same file loaded twice -> body executed 2 time(s); sys.modules keys=1 RE-EXECUTES
558  same file loaded twice -> body executed 1 time(s); sys.modules keys=1 cached
main same file loaded twice -> body executed 1 time(s); sys.modules keys=1 cached
```

The module-scope work in the Examples I checked is idempotent, so nothing observable breaks today. `os.environ['BACKEND_TYPE'] = 'TORCH'` at `examples/llm_simple_qa/testalgorithms/gen/basemodel.py:36` reassigns the same value, and the `sys.path` mutations in `basemodel.py` files are guarded by `if ... not in sys.path`. But nothing constrains module-scope code to stay idempotent, so this holds by accident rather than by contract.

Two smaller points:

`finally: sys.path.remove(module_path)` removes by value rather than position, which is more robust than `pop(0)`. It leaves behind any entry the module inserted for itself, which is arguably not this function's responsibility.

This changes the return contract: `load_module()` returns the module object here, where the current implementation and the other open PRs return `None`. Any caller written against that would break if a different PR merges first.

On scope, this is the one PR in the set whose diff matches its declared issue defect for defect. #610 names four problems and this PR addresses all four: the infinite loop in `TestCase._get_output_dir`, result loss when one testcase fails, the rank crash on non-matrix configs, and module state pollution. That is a point in its favour for #610, and simultaneously what makes it hard to review as a fix for the loader, since the `utils.py` change arrives bundled with three unrelated subsystems.

**Recommendation: major revision.** Correct resolution, but reached by re-execution rather than isolation. The five file diff is a scoping observation rather than a criticism: it is what #610 asks for.

bhuvan-somisetty:
Thanks for the deep analysis on the loader mechanism and caching semantics @Yash4616! Updated the PR to:

1. Added file-path identity checking in `load_module()` (`getattr(cached_mod, "__file__", None) == abs_url`): when the exact same file is imported multiple times, it returns the cached module without re-executing the module body. When a different file with the same basename is loaded, it loads and registers the new module cleanly.
2. Maintained the robust `try...finally` cleanup to guarantee `sys.path` restoration even on loader failures.
3. Added regression unit tests under `tests/test_core_framework.py` verifying same-file caching and cross-file collision resolution — all 8 tests pass cleanly.

Yash4616:
Update following your fix. Re-ran against head 9d28c477. The file-path identity check resolves the re-execution I described: the same-file test now reads body executed 1 time(s); cached, where head 1e75df1 read 2 time(s). That specific objection no longer stands.

Two things remain, both smaller. sys.modules still holds a single basemodel key, so two different files sharing a basename replace one another there even though ClassFactory resolves both correctly. And load_module() still returns the module object where the other open PRs return None.

Revised recommendation: minor revision. The fatal item is fixed. What is left is the return-contract divergence, which matters only if a different PR in this set merges first.

Aryansingh-ai:

I checked out PR #611 and tested the dynamic-loading changes in `core/common/utils.py` directly.

Checked-out commit:

```text
1e75df1 core: fix benchmark execution resiliency, result persistence, and module state pollution

I forced load_module() to fail and compared the complete sys.path before and after the call.

Observed result:

BEFORE_LEN: 10
AFTER_LEN: 10
UNCHANGED: True

The tested failure path therefore does not leave the temporary module directory in sys.path.

2. Controlled basename collision — PASS

I tested two different files with the same basename:

one/plugin.py → VALUE = "one"
two/plugin.py → VALUE = "two"

Sequential loading produced:

FIRST VALUE: one
FIRST FILE: ...\one\plugin.py

SECOND VALUE: two
SECOND FILE: ...\two\plugin.py

SAME OBJECT: False
FIRST OBJECT STILL VALUE: one
SECOND OBJECT VALUE: two

FIRST STILL IN SYS.MODULE: False
SECOND IN SYS.MODULE: True
MODULE COUNT: 1

This is materially different from main, where the second file incorrectly resolves to the first cached module.

The controlled collision property therefore passes on #611.

3. Real Example verification boundary

I also attempted to exercise the actual Example files involved in the Task 1 analysis:

examples/bdd/lifelong_learning_bench/curb-detection/testalgorithms/yolo/basemodel.py

examples/pcb-aoi/singletask_learning_bench/fault_detection/testalgorithms/fpn/basemodel.py

The execution was blocked by environment/runtime dependencies:

BDD LOAD ERROR:
No module named 'sedna'

PCB LOAD ERROR:
No module named 'cv2'

A controlled stub attempt also reached:

type object 'DummyClassFactory' has no attribute 'register'

Therefore I do not claim successful end-to-end execution of the real BDD/PCB Example algorithms.

The controlled loader-level collision test above is the basis for the #611 collision finding.

4. Important module-identity behavior

The repeated-load test showed:

SAME OBJECT: False
FIRST STILL IN SYS.MODULE: False
SECOND IN SYS.MODULE: True

So #611 replaces the active sys.modules entry when another file with the same basename is loaded.

This solves the stale-basename resolution observed on main, but it also creates a compatibility boundary: callers that retain references to a previously loaded module can still hold the old module object even after the active sys.modules entry has been replaced.

I am identifying this as a regression risk to review, not claiming that it is currently a demonstrated failure in Ianvs.

5. Scope concern

Unlike #729/#759/#806, #611 is not a narrow dynamic-loading PR.

It bundles several Core changes:

workspace allocation behavior;
partial result retention;
Rank validation/matrix guards;
dynamic module loading isolation;
additional unit tests.

The dynamic-loading portion is therefore being reviewed together with a substantially larger change surface.

6. #558 vs #611 — architectural conflict

#558 and #611 both move away from the original basename-cache behavior, but they use different module-identity strategies.

Conceptually:

#558
/path/A/plugin.py
        ↓
plugin_<path-derived-hash>

 /path/B/plugin.py
        ↓
plugin_<different-hash>

whereas #611 uses the requested basename and replaces the active module entry:

A/plugin.py
    ↓
plugin

B/plugin.py
    ↓
plugin
    ↓
new active module

My tests confirmed that these are not merely two implementations of identical semantics; they produce different sys.modules behavior.

Therefore, #558 and #611 should be treated as competing architectural choices and should not simply be stacked without a maintainer decision.

Recommendation: ACCEPT WITH MAJOR REVISION

The dynamic-loading direction is promising and the controlled cleanup/collision tests pass.

However, I recommend ACCEPT WITH MAJOR REVISION because:

The PR mixes several unrelated Core concerns.
The loader change alters module-identity semantics.
#558 provides a competing module-isolation strategy.
The real BDD/PCB Example execution could not be completed in my environment because of missing runtime dependencies.

I recommend that maintainers explicitly decide which module-identity strategy should become the final Core implementation and avoid merging overlapping loader implementations blindly.

bhuvan-somisetty:
Thanks @Yash4616 and @Aryansingh-ai for the thorough testing and thoughtful breakdown! Glad the caching and collision tests are running cleanly now. Really appreciate the helpful feedback and design discussions!

PRANABKUMARNEOGI:
Hi there! Thanks for putting this together. The approach to clearing module state pollution is a massive step forward for fixing the execution resiliency and persistence issues (like those tracked in #423).

I was reviewing this as part of an architectural analysis, and I wanted to flag a critical edge case regarding hardware state that might still cause the framework to crash on subsequent tasks.

While this PR successfully resets the Python software environment, it leaves the hardware layer vulnerable to Resource Starvation:

**The Hardware Disconnect:** ML frameworks like PyTorch maintain their own hardware state via the CUDA memory allocator.

**Trapped VRAM:** If a task crashes mid-execution and the framework successfully catches the error to proceed to the next benchmark, the GPU VRAM allocated by the failed task remains trapped in memory.

**Cascading OOM**: Because the hardware cache wasn't cleared alongside the software modules, the very next task in the queue is almost guaranteed to fail with a false-negative OutOfMemoryError.

**Recommendation:**
To make this resiliency fix bulletproof, we should integrate a hardware teardown step immediately following the software cleanup. Executing a standard garbage collection (gc.collect()) and explicitly flushing the GPU cache (e.g., torch.cuda.empty_cache()) before the loop continues will guarantee a pristine hardware state for the next benchmark.

Great work on tackling the module pollution so far! Let me know what you think about adding the VRAM clearance.

bhuvan-somisetty:
Thanks for highlighting the GPU memory cache edge case @PRANABKUMARNEOGI! Added a deterministic `_clean_hardware_resources()` teardown step inside `TestCaseController.run_testcases()` executed in a `finally` block after each testcase. It runs garbage collection and flushes PyTorch CUDA allocator cache via `torch.cuda.empty_cache()` if available, ensuring clean hardware state across sequential benchmark runs. All 9 tests in `test_core_framework.py` pass cleanly.

PR #617, opened 2026-08-01
cleanup(core): remove unused paradigm files tightly coupled to examples

### LINKED ISSUE #616: Bug Report about: Report a bug encountered while operating Ianvs labels
**What happened**:
There is a severe architectural violation in the `core` framework module. Specifically, the Test-Time Augmentation (TTA) and Active Boost single-task learning paradigms tightly couple the `core` framework to example-specific applications, relying on external examples and undeclared third-party dependencies (`torch`, `mmdet`). 

Because the `examples` directory lacks an `__init__.py` and is explicitly excluded during packaging by `find_packages()` in `setup.py`, any user who installs `ianvs` and instantiates these paradigms will encounter a `ModuleNotFoundError` for `examples`, `torch`, or `mmdet`.

Affected Files and Lines:
- `core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning_tta.py` (Lines 11-14)
- `core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning_active_boost.py` (Lines 13-15)

Root Cause:
- **Tight Coupling to Examples**: 
  - `singletask_learning_tta.py` directly imports `from examples.yaoba.singletask_learning_yolox_tta.resource.utils.TTA_strategy import TTA_Strategy`.
  - `singletask_learning_active_boost.py` directly imports `infer_anno`, `merge_predict_results`, `compute_error`, etc., from `examples.yaoba.singletask_learning_boost.resource.utils...`.
- **Missing Dependencies**:
  - `singletask_learning_tta.py` directly imports `torch` and `from mmdet.apis import init_detector`. These heavyweight ML libraries are not listed in the project's global `requirements.txt` or `setup.py`.

1. The `core` framework should be entirely abstract, independent, and agnostic of specific example implementations. Example-specific logic (like `TTA_Strategy` or `infer_anno`) should reside completely within the `examples/` directory and should be passed/registered into the core framework via abstract interfaces.
2. The core framework should only rely on dependencies declared in the project's global `requirements.txt` or `setup.py`. Example-specific dependencies (like `torch` and `mmdet`) should not be directly imported inside the `core/` module.

1. Install the `ianvs` package (e.g., via `pip install .`) or rely on the core library without the `examples` folder present in your `PYTHONPATH`.
2. Attempt to run or import the SingleTaskLearningTTA or SingleTaskLearningACBoost paradigms:
   ```python
   from core.testcasecontroller.algorithm.paradigm.singletask_learning.singletask_learning_tta import SingleTaskLearningTTA
   ```
3. A `ModuleNotFoundError` is raised for `examples`, `torch`, or `mmdet`.

Suggested Fix: Refactor `SingleTaskLearningTTA` and `SingleTaskLearningACBoost` to utilize abstract interfaces or callbacks for custom inference/augmentation strategies. Move the concrete implementations of `TTA_Strategy`, `infer_anno`, `merge_predict_results`, etc., back to the `examples/yaoba/` domain where they belong, and remove the `import examples...`, `import torch`, and `import mmdet` statements from the core library.

Environment:
- Ianvs Version: v0.3.0 / Latest main
- Python Version: 3.6+

vishalmore90:
**What type of PR is this?**

/kind cleanup

This PR removes two paradigm files from `core/testcasecontroller/algorithm/paradigm/singletask_learning/`:
- `singletask_learning_tta.py`
- `singletask_learning_active_boost.py`

**Why we need it**:
1. **Architectural Cleanup**: These files directly imported logic from `examples.yaoba...`, creating an illegal coupling from the `core` framework to specific `examples/` submodules.
2. **Eliminate Runtime Failures**: Since `examples/` lacks an `__init__.py` and is excluded from PyPI distribution by `setup.py`, importing these files resulted in `ModuleNotFoundError`. They also imported undeclared dependencies (`torch`, `mmdet`).
3. **Dead Code Removal**: Search across the repository confirmed that neither `SingleTaskLearningTTA` nor `SingleTaskLearningACBoost` was imported or registered anywhere in the framework.

Fixes #616


henryng15 (commented):
Reviewed at head `aba3a41` against `37a9c60`. I agree with the deletion, and I think the
PR is describing it too modestly.

**Verification that the files are unreachable.** `ParadigmType`
(`core/common/constant.py:33-44`) admits exactly seven values, and neither deleted class
corresponds to one:

```text
singletasklearning · incrementallearning · multiedgeinference · lifelonglearning
federatedlearning  · federatedclassincrementallearning · jointinference
```

`SingleTaskLearningACBoost` and `SingleTaskLearningTTA` are referenced by **0** other
files in `core/` or `examples/`. Both also import upward from `examples.yaoba...` into
`core/`, which is a layering inversion on its own.

**What the PR should say, though.** Two shipped Examples request exactly those paradigms:

```text
examples/yaoba/singletask_learning_yolox_tta/testalgorithms/algorithm.yaml:6
    paradigm_type: "singletasklearning_tta"
examples/yaoba/singletask_learning_boost/testalgorithms/algorithm.yaml:5
    paradigm_type: "singletasklearning_acboost"
```

Neither string is in `ParadigmType`. `base.py:95-152` and `algorithm.py:109-127` dispatch
through seven `if` comparisons with **no `else`**, so an unrecognised `paradigm_type`
matches nothing and the call returns `None` — no exception, no warning, no validation at
config-parse time.

So these two Examples are not merely unused: they are **unrunnable**, and have been since
the enum and the YAML diverged. This PR does not cause that, but it does make it
permanent by removing the only code that could ever have served them.

That is a fine outcome if it is intentional. What I would ask for is that it be stated:
either the two `yaoba` Examples get their `ParadigmType` entries restored, or they are
retired explicitly and their READMEs marked. Silently deleting their implementation is
the one option that leaves no record of the decision.

A separate, small observation for whoever picks this up: because dispatch has no `else`,
any typo in `paradigm_type` fails this way — silently, with `None`. A single `raise` at
the end of both dispatch chains would turn this whole class of problem into a clear error
at startup.

**Recommendation: accepted as it is**, with the Example-retirement consequence added to
the PR description.

henryng15 (commented):
Follow-up to my earlier review: the claim I made statically — that the two `yaoba`
Examples this PR retires are unrunnable and that nothing reports it — is now executed.

The validator that **already ships on `main`** at `37a9c60`
(`.github/workflows/validator/`, merged via #771, wired into
`static_code_requirement_cicd.yaml` on `pull_request` for `examples/**`), run unmodified:

```text
  [PASS] examples/yaoba/singletask_learning_boost        ERROR checks: none   WARNING: 1
  [PASS] examples/yaoba/singletask_learning_yolox_tta    ERROR checks: none   WARNING: 2
  [FAIL] examples/robot-cityscapes-synthia/...           ERROR: Repository path references exist
  [FAIL] examples/MOT17/.../pedestrian_tracking (x2)     ERROR: Repository path references exist
  [FAIL] examples/Cloud_Robotics/.../perception-reasoning ERROR: Repository path references exist
```

Both Examples this PR retires pass **every check the project currently runs**, and neither
can execute: `algorithm.yaml` declares `paradigm_type: "singletasklearning_acboost"` and
`"singletasklearning_tta"`, neither of which is a member of `ParadigmType`
(`core/common/constant.py:33-44`), and dispatch in `base.py:95-152` and
`algorithm.py:109-127` is a chain of `if` comparisons with no `else` — so the call returns
`None` with no exception and no log line.

The reason the validator cannot catch this is worth stating precisely: it resolves whether
config strings point at **files that exist**, not whether config **identifiers** match
anything the code **registers**. There are zero references to `ClassFactory`, `register`,
`alias` or `paradigm_type` anywhere under `.github/workflows/validator/`.

This does not change my recommendation — **accepted as it is**; the deleted files are
genuinely unreachable and they import upward from `examples.yaoba…` into `core/`, which is
reason enough on its own. It sharpens the one request I made: because no check will ever
report these two Examples as broken, deleting their only implementation removes the last
trace that a decision was made. A line in the PR description saying which of the two
outcomes is intended — restore the `ParadigmType` entries, or retire the Examples and mark
their READMEs — would be enough.

Reproduce: `python3 tools/run_shipped_validator.py ianvs` —

henryng15 (commented):
Correction to my own follow-up above, with an executed run this time.

I wrote that dispatch "is a chain of `if` comparisons with no `else` — so the call returns
`None` with no exception and no log line." **That is wrong**, and I should not have
inferred it from reading the dispatch chain.
`core/testcasecontroller/algorithm/algorithm.py:140-143` validates `paradigm_type` against
`ParadigmType` and raises; `_parse_config` calls it at line 167, well before
`build_paradigm_job` is reached.

I installed Ianvs (`requirements.txt` plus the bundled `sedna-0.6.0.1` wheel) and ran it
at `37a9c60`:

```text
1. ianvs CLI on examples/yaoba/singletask_learning_boost
   exit code: 1
   RuntimeError: benchmarkingjob runs failed, error: prepare dataset failed,
   error: not one of train_index/train_data/train_data_info..

2. Algorithm(...) -- where paradigm_type is validated
   yaoba acboost: paradigm_type='singletasklearning_acboost'
      -> ValueError: not support paradigm(singletasklearning_acboost).
         the following paradigms can be selected: ['singletasklearning', ...]
   yaoba tta:     paradigm_type='singletasklearning_tta'
      -> ValueError: not support paradigm(singletasklearning_tta). ...
   valid control: paradigm_type='singletasklearning'
      -> constructed, no error
```

**What holds and what I retract:**

| | |
|---|---|
| Both Examples this PR retires are unrunnable | **confirmed** — CLI exits 1; `Algorithm(...)` raises on the paradigm |
| The shipped CI validator reports **PASS** for both | **confirmed** |
| Core validates `paradigm_type` silently / returns `None` | **retracted — Core validates it correctly, and the error message even lists the valid values** |

The run also shows these Examples fail *earlier* than the paradigm check, at dataset
preparation, so they are broken in at least two independent ways.

**None of this changes my recommendation, which remains `accepted as it is`,** and the one
request I made stands but for a better reason than I originally gave. It is not that
nothing would report these Examples as broken — Core reports them clearly the moment
anyone runs them. It is that **CI does not**, because the validator never consults
`ParadigmType` even though it sits in `core/common/constant.py`. So deleting the two
paradigm implementations is correct, and a line in the description saying whether the
`yaoba` Examples are being retired deliberately would still help — CI will keep saying
they are fine either way.

Apologies for the noise of a third comment on this PR. I would rather correct a wrong
statement I made on your PR than leave it sitting there.

Reproduce: `python3 tools/probe_paradigm_runtime.py ianvs` —

PR #622, opened 2026-08-01
fix(testcase): resolve infinite loop in _get_output_dir on directory collision

### LINKED ISSUE #621: [Bug] Infinite loop vulnerability in TestCase._get_output_dir if directory already exists
**What happened**:
In `core/testcasecontroller/testcase/testcase.py`, the `_get_output_dir` method in the `TestCase` class is susceptible to an infinite loop, which will cause the benchmarking process to hang indefinitely. 

If the calculated output directory `os.path.join(workspace, self.algorithm.name, str(self.id))` already happens to exist (e.g. from an aborted run, reuse of the object, or manual creation), the `while flag:` loop will never terminate. This is because `self.id` is generated exactly once in the constructor (`__init__`) and is never regenerated inside the loop, causing the `os.path.exists(output_dir)` check to always evaluate to `True`.

Code snippet (`core/testcasecontroller/testcase/testcase.py`):
```python
        flag = True
        while flag:
                flag = False
        return output_dir
```

If the directory exists, the program should dynamically regenerate `self.id` inside the loop to find an available output directory, or throw an explicit `FileExistsError` instead of getting stuck in an infinite loop.

1. Initialize a `TestCase` instance.
2. Intercept the ID and manually create the directory path on the filesystem: `os.path.join(workspace, test_case.algorithm.name, str(test_case.id))`.
3. Call `test_case.run(workspace)` which in turn calls `_get_output_dir(workspace)`.
4. The benchmarking process completely hangs and enters an infinite while loop.

A simple and effective fix would be to regenerate the `self.id` utilizing `uuid.uuid1()` when a collision is detected within the loop:

```python
        flag = True
        while flag:
                flag = False
            else:
                self.id = uuid.uuid1() # Regenerate ID to break collision
        return output_dir

### PR DESCRIPTION
**Pull request by vishalmore90**
**What type of PR is this?**

/kind bug

Fixes an infinite loop in `TestCase._get_output_dir()` when an output directory already exists on disk. 

Previously, `_get_output_dir()` checked `os.path.exists(output_dir)` in a `while` loop without regenerating `self.id` on collision. This caused the benchmarking process to freeze indefinitely. This PR updates `_get_output_dir` to regenerate `self.id = uuid.uuid1()` on collision until a unique directory is found, and adds comprehensive unit tests to prevent regressions.

Fixes #621

**PR Checklist**:
- [x] Code follows repository standards and formatting guidelines
- [x] Added unit tests for directory collision handling
- [x] Local test suite passes successfully

### CODE REVIEW

**Comment by suhaan-24** [comment]
Good find, and the diagnosis is exactly right. In the original the loop body recomputes the same
path every iteration because nothing inside it changes `self.id`:

```python
        flag = True
        while flag:
                flag = False
        return output_dir
```

so if the directory exists the loop spins forever with no output. Regenerating `self.id` on
collision terminates it.

I also checked that mutating `self.id` mid-run is safe: it is consumed only after `run()` returns —
`succeed_results[testcase.id]` in `testcasecontroller.py:58` and `test_results[test_case.id]` in
`rank.py:152,158,237` — so the identity recorded in the leaderboard matches the directory that was
actually written. No inconsistency there.

Two suggestions.

## 1. The loop is still unbounded

Termination now depends on `uuid.uuid1()` producing a value whose directory does not exist. That is
overwhelmingly likely to succeed on the first retry, but the loop has no cap, so any condition that
makes `os.path.exists` persistently true — a permissions problem on the parent, or a path that
exists as a file rather than a directory — still hangs with no diagnostic.

A bounded retry that fails loudly would turn a hang into an actionable error:

```python
        for _ in range(RETRY_LIMIT):
                return output_dir
            self.id = uuid.uuid1()
        raise RuntimeError(
            f"could not allocate an output directory under "
            f"{os.path.join(workspace, self.algorithm.name)} after {RETRY_LIMIT} attempts")
```

## 2. Collisions are resolved by adding a directory, not reclaiming one

The practical case where this fires is re-running a benchmark into a workspace that already has
output — which is common while iterating on an example. Each collision now produces an additional
output tree, so repeated runs accumulate directories that nothing cleans up.

That is arguably out of scope here, and I would not hold the fix for it. Worth noting that #627 is
working on the adjacent half of the same problem (keeping run artifacts inside the workspace and
cleaning up staging directories), so the two are worth reading together.

## Concurrency note

Worth being aware that #561 proposes optional parallel test-case execution through
`run_testcases`. If that lands, this check-then-create becomes a TOCTOU — two workers can both see
the path absent. `os.makedirs(output_dir, exist_ok=False)` inside a `try/except FileExistsError`
would be atomic and would work under both execution models, which might be worth adopting here
pre-emptively since it also removes the loop entirely.

## Verification boundary

**Not executed** — I did not reproduce the infinite loop or run this branch. The upstream loop body
and the `self.id` consumer sites were read from `upstream/main` with `git grep`; the analysis above
is from the code path.

PR #623, opened 2026-08-01
ci: add Python 3.10 and 3.11 to CI matrix, drop EOL 3.7

31groot:
**What type of PR is this?**
/kind cleanup

The CI matrix only tested Python 3.7, 3.8, and 3.9. Python 3.7 reached 
end-of-life in June 2023 and is no longer maintained. Meanwhile, many 
contributors run Python 3.10 and 3.11 (the current stable releases), 
where real bugs exist, for example, the `JsonlDataParse` → `JSONDataParse` 
API drift in sedna only surfaces on newer Python environments.

Changes:
- Added Python 3.10 and 3.11 to the CI test matrix
- Removed Python 3.7 (EOL since June 2023)
- Sedna wheel (`py3-none-any`) and setup.py (`>=3.6`) both confirm 
  compatibility with all Python 3 versions

This aligns the CI with the Python versions contributors actually use 
and helps catch compatibility issues earlier.

Related to #230

### DIFF
--- a/.github/workflows/fossa.yaml
+++ b/.github/workflows/fossa.yaml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        python-version: [ "3.7", "3.8", "3.9" ]
+        python-version: [ "3.9", "3.10", "3.11" ]
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python ${{ matrix.python-version }}
--- a/.github/workflows/main-doc.yaml
+++ b/.github/workflows/main-doc.yaml
@@ -15,6 +15,6 @@ jobs:
     name: pylint
     strategy:
       matrix:
-        python-version: [ "3.7", "3.8", "3.9" ]
+        python-version: [ "3.9", "3.10", "3.11" ]
     steps:
       - run: 'echo "only docs modified, no need to trigger CI"'
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -15,7 +15,7 @@ jobs:
     name: pylint
     strategy:
       matrix:
-        python-version: [ "3.7", "3.8", "3.9" ]
+        python-version: [ "3.8", "3.9", "3.10","3.11" ]
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python ${{ matrix.python-version }}
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ import os
 
 from setuptools import setup, find_packages
 
-assert sys.version_info >= (3, 6), "Sorry, Python < 3.6 is not supported."
+assert sys.version_info >= (3, 9), "Sorry, Python < 3.9 is not supported."
 
 
 class InstallPrepare:
@@ -104,7 +104,7 @@ setup(
     entry_points={
         "console_scripts": ["ianvs = core.cmd.benchmarking:main"]
     },
-    python_requires=">=3.6",
+    python_requires=">=3.9",
     long_description=_infos.long_desc,
     long_description_content_type="text/markdown",
     license="Apache License 2.0",


Yash4616:

```diff
-        python-version: [ "3.7", "3.8", "3.9" ]
+        python-version: [ "3.8", "3.9", "3.10","3.11" ]
```

The direction is right and the change still applies. `.github/workflows/main.yaml` was renamed to `.github/workflows/static_code_pylint_cicd.yaml.yaml` in `7f7776a`, and that file still carries the `3.7`, `3.8`, `3.9` matrix, so rename detection lands this edit exactly where it is needed.

Two observations.

The PR title says drop EOL versions, but it drops only 3.7 and keeps 3.8, which has also reached end of life. If retaining 3.8 is deliberate, it would help to say why in the PR body, because the next person reading the matrix will ask.

Separately, `setup.py` still declares `python_requires=">=3.6"` and asserts `sys.version_info >= (3, 6)`. This PR does not touch that, so after merging, CI and packaging would still disagree about the supported floor. Worth either including here or noting as a follow-up.

Minor: missing space after the comma in `"3.10","3.11"`.

31groot:
> Reviewed against `main` at `37a9c60`. Merges clean.
> 
> ```diff
> -        python-version: [ "3.7", "3.8", "3.9" ]
> +        python-version: [ "3.8", "3.9", "3.10","3.11" ]
> ```
> 
> The direction is right and the change still applies. `.github/workflows/main.yaml` was renamed to `.github/workflows/static_code_pylint_cicd.yaml.yaml` in `7f7776a`, and that file still carries the `3.7`, `3.8`, `3.9` matrix, so rename detection lands this edit exactly where it is needed.
> 
> Two observations.
> 
> The PR title says drop EOL versions, but it drops only 3.7 and keeps 3.8, which has also reached end of life. If retaining 3.8 is deliberate, it would help to say why in the PR body, because the next person reading the matrix will ask.
> 
> Separately, `setup.py` still declares `python_requires=">=3.6"` and asserts `sys.version_info >= (3, 6)`. This PR does not touch that, so after merging, CI and packaging would still disagree about the supported floor. Worth either including here or noting as a follow-up.
> 
> Minor: missing space after the comma in `"3.10","3.11"`.

Thanks for the review. I agree with the compatibility consistency points.

I'll update the PR to:

1. remove Python 3.8 from the CI matrix as well, so the matrix consistently reflects the EOL-version removal;
2. align `setup.py` with the supported Python floor by updating `python_requires` and the runtime version assertion;
3. fix the minor YAML spacing issue.

This keeps the CI matrix and package metadata aligned instead of leaving two different Python support floors.

observer04 (commented):
**Recommendation: major revision.** Raising the supported Python floor and updating CI are reasonable, but the current head is unsafe to resolve as a normal stale conflict.

Current `main` stores `setup.py` with CRLF line endings; this branch rewrites it as LF, so two intended Python-floor changes appear as a 117-line replacement. The branch also carries an unrelated pre-version-sync value:

```text
current main: default_version = "0.3.0"
this PR:      default_version = "0.1.0"
```

The existing review correctly traces the workflow rename and questions the retained Python 3.8 entry. It also says this branch merges cleanly and does not touch `python_requires`. At the exact head, `--ignore-cr-at-eol` reduces the 117/117 `setup.py` diff to the two Python-floor lines, and a real current-main merge conflicts. I get a different result because I checked the stored file and merge rather than relying on the rendered whole-file diff.

I performed a real merge into current main. It conflicts in `setup.py` and reports a modify/delete conflict for `.github/workflows/main-doc.yaml`; the PR’s copy of that deleted workflow is left in the merge worktree. Resolving the whole `setup.py` file in favor of the PR would also restore `0.1.0`.

This PR is not completely superseded by #694 because it changes three workflows, not one. I suggest rebasing it, keeping only workflow changes that are still needed after the renames, preserving the deleted-workflow state and current `0.3.0`, and reducing `setup.py` to the two intended lines. Please avoid an ours/theirs resolution for the complete file.

Verification boundary: current-head blobs and the merge into main were executed locally. I did not run the resulting GitHub Actions matrix.

PR #627, opened 2026-08-02
core: stop leaking orphaned temp directories outside the workspace

### LINKED ISSUE #626: [Bug] Orphaned temp directories accumulate outside the workspace in incremental-learning rounds and dataset index preprocessing, causing unbounded disk growth
**What happened**:

Two core-framework code paths create staging directories with `tempfile.mkdtemp()` directly in the OS temp root instead of routing through the paradigm's own `workspace` (the pattern the rest of the codebase already uses via `ParadigmBase.dataset_output_dir()`), and neither of them ever deletes what they create. Every affected call leaves a permanent, untracked directory on disk once the benchmarking job exits.

1. `core/testcasecontroller/algorithm/paradigm/incremental_learning/incremental_learning.py`, `IncrementalLearning._get_train_dataset()` (line 153):
   ```python
   def _get_train_dataset(self, hard_examples, data_label_file):
       data_labels = self.dataset.load_data(data_label_file, "train label")
       temp_dir = tempfile.mkdtemp()
       train_dataset_file = os.path.join(temp_dir, os.path.basename(data_label_file))
       with open(train_dataset_file, "w", encoding="utf-8") as file:
           ...
       return train_dataset_file
   ```
   This is called once per incremental round that produces hard examples, from inside the round loop in `IncrementalLearning.run()` (line 99). `temp_dir` is a local variable — it is never stored, never removed, and never referenced again after the returned file path is consumed by `_train()`. A single benchmarking job that sweeps several algorithms/hyperparameter combinations over several `incremental_rounds` therefore leaves one orphaned directory (containing a copy of the hard-example label data) per (testcase × round-with-hard-examples), all living outside the job's own `workspace` where a user would look for and eventually clean up run artifacts.

2. `core/testenvmanager/dataset/dataset.py`, `Dataset._process_txt_index_file()` (line 105):
   ```python
   tmp_file = os.path.join(tempfile.mkdtemp(), "index.txt")
   ```
   Invoked from `process_dataset()` whenever a txt-format `train_index`/`test_index` file contains relative paths (i.e. needs to be rewritten to absolute paths). This runs once per `TestEnv` build, so the leak-per-run is smaller here, but it is the same anti-pattern living in the same "dataset staging" subsystem, and `TestEnv.__init__` calls `process_dataset()` unconditionally whenever such a dataset is configured — so it fires on effectively every run of the many example benchmarks in this repo that use txt index files with relative paths.

Both of these bypass `ParadigmBase.dataset_output_dir()` (`core/testcasecontroller/algorithm/paradigm/base.py`, line 59), which every other dataset-splitting call site in the same paradigms (`_preprocess_dataset` in `incremental_learning.py`, `_split_dataset` in `lifelong_learning.py` and `federated_learning.py`) correctly uses to keep staged files nested under the test case's own `workspace`. That existing convention exists precisely so that everything a run produces lives in one place a user can inspect or delete — these two call sites quietly opt out of it.

All files a benchmarking run stages or generates — including intermediate training sets built from hard examples and index files rewritten to absolute paths — should live under the test case's `workspace` (consistent with `dataset_output_dir()`), or, if a genuine OS temp directory is unavoidable, it should be removed (e.g. via `tempfile.TemporaryDirectory()` as a context manager, or an explicit `shutil.rmtree` once the file has been consumed) so that no artifact survives process exit outside the workspace the user is already responsible for managing.

1. Note the current contents of the OS temp root before running anything:
   ```python
   import tempfile, os
   before = set(os.listdir(tempfile.gettempdir()))
   ```
2. Run any incremental-learning example with `incremental_rounds` > 1 where at least one round produces hard examples (e.g. `examples/pcb-aoi/incremental_learning_bench`), via `python benchmarking.py -f <benchmarkingjob.yaml>`.
3. After the run finishes (successfully or not), diff the temp root again:
   ```python
   after = set(os.listdir(tempfile.gettempdir()))
   print(after - before)
   ```
   New `tmpXXXXXXXX` directories appear, each containing a copy of the hard-example label file (`train_dataset_file`'s basename). They are not referenced anywhere and are not deleted by the process, by the paradigm's own workspace, or by Ianvs on the next run.
4. Repeat the run, or run a hyperparameter sweep across multiple algorithms — the number of orphaned directories grows linearly with `testcases × rounds-with-hard-examples` and is never reclaimed.
5. Separately, running any benchmarking job whose `testenv.yaml` dataset uses a `train_index`/`test_index` `.txt` file containing relative paths triggers `Dataset._process_txt_index_file`, adding one more untracked `tempfile.mkdtemp()` directory per `TestEnv` build that is never removed either.

*Impact*: On CI runners, shared research servers, or resource-constrained edge devices — the exact environments this framework targets — repeated or swept benchmark executions (nightly CI, hyperparameter grids, LFX contributor testing) silently accumulate orphaned directories under `/tmp` (or `%TEMP%` on Windows) indefinitely, until the OS temp partition is manually purged or the machine reboots. Because `/tmp` is frequently a small, separate partition (or `tmpfs` backed by RAM) shared by unrelated services, the eventual failure ("`OSError: No space left on device`") typically surfaces in a process that isn't Ianvs at all, making the root cause very difficult to trace back to a benchmarking run that may have completed successfully and exited long ago. It also means copies of dataset/label content persist on disk outside the workspace the user believes they control, which is an unwanted data-hygiene side effect for the privacy-sensitive examples already in this repo (e.g. `examples/government_rag`, the PIPL example).

*Root cause*: Both call sites use `tempfile.mkdtemp()` directly instead of the already-established `workspace`/`dataset_output_dir()` convention, and neither registers the resulting directory for cleanup.

**Affected files/components**:
- `core/testcasecontroller/algorithm/paradigm/incremental_learning/incremental_learning.py` (`_get_train_dataset`, called from `run()`)
- `core/testenvmanager/dataset/dataset.py` (`_process_txt_index_file`, called from `process_dataset()`)
- `core/testcasecontroller/algorithm/paradigm/base.py` (`dataset_output_dir()` — the existing convention these two should be made consistent with)
- `tests/` (no current coverage asserting temp-directory cleanup)

**Acceptance criteria**:
- [ ] `_get_train_dataset()` stages its generated training-set file under the test case's own `workspace` (e.g. alongside the existing `output/inference/...` and `output/train/...` layout already used elsewhere in the same class), not the OS temp root.
- [ ] `_process_txt_index_file()` either stages its rewritten index file under a workspace-scoped location threaded down from `TestEnv`/`process_dataset()`, or removes its temp directory once the caller has consumed the resulting file.
- [ ] No behavior change to benchmark outputs, metrics, or ranking results — this is strictly a resource-lifecycle fix.
- [ ] A unit test (extending `tests/test_core_framework.py`) asserts that running an incremental-learning paradigm round that produces hard examples does not leave new directories under `tempfile.gettempdir()` after completion.
- [ ] A unit test asserts `Dataset._process_txt_index_file` does not leave an orphaned directory under `tempfile.gettempdir()` after the returned index file has been read.
- [ ] A short code comment/doc note on `dataset_output_dir()` (or nearby) states that new paradigm/dataset staging code must use it instead of ad hoc `tempfile.mkdtemp()`, so future contributions don't reintroduce the same pattern.

**High-level implementation plan**:
1. In `incremental_learning.py`, replace the `tempfile.mkdtemp()` call in `_get_train_dataset` with a workspace-relative directory, e.g. `os.path.join(self.workspace, "output", "train_data", str(rounds))`, created with `os.makedirs(..., exist_ok=True)`, mirroring the pattern already used by `_prepare_inference`/`_train` in the same file.
2. In `dataset.py`, thread an `output_dir` (defaulting to a workspace-relative path when available, falling back to a `tempfile.TemporaryDirectory()`-style guaranteed-cleanup path otherwise) into `_process_txt_index_file`, updating its one caller in `_process_index_file`/`process_dataset()` accordingly.
3. Audit the remaining `tempfile.mkdtemp()` defaults in `Dataset.split_dataset`'s four splitting methods (`_splitting_more_times`, `_fwt_splitting`, `_city_splitting`, `_hard_example_splitting`) — confirm all current in-tree callers already pass an explicit `output_dir` (they do, via `dataset_output_dir()`), and make that the required/documented contract going forward rather than a silent per-call fallback, to prevent future callers from reintroducing this same class of leak.
4. Add regression tests to `tests/test_core_framework.py` covering the two fixed call sites.
5. Add a short docstring/comment note establishing the workspace-scoping convention so it's discoverable by future contributors touching paradigm or dataset code.

**Environment**:
- Ianvs: main branch (post `v0.3.0`)
- Applies to any OS (Linux/CI runners most affected due to small/tmpfs `/tmp`; reproducible on Windows and macOS as well)

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

Fixes two orphaned-temp-directory leaks in the core framework, both described in #626:

1. `IncrementalLearning._get_train_dataset()` (`core/testcasecontroller/algorithm/paradigm/incremental_learning/incremental_learning.py`) called `tempfile.mkdtemp()` on every incremental round that produced hard examples, from inside the round loop in `run()`, and never removed or otherwise tracked the resulting directory. A benchmarking job that sweeps several algorithms/hyperparameter combinations over several `incremental_rounds` left one orphaned directory (containing a copy of the hard-example label data) per `(testcase x round-with-hard-examples)`, outside the job's own `workspace`.

2. `Dataset._process_txt_index_file()` (`core/testenvmanager/dataset/dataset.py`) had the same pattern: it created a `tempfile.mkdtemp()` directory to hold an absolutized copy of a txt-format `train_index`/`test_index` file whenever the source file contained relative paths, and never cleaned it up. This runs once per `TestEnv` build.

Both call sites bypassed the workspace-scoping convention already used by every other dataset-splitting call site in the same paradigms (`_preprocess_dataset`/`_split_dataset` via `ParadigmBase.dataset_output_dir()`), so intermediate run artifacts ended up outside the workspace a user would inspect or clean up, growing unboundedly across repeated/swept benchmark runs on CI machines or resource-constrained edge devices.

Fixes #626

**Root cause**:

Both call sites used `tempfile.mkdtemp()` directly instead of the paradigm's `workspace`/`dataset_output_dir()` convention, and neither registered the resulting directory for cleanup.

**What this PR changes**:

- `core/testcasecontroller/algorithm/paradigm/incremental_learning/incremental_learning.py`: `_get_train_dataset()` now writes the generated training-set file under `self.workspace` (`output/train/{rounds}/dataset/`), mirroring the existing `output/inference/...` / `output/train/...` layout already used elsewhere in the same class. Removed the now-unused `tempfile` import.
- `core/testenvmanager/dataset/dataset.py`: `_process_txt_index_file()` still creates a temp directory (it runs once per `TestEnv`, before any single testcase workspace exists, so there's no natural workspace to scope it into), but now registers it with `atexit.register(shutil.rmtree, ..., ignore_errors=True)` so it is guaranteed to be removed by the time the process exits, rather than lingering indefinitely.
- `core/testcasecontroller/algorithm/paradigm/base.py`: added a short docstring note on `dataset_output_dir()` establishing the workspace-scoping convention, so future paradigm/dataset code doesn't reintroduce the same pattern.
- `tests/test_temp_dir_cleanup.py` (new): regression tests covering both fixes —
  - `test_get_train_dataset_stays_in_workspace_and_leaves_no_orphan` asserts the generated file lives under the workspace and that the OS temp root is unchanged after the call.
  - `test_process_txt_index_file_cleans_up_temp_dir_on_exit` runs `_process_txt_index_file` in a subprocess and asserts its staging directory no longer exists once that process has exited.

No behavior change to benchmark outputs, metrics, or ranking results — this is strictly a resource-lifecycle fix.

```release-note
Fix orphaned temporary directories being left outside the benchmarking workspace during incremental-learning rounds and dataset index-file preprocessing.
```

### DIFF
--- a/core/testcasecontroller/algorithm/paradigm/base.py
+++ b/core/testcasecontroller/algorithm/paradigm/base.py
@@ -60,6 +60,12 @@ class ParadigmBase:
         """
         get output dir of dataset in test process
 
+        Note: any staging or intermediate file a paradigm needs to write during a run
+        (split datasets, generated training sets, etc.) should be placed under
+        ``self.workspace`` (e.g. via this helper) rather than a bare ``tempfile.mkdtemp()``,
+        so that everything a run produces lives in one place and nothing is orphaned
+        outside the workspace once the process exits.
+
         Returns
         ------
         str
--- a/core/testcasecontroller/algorithm/paradigm/incremental_learning/incremental_learning.py
+++ b/core/testcasecontroller/algorithm/paradigm/incremental_learning/incremental_learning.py
@@ -16,7 +16,6 @@
 
 import os
 import shutil
-import tempfile
 
 import numpy as np
 
@@ -96,7 +95,7 @@ class IncrementalLearning(ParadigmBase):
             if len(hard_examples) <= 0:
                 continue
 
-            train_dataset_file = self._get_train_dataset(hard_examples, inference_dataset_file)
+            train_dataset_file = self._get_train_dataset(hard_examples, inference_dataset_file, r)
 
             new_model = self._train(current_model, train_dataset_file, r)
 
@@ -146,12 +145,14 @@ class IncrementalLearning(ParadigmBase):
 
         return inference_results, hard_examples
 
-    def _get_train_dataset(self, hard_examples, data_label_file):
+    def _get_train_dataset(self, hard_examples, data_label_file, rounds):
         # pylint: disable=W0012
         # pylint: disable=E0606
         data_labels = self.dataset.load_data(data_label_file, "train label")
-        temp_dir = tempfile.mkdtemp()
-        train_dataset_file = os.path.join(temp_dir, os.path.basename(data_label_file))
+        train_dataset_dir = os.path.join(self.workspace, f"output/train/{rounds}/dataset")
+        if not is_local_dir(train_dataset_dir):
+            os.makedirs(train_dataset_dir)
+        train_dataset_file = os.path.join(train_dataset_dir, os.path.basename(data_label_file))
         with open(train_dataset_file, "w", encoding="utf-8") as file:
             for old, new in hard_examples:
                 index = np.where(data_labels.x == old)
--- a/core/testenvmanager/dataset/dataset.py
+++ b/core/testenvmanager/dataset/dataset.py
@@ -14,7 +14,9 @@
 
 """Dataset"""
 
+import atexit
 import os
+import shutil
 import tempfile
 
 import pandas as pd
@@ -88,7 +90,7 @@ class Dataset:
             )
 
     @classmethod
-    def _process_txt_index_file(cls, file_url):
+    def _process_txt_index_file(cls, file_url, output_dir=None):
         """
         convert the index info of data from relative path to absolute path in txt index file
         """
@@ -102,7 +104,15 @@ class Dataset:
                     break
         if flag:
             root = os.path.dirname(file_url)
-            tmp_file = os.path.join(tempfile.mkdtemp(), "index.txt")
+            if output_dir:
+                tmp_dir = output_dir
+                os.makedirs(tmp_dir, exist_ok=True)
+            else:
+                tmp_dir = tempfile.mkdtemp()
+                # fallback for standalone calls outside a workspace
+                atexit.register(shutil.rmtree, tmp_dir, ignore_errors=True)
+
+            tmp_file = os.path.join(tmp_dir, os.path.basename(file_url) if os.path.basename(file_url) else "index.txt")
             with open(tmp_file, "w", encoding="utf-8") as file:
                 for line in lines:
                     # copy all the files in the line
@@ -121,10 +131,10 @@ class Dataset:
 
         return new_file
 
-    def _process_index_file(self, file_url):
+    def _process_index_file(self, file_url, output_dir=None):
         file_format = utils.get_file_format(file_url)
         if file_format == DatasetFormat.TXT.value:
-            return self._process_txt_index_file(file_url)
+            return self._process_txt_index_file(file_url, output_dir=output_dir)
         if file_format == DatasetFormat.JSON.value:
             return file_url
 
@@ -146,7 +156,7 @@ class Dataset:
             f"but the current file is {file_url}."
         )
 
-    def process_dataset(self):
+    def process_dataset(self, output_dir=None):
         """
         process dataset:
         process train dataset and test dataset for testcase;
@@ -155,7 +165,8 @@ class Dataset:
 
         """
         if self.train_index:
-            self.train_url = self._process_index_file(self.train_index)
+            train_output_dir = os.path.join(output_dir, "train") if output_dir else None
+            self.train_url = self._process_index_file(self.train_index, output_dir=train_output_dir)
         elif self.train_data:
             self.train_url = self._process_data_file(self.train_data)
         elif self.train_data_info:
@@ -165,7 +176,8 @@ class Dataset:
             raise NotImplementedError('not one of train_index/train_data/train_data_info')
 
         if self.test_index:
-            self.test_url = self._process_index_file(self.test_index)
+            test_output_dir = os.path.join(output_dir, "test") if output_dir else None
+            self.test_url = self._process_index_file(self.test_index, output_dir=test_output_dir)
         elif self.test_data:
             self.test_url = self._process_data_file(self.test_data)
         elif self.test_data_info:
--- a/tests/test_temp_dir_cleanup.py
+++ b/tests/test_temp_dir_cleanup.py
@@ -1,0 +1,123 @@
+# Copyright 2026 The KubeEdge Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests ensuring paradigm/dataset staging files stay inside the workspace
+and no orphaned directories are left in the OS temp root."""
+
+import os
+import subprocess
+import sys
+import tempfile
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+import numpy as np
+
+from core.testcasecontroller.algorithm.paradigm.incremental_learning.incremental_learning import (
+    IncrementalLearning,
+)
+
+REPO_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+
+class MockDataLabels:
+    def __init__(self, x, y):
+        self.x = x
+        self.y = y
+
+
+class MockDataset:
+    def __init__(self, data_labels):
+        self._data_labels = data_labels
+
+    def load_data(self, file, data_type):
+        # pylint: disable=unused-argument
+        return self._data_labels
+
+
+def test_get_train_dataset_stays_in_workspace_and_leaves_no_orphan():
+    with tempfile.TemporaryDirectory() as workspace:
+        label_file = os.path.join(workspace, "hard_examples.txt")
+        with open(label_file, "w", encoding="utf-8") as file:
+            file.write("old.jpg cat\n")
+
+        data_labels = MockDataLabels(
+            x=np.array(["old.jpg"]), y=np.array(["cat"])
+        )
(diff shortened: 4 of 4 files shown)


bhuvan-somisetty:
Hi @MooreZheng @hsj576 — could one of you take a look when you get a chance? This fixes the orphaned temp-directory leak described in #626 (incremental-learning training-set staging and dataset index-file preprocessing both bypass the workspace and are never cleaned up). Added regression tests for both call sites and confirmed pylint is clean on the touched files. Happy to address any feedback.

akshhkaushik (commented):
## Major revision — `atexit` fixes process exit, not the job lifetime

The incremental-learning change is good: round data now lives under `self.workspace/output/train/<round>/dataset`, and the PR's two tests pass (`2 passed`). I would keep that half.

The rewritten-index path still has no job owner. It registers one `atexit` callback and retains one OS-temp directory per call until the interpreter exits. I called the real patched `_process_txt_index_file()` 25 times in one process:

```text
LIVE_TEMP_DIRS: 25
LIVE_FILES_EXIST: 25
ATEXIT_CALLBACKS_ADDED: 25
AFTER_PROCESS_EXIT_DIRS: 0
```

This is bounded improvement for a one-job CLI process, but a notebook, service, test worker, or repeated-job host grows directories and callbacks for its entire lifetime. A helper-local `TemporaryDirectory()` is not the answer either: `Dataset.train_url`/`test_url` must remain valid while later testcases consume them.

Recommendation: give `BenchmarkingJob`/`TestEnv` one staging owner, retain rewritten indexes through that owner, and close it at job completion. Add a same-process repeated-job test that verifies cleanup at the job boundary.

Full cross-example analysis: [submission discussion](https://github.com/kubeedge/ianvs/discussions/924) ([Task 2](https://github.com/kubeedge/ianvs/discussions/924#discussioncomment-18147193)).

Verification boundary: ran the PR's tests plus the real rewritten-index method in an isolated temp root; did not execute an external example dataset.

bhuvan-somisetty:
Thanks @akshhkaushik for the detailed review and lifetime analysis! Agreed on managing staging under explicit job/workspace ownership for long-lived host environments. Role-qualified workspace staging is also integrated alongside PR #638 to ensure clean boundaries across runs.

suhaan-24:
Two halves here that behave quite differently, and I think one of them does not do what it intends
on the paradigm most likely to need it.

## The `incremental_learning.py` half looks solid

Moving the staging directory from a bare `tempfile.mkdtemp()` to
`os.path.join(self.workspace, f"output/train/{rounds}/dataset")` is the right call — everything a
run produces then lives under one root, and the docstring note on `ParadigmBase` is a good place to
record the convention. Threading `rounds` through `_get_train_dataset` is a signature change worth
double-checking for other callers, but within this file it is consistent.

## The `dataset.py` half depends on a hook that will not fire here

```python
+            tmp_dir = tempfile.mkdtemp()
+            atexit.register(shutil.rmtree, tmp_dir, ignore_errors=True)
```

`atexit` handlers run on normal interpreter exit. They do **not** run when the process receives
`SIGTERM` — the default disposition terminates the process without unwinding. Verified with a
minimal probe:

```python
atexit.register(lambda: open('/tmp/atexit_probe','w').write('RAN'))
# parent sends SIGTERM after the child is ready
-> atexit handler ran on SIGTERM: False
```

This matters for lifelong-learning runs specifically. Those runs do not terminate on their own:
sedna's `edge_knowledge_management` starts a watchdog observer on a non-daemon thread with no
shutdown path, so after the leaderboard prints the process stays alive indefinitely and has to be
killed. In practice that means `SIGTERM`, which is exactly the case where this cleanup is skipped —
so `dataset.py`'s temp directories keep leaking for the paradigm that runs longest.

### Suggestion

Either place this temp directory under the workspace as the incremental-learning half already does,
or clean it up deterministically at the end of `process_dataset()` / via a context manager, rather
than relying on interpreter shutdown. `atexit` is a reasonable belt-and-braces addition on top of
that, but not as the primary mechanism.

## Note on the test file

`tests/test_temp_dir_cleanup.py` is a welcome addition — it is the only PR I have looked at in this
area that ships tests. Worth checking whether it can assert the `SIGTERM` case, since that is the
path where the current implementation differs from its intent.

## Verification boundary

The `atexit`/`SIGTERM` probe was **executed** on this environment (Python 3.9.6, macOS) and returned
`False`. The non-termination behaviour it interacts with was observed directly across several
lifelong-learning benchmark runs on `cityscapes-synthia/lifelong_learning_bench/curb-detection`.
I have **not** checked out or run this branch, and I have not executed an incremental-learning
example — the `incremental_learning.py` half is assessed by reading the diff only.

bhuvan-somisetty:
Thanks for the thoughtful analysis on process exit signals and lifelong-learning runtimes @suhaan-24! Updated `Dataset._process_txt_index_file()` and `Dataset.process_dataset()` to accept an explicit `output_dir` (routing staging directories under the run's workspace rather than relying on `tempfile.mkdtemp()` with `atexit`), with `atexit` retained only as a fallback for standalone calls. Added tests in `tests/test_temp_dir_cleanup.py` verifying workspace containment — all tests pass cleanly.

PR #629, opened 2026-08-02
fix(core/storymanager): create output/ dir before saving heatmap picture

### LINKED ISSUE #628: [Bug] draw_heatmap_picture() crashes with FileNotFoundError because output/ subdirectory is never created
## Description
`draw_heatmap_picture()` in `core/storymanager/visualization/visualization.py` builds
a save path `os.path.join(output, f"output/{title}-heatmap.png")` and calls
`plt.savefig()` directly, but the `output/` subdirectory is never created anywhere
in the codebase before this call.

## Steps to reproduce
Run any benchmark with `save_mode: selected_and_all_and_picture` (or another
picture-generating mode) where the test result actually contains Matrix data
(e.g. a lifelong/incremental learning benchmark). `Rank._draw_pictures()` calls
`draw_heatmap_picture(test_case.output_dir, key, matrix[key])`, and
`test_case.output_dir` has no `output/` subfolder inside it.

## Actual behavior
FileNotFoundError: [Errno 2] No such file or directory: '.../output/<key>-heatmap.png'

Reproduced directly:
import matplotlib; matplotlib.use('Agg')
import tempfile, os
d = tempfile.mkdtemp()
plt.figure(); plt.plot([1,2,3])
plt.savefig(os.path.join(d, 'output', 'test-heatmap.png'))  # FileNotFoundError

## Expected behavior
The `output/` directory should be created (if missing) before saving.

## Note
Distinct from #559/#560/#611, which guard against missing/non-dict Matrix data —
this crashes even when Matrix data is present and valid.

ajmani-x:
## What does this PR do?
Creates the `output/` subdirectory (if missing) before `plt.savefig()` in
`draw_heatmap_picture()`, `core/storymanager/visualization/visualization.py`.

## Why is this important?
The function built a save path pointing inside an `output/` subdirectory
of the test case's workspace, but nothing in the codebase ever creates
that directory. Any benchmark run using a picture-generating `save_mode`
(e.g. `selected_and_all_and_picture`) crashes with `FileNotFoundError`
once results actually contain Matrix data — a different failure mode
than the missing/non-dict Matrix guards already fixed in #559/#560/#611.

## How to test?
Reproduced directly:
```python
import matplotlib; matplotlib.use('Agg')
import matplotlib.pyplot as plt, tempfile, os
d = tempfile.mkdtemp()
plt.figure(); plt.plot([1,2,3])
plt.savefig(os.path.join(d, 'output', 'test-heatmap.png'))  # FileNotFoundError before fix
```
Confirmed the same reproduction succeeds after the fix, with the file
correctly written to `<output_dir>/output/<title>-heatmap.png`.

## Issues fixed
Closes #628

## Type of change
- [x] Bug fix

## Checklist
- [x] I've tested this locally
- [x] My code follows the project style


azaynul10:
The diagnosis is right and the patch is right: the path was built with an `output/`
segment that nothing creates, and `os.path.join(output, "output")` plus a `makedirs`
is the correct repair. Two things about it that are not visible from this diff
alone.

## The patched line is not reachable from `Rank` today

`_draw_pictures` looks the metric up as `test_result.get("Matrix")`
(`core/storymanager/rank/rank.py:238`), and every stage that produces it spells it
`MATRIX`  `SystemMetricType.MATRIX = "MATRIX"` at `core/common/constant.py:96`,
`lifelong_learning.py:67/165/261`, `metrics.py:127`, and the three shipped testenvs
that declare it (`testenv-robot-small.yaml:37`, `testenv-robot.yaml:37`,
`robot-cityscapes-synthia/.../testenv.yaml:36`). Scanning every literal spelling
under `core/` at `37a9c60` returns exactly two, and they disagree:

```
MATRIX   core/common/constant.py:96              MATRIX = "MATRIX"
Matrix   core/storymanager/rank/rank.py:238      matrix = test_result.get("Matrix")
```

So `.get("Matrix")` returns `None`, and `rank.py:239` raises `AttributeError`
before `draw_heatmap_picture` is ever called. Running the real
`Rank._draw_pictures` with a result dict keyed the way `testcase.py:115` keys it:

```
result stores 'MATRIX' -> AttributeError: 'NoneType' object has no attribute 'keys'
result stores 'Matrix' -> FileNotFoundError: .../output/BWT-heatmap.png
same, with output/ present -> no error: completed
```

The middle row is this issue, and it needs the key nothing writes. I think that is
why the reproduction in the description calls `plt.savefig` directly rather than
running a benchmark  the framework path cannot reach the line. That is a fact
about the tree, not a criticism of the repro.

## If #560 merges first, this patch becomes dead code

#560 adds `if not matrix: continue` above the loop. Since the lookup never matches,
that branch is taken on every test case on every run. Running #560's loop body
verbatim with the drawing call counted:

```
result stores 'MATRIX'   -> completed, draw_heatmap_picture reached 0 time(s)
result stores 'Matrix'   -> completed, draw_heatmap_picture reached 1 time(s)
```

So with both PRs merged, `#628` and `#559` would both close, CI would be green, and
`selected_and_all_and_picture` would still produce no picture  silently, with no
error at all. I have raised the same point on #560; it is the ordering that is the
problem, not either patch.

The smallest thing that makes this patch live is binding the lookup to the enum:

```python
from core.common.constant import SystemMetricType

matrix = test_result.get(SystemMetricType.MATRIX.value)
```

No open PR does that. With it in place, this patch is exactly what is needed next.

## Two smaller notes on the same function

**`exist_ok=True` would match the rest of the tree.** The check-then-create pair is
racy if two test cases share a workspace. `core/.../joint_inference.py:71` already
uses `os.makedirs(inference_output_dir, exist_ok=True)`, so:

```python
os.makedirs(os.path.join(output, "output"), exist_ok=True)
```

is one line, matches existing style, and drops the race.

**The figure is never closed.** `visualization.py:33` opens a figure and line 44
calls `plt.show()`, with no `plt.close()`. `_draw_pictures` calls this once per
matrix key per test case, so figures accumulate for the lifetime of the run and
matplotlib starts warning past twenty. Under the non-interactive backend a headless
run uses, `plt.show()` also emits:

```
UserWarning: FigureCanvasAgg is non-interactive, and thus cannot be shown
```

on every call. Since this PR is already touching the save path, `plt.close()` in
place of `plt.show()` would fix both, though I appreciate that widens the diff and
is reasonable to leave for a follow-up.

## Boundary

Executed against `37a9c60`: the spelling scan across every `.py` under `core/`, the
metric-name scan across every `.yaml` under `examples/`, the three
`Rank._draw_pictures` runs above, and #560's loop body with a counter. Not executed:
a full benchmark run. The only shipped config selecting this mode,
`examples/robot/lifelong_learning_bench/semantic-segmentation/benchmarkingjob-sam.yaml`,
needs the robot corpus and RFNet/SAM weights, and its line 10 points at a testenv
path that is not in the tree.

One small correction offered in good faith: the description says this is "a
different failure mode than the missing/non-dict Matrix guards **already fixed** in
#559/#560/#611". #559 is an issue rather than a fix, and #560 and #611 are both
still open, so nothing in that list has landed yet. The distinction the sentence
draws is correct and worth keeping  it is the most useful observation on either
thread it is only the "already fixed" that reads as settled.

haneul-24:
I reviewed the change against Issue #628.

The PR correctly fixes the `FileNotFoundError` by creating the `output/` directory before calling `plt.savefig()`. However, I would not recommend merging this PR as the final fix because PR #753 uses `os.makedirs(..., exist_ok=True)`, which handles both cases when the directory already exists or does not exist. PR #753 also adds `plt.close()` after showing the figure, which helps close the figure properly.

Since both PRs modify the same `draw_heatmap_picture()` function for the same issue, merging both would duplicate the fix and could cause unnecessary conflicts.

**Recommendation: Reject #629  in favor of #753 instead.**

PR #631, opened 2026-08-02
fix(imagenet): update multiedge_inference_bench requirements to insta…

kai11017:
…llable versions

- Bump torch 1.13.1 -> 2.2.2 (first release with Python 3.12 wheels)
- Bump torchvision 0.14.1 -> 0.17.2 (matches torch 2.2.x)
- Bump numpy 1.21.6 -> 1.26.4 (last NumPy 1.x; required by torch 2.2 + onnxruntime 1.17)
- Bump onnx 1.14.1 -> 1.16.2 (has cp312 wheel; fixes slow graph-partition noted in README)
- Bump onnxruntime 1.14.1 -> 1.17.3 (cp312 wheel available)
- Bump onnxruntime-gpu 1.14.1 -> 1.17.1 (cp312 wheel available)
- Bump tokenizers 0.13.3 -> 0.15.2 (required by transformers 4.37+)
- Bump transformers 4.30.2 -> >=4.37.2,<4.40 (retains internal ViT symbols used by vit.py)
- Add onnxscript>=0.1.0 (required by torch.onnx.export() on PyTorch 2.x automatic path)

All versions confirmed installable via pip dry-run on Python 3.12.

**What type of PR is this?**


kai11017:
/cc @MooreZheng @Poorunga

This PR fixes the uninstallable dependency pins in the multiedge_inference_bench 
example (torch==1.13.1 etc. are no longer on PyPI for Python 3.10+). 
Happy to make any changes you suggest!

31groot:
### Request changes — `pandas==1.3.5` contradicts the PR's Python 3.12 installability claim

I independently checked the submitted requirements against the compatibility claim in this PR.

The PR body states:

> “All versions confirmed installable via pip dry-run on Python 3.12.”

However, the submitted `requirements.txt` still contains:

```text
pandas==1.3.5
```

I reproduced the dependency-resolution check independently using pip's wheel-only resolver for CPython 3.12:

```text
python -m pip download \
  --only-binary=:all: \
  --no-deps \
  --dest <tmpdir> \
  pandas==1.3.5 \
  --python-version 3.12 \
  --platform manylinux_2_17_x86_64 \
  --implementation cp \
  --abi cp312
```

The resolver failed with:

```text
ERROR: Could not find a version that satisfies the requirement pandas==1.3.5
ERROR: No matching distribution found for pandas==1.3.5
```

So although the Torch / torchvision / ONNX pins were modernized, the submitted requirements are still not installable as declared on Python 3.12.

This is distinct from the existing discussion on this PR, which focuses on the older `torch==1.13.1`-era pins being unavailable for newer Python versions. That discussion does not cover the remaining `pandas==1.3.5` constraint.

Because Python 3.12 installability is an explicit goal and verification claim of this PR, I recommend treating this as a correctness issue rather than a maintenance concern.

Please update the pandas constraint to a version with a Python 3.12-compatible distribution and rerun the complete requirements-resolution check for the target Python version.

**Verification boundary:** I reproduced the dependency-resolution failure for the submitted `pandas==1.3.5` pin using pip's CPython 3.12 wheel resolver. I did not run the full ImageNet benchmark or install the complete CUDA/runtime stack.

PR #632, opened 2026-08-02
cleanup: remove dead code in JointInference.set_config()

iron-prog:
**What type of PR is this?**

/**What type of PR is this?**

/kind cleanup

Removes dead code from `JointInference.set_config()`.

- Removes a stale commented-out `load_data()` call.
- Removes an unnecessary `hasattr(self.dataset, "test_url")` fallback since `test_url` is always initialized during `Dataset` construction.
- Replaces the fallback with a direct call to `self.dataset.load_data(self.dataset.test_url, "inference")`.

This is a cleanup only and does not change behavior. Verified by running the `jointinference` example before and after the change.

Fixes #


henryng15 (commented):
Reviewed at head `ef4d9ec` against `37a9c60`. The change is correct — but the PR asserts
the removed branch is dead without showing it, and "removes a defensive fallback" is not
something a reviewer should accept on assertion. Here is the verification.

The PR replaces

```python
source = self.dataset.test_url if hasattr(self.dataset, 'test_url') else self.dataset.test_data_info
self.inference_dataset = self.dataset.load_data(source, "inference")
```

with a direct `self.dataset.test_url`. The question is whether `hasattr` can ever be
`False`.

It cannot. `core/testenvmanager/dataset/dataset.py:46` sets

```python
self.test_url: str = ""
```

unconditionally in `__init__`, so the attribute exists on every `Dataset` instance before
any config is parsed. Lines 168-175 then assign it from `test_index`, `test_data` or
`test_data_info`, or raise `NotImplementedError` if none is present — so the object is
never handed to `JointInference` in a state where `test_url` is absent.

One further check, because it is the case that would bite: the original used `hasattr`,
not truthiness. An empty `test_url` (`""`) therefore took the `test_url` branch before
this PR and takes it after. The behaviour is identical, not merely equivalent in the
common case.

The import reordering and blank-line changes in the same diff are unrelated to the
described fix. They are harmless, but splitting them out would make the behavioural part
of this PR reviewable on its own.

**Recommendation: accepted as it is.**

ajmani-x:
Reviewed this as part of an LFX Term 3 pre-test (full review in context here: https://github.com/kubeedge/ianvs/discussions/972#discussioncomment-18186183). The cleanup itself looks correct and safe — the `hasattr(self.dataset, "test_url")` fallback does appear dead if `Dataset.__init__` always sets `test_url` (I did not verify this invariant across every `Dataset` construction path, flagging that as unverified). One thing worth making explicit: the PR description says "Verified by running the jointinference example before and after the change" — which Sedna install was that run against? I found that Ianvs Core's `JointInference(...)` call (in `base.py`, just above where this PR's `set_config()` runs) only succeeds against the exact vendored wheel in `resources/third_party/`, not against the official `kubeedge/sedna` `main` branch or `pip install sedna` — so this verification claim is Sedna-install-dependent in a way that isn't stated. Worth a one-line note on which install was used, so future readers don't assume this PR validates the paradigm works with any correctly-named `sedna` install.

PR #640, opened 2026-08-03
fix(dataset): preserve non-path labels in txt index file preprocessing

### LINKED ISSUE #639: [PRE-TASK] Path/Label Disambiguation Gap in Dataset._process_txt_index_file() Risks Silent Data Corruption
## Background

Ianvs's dataset preprocessing subsystem (`core/testenvmanager/dataset/dataset.py`) rewrites relative
paths inside `.txt`-format index files (`train_index` / `test_index`) to absolute paths before a
benchmarking job runs, via `Dataset._process_txt_index_file()`. This code path is exercised by
`process_dataset()` on every `TestEnv` build whenever a txt-format index file with relative paths is
configured — which currently includes at least `examples/pcb-aoi` (both `singletask_learning_bench`
and `incremental_learning_bench`) and `examples/RoboDK Palletizing/singletask_learning_bench`.

The function (`core/testenvmanager/dataset/dataset.py`, lines 105-119) reads each line of the index
file, splits it on whitespace, and applies `os.path.abspath(os.path.join(root, word))` to **every
resulting token, unconditionally**:

```python
for i in range(length):
    file.writelines(
        f"{os.path.abspath(os.path.join(root, words[i]))}"
    )
```

For the examples I was able to verify against their documented data formats, every token on an index
line happens to already be a file path (e.g. PCB-AoI's index format is
`./JPEGImages/xxx.jpeg ./Annotations/xxx.xml` — an image path followed by an annotation-file path), so
`abspath()` is currently applied only to genuine paths in those cases and produces no visible failure
today. I want to be upfront about that: I could not reproduce an actual crash or visibly wrong output
in any example currently in the repository, and I don't want to overstate this as an active outage.

What I can show is that the function itself has no logic to distinguish a path-like token from a
non-path token (e.g. a bare numeric class label, as is conventional in classification-style index
files such as `image.jpg 1`). If any current or future example's index file mixes a real path with a
non-path label — a very ordinary index-file convention elsewhere in the ML ecosystem — this loop will
silently rewrite that label into an absolute filesystem path (e.g. label `1` becomes
`/abs/path/to/dataset/1`) with no exception raised anywhere. This is strictly worse than a crash: it
produces a dataset that loads successfully but trains or evaluates against corrupted ground truth,
which would be very difficult for a user to detect without independently inspecting the rewritten
index file.

Reproduction of the underlying logic gap (not tied to a specific example's data):
```python
words = ["./JPEGImages/img01.jpg", "1"]   # path + bare class label
for w in words:
    print(os.path.abspath(os.path.join(root, w)))
# ./JPEGImages/img01.jpg -> /abs/root/JPEGImages/img01.jpg   (correct)
# "1"                    -> /abs/root/1                       (silently wrong)
```

## Goals

- Fix `_process_txt_index_file()` so it only rewrites tokens that are actually meant to be paths,
  rather than blindly applying `os.path.abspath()` to every whitespace-delimited token on a line.
- Add a regression test asserting that a non-path token (e.g. a bare label) in an index file is
  preserved unchanged after processing, alongside the existing path-rewriting behavior for genuine
  path tokens.
- Document the expected index-file token format (path-only vs. path+label) for contributors adding
  new examples, so future example authors know which convention this function safely supports today.

## Scope

This is a core-framework fix in the dataset preprocessing subsystem, not an example-specific patch, since
`_process_txt_index_file()` is shared by every example using txt-format index files with relative paths.
Expected users: any current or future Ianvs example author whose index files use a mix of path and
non-path tokens per line.

**Relationship to existing issues**: while investigating this function I found that issue #626
(opened earlier today by another contributor) analyzes the same function, `_process_txt_index_file()`,
and identifies a real, distinct bug in it — an unmanaged `tempfile.mkdtemp()` resource leak. That issue
does not mention the path/label disambiguation gap described here; the two are independent defects in
the same function (one about resource lifecycle, this one about data correctness) and the fixes should
not conflict, though I flag #626 explicitly here since a reviewer will likely want both applied together
in one pass over this function. I have not found any other open issue describing the path/label
disambiguation problem specifically.

## Detailed design

**Architecture**: The fix is entirely contained within `core/testenvmanager/dataset/dataset.py`. No
changes to `core/testcasecontroller/`, `core/storymanager/`, or any example code are required, since
`_process_txt_index_file()` is a self-contained preprocessing step invoked from `_process_index_file()`
/ `process_dataset()` before a test case ever runs.

**Module details**: The revision replaces the unconditional `os.path.abspath()` call on every token
with a per-token check — a simple approach is to only rewrite a token if it already looks like a
relative filesystem reference (e.g. contains `os.sep`, or matches a configurable set of known
extensions), leaving bare tokens (labels) untouched. An alternative, more explicit approach is to add
an optional `path_columns` parameter to `split_dataset()`/the index-file config, letting example authors
declare which columns are paths versus labels — more flexible, but a larger API surface change. I'd
propose starting with the simpler heuristic fix plus the regression test, and raising the explicit
`path_columns` design as a follow-up if maintainers want stronger guarantees. No change to core module
(`core/testcasecontroller/algorithm/`) is needed since path resolution happens entirely upstream of
algorithm execution.

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

Fixes the path/label disambiguation gap in `Dataset._process_txt_index_file()` (`core/testenvmanager/dataset/dataset.py`).

Previously, `_process_txt_index_file()` read each line of a `.txt` index file and unconditionally applied `os.path.abspath(os.path.join(root, word))` to every whitespace-separated token. When index files include non-path tokens such as bare numeric class labels (e.g. `./images/img1.jpg 1`), the numeric label `1` was silently rewritten into an absolute filesystem path `/path/to/root/1`.

Changes Introduced:
1. **Token-Level Path Identification (`_is_path_token`)**:
   Added `@classmethod _is_path_token(cls, word, root)` to check whether a token is a file path (contains `/` or `\`, starts with `.`, exists relative to `root`, or is an absolute path).
2. **Preserve Non-Path Labels**:
   Updated `_process_txt_index_file()` to rewrite only relative file path tokens to absolute paths while preserving non-path tokens (like class labels `"0"`, `"1"`, `"cat"`) as-is.
3. **Unit Tests**:
   Added `tests/test_txt_index_file.py` to verify path resolution for relative file paths and preservation of non-path labels.

Fixes #639

NONE

```release-note
Fix Dataset._process_txt_index_file() to preserve non-path class labels when rewriting relative file paths to absolute paths in text index files.
```

### DIFF
--- a/core/testenvmanager/dataset/dataset.py
+++ b/core/testenvmanager/dataset/dataset.py
@@ -88,34 +88,60 @@ class Dataset:
             )
 
     @classmethod
+    def _is_path_token(cls, word, root):
+        if not word:
+            return False
+        if os.path.isabs(word):
+            return True
+        if "/" in word or "\\" in word or word.startswith("."):
+            return True
+        ext = os.path.splitext(word)[1].lower()
+        if ext and len(ext) > 1 and ext[1:].isalnum() and not ext[1:].isdigit():
+            return True
+        if os.path.isfile(os.path.join(root, word)):
+            return True
+        return False
+
+    @classmethod
     def _process_txt_index_file(cls, file_url):
         """
         convert the index info of data from relative path to absolute path in txt index file
         """
+        import re
+
         flag = False
         new_file = file_url
+        root = os.path.dirname(file_url)
         with open(file_url, "r", encoding="utf-8") as file:
             lines = file.readlines()
             for line in lines:
-                if not os.path.isabs(line.split(" ")[0]):
-                    flag = True
+                stripped_line = line.strip()
+                if not stripped_line:
+                    continue
+                tokens = [t for t in re.split(r"[\t ]+", stripped_line) if t]
+                for token in tokens:
+                    if cls._is_path_token(token, root) and not os.path.isabs(token):
+                        flag = True
+                        break
+                if flag:
                     break
         if flag:
-            root = os.path.dirname(file_url)
             tmp_file = os.path.join(tempfile.mkdtemp(), "index.txt")
             with open(tmp_file, "w", encoding="utf-8") as file:
                 for line in lines:
-                    # copy all the files in the line
-                    line = line.strip()
-                    words = line.split(" ")
-                    length = len(words)
-                    words[-1] = words[-1] + "\n"
-                    for i in range(length):
-                        file.writelines(
-                            f"{os.path.abspath(os.path.join(root, words[i]))}"
-                        )
-                        if i < length - 1:
-                            file.writelines(" ")
+                    stripped_line = line.strip()
+                    if not stripped_line:
+                        file.write("\n")
+                        continue
+                    delimiter = "\t" if "\t" in line else " "
+                    tokens = [t for t in re.split(r"[\t ]+", stripped_line) if t]
+                    processed_tokens = []
+                    for token in tokens:
+                        if cls._is_path_token(token, root) and not os.path.isabs(token):
+                            processed_tokens.append(os.path.abspath(os.path.join(root, token)))
+                        else:
+                            processed_tokens.append(token)
+                    file.write(delimiter.join(processed_tokens) + "\n")
 
             new_file = tmp_file
 
--- a/tests/test_txt_index_file.py
+++ b/tests/test_txt_index_file.py
@@ -1,0 +1,110 @@
+# Copyright 2026 The KubeEdge Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import sys
+import tempfile
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+from core.testenvmanager.dataset import Dataset
+
+
+def test_process_txt_index_file_path_label_disambiguation():
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        index_file = os.path.join(tmp_dir, "train_index.txt")
+        lines = [
+            "./images/img01.jpg 1\n",
+            "./images/img02.jpg 0\n",
+            "./images/img03.jpg ./annotations/ann03.xml\n"
+        ]
+        with open(index_file, "w", encoding="utf-8") as f:
+            f.writelines(lines)
+
+        processed_file = Dataset._process_txt_index_file(index_file)
+
+        assert os.path.exists(processed_file)
+        with open(processed_file, "r", encoding="utf-8") as f:
+            processed_lines = [l.strip() for l in f.readlines()]
+
+        expected_img01_path = os.path.abspath(os.path.join(tmp_dir, "./images/img01.jpg"))
+        expected_img02_path = os.path.abspath(os.path.join(tmp_dir, "./images/img02.jpg"))
+        expected_img03_path = os.path.abspath(os.path.join(tmp_dir, "./images/img03.jpg"))
+        expected_ann03_path = os.path.abspath(os.path.join(tmp_dir, "./annotations/ann03.xml"))
+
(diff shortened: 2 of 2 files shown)


bhuvan-somisetty:
Hi @MooreZheng @hsj576 @Poorunga, could you please review this PR when you get a chance? Thanks!

MukandKrishna (commented):
**Recommendation: major revision.** The shared-layer location and added tests are good, but I reproduced three cases the heuristic does not handle safely.

1. Current CIFAR100 rows are tab-delimited (`filename<TAB>numeric-label` in `examples/cifar100/utils.py:54`). This PR still uses `split(" ")`, so:

```text
INPUT='cifar100_train_index_0.npy\t0\n'
RETURNED_NEW_FILE=False
OUTPUT='cifar100_train_index_0.npy\t0\n'
```

The feature path remains relative.

2. `_is_path_token()` makes semantics depend on local filesystem contents. With an image and string label `cat`, plus an unrelated directory named `cat`:

```text
INPUT='./images/img.jpg cat\n'
OUTPUT='<TEMP>/images/img.jpg <TEMP>/cat\n'
```

The non-path label is silently converted solely because `root/cat` exists.

3. If field zero is absolute and field one is a relative annotation path, the initial scan never sets `flag`, so the original file is returned and field one remains relative.

These are data-contract problems rather than missing heuristic clauses. Please consider explicit path-column metadata, generic whitespace tokenization, and column-count validation. Tests should add the checked-in CIFAR100 tab format, a label colliding with a local filename or directory, and an absolute-first/relative-second row.

Executed against PR head `258827bd2bc4b3a20b94ba3835fde4f5b571fee8`; full Example training was not run. Related reproduction context: [Task 1](https://github.com/kubeedge/ianvs/discussions/921#discussioncomment-18145877) and the [Issue #639 follow-up](https://github.com/kubeedge/ianvs/issues/639#issuecomment-5408089451).

bhuvan-somisetty:
Thanks for the thorough review and test cases @MukandKrishna! Updated `_process_txt_index_file()` to:
1. Support both tab (`\t`) and space delimiters, preserving the original line formatting (including CIFAR100 index files).
2. Use `os.path.isfile` and extension-aware path token inspection rather than directory existence, avoiding false positive label matches.
3. Check all tokens across the row so relative secondary paths (e.g. annotations) are converted even when column zero is already absolute.
4. Added tests in `tests/test_txt_index_file.py` covering CIFAR100 tab rows, absolute-first/relative-second rows, and label collision edge cases — all 4 tests pass cleanly.

MukandKrishna:
> Thanks for the thorough review and test cases @MukandKrishna! Updated `_process_txt_index_file()` to:
> 
> 1. Support both tab (`\t`) and space delimiters, preserving the original line formatting (including CIFAR100 index files).
> 2. Use `os.path.isfile` and extension-aware path token inspection rather than directory existence, avoiding false positive label matches.
> 3. Check all tokens across the row so relative secondary paths (e.g. annotations) are converted even when column zero is already absolute.
> 4. Added tests in `tests/test_txt_index_file.py` covering CIFAR100 tab rows, absolute-first/relative-second rows, and label collision edge cases — all 4 tests pass cleanly.

Thanks for the update, @bhuvan-somisetty. I independently rechecked the updated head [`360947f`](https://github.com/kubeedge/ianvs/commit/360947f92392b9f7c17e325c89951041ca49ef29).

Verified on this head:

- All four test functions in `tests/test_txt_index_file.py` pass when executed directly.
- The CIFAR100 tab-delimited row preserves the tab separator while converting the `.npy` path.
- A directory named like a label no longer causes the label to be rewritten as a path.
- A relative secondary path is converted when the first path is already absolute.
- The original numeric-label preservation case continues to pass.

One non-blocking precision note: the implementation preserves the delimiter choice for ordinary single-tab or single-space rows, but mixed or repeated whitespace is normalized. For example, `"  \t"` becomes `"\t"`, so formatting is not preserved byte-for-byte. This does not affect the demonstrated CIFAR100 tab format.

Verification boundary: I reran `tests/test_txt_index_file.py` using `pytest 9.1.1` in a dedicated Python 3.14 environment. Result: **4 passed in 0.11s**. The repository does not declare `pytest`, and unused external imports required by the `Dataset` module were isolated for this focused unit-test run. I also reran the independent reproduction. I did not run the full repository test suite.

On this evidence, the three functional findings from my original review are addressed at `360947f`.

bhuvan-somisetty:
Thank you @MukandKrishna for the thorough independent re-testing and verification across all the test scenarios! Appreciate your detailed reviews and insights.

PR #642, opened 2026-08-03
fix: correct __all__ mismatch in Cloud_Robotics cloud_model.py

### LINKED ISSUE #641: `__all__` in `cloud_model.py` references non-existent name `QwenSemanticSegmentation`
**What happened**:

`examples/Cloud_Robotics/cloud-edge-collaborative-inference_bench/perception-reasoning/testalgorithms/cloud_model.py` defines:

```python
__all__ = ["QwenSemanticSegmentation"]

iron-prog:
**What type of PR is this?**

/kind bug

`cloud_model.py` exported `QwenSemanticSegmentation` in `__all__`, but that
symbol does not exist in the module. The actual exported class is
`CloudModel`.

This PR updates `__all__` to reference `CloudModel`, keeping the module
exports consistent with the implementation. This is a non-functional change
that removes an incorrect export.

Fixes #641


henryng15 (commented):
Reviewed at head `8a37653` against `37a9c60`.

The change is correct — `cloud_model.py` defines `CloudModel`, not
`QwenSemanticSegmentation`, so pointing `__all__` at the real name is right.

One note on impact, since the PR is filed `/kind bug`.

**`__all__` is not read by anything in Ianvs.** Verified at `37a9c60`: `core/` contains
**0** star-imports and **0** references to `__all__`, and nothing in `examples/`
star-imports this module. Core resolves the class by ClassFactory alias —
`core/testcasecontroller/metrics/metrics.py:173` and
`core/testcasecontroller/algorithm/module/module.py:116,130` — which this module already
supplies correctly.

So issue #641 describes a genuine inconsistency with no runtime consequence, and this PR
has zero regression risk. Both are good reasons to merge it; both also argue for
`kind/cleanup` over `kind/bug`, so the Example is not counted as broken when it is not.

For context, this pattern is repository-wide rather than specific to this file: 30
`__all__` declarations across 10 Examples are malformed, and none is load-bearing.
`examples/robot-cityscapes-synthia/.../erfnet/task_definition_by_domain.py` is a neat
example — `__all__` names `TaskDefinitionByDomain` while the class is
`TaskDefinitionByOrigin`, and the Example works anyway because the ClassFactory alias is
correct.

Reproduce: `python3 tools/probe_all_is_inert.py ianvs` —

**Recommendation: minor revision** — description and label only.

PR #643, opened 2026-08-03
fix(examples): replace hardcoded absolute paths with relative paths

### LINKED ISSUE #515: bug: hardcoded absolute paths (/home/*) across multiple examples make benchmarks non-portable

priyam99:
Fixes #515

**What type of PR is this?**
/kind bug

Multiple examples contained hardcoded absolute developer paths (e.g. \/home/wyd/\, \/home/icyfeather/\, \/home/shijing.hu/\) across YAML config files and Python entrypoints, breaking portability when benchmarking jobs were executed in different environments.

This PR replaces all hardcoded absolute paths across \examples/\ with workspace-relative paths (\./...\).

Fixes #515


diegodevxd:
The PR title/body say all hardcoded absolute paths under `examples/` are replaced, but that is not true for the PR head. A scan still finds active private-root references, for example RoboDK's `/root/ianvs/project/...` config/dataset paths and Cloud VLA's `/inspire/hdd/...` values in `benchmarkingjob.yaml`, `testenv.yaml`, and its algorithm configuration/code. This is not a request to rewrite every external asset string blindly; it is a mismatch between the claimed scope and the executable surface.

Please either narrow the claim and add an acceptance test for the exact `/home/` inventory, or classify and cover the remaining active private-root families. A table-driven config-closure test for one changed benchmark (`benchmarkingjob â†’ testenv â†’ algorithm â†’ module`) would also guard the root-relative `./examples/...` convention against bulk-edit regressions.

Static evidence only; full examples require external assets.

diegodevxd:
Follow-up from local verification of this PR head `b3fc22c1d13ef3c0fa7c2387b4b8d45a4b4b3270`: the `/home/` cleanup is measurable, but the repair boundary remains too narrow. A raw scan finds `/root/` and `/inspire/` families unchanged; I then manually verified four active checked-in RoboDK YAML operands under `/root/ianvs/...` (job testenv, job algorithm, and two metric URLs). They are unchanged on this branch and fail structural closure before any external asset is needed.

Please add a policy-based regression fixture covering a checked-in `/root/` or `/inspire/` reference and a separately declared external dataset/model path. That would prevent future scans from fixing one private-root family while leaving equivalent repository references behind.

Karan825:
Independent review of #643

Recommendation: major revision

The PR correctly addresses the portability problem in #515 by replacing developer-specific absolute paths with relative paths across multiple Examples.

However, there is a path-resolution concern that is not addressed by the PR.

In the BDD Example, #643 changes developer-specific absolute paths such as:

    /home/liyunzhe/Mobile-Inference/algorithm/labels/...

to:

    ./algorithm/labels/...

Ianvs Core currently resolves local relative paths through `os.path.isfile(url)`, which uses the process working directory.

I independently reproduced this behavior: the same relative path returned `True` from the repository root and `False` after changing the working directory.

Therefore, the PR removes the machine-specific absolute path but does not establish a deterministic base directory for the replacement relative path. The result can remain dependent on how the benchmark is invoked.

I could not perform the full BDD benchmark because the referenced external dataset artifacts are not present in the repository. The finding is therefore based on static code-path analysis plus an independent path-resolution reproduction.

I recommend defining an explicit path-resolution contract and adding regression tests that verify configuration paths independently of the caller's working directory.

A separate repository-policy issue is also visible on this PR: GitHub currently marks the commit message as invalid, which should be corrected before merge.

iron-prog:
## Review from LFX Term 3 pretest — full analysis in Discussion #883

### Correction first

An earlier pass of my analysis in Discussion #883 incorrectly associated this PR with RoboDK / `cloud_VLA_finetune` hardcoded paths. I re-checked the actual PR head and diff (`b3fc22c1d13ef3c0fa7c2387b4b8d45a4b4b3270`) and am correcting that here: the relevant scope is the hardcoded-path changes in the examples touched by this PR, including the CIFAR-100 and `llm_simple_qa` changes visible in the actual diff. The earlier broader attribution should not be used.

### What I checked

I checked the PR at its actual head and validated the changed configuration rather than relying only on the PR description.

I constructed the `BenchmarkingJob` from the changed `benchmarkingjob.yaml` successfully, then called the real `TestEnv.prepare()` against the changed `testenv.yaml`.

The preparation path fails in dataset processing with:

```text
RuntimeError: prepare dataset failed, error: not one of train_index/train_data/train_data_info

PR #645, opened 2026-08-04
fix(core/lifelong): resolve Sedna API mismatches and module object instantiation crashes

### LINKED ISSUE #470: [Bug] lifelong_learning.py and module.py crash with multiple Sedna API mismatches when running robot-cityscapes-synthia example
## Environment

| Item | Version |
|---|---|
| OS | macOS 14 (Darwin 23.6.0) |
| Python | 3.12.7 |
| Sedna | 0.6.0.1 |
| PyTorch | 2.12.0 |
| ianvs branch | main |

## Example

```
```

## Run Command

```bash
```

## Description

Running the `robot-cityscapes-synthia` lifelong learning benchmark causes a cascade of crashes in two ianvs core files: `core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py` and `core/testcasecontroller/algorithm/module/module.py`. Each crash is a different mismatch between what these files expect from Sedna and what the installed Sedna (0.6.0.1) actually provides.

The errors appear in sequence as the pipeline progresses and each one blocks the next stage of execution.

---

## Error 1 — `module.py` passes STP module instances; Sedna expects dicts

**Triggered at:** Task definition/allocation setup, before training begins.

```
  File ".../sedna/algorithms/seen_task_learning/seen_task_learning.py", line 160
    method_name = self.task_definition.get("method", ...)
AttributeError: 'TaskDefinitionByOrigin' object has no attribute 'get'
  error: 'TaskDefinitionByOrigin' object has no attribute 'get'
```

**Location:** `core/testcasecontroller/algorithm/module/module.py`

`module.py` instantiates STP-class modules (task_definition, task_allocation, etc.) and passes the live Python objects to Sedna. Sedna's `SeenTaskLearning` internally calls `.get("method")` on whatever it receives, expecting a dict of the form `{"method": "ClassName", "param": {...}}`. The existing code only returns a dict for `ClassType.HEM`; all STP modules are passed as instances, causing this crash.

---

## Error 2 — `job.my_evaluate()` does not exist in Sedna

**Triggered at:** First evaluation step after round-0 training.

```
  File ".../lifelong_learning.py", line 419, in my_eval
    edge_task_index, tasks_detail, res = job.my_evaluate(eval_dataset, metrics=metric_func)
AttributeError: 'LifelongLearning' object has no attribute 'my_evaluate'.
  Did you mean: 'evaluate'?
  error: 'LifelongLearning' object has no attribute 'my_evaluate'
```

**Location:** `core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py` — `my_eval()` method.

`my_evaluate` does not exist in the Sedna `LifelongLearning` API. The available methods are `train`, `update`, `evaluate`, and `inference`. Note: PR #301 fixed a different occurrence of `my_evaluate` at line 284 but left this call at line 419 untouched.

---

## Error 3 — `job.train()` return value treated as a file path; Sedna returns a metrics dict

**Triggered at:** After round-0 training completes, when passing the task index to the next round.

```
  File ".../lifelong_learning.py", line 357, in _train
    os.environ["CLOUD_KB_INDEX"] = cloud_task_index
TypeError: str expected, not dict
  error: str expected, not dict
```

**Location:** `core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py` — `_train()` method.

`_train()` stores the return value of `job.train()` as `cloud_task_index` and passes it as a file path to subsequent stages. Current Sedna (0.6.0.1) `LifelongLearning.train()` returns a training metrics dict (`{'accuracy': 0.0, ...}`), not a path. The actual task index is saved to disk at `{OUTPUT_URL}/index.pkl` during training.

---

## Error 4 — `UpdateStrategyDefault` not registered in Sedna KM class factory

**Triggered at:** When ianvs registers the lifelong learning job, Sedna internally looks up the default task update strategy.

```
  File ".../sedna/common/class_factory.py", line 108, in get_cls
    raise RuntimeError(f"can't find class type {class_type} class name {t_cls_name}"
                        " in class registry")
RuntimeError: can't find class type knowledge_management class name UpdateStrategyDefault
  error: can't find class type knowledge_management class name UpdateStrategyDefault
```

**Location:** `core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py` — at `build_paradigm_job()` time.

`Sedna`'s `LifelongLearning.__init__()` defaults `task_update_decision` to `{"method": "UpdateStrategyDefault"}` and later resolves it via `ClassFactory.get_cls(ClassType.KM, "UpdateStrategyDefault")`. No class is registered under this name in the installed Sedna 0.6.0.1.

---

## Error 5 — `UpdateStrategyByFinetune` (nearest substitute) requires an `attribute` parameter that is never provided

**Triggered at:** Immediately after the KM registry lookup, when Sedna instantiates the update strategy.

```
  File ".../task_update_decision_finetune.py", line 39, in __init__
    self.attribute = kwargs.get("attribute").split(", ")
AttributeError: 'NoneType' object has no attribute 'split'
  error: 'NoneType' object has no attribute 'split'
```

**Location:** `sedna/algorithms/seen_task_learning/task_update_decision/task_update_decision_finetune.py`, triggered by ianvs calling into Sedna's KM layer.

`UpdateStrategyByFinetune.__init__()` unconditionally calls `.split()` on the result of `kwargs.get("attribute")`. When no `attribute` kwarg is passed, `get()` returns `None` and the split crashes. The ianvs pipeline never supplies this parameter.

---

## Error 6 — `job.my_inference()` does not exist in Sedna

**Triggered at:** Final inference step on the test set after all training rounds complete (no-inference mode branch, line 167).

```
  File ".../lifelong_learning.py", line 167, in run
    test_res = job.my_inference(inference_dataset, **kwargs)
AttributeError: 'LifelongLearning' object has no attribute 'my_inference'
  error: 'LifelongLearning' object has no attribute 'my_inference'
```

**Location:** `core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py` — `run()`, `no-inference` mode branch.

---

## Error 7 — `job.inference_2()` does not exist in Sedna

**Triggered at:** Per-sample inference inside `_inference()`, used by the `hard-example-mining` and default mode branches (line 345).

```
  File ".../lifelong_learning.py", line 345, in _inference
    res, is_unseen_task, _ = job.inference_2(data, **kwargs)
AttributeError: 'LifelongLearning' object has no attribute 'inference_2'
  error: 'LifelongLearning' object has no attribute 'inference_2'
```

**Location:** `core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py` — `_inference()` method.

`inference_2` does not exist in the Sedna `LifelongLearning` API. The installed Sedna 0.6.0.1 exposes `inference()` which returns `(res, is_unseen_task, tasks)`.

---

## Files Involved

- `core/testcasecontroller/algorithm/module/module.py`
- `core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py`

priyam99:
Fixes #470

**What type of PR is this?**
/kind bug

Running lifelong learning benchmarks (such as `robot-cityscapes-synthia`) triggered a cascade of crashes in `core/testcasecontroller/algorithm/module/module.py`, `base.py`, and `lifelong_learning.py` due to Sedna API contract mismatches:

1. **Module Instance Specification Fix** (`module.py`): Return dict specifications `{"method": name, "param": hyperparameters}` for Sedna module types (`HEM`, `STP`, `KM`, `UTP`, `UTD`) when custom modules are loaded instead of returning raw Python class instances. This resolves `AttributeError: 'TaskDefinitionByOrigin' object has no attribute 'get'`.
2. **Task Update Decision Fallbacks** (`base.py`): Provide default `task_update_decision = {"method": "UpdateStrategyByFinetune", "param": {"attribute": "accuracy"}}` when unconfigured or missing `attribute`, preventing Sedna lookup failures for `UpdateStrategyDefault` and attribute split crashes.
3. **Sedna API Method Alignment** (`lifelong_learning.py`):
   - Replace undefined `job.my_evaluate(...)` with valid Sedna `job.evaluate(...)`.
   - Guarantee return of string file path (`index.pkl`) when Sedna `job.train()` or `job.evaluate()` returns metric dictionaries.
   - Replace undefined `job.my_inference(...)` and `job.inference_2(...)` with standard Sedna `job.inference(...)`.

Fixes #470


rajdhruvsingh:
Read the full diff against upstream/main@37a9c60a9747af0cfe3170f84249bc349c56e8d5 and executed the patched code in isolation. Split verdict: most of this PR should land; one hunk should not, in its current form.

Genuinely valuable and correct (should be preserved):
- module.py dict-spec fix — passes {"method": ..., "param": ...} dicts where Sedna expects them instead of live instances. Blast radius verified: 9 of the 10 shipped lifelong-learning algorithm.yaml configs declare a task_definition module, so this fix reaches nearly every LL example, not just robot-cityscapes-synthia where #470 reproduced it.
- base.py default for task_update_decision — verified that all 10 LL algorithm.yaml files omit this key entirely, so the default fires everywhere.
- job.train() fallback — isinstance(res, str) check with os.path.join(train_output_dir, "index.pkl") correctly handles Sedna returning a metrics dict (the TypeError: str expected, not dict from #470 Error 3).

These three fixes are same-problem/same-solution as #470's documentation — correct, but not novel; landing them standalone would decouple them from the problematic hunk below.

The my_eval hunk has a fatal contract flaw:
eval_result = job.evaluate(eval_dataset, metrics=metric_func)
if isinstance(eval_result, (tuple, list)) and len(eval_result) == 2:
    tasks_detail, res = eval_result
else:
    tasks_detail, res = [], eval_result
Two independently verifiable problems:
1. The isinstance guard protects a shape the API cannot produce. Sedna v0.6.0's evaluate() ends return callback_func(res) if callback_func else res — a single value in both branches; tasks_detail is consumed entirely inside _task_evaluation() and never surfaced to callers. The guarded 2-tuple is unreachable by construction, while the method then returns a hard-coded legacy 3-value shape that diverges from the real contract.
2. Executed isolation shows the returned shape carries an empty task list: calling the patched my_eval against a spec'd mock yields ('<workspace>/output/eval/1/index.pkl', [], {'accuracy': 0.83}) — tasks_detail observed as [], which is precisely the input that later drives task_avg_score['accuracy']/i toward ZeroDivisionError downstream.

Recommendation: split the PR. Land module.py + base.py + _train standalone (they're ready); rework my_eval to align with the single-value return the installed Sedna actually exposes, and note that any such rework must also update the five run() call sites' unpacking in the same commit — partial adoption leaves the file internally inconsistent.

suhaan-24:
Reviewed against `sedna 0.6.0.1` on Python 3.9.6 / torch 2.8.0 / macOS Apple M4, CPU+MPS, no CUDA.
I ran this PR rather than only reading it; results and boundaries below.

## The core fix works — executed

I checked this branch out into an isolated worktree and ran
`examples/cityscapes-synthia/lifelong_learning_bench/curb-detection` against it.

**As submitted, the PR cannot be exercised at all.** It fails at config load, before any line it
changes is reached:

```
RuntimeError: not found testenv config file(./examples/curb-detection/lifelong_learning_bench/testenv/testenv.yaml) in local
```

The Example's `benchmarkingjob.yaml` still points at a directory that no longer exists. That is
not this PR's fault, but it does mean the PR cannot be validated end-to-end without also
repairing the Example layer.

**With only the Example-local repairs applied on top** (stale config paths corrected,
`train_url`/`test_url` renamed to `train_index`/`test_index`, dataset path repointed — no other
change), the `module.py` fix demonstrably works: the log contains **zero** occurrences of
`'TaskDefinitionByOrigin' object has no attribute 'get'`, and execution reaches
`Model loaded successfully` and enters `_train`. That is a real improvement and it is worth
stating clearly.

Execution then stops here:

```
[INFO] lifelong_learning.py(316) - Download last task index from /tmp/cloud_task/index.pkl to index.pkl
  File ".../paradigm/lifelong_learning/lifelong_learning.py", line 369, in _train
Exception: index.pkl not exists!
```

Sedna took the `update()` branch on the *first* training call — the
`HAS_COMPLETED_INITIAL_TRAINING` off-by-one. This PR does not touch that, so on its own it does
not make a lifelong-learning Example runnable. Not a defect in the change; useful for sequencing
it against the PR that does fix the flag.

## Two concerns in the diff

### 1. `_eval` returns an index that is not edge-deployable

```python
eval_res = job.evaluate(eval_dataset, metrics=metric_func)
if isinstance(eval_res, str) and eval_res:
    edge_task_index = eval_res
else:
    edge_task_index = os.path.join(eval_output_dir, "index.pkl")
```

The motivation is right — sedna 0.6.0.1 returns metrics, not a path. But the index written to
`output/eval/1/` is missing the field the edge deployment path dereferences. Loading the
artifacts from a completed run with `sedna.common.file_ops.FileOps.load`:

```
--- eval/1/index.pkl ---   keys=['seen_task', 'unseen_task']          <-- no 'create_time'
  entry='real_semantic_segamentation_model'  samples.data_url=None
  entry='sim_semantic_segamentation_model'   samples.data_url=None

--- train/0/index.pkl ---  keys=['seen_task', 'unseen_task', 'create_time']
  entry='real_semantic_segamentation_model'  samples.data_url='.../train_real_semantic_segamentation_model.sample'
```

`EdgeKnowledgeManagement.save_task_index()` does:

```python
_task.samples.data_url = FileOps.download(_task.samples.data_url, sample_dir)
```

and `FileOps.download(None, ...)` raises
`AttributeError: 'NoneType' object has no attribute 'startswith'`.

This is latent rather than immediately visible: `eval/2/index.pkl` *does* carry `data_url`
(it points back at the `train/0` samples), so only the first round is affected and a smoke test
that checks the final leaderboard will not catch it. Suggest either checking deployability before
returning the eval index and falling back to the train/cloud index, or returning the cloud index
outright and logging the decision.

### 2. The `task_update_decision` fallback resolves to a class that raises

```python
task_update_decision = {"method": "UpdateStrategyByFinetune", "param": {"attribute": "accuracy"}}
```

The `attribute` default is genuinely needed — `UpdateStrategyByFinetune.__init__` does
`kwargs.get("attribute").split(", ")`, which raises on `None`. But the resulting
`self.attribute = ["accuracy"]` is then compared against `task.meta_attr`, and `__call__` does:

```python
task = tasks[[i for i, task in enumerate(tasks) if
              task.meta_attr == self.attribute[0]]][0]
```

That indexes a Python list with a list. When no task matches — which is the case for any Example
whose `meta_attr` values are not `accuracy`, e.g. curb-detection's `real`/`sim` — it raises
`TypeError: list indices must be integers or slices, not list`. The same method also filters on
a hardcoded city (`self.city = kwargs.get("city", "berlin")`).

Blast radius, measured against `upstream/main`:

```
$ git grep -l 'task_update_decision' upstream/main -- 'examples/**/*.yaml'
(none)
$ git grep -l 'lifelonglearning' upstream/main -- 'examples/**/*.yaml' | wc -l
11
```

No Example configures `task_update_decision`, and eleven lifelong-learning algorithm YAMLs exist —
so this fallback is reached by all of them. Worth considering whether an explicit, working
strategy module belongs in the Examples instead, with the Core fallback failing loudly rather
than routing into a method that cannot succeed.

## Smaller note

The `module.py` change omits `param` entirely when `hyperparameters` is empty, rather than
passing an empty dict. That is a nice touch and worth keeping.

## Verification boundary

Executed: this PR as submitted, and with the Example-local layer applied, on
`cityscapes-synthia/lifelong_learning_bench/curb-detection` (CPU/MPS, no GPU, no gated dataset).
The index inspection, `FileOps.download(None)` behaviour, and the `git grep` counts are executed
probes. I have **not** run `robot-cityscapes-synthia`, the Example this PR names — its `mdil-ss`
dataset has no resolvable public link — so my confirmation is from a second Example that shares
the same Core path, not a reproduction of the original report.

phantom-712:
The proposed direction is appropriate for the installed Sedna surface, but the
patch should be reviewed as several contracts rather than as one mechanical
rename.

1. Strategy configuration types (`HEM`, `STP`, `KM`, `UTP`, and `UTD`) need
   dictionary specifications with `method` and `param` keys because Sedna
   consumes them with `.get(...)`. BaseModel and aggregation must retain their
   runtime object contracts; converting every module result to a dictionary
   would repair one consumer while breaking another.
2. The public Sedna `evaluate()` method returns a single result value. A
   fallback that conditionally unpacks two or three values is not a safe
   compatibility adapter. The caller must obtain task-index state from the
   persisted job/output contract and keep any task-detail handling at the
   correct boundary.
3. `train()` returning a metrics dictionary cannot be passed directly to an
   environment variable or file-loading function that expects a path. The
   implementation must distinguish the observable metrics result from the
   index file written by the job.
4. The update-strategy default must be tested with the exact Sedna artifact,
   including the required `attribute` parameter. A registry alias that only
   changes the lookup name is not sufficient if the selected implementation
   then dereferences a missing parameter.

Recommendation: major revision unless the patch is split into independently
tested module, update-strategy, training-index, and public-API changes. The
verification boundary is the exact Sedna wheel plus contract fixtures. The repaired
composed Robot RFNet simple configuration was also run against the
official Cloud-Robotics resource; it returned Ianvs code 0 after five rounds
and produced the expected metric and rank artifacts. That run supports the
adapter composition only and does not prove that PR #645 passed unchanged.

swikarthanekar (commented):
Reviewed #645 against `main @ 37a9c60` with the bundled Sedna 0.6.0.1 wheel (`resources/third_party/sedna-0.6.0.1-py3-none-any.whl`, SHA256 verified).

I ran the published `robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation` benchmark on the PR branch. It dies during config parsing at the same stale testenv path as main:

`./examples/class_increment_semantic_segmentation/lifelong_learning_bench/testenv/testenv.yaml`

The changed Core lifecycle code is not reached in that launch. I keep that runtime attempt separate from the source review below.

Source-level, the migration lines up with bundled Sedna 0.6.0.1:

- legacy lifecycle calls removed from `lifelong_learning.py` (`my_evaluate`, `my_inference`, `inference_2`)
- `module.py` serializes STP/KM/UTP/UTD modules into the dict shape Sedna expects
- `base.py` adds a default `task_update_decision` with `UpdateStrategyByFinetune` and `attribute: accuracy`
- `_train`, `_eval`, and `my_eval` adapt non-path Sedna returns toward Ianvs index handling
- inference uses `job.inference()` with 3-tuple unpack

That targets the shared Core contract behind #470, not a one-off Example workaround. Module representation and lifecycle rename look headed in the right direction.

Before merge I would still want code changes around:

1. **`my_eval` / empty tasks:** if `evaluate()` returns metrics while `tasks_detail` is empty, downstream averaging may still hit a divide-by-zero path.
2. **KM fallback:** default `UpdateStrategyByFinetune` with `attribute: accuracy` may not match every Example's task metadata.
3. **Eval index deployability:** first-round eval index needs to stay usable downstream, including fields like `data_url` where required.
4. **Tests:** broad Core surface change with no focused coverage on module serialization or return-shape branches.

I executed a benchmark launch on the #645 tree and inspected the bundled Sedna artifact. The run stops at stale testenv config before changed Core code runs. I also reviewed the Core diff statically. I am not claiming #645 passes or fails end to end from that launch.

Stub-job tests for module serialization, evaluate return adaptation, empty-task handling, and inference tuple unpack would cut merge risk.

**Major revision.** Direction looks appropriate, but the correctness items above need code fixes, not just docs. Address empty-`tasks_detail` in the eval path, revisit KM fallback for non-`accuracy` tasks, confirm eval index deployability or document fallback, and add focused regression tests. Merge metadata, if any, is secondary.

PR #649, opened 2026-08-04
fix(examples): correct bleu4 and rouge metric parameter order

### LINKED ISSUE #646: BLEU-4 metric receives predictions and ground truth in reversed order (federated-llm/fedllm-peft)
**What happened**:

In `examples/federated-llm/fedllm-peft/testenv/bleu4_metric.py` the metric is defined as:

    def bleu4_metric(y_pred, y_true, **kwargs):

But the core calls metrics positionally as `metric_func(test_dataset.y, paradigm_result)` — ground truth first, prediction second (`core/testcasecontroller/testcase/testcase.py`, line 117).

So `y_pred` actually receives the ground truth and `y_true` receives the prediction. BLEU is then computed with predictions and references swapped. Since BLEU is asymmetric, the reported score is wrong.

BLEU-4 should be computed with predictions and references in the correct order, matching how the core passes the data.

Look at the function signature in `testenv/bleu4_metric.py` and compare it with the positional call in `core/testcasecontroller/testcase/testcase.py` line 117. The parameter order is reversed.

The three ROUGE metrics in the same folder share the same reversed signature, but ROUGE is symmetric so their values are unaffected. Only BLEU is impacted. The fix is one line — swapping the two parameters:

    def bleu4_metric(y_true, y_pred, **kwargs):

Related: #492.

ahsan-zamil:
**What type of PR is this?**
/kind bug

Correct the parameter signature of custom metrics in `examples/federated-llm/fedllm-peft/testenv/` from `(y_pred, y_true)` to `(y_true, y_pred)`. 

The core framework (`core/testcasecontroller/testcase/testcase.py` line 117) calls metric functions positionally as `metric_func(test_dataset.y, paradigm_result)` — ground truth first, predictions second. Because of the reversed signatures, `y_pred` received the ground truth and `y_true` received predictions. 

Since BLEU is asymmetric, this resulted in incorrect BLEU-4 scores. The three ROUGE metrics in the same folder shared the same reversed signature but were unaffected due to ROUGE being symmetric. We corrected all four metrics for correctness and consistency.

Fixes #646

No


blackdragoon26:
Hi,

The argument orientation is corrected, but sample pairing still depends on dictionary insertion order. With truth keys inserted as `a,b` and prediction keys as `b,a`, the callback pairs `prediction-B` with `reference-A`. Please validate matching key sets and build both lists from one shared key order. This overlaps #770, so only one implementation should proceed.

Links: [Issue #646](https://github.com/kubeedge/ianvs/issues/646) · [#649 callback](https://github.com/kubeedge/ianvs/blob/c24707654c5f765affceaf78936f9432465d6ad0/examples/federated-llm/fedllm-peft/testenv/bleu4_metric.py#L22-L26)

PR #650, opened 2026-08-04
fix(examples): resolve KeyError and ZeroDivisionError in cloud-edge LLM testenv

### LINKED ISSUE #338: KeyError crashes in cloud-edge-collaborative-inference-for-llm benchmark
### Description
When running the cloud-edge-collaborative-inference-for-llm benchmark, the system crashes with KeyError exceptions during metric computation, preventing successful benchmark completion.

### Error Messages
```
KeyError: 'completion'
KeyError: 'usage'
```

### Root Cause
1. **Response parsing issue**: `Response.from_dict()` in `result_parser.py` uses direct dictionary access (`response["key"]`) which throws KeyError when response structure is incomplete or has missing keys
2. **Division by zero**: `throughput.py` calculates `1 / average_itl` without checking if `average_itl` is zero, causing ZeroDivisionError when internal_token_latency metrics are not populated

### Steps to Reproduce
1. Set up the cloud-edge-collaborative-inference-for-llm benchmark
2. Run: `ianvs -f benchmarkingjob.yaml`
3. Observe KeyError or ZeroDivisionError during inference/metric computation

### Expected Behavior
- Benchmark should run successfully even when response structures are incomplete
- Metrics should handle zero values gracefully
- Should provide default/fallback values instead of crashing

### Environment
- **Ianvs version**: main branch (latest)
- **Example**: cloud-edge-collaborative-inference-for-llm
- **Python version**: 3.12
- **OS**: Linux

### Impact
- Prevents running the benchmark successfully
- Blocks testing of LLM inference algorithms
- Affects reproducibility of results

---

ahsan-zamil:
**What type of PR is this?**
/kind bug

Guard the LLM collaborative inference benchmark evaluation metrics against KeyError and ZeroDivisionError crashes:
1. Refactored `Response.from_dict` in `result_parser.py` to safely parse incomplete response structures using `.get()` with safe fallback defaults.
2. Added list-length and zero-latency guards to `throughput.py`, `internal_token_latency.py`, `time_to_first_token.py`, `edge_ratio.py`, and `accuracy.py` to prevent ZeroDivisionErrors when prediction lists are empty or latencies are zero.
3. Standardized line endings and module docstrings to ensure a perfect 10.00/10 pylint score.

Fixes #338

**Does this PR introduce a user-facing change? ****:
No


31groot:
### Request changes : `accuracy()` can overstate scores on partial predictions

**Review depth: Analysis**

The new empty-prediction guard fixes the `ZeroDivisionError`, but there is still a correctness issue when the number of predictions differs from the number of ground-truth labels.

The comparison currently uses:

```python id="r6h4mp"
same_elements = [
    y_pred_list[i] == y_true_list[i]
    for i in range(len(y_pred_list))
]
```

Because the loop is based on `len(y_pred_list)`, missing predictions are simply ignored.

I reproduced this against PR head `7d8f35e`:

```text id="k1c8vz"
y_true length: 2
y_pred length: 1
reported accuracy: 100.0
LENGTH-MISMATCH ACCURACY BUG REPRODUCED
```

With two expected answers (`A`, `B`) and only one prediction (`A`), the metric
reports **100%**, even though only 1/2 predictions were produced.

This means partial results can still produce an over-optimistic benchmark score,
despite the PR's goal of handling incomplete evaluation results.

**Recommendation:** explicitly handle length mismatches and account for missing
predictions in the accuracy calculation. Please add a regression test covering
fewer predictions than ground-truth labels.

**Verification:** Reproduced locally against the exact PR head `7d8f35e` using
the real `accuracy.py` implementation. The full benchmark was not run.

PR #651, opened 2026-08-04
fix(examples): declare __all__ as a list instead of a parenthesised string

akshita317:
**What type of PR is this?**

/kind bug
/kind cleanup

Eighteen example modules declare their public name like this:

```python
__all__ = ('accuracy')
```

Parentheses without a trailing comma do not build a tuple — `('accuracy')` is just the string `"accuracy"`. Since strings are iterable, `from module import *` walks it **character by character**, looking for names `'a'`, `'c'`, `'c'`, `'u'`, `'r'`, `'a'`, `'c'`, `'y'`, each of which raises `AttributeError`:

```python
>>> __all__ = ('accuracy')
>>> type(__all__), list(__all__)
(<class 'str'>, ['a', 'c', 'c', 'u', 'r', 'a', 'c', 'y'])
```

The intended form is `('accuracy',)` or, as used here for readability, `["accuracy"]`.

**Scope — what this does *not* claim**

Ianvs is unaffected today, and I would rather say so plainly than overstate this. `core/common/utils.py` loads these modules with `importlib.import_module`, which does not consult `__all__`, and `get_metric_func` then resolves the metric through its `ClassFactory` alias rather than the module namespace. So:

- no benchmark result changes
- no example that runs today stops running

This is a latent defect, not a live failure. It still misrepresents each module's public surface, breaks `import *` for anyone who tries it, and is exactly the class of thing a static-analysis gate over the examples ought to be able to depend on.

**How it was found**

With an AST sweep over every `.py` file in the tree rather than by hand, so the set is complete. After this change, no module binds `__all__` to a string except one, noted below.

None open for these files. Related in kind to #641 / PR #642, which corrects a `__all__` name mismatch in `Cloud_Robotics/.../cloud_model.py` — I have deliberately left that file untouched to avoid conflicting with that PR.

Each occurrence is replaced with a single-element list naming the symbol the module already defines. **No function, class, or `ClassFactory` registration is renamed**, so there is no behavioural change anywhere.

One file is deliberately excluded: `examples/imagenet/multiedge_inference_bench/testenv/peak_memory.py` declares `__all__ = ('peak_memory')` but the function it defines is named `peak_power` (it registers under `alias="peak_memory"` and returns `peak_mem`, so it does compute peak memory — the function name looks like a copy-paste from `peak_power.py`). Correcting that is a naming decision rather than a syntax fix, so I left it for you. Happy to follow up with it, or with the other `__all__` name mismatches the sweep turned up, if you would like them handled — separately I found roughly twenty more files where `__all__` names a symbol that does not exist in the module, and I can open an issue listing them if that would be useful.

**Testing**

Every modified file still parses (`ast.parse`), and the AST sweep confirms the only remaining string-valued `__all__` is the `peak_memory.py` case described above. The change is one line per file and touches no runtime path.


akshita317:
/assign @MooreZheng
/cc @hsj576

Gentle ping on this one — it has been open since early August without a review, and it should be quick to check.

**What it does:** 18 example modules write `__all__ = ('accuracy')`. Without a trailing comma that is a plain string, not a tuple, so `from module import *` iterates it character by character. Each one becomes a single-element list.

**Why it is safe:** Ianvs loads these modules through `importlib.import_module`, which never consults `__all__`, and metrics resolve via the `ClassFactory` alias rather than the module namespace. So no benchmark result changes — this is a latent correctness fix, not a behavioural one.

**Scope:** only the broken string form is touched. The 17 places that already write a correct single-element tuple (`('TaskRemodeling',)`) are left alone, so there is no style churn mixed in.

One deliberate exclusion, in case it looks like an oversight: `examples/imagenet/multiedge_inference_bench/testenv/peak_memory.py` also uses the string form, but its `__all__` names `peak_memory` while the function it defines is called `peak_power`. Fixing it mechanically would make `__all__` point at a name the module does not define, so I left it for a maintainer to decide whether the function should be renamed. Happy to include it in this PR or open a separate one, whichever you prefer.

DCO passes and the branch merges cleanly.

henryng15 (commented):
Reviewed at head `793718f` against `37a9c60`.

The Python is correct: `__all__ = ('accuracy')` is the string `"accuracy"`, not a tuple,
and `from module import *` would walk it character by character. The 18 modules are worth
tidying and I would merge this.

One correction to the rationale, because it affects how the PR should be labelled.

**Nothing star-imports these modules, and Ianvs never reads `__all__`.** Checked at
`37a9c60`:

```text
star-imports in core/                          : 0
references to __all__ anywhere in core/        : NONE
Example modules with a malformed __all__       : 30
  of those, ever consumed by `from X import *` : 0
```

Core resolves metric and algorithm modules by ClassFactory alias, not by `__all__`:

```python
# core/testcasecontroller/metrics/metrics.py:172-175
load_module(url)
metric_func = ClassFactory.get_cls(type_name=ClassType.GENERAL, t_cls_name=name)
```

`core/testcasecontroller/algorithm/module/module.py:116` and `:130` do the same. There is
no code path on this commit through which the described failure can occur.

So the described bug is real Python but not a reachable Ianvs defect, and this PR changes
no runtime behaviour. That is worth knowing for two reasons: it means the change carries
**zero regression risk** (a genuine argument for merging it), and it means the PR is
`kind/cleanup` rather than `kind/bug`, which currently affects how it is prioritised
against PRs that do change behaviour.

Reproduce: `python3 tools/probe_all_is_inert.py ianvs` —

**Recommendation: minor revision** — no code change needed; re-label and adjust the
description so the next reader is not looking for a runtime symptom that cannot appear.

PR #652, opened 2026-08-04
ci(pipeline): modernize python matrix, update actions, and add example validation

### LINKED ISSUE #411: [CI/CD] Modernize CI Pipeline: Update Python Matrix, Actions Versions, and Add Example Validation
## Problem

The current CI pipeline (`main.yaml`) has several structural issues that reduce its effectiveness as a quality gate for the project.

### 1. EOL Python Versions in Test Matrix

The CI tests against Python 3.7, 3.8, and 3.9:

| Version | EOL Date | Status |
|---------|----------|--------|
| Python 3.7 | June 2023 | EOL |
| Python 3.8 | October 2024 | EOL |
| Python 3.9 | October 2025 | EOL |

Meanwhile, `setup.py` specifies `python_requires=">=3.6"` (Python 3.6 reached EOL in December 2021).

Testing against EOL interpreters means CI passes on versions that no longer receive security patches, while compatibility with currently-supported Python versions (3.10, 3.11, 3.12) is never validated.

### 2. Outdated GitHub Actions

- `actions/checkout@v3` — current stable release is `v4` (Node.js 16 runtime in v3 has been deprecated by GitHub)
- `actions/setup-python@v3` — current stable release is `v5`

GitHub has been issuing deprecation warnings for Node.js 16-based actions since September 2023.

### 3. No Example Validation

The CI only runs `pylint` on `core/`. There is no validation that:
- Example YAML configuration files are parseable
- Example Python files have no syntax errors
- Core `ianvs` package can be imported successfully after installation

This means PRs that introduce broken imports, malformed YAML configs, or syntax errors in examples merge without any CI detection. This directly contributes to the pattern documented in #230 where historical examples accumulate silent failures over time.

To illustrate, running a simple YAML validation across all 37 `benchmarkingjob.yaml` files locally reveals that `examples/cifar100/fci_ssl/fedavg/benchmarkingjob.yaml` has a pre-existing syntax error on line 17 (`algorithms:conda` instead of `algorithms:`). This error exists on `main` today and would have been caught by an automated validation step.

### 4. Hardcoded pip Version

Line 29 pins `pip==24.0` explicitly. This prevents pip from receiving bug fixes and security patches, and will become increasingly outdated as the pip project releases new versions.

## Proposed Changes

1. **Update Python matrix** from `["3.7", "3.8", "3.9"]` to `["3.9", "3.10", "3.11", "3.12"]`
   - 3.9 is retained as the minimum supported version for backward compatibility with existing user environments, even though it reached EOL in October 2025
   - 3.10 through 3.12 are currently supported by the Python Software Foundation
2. **Update `python_requires`** in `setup.py` from `">=3.6"` to `">=3.9"`
3. **Update GitHub Actions** to current stable versions (`checkout@v4`, `setup-python@v5`)
4. **Replace hardcoded pip pin** with `pip install --upgrade pip`
5. **Simplify pylint invocation** — since Python 3.9 is now the minimum, the latest pylint (which supports `--max-positional-arguments`) will be installed across all matrix entries, so the conditional workaround from #157 can be applied universally
6. **Add an example validation job** that:
   - Validates all example `benchmarkingjob.yaml` files are parseable
   - Checks that example Python files have no syntax errors
   - Verifies core `ianvs` package can be imported after installation
7. **Fix the pre-existing YAML syntax error** in `examples/cifar100/fci_ssl/fedavg/benchmarkingjob.yaml`

## Relationship to #230

This addresses the CI pipeline component of the Comprehensive Example Restoration initiative:

> "Build a CI pipeline testing examples with GitHub Actions against multiple Python versions, critical Ianvs/upstream updates, and block PRs that break validated examples"

Happy to submit a PR for this.

priyam99:
Fixes #411

Modernizes the CI pipeline with active Python versions 3.9-3.12, updates actions to checkout@v4 and setup-python@v5, upgrades pip, and adds automated example validation to prevent broken configs from merging into main.


priyam99:
This one is worth a maintainer opinion before it goes further, since it changes CI policy rather than just fixing a bug.

It moves the pylint matrix from 3.7-3.9 to 3.9-3.12 and bumps `python_requires` in setup.py accordingly, so it's effectively a decision to drop 3.7/3.8 support. If that's not a change you want yet, I can drop that part and keep only the action version bumps and the example validation script.

The validation script itself is deliberately minimal — it only checks that example YAML parses and example Python compiles. It won't catch runtime or config errors, but it would have caught the malformed `algorithms:conda` line in the fedavg benchmarkingjob.yaml that's also fixed here.

/assign @moorezheng

priyam99:
Following up on my earlier note — I've narrowed this PR so it no longer touches Python support policy.

Removed from this PR:
- the matrix change (3.7/3.8/3.9 → 3.9–3.12)
- the `python_requires` bump in setup.py

`setup.py` is now unchanged, and the matrix, the `pip==24.0` pin (needed for 3.7) and the version-conditional pylint invocation are all restored to their current values on main.

What's left is only the non-policy part: `checkout@v3→v4`, `setup-python@v3→v5`, the example syntax validator, and the malformed `algorithms:conda` key in the fedavg benchmarkingjob.yaml.

Dropping 3.7/3.8 is a maintainer decision rather than mine to make in a CI cleanup, so I'd rather raise it separately under #411 if and when you want it. Happy to open that as its own PR.

Yash4616:
Reviewed against `main` at `37a9c60`. This one conflicts against `main` as it stands.

Three separate changes are bundled here: the action version bumps, a whitespace fix in a pip invocation, and a new `Validate Ianvs Examples` step calling `.github/scripts/validate_examples.py`.

On the action bumps, `checkout@v3` to v4 and `setup-python@v3` to v5: `main` currently uses `actions/checkout@v7` and `actions/setup-python@v7` throughout, so those two lines would move the workflow backwards. They can be dropped without affecting the rest of the PR.

On the validator, this is the part worth a maintainer decision. There is already a validation framework merged under `.github/workflows/validator/`, and two other open PRs each add a further example-config validator in a different location. That makes four implementations of broadly the same check in four places. Before adding a fourth entry point at `.github/scripts/`, it is worth establishing whether this extends the merged framework or replaces part of it.

I also checked how this interacts with the other open work in this area. It is the most entangled of the set, conflicting with six of the seven other open CI PRs, all in `static_code_pylint_cicd.yaml.yaml`.

Recommendation: major revision. Drop the action bumps, and reconcile the validator against the merged framework before landing a fourth implementation.

PR #655, opened 2026-08-05
feat(llm): enable model quantization support in vllm backend

### LINKED ISSUE #372: Quantization parameter not working in vLLM backend (cloud-edge LLM example)
While going through `vllm_llm.py`, I noticed the quantization parameter is commented out with a TODO:

```python
#quantization=self.quantization # TODO need to align with vllm API
Looked into why this was disabled - BaseLLM._parse_kwargs() never parses quantization from kwargs, so self.quantization is undefined when VllmLLM._load() tries to use it.

Other examples like examples/PIPL/edge-cloud_collaborative_learning_bench already use quantization in their configs, and the original proposal also mentions it as a planned feature.

Checked the vLLM docs - the quantization parameter works directly with values like bitsandbytes, awq, gptq.

Fix:

Parse quantization in BaseLLM._parse_kwargs()
Conditionally pass it to vLLM in VllmLLM._load()
Happy to submit a PR for this.

priyam99:
Fixes #372

Parses the quantization keyword argument in BaseLLM and passes it to the vLLM engine initialization in VllmLLM to enable model quantization methods (AWQ, GPTQ, BitsAndBytes, FP8) in cloud-edge LLM benchmarks.


priyam99:
Small change — the `quantization` kwarg was being accepted by `BaseLLM` but never threaded through to the vLLM engine init, so AWQ/GPTQ/BitsAndBytes/FP8 configs had no effect. This passes it through.

Let me know if you'd like the supported methods validated explicitly rather than passed straight to vLLM.

/assign @moorezheng

Shikhar-Kesharwani:
### Review — supplementary configuration-validation finding

The `quantization` value is now correctly threaded from `BaseLLM` into the vLLM engine, which fixes the original issue where the parameter was accepted but ignored.

One additional concern is that `BaseLLM._parse_kwargs()` accepts any string as `quantization`, while `VllmLLM._load()` passes that value directly to the installed vLLM `LLM` constructor without validating the supported quantization modes.

This means a typo or unsupported value such as `awqq`, `gptq4`, or a method unavailable in the installed vLLM version is not rejected at Ianvs configuration parsing time. The resulting failure is deferred until model construction, where the backend-specific exception may be much less clear and can occur after other benchmark setup has already happened.

Because this is a benchmark configuration parameter, I recommend validating the accepted values at the Ianvs boundary (for example, against the set supported by the repository's targeted vLLM version) and producing a clear configuration error for unsupported values. If the supported set intentionally varies by vLLM version, the validation should at least reject empty/unknown values and document that backend-specific validation is delegated to vLLM.

This is separate from the bug fixed by this PR: the current patch ensures the configured value reaches vLLM; the remaining concern is that invalid configuration values are still accepted by Ianvs and only fail later during backend initialization.

PR #588, opened 2026-07-29
fix(core/lifelong): guard task_avg_score division against ZeroDivisionError

ajmani-x:
## What does this PR do?
Guards the four `task_avg_score['accuracy'] / i` division sites in
`core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py`
(no-inference and hard-example-mining modes) against ZeroDivisionError when
`tasks_detail` is empty.

## Why is this important?
When `job.evaluate()` returns no task-level details (e.g. round 0 before the
task knowledge base is populated), the counter `i` stays 0 and the division
crashes with `ZeroDivisionError: float division by zero`. This blocks any
lifelong learning benchmark that hits an empty-detail round, including
robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation.

## How to test?
Reproduced the bug pattern with a minimal script mirroring the loop
(`tasks_detail = []` → division by `i=0` crashes; confirmed the guarded
version returns `0.0` instead). Ran `python -m py_compile` on the changed
file; no other callers of these code paths were affected.

## Issues fixed
Closes #462

## Type of change
- [x] Bug fix

## Checklist
- [x] I've tested this locally
- [x] My code follows the project style

Issue #285, opened 2025-11-06
[Core Framework] ValueError in evaluate() - incorrect return value unpacking


abhishek-8081:
## Description
The `evaluate()` method returns a single dictionary, but the code attempts to unpack it into three variables, causing a `ValueError` that blocks evaluation in all lifelong learning examples.

## Environment
- **File:** `core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py`
- **Lines:** 108, 419
- **Severity:** Critical
- **Affects:** All lifelong learning examples

## Error
```python
ValueError: not enough values to unpack (expected 3, got 1)
```

## Root Cause

**Current code (incorrect):**
```python
# Lines 108 and 419
edge_task_index, tasks_detail, res = job.evaluate(eval_dataset, metrics=metric_func)
```

The `evaluate()` method returns a single dictionary:
```python
{"accuracy": 0.012, "mIoU": 0.010, "fwIoU": 0.018}
```

Attempting to unpack one value into three variables causes the ValueError.

## Solution

**Proposed fix:**
```python
# Lines 108 and 419
res = job.evaluate(eval_dataset, metrics=metric_func)
```

**Additional fix required (lines 120-129):**
```python
# Current (broken):
for detail in tasks_detail:  # tasks_detail is undefined after fix
    ...

# Fixed:
task_avg_score['accuracy'] = res.get('accuracy', 0.0) if isinstance(res, dict) else 0.0
```

## Testing
- ✅ **Verified on:** Robot Semantic Segmentation example
- ✅ **Environment:** Ubuntu 24.04, Python 3.12
- ✅ **Result:** Evaluation completes successfully, metrics calculated correctly

## Impact
**Before:** Evaluation fails, no metrics generated, benchmarking incomplete  
**After:** Full evaluation pipeline works, all metrics calculated

## Affected Examples
- Robot Semantic Segmentation
- Cityscapes-Synthia Semantic Segmentation
- Cityscapes-Synthia Curb Detection
- Robot-Cityscapes Class Increment Learning
- Robot SAM Annotation

## Context
Discovered during LFX Mentorship #48: "Comprehensive Example Restoration for Ianvs"

**Note:** This issue is independent from but related to the method name error at line 419. Both must be fixed for evaluation to work.

---

PR #657, opened 2026-08-05
fix(core): resolve my_evaluate/my_inference AttributeError and ZeroDivisionError in lifelong learning paradigm

### LINKED ISSUE #284: [Core Framework] AttributeError: 'LifelongLearning' object has no attribute 'my_evaluate'
## Description
The lifelong learning framework contains a critical typo in the evaluation method name. The code calls `my_evaluate()` which does not exist, instead of the correct method `evaluate()`. This bug **prevents ALL lifelong learning examples from completing evaluation**, regardless of the specific algorithm or dataset used.

## Affected Component
- **File:** `core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py`
- **Line:** 419
- **Component:** Core Lifelong Learning Framework
- **Framework Version:** Current main branch

## Severity
🔴 **CRITICAL** - Blocks execution of all lifelong learning paradigm examples

## Affected Examples
This bug affects **ALL** examples using the lifelong learning paradigm:

| Example | Path | Paradigm |
|---------|------|----------|
| Robot Semantic Segmentation | `examples/robot/lifelong_learning_bench/semantic-segmentation/` | ✅ Lifelong Learning |
| Cityscapes-Synthia Semantic Segmentation | `examples/cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/` | ✅ Lifelong Learning |
| Cityscapes-Synthia Curb Detection | `examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/` | ✅ Lifelong Learning |
| Robot-Cityscapes Class Increment | `examples/robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/` | ✅ Lifelong Learning |
| Robot SAM Annotation | `examples/robot/lifelong_learning_bench/sam_annotation/` | ✅ Lifelong Learning |

**Why ALL examples are affected:** Every example configured with `paradigm_type: lifelong_learning` imports and executes code from `lifelong_learning.py` line 419.

## Error Message
```python
  File "core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py", line 419
    edge_task_index, tasks_detail, res = job.my_evaluate(eval_dataset, metrics=metric_func)
AttributeError: 'LifelongLearning' object has no attribute 'my_evaluate'
```

### 1. Setup
```bash
# Clone repository
cd ianvs

# Install Ianvs
python -m pip install -e .
```

### 2. Run ANY Lifelong Learning Example
```bash
# Example: Robot Semantic Segmentation
cd examples/robot/lifelong_learning_bench/semantic-segmentation/

# Download dataset (follow example README)
# ...

# Run benchmark
ianvs -f benchmarkingjob-simple.yaml
```

### 3. Observe Failure
The benchmark will execute training successfully but **fail during evaluation phase** with the error shown above.

### Current Code (Broken)
```python
# File: core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py
# Line: 419

edge_task_index, tasks_detail, res = job.my_evaluate(eval_dataset, metrics=metric_func)
```

**Problem:** The method name `my_evaluate` does not exist in the `LifelongLearning` class or its parent classes.

### Correct Method Name
The actual method defined in the codebase is `evaluate()`, not `my_evaluate()`.

### Fix
```python
# File: core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py
# Line: 419

# Change from:
edge_task_index, tasks_detail, res = job.my_evaluate(eval_dataset, metrics=metric_func)

# To:
res = job.evaluate(eval_dataset, metrics=metric_func)
```

**Note:** This also fixes a related unpacking issue (the method returns 1 value, not 3).

### Tested On
- ✅ **Robot Semantic Segmentation Example**
  - Environment: Ubuntu 24.04, Python 3.12, CPU-only
  - Training: SUCCESS
  - Evaluation: SUCCESS (after fix)
  - Metrics calculation: SUCCESS

### Expected Results for Other Examples
Since this is a framework-level typo, the fix should work identically for all lifelong learning examples.

### Before Fix
- ❌ **0%** of lifelong learning examples can complete evaluation
- ❌ No metrics can be calculated
- ❌ Benchmarking process fails at evaluation phase
- ❌ Single-round execution incomplete
- ❌ Multi-round execution impossible

### After Fix
- ✅ **100%** of examples can complete evaluation (verified on Robot example)
- ✅ Metrics calculation works correctly
- ✅ Benchmarking completes successfully
- ✅ Single-round execution: FULL SUCCESS
- ✅ Multi-round execution: Possible (tested on Robot example)

## Additional Context

**Mentor:** Zimu Zheng 
**Developer:** Abhishek Kumar

### Related Issues
- This may be related to #132 (LifeLongLearning algorithm unable to run)
- Part of comprehensive debugging that identified 14 total bugs in lifelong learning framework

### Files Modified (Proposed)
```
core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py
  - Line 419: Change method name from my_evaluate to evaluate
  - Line 419: Fix return value unpacking (3 values → 1 value)
```

## Proposed PR

A pull request will be submitted with:
- ✅ Fix for this specific bug
- ✅ Test results from Robot example
- ✅ Code review ready
- ✅ Link to this issue

### LINKED ISSUE #461: [Core Framework][no-inference] AttributeError: 'LifelongLearning' object has no attribute 'my_inference'
## Description

The `no-inference` mode of the lifelong learning paradigm controller calls `job.my_inference()` at line 158 of `lifelong_learning.py`. This method does not exist in the Sedna `LifelongLearning` API. The call raises an `AttributeError` that prevents the final test inference from ever running, meaning no test results are produced even when all preceding training and evaluation rounds complete successfully.

This bug is the **third sequential crash** in the `no-inference` mode cascade and only becomes visible after fixing #284 and #285. Because it requires two prior fixes to surface, it has not been reported before.

## Affected Component

- **File:** `core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py`
- **Line:** 158
- **Mode:** `no-inference` only
- **Example:** `robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation`

## Affected Examples

Any lifelong learning example running in `no-inference` mode. Currently, `robot-cityscapes-synthia` is the only active example using this mode.

## Error

```
AttributeError: 'LifelongLearning' object has no attribute 'my_inference'
  File "core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py", line 158
    test_res = job.my_inference(inference_dataset, **kwargs)
```

## Root Cause

**Offending code (`lifelong_learning.py:154-158`):**

```python
job = self.build_paradigm_job(ParadigmType.LIFELONG_LEARNING.value)
inference_dataset = self.dataset.load_data(self.dataset.test_url, "eval",
                                           feature_process=_data_feature_process)
kwargs = {}
test_res = job.my_inference(inference_dataset, **kwargs)  # line 158 — method does not exist
```

The Sedna `LifelongLearning` class (v0.6.0.1, `sedna/core/lifelong_learning/lifelong_learning.py`) exposes exactly four public methods:

```
train      (line 217)
update     (line 284)
evaluate   (line 366)
inference  (line 414)
```

There is no `my_inference` method. The name `my_inference` follows the same erroneous pattern as `my_evaluate` in issue #284, where a `my_` prefix was added that does not match the Sedna API.

## Comparison: correct pattern vs incorrect pattern

```python
# lifelong_learning.py:392 — correct pattern in _eval() (normal mode)
edge_task_index = job.evaluate(eval_dataset, metrics=metric_func)

# lifelong_learning.py:419 — incorrect pattern in my_eval() (no-inference mode)
edge_task_index, tasks_detail, res = job.my_evaluate(eval_dataset, metrics=metric_func)
# ^^^^^^^^^^^^ fixed by #284 / PR #301

# lifelong_learning.py:158 — same pattern, inference call (no-inference mode)
test_res = job.my_inference(inference_dataset, **kwargs)
# ^^^^^^^^^^^^ THIS ISSUE — not fixed by any existing PR
```

## Proposed Fix

Replace `job.my_inference` with `job.inference`:

```python
# lifelong_learning.py:158 — corrected
test_res = job.inference(inference_dataset, **kwargs)
```

The `inference()` method in Sedna accepts the full dataset in one call, which matches the calling convention already used in the no-inference mode body (passing `inference_dataset` as a complete object, not per-sample).

## Why This Was Not Caught by Existing PRs

PRs #286, #373, and #301 all address the `my_evaluate` bug (#284) and the value-unpacking bug (#285). None of them reach line 158 because the cascade stops at line 419 (the `my_evaluate` call) before reaching the final inference block. This bug only surfaces after both prior fixes are applied.

## Related Issues and PRs

| Reference | Relation |
|-----------|----------|
| #284 | Preceding crash in the same cascade (my_evaluate) |
| #285 | Preceding crash in the same cascade (unpack mismatch) |
| #301 | Open PR fixing #284 and #285 — must be merged before this bug is reachable |
| #79  | Root tracking issue for robot-cityscapes-synthia example |

### LINKED ISSUE #462: [Core Framework][no-inference] ZeroDivisionError: division by zero in task average accuracy computation

### LINKED ISSUE #465: [robot-cityscapes-synthia] No baseline performance published — BWT, FWT, MATRIX, and task_avg_acc metrics never validated end-to-end
## Description

The `robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation` example configures six evaluation metrics in `testenv.yaml`:

```yaml
metrics:
  - name: "accuracy"
  - name: "samples_transfer_ratio"
  - name: "task_avg_acc"
  - name: "BWT"
  - name: "FWT"
  - name: "MATRIX"
```

Of these, \`BWT\`, \`FWT\`, \`MATRIX\`, \`task_avg_acc\`, and \`samples_transfer_ratio\` are Ianvs-specific lifelong learning system metrics defined in \`core/testcasecontroller/metrics/metrics.py\`. They are designed to measure forward and backward knowledge transfer across incremental training rounds.

**None of these metrics have ever been computed for this example.** The Ianvs lifelong learning leaderboard contains zero entries for the MDIL-SS 3-domain benchmark.

## Direct Evidence — Leaderboard Verified Empty

The published leaderboard at \`docs/leaderboards/leaderboard-in-semantic-segmentation-of-Cloud-Robotics/leaderboard-of-lifelong-learning.md\` contains exactly one row:

| rank | algorithm | accuracy | BWT | FWT | task_definition-origins | task_allocation-origins |
|------|-----------|----------|-----|-----|------------------------|------------------------|
| 1 | rfnet_lifelong_learning | 0.5206 | 0.0424 | 0.0230 | ['front', 'garden'] | ['front', 'garden'] |

This result is from \`examples/robot/lifelong_learning_bench\` (2-domain, \`rfnet\`), **not** from \`robot-cityscapes-synthia\`. The MDIL-SS example (\`erfnet_lifelong_learning\`, origins: \`["Cityscapes", "Synthia", "Cloud-Robotics"]\`) has no entry in the leaderboard. No accuracy, BWT, or FWT value has ever been recorded for it.

## Root Cause

The example has never successfully completed an end-to-end run due to a sequential cascade of bugs that block execution before the metrics computation stage is reached:

| Bug | Issue | Blocks |
|-----|-------|--------|
| 8 broken YAML configuration paths | [#79](https://github.com/kubeedge/ianvs/issues/79) | Launch — Ianvs cannot start |
| \`job.my_evaluate()\` does not exist in Sedna API | [#284](https://github.com/kubeedge/ianvs/issues/284) | All evaluation in no-inference mode |
| Incorrect 3-value unpack of \`job.evaluate()\` return | [#285](https://github.com/kubeedge/ianvs/issues/285) | Metric score collection |
| \`job.my_inference()\` does not exist in Sedna API | [#461](https://github.com/kubeedge/ianvs/issues/461) | Final test inference |
| ZeroDivisionError in task average accuracy | [#462](https://github.com/kubeedge/ianvs/issues/462) | task_avg_acc computation |

Until all of the above are fixed, the benchmark cannot reach the metrics computation step.

## Why This Matters Beyond the Example

\`BWT\`, \`FWT\`, and \`MATRIX\` metrics are shared infrastructure used by every lifelong learning example in the repository. \`robot-cityscapes-synthia\` is the designated reference implementation for these metrics. If this example never runs, there is no validation that the metric computation code in \`core/testcasecontroller/metrics/metrics.py\` produces correct output for any 3-domain incremental scenario.

The only existing leaderboard entry uses 2 domains and a different algorithm. It does not exercise the BWT/FWT computation paths that involve 3-domain cross-evaluation matrices.

## Expected Outcome

After fixing issues #79, #284, #285, #461, and #462, a successful end-to-end run should:

1. Produce the first leaderboard entry for \`erfnet_lifelong_learning\` with Cityscapes/Synthia/Cloud-Robotics origins.
2. Publish baseline values for \`accuracy\`, \`BWT\`, \`FWT\`, \`MATRIX\`, \`task_avg_acc\`, and \`samples_transfer_ratio\`.
3. Serve as the first end-to-end validation of the BWT/FWT metric infrastructure for a 3-domain lifelong learning benchmark.

## Reference

MDIL-SS paper (WACV 2022): *"Multi-Domain Incremental Learning for Semantic Segmentation"*
Published accuracy values for Cityscapes, SYNTHIA, and Cloud-Robotics domains are available in the paper and can serve as validation targets once the example runs end-to-end.

## Related Issues

| Reference | Relation |
|-----------|----------|
| [#79](https://github.com/kubeedge/ianvs/issues/79) | Path failures block launch |
| [#284](https://github.com/kubeedge/ianvs/issues/284) | Must fix before metrics can be collected |
| [#285](https://github.com/kubeedge/ianvs/issues/285) | Must fix before metrics can be collected |
| [#461](https://github.com/kubeedge/ianvs/issues/461) | Must fix before final inference runs |
| [#462](https://github.com/kubeedge/ianvs/issues/462) | Must fix before task_avg_acc computes |

ahsan-zamil:
**What type of PR is this?**
/kind bug

Resolves several critical core framework blocker bugs in the lifelong learning paradigm:
1. Adds `hasattr(job, "my_inference")` check to dynamically fall back to `job.inference()` if `my_inference()` is not implemented in the installed version of Sedna.
2. Adds `hasattr(job, "my_evaluate")` check to dynamically fall back to `job.evaluate()` if `my_evaluate()` is not implemented.
3. Guards against `ZeroDivisionError` in the task average accuracy calculation when `i == 0` during accuracy aggregation.

These fixes resolve the core blocker chain preventing lifelong learning examples (such as `robot-cityscapes-synthia`) from completing end-to-end execution and collecting baseline metrics.

Fixes #465
Fixes #284
Fixes #461
Fixes #462


AdityaKumarSethia:
Reviewed against the same #462 crash cluster covered in #588, #496, and #755 (all open, cross-analysis in [Discussion #873](https://github.com/kubeedge/ianvs/discussions/873)).

This PR's `hasattr(job, "my_inference")` / `hasattr(job, "my_evaluate")` fallback looks like a reasonable, separate fix for a different bug (an `AttributeError` when a job object doesn't implement the expected method name), not evaluated further here since it's outside this review's scope.

On the `ZeroDivisionError` specifically: this PR guards 1 of the 4 identical crash sites in `lifelong_learning.py` (the one at `main`'s line 151). The other 3 (lines 129, 227, 250 on `main`) still have the exact unguarded `task_avg_score['accuracy']/i` pattern, confirmed present in this PR's own branch via `grep`, and confirmed to still crash via direct execution of the real, unmodified `run()` method (only I/O-heavy dependencies stubbed):

```python
lt = object.__new__(LifelongLearning)
lt.incremental_rounds = 1
lt.system_metric_info = {}
lt.model_eval_config = {'model_metric': {'mode': 'no-inference'}}
lt.cloud_task_index = None
lt._split_dataset = lambda splitting_dataset_times: {0: ('train.txt', 'eval.txt'), 1: ('train.txt', 'eval.txt')}
lt._train = lambda cloud_task_index, train_file, r: cloud_task_index
lt.my_eval = lambda cloud_task_index, eval_file, r: (None, [], {})

lt.run()
```
Output:
```
REAL run() METHOD, CONFIRMED CRASH AT LINE 129: float division by zero
```
Recommendation: if this PR is intended to resolve #462, it needs the same guard applied to the other 3 sites, currently it only fixes the specific site the reporter happened to hit first. Also worth flagging: this makes 4 open PRs now touching some or all of these 4 lines (#588, #496, #755, and this one), with none referencing each other. #496 is currently the most complete (all 4 sites, plus a related `metrics.py` fix); recommend consolidating around that one rather than adding a 5th partial fix.

rajdhruvsingh:
Executed this PR's patched my_eval against a mock constrained to Sedna v0.6.0.1's real API surface (spec=['evaluate','get_parameters','inference','model_path','parameters','report_task_info','train','update']). The hasattr-guard design is forward-compatible in principle — against real Sedna, hasattr(job, "my_evaluate") is False and control reaches the else branch — but the implementation has two independent defects, plus coverage and hygiene problems.

Defect 1 (executed): the else branch keeps the legacy 3-value unpack:
edge_task_index, tasks_detail, res = job.evaluate(eval_dataset, metrics=metric_func)
Real Sedna's evaluate() returns a single value (return callback_func(res) if callback_func else res). Execution through the fallback path raises:
ValueError: not enough values to unpack (expected 3, got 1)
i.e., the guard routes around the original AttributeError into the same class of crash documented in #285.

Defect 2 (diff + static trace; boundary stated): the my_inference else branch assigns without unpacking:
test_res = job.inference(inference_dataset, **kwargs)
run() ends return test_res, self.system_metric_info (line 299), so callers receive ((res, is_unseen_task, tasks), system_metric_info) instead of the flat (res, system_metric_info) every other mode branch produces. Not executed against a live caller of run() — flagged as static-trace evidence, not a confirmed runtime crash.

ZeroDivisionError coverage is 1 of 4 sites. The if i > 0: ... else: 0.0 guard is applied only at the no-inference tail division (main line 151). The inner-loop site (line 129) and both hard-example-mining sites (lines 227, 250 — undocumented in #462 itself) remain unguarded in the patched file. The commit message claims general resolution; the diff supports a quarter of it.

Hygiene: the branch bundles this single-file Core fix with three unrelated commits touching cloud-edge-collaborative-inference-for-llm/ and federated-llm/fedllm-peft/, including new files absent from main (benchmarkingjob-mmlu.yaml, testenv-mmlu.yaml) — worth splitting regardless of the code outcome.

Suggestion for a clean v2: standalone Core PR; single-value else branch (test_res, _, _ = job.inference(...) pattern already proven in this file's history); identical guard at all four division sites.

AdityaKumarSethia:
This is a real and important extension of the finding, thank you. I'd only checked how many of the 4 division sites were guarded; I hadn't traced whether the `hasattr` fallback branch actually works against Sedna's real return signature. A guard that routes an `AttributeError` into a `ValueError` on the very next line is a materially worse outcome than what I'd flagged, since it makes the PR look more complete than it is. Between the two of us this is now a clear case for major revision, not partial credit, appreciate you digging into the Sedna API surface directly rather than assuming the fallback path was safe.

suhaan-24:
Static review against `sedna 0.6.0.1` / Python 3.9.6. I have **not** executed this branch.

## 1. Large overlap with #697

This PR shares **19 changed files** with #697, by the same author, including
`core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py` and most of
the `cloud-edge-collaborative-inference-for-llm` tree:

```
$ comm -12 <(git diff --name-only $(git merge-base upstream/main pr/657) pr/657 | sort) \
           <(git diff --name-only $(git merge-base upstream/main pr/697) pr/697 | sort) | wc -l
19
```

Whichever merges second will conflict on all of them. Worth deciding which PR owns which concern
and rebasing the other, rather than resolving 19 files at merge time.

## 2. The `ZeroDivisionError` guard publishes a fabricated metric

```python
if i > 0:
    task_avg_score['accuracy'] = task_avg_score['accuracy'] / i
else:
    task_avg_score['accuracy'] = 0.0
```

This stops the crash, but `0.0` then flows into `system_metric_info` and onto the leaderboard,
where it is indistinguishable from a genuine measurement of zero accuracy. The `i == 0` case means
*no tasks were scored* — which is a different condition from *tasks scored zero*.

Suggest either propagating `None`/`nan` so the leaderboard can render it as "not measured", or
raising with a message naming the empty task set. Silently reporting `0.0` risks a run being
recorded as a valid poor result when it actually produced no result at all.

## 3. The Sedna method renames

Replacing `my_evaluate` / `my_inference` with `evaluate` / `inference` is correct for
sedna 0.6.0.1 — the installed class exposes exactly:

```
['evaluate', 'get_parameters', 'inference', 'model_path', 'parameters', 'report_task_info', 'train', 'update']
```

Two notes. First, #645 makes the same substitution in the same file, so these will collide.
Second, `my_inference` and `my_evaluate` are reachable from the `no-inference` and
`hard-example-mining` paths, whose semantics differ from the default path — mapping them onto
`inference`/`evaluate` is a behavioural decision, not a mechanical rename, and would benefit from
a maintainer confirming the intended semantics of those two modes rather than being folded in
with the `inference_2` rename (which does look like a straightforward stale name).

## 4. Scope

19 files spanning Core and four Examples in one diff. Splitting the Core change out would let it
be reviewed and merged on its own merits and would remove most of the overlap with #697.

## Verification boundary

**Not executed.** Findings are from the diff, the file-list intersection with #697, and
introspection of the installed sedna class. No claim is made about runtime behaviour of this
branch.

phantom-712:
The `hasattr` idea is useful only if both branches obey the same normalized
contract. Against Sedna 0.6.0.1, the public surface has `evaluate()` and
`inference()`, not the legacy names.

For evaluation, the fallback must not retain the legacy three-value unpack:

```python
result = job.evaluate(eval_dataset, metrics=metric_func)
```

The public method returns one result value. Task-index state and task-detail
state must be obtained through the job/output contract used by the current
Sedna version, then normalized once for the Ianvs controller.

For inference, the public method's tuple is not automatically the same shape
as the value returned by every Ianvs mode. The adapter must explicitly map the
result, unseen-task flag, and task collection to the caller's expected result;
otherwise a branch can return a nested tuple and silently break downstream
metric or reporting code even though the AttributeError disappears.

The task-average guard also needs to cover all four accuracy-aggregation sites,
not only the first traceback location. Returning `0.0` for an empty task set
should be treated deliberately: it is numerically indistinguishable from a
real zero score. A documented `None`/`nan` policy or an explicit "not measured"
path is safer if the result reaches a leaderboard.

Recommendation: major revision. Split unrelated example changes from the Core
adapter, normalize the exact public return shapes, cover all four division
sites, and add tests for absent legacy methods plus empty task details. The repaired
composed Robot RFNet simple configuration was also run against the
official Cloud-Robotics resource. Ianvs returned code 0 after five rounds and
produced metric, matrix, and rank artifacts. This is composition evidence only
and does not prove that PR #657 passed unchanged. The observed accuracy was
`0.002232849365268865` for one configured algorithm, so it is not a comparative
benchmark claim.

PR #658, opened 2026-08-05
fix(llm): prevent keyerror and zero division in collaborative inference metrics

### LINKED ISSUE #338: KeyError crashes in cloud-edge-collaborative-inference-for-llm benchmark

priyam99:
Fixes #338

Refactors Response.from_dict() to use safe dictionary access with default fallbacks and adds zero-division guards across LLM metric functions to prevent crashes during benchmark execution.


priyam99:
This touches 10 metric files in the collaborative inference example. The pattern throughout was the same: `Response.from_dict()` assumed every key was present, and the ratio/throughput/latency metrics assumed a non-zero denominator — both false when the edge model returns an empty or partial completion.

Changes are mechanical and repetitive by nature, so it should read quickly. Happy to split it per-metric if that's easier to review.

/assign @moorezheng

PR #660, opened 2026-08-05
fix: validate rank configuration field types

### LINKED ISSUE #659: Rank config validation accepts invalid field types
**What happened**:

`Rank._check_fields()` currently validates several configuration fields using conditions like:

```python
if not self.sort_by and not isinstance(self.sort_by, list):
```

Because these checks use `and`, invalid non-empty values with the wrong type can pass validation. For example:

```python

Rank({
    "sort_by": {},
    "visualization": {
        "mode": "selected_only",
        "method": "print_table",
    },
    "selected_dataitem": {
        "paradigms": ["all"],
        "modules": ["all"],
        "hyperparameters": ["all"],
        "metrics": ["all"],
    },
    "save_mode": "selected_and_all",
})
```

Here, `sort_by` is a `dict` instead of a `list`, but because it is non-empty, `not self.sort_by` evaluates to `False` and no `ValueError` is raised. The same validation pattern affects `visualization` and `selected_dataitem`.

There is also an inconsistency for `save_mode`: it is used as a string value (e.g. `"selected_and_all"` or `"selected_only"`), but the validation error states that it must be a list.

`Rank._check_fields()` should reject values that are either missing/empty or of the wrong type.

- `sort_by` should be a non-empty list.
- `visualization` should be a dict.
- `selected_dataitem` should be a dict.
- `save_mode` should be a string.

Invalid values should raise `ValueError` immediately instead of being accepted.

Run the following from the repository root:

```python

Rank({
    "sort_by": {},
    "visualization": {
        "mode": "selected_only",
        "method": "print_table",
    },
    "selected_dataitem": {
        "paradigms": ["all"],
        "modules": ["all"],
        "hyperparameters": ["all"],
        "metrics": ["all"],
    },
    "save_mode": "selected_and_all",
})
```

**Expected:** `ValueError` because `sort_by` is not a list.

**Actual:** No validation error is raised.

The same issue can be reproduced by providing incorrect non-empty types for `visualization`, `selected_dataitem`, or `save_mode`.

This appears to be caused by using `and` instead of `or` in the presence/type validation checks. Updating the validation logic and validating `save_mode` as a string instead of a list would resolve the issue. Adding focused unit tests for `Rank` configuration validation would also help prevent regressions.

iron-prog:
**What type of PR is this?**

/kind bug

Fixes validation in `Rank._check_fields()` so configuration fields reject invalid non-empty values with the wrong type instead of silently passing validation.

This PR:
- Corrects the presence/type validation logic for `sort_by`, `visualization`, and `selected_dataitem`.
- Validates `save_mode` as a string, matching how it is used throughout the codebase.
- Adds unit tests covering valid configurations and invalid field types to prevent regressions.

Fixes #659


hira299:
**Review:** Verified diff against the upstream branch: The PR comprehensively fixes all four inverted boolean guards (`sort_by`, `visualization`, `selected_dataitem`, `save_mode`) in `_check_fields()`, correctly updates `save_mode`'s type check from `list` to `str` (matching how it is used in `save()`), and adds unit test coverage in `tests/storymanager/test_rank.py` and `tests/conftests.py`.

**Verified no regression:** All `benchmarkingjob.yaml` files in the repository that declare `save_mode` use `"selected_and_all"` (a non-empty string), which continues to pass the new validation guard cleanly.

**One residual, non-blocking observation:** The fix validates that `save_mode` is a non-empty string, but does not check if the value is one of the valid string options (`"selected_and_all"`, `"selected_only"`, `"selected_and_all_and_picture"`). An invalid string like `"selected_al"` would pass `_check_fields()` and silently no-op in `save()`. This is worth a follow-up issue, not a blocker on this PR.

**Recommendation: Accept as-is.**

PR #661, opened 2026-08-05
fix(examples): MATPLOTLIB_AVAILABLE and OPEN3D_AVAILABLE always False — add conditional imports to enable visualization in pose-estimation-llio

### LINKED ISSUE #635: MATPLOTLIB_AVAILABLE/OPEN3D_AVAILABLE never set — visualization silently disabled in pose-estimation-llio
**What happened**:

Image

`examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py`
declares two feature-detection flags at module level:

```python
MATPLOTLIB_AVAILABLE = False
OPEN3D_AVAILABLE = False
```

These flags are never set to True, so visualization and point-cloud features are silently disabled even when matplotlib and open3d are installed. Additionally, enabling either flag causes runtime failures because required modules (plt, matplotlib, o3d, Ellipse, PatchCollection) are never imported.

**How to reproduce**:

Arijit429:
/kind bug

`MATPLOTLIB_AVAILABLE` and `OPEN3D_AVAILABLE` flags in `pose-estimation-llio/testalgorithms/llio_fusion/utils.py` were hardcoded to `False` at module level:

```python
MATPLOTLIB_AVAILABLE = False
OPEN3D_AVAILABLE = False
```

This meant visualization and point-cloud features were permanently disabled even when `matplotlib` and `open3d` are installed. Worse, functions like `visualize()`, `plot_gaussian()`, `velo2downpcd()` reference `plt`, `o3d`, `Ellipse`, and `PatchCollection` — none of which were ever imported — causing runtime failures if the flags were manually set to `True`.

This PR replaces the hardcoded flags with proper `try/except` import blocks:

```python
try:
    import matplotlib
    from matplotlib.patches import Ellipse
    from matplotlib.collections import PatchCollection
    MATPLOTLIB_AVAILABLE = True
except ImportError:
    MATPLOTLIB_AVAILABLE = False

try:
    import open3d as o3d
    OPEN3D_AVAILABLE = True
except ImportError:
    OPEN3D_AVAILABLE = False
```

Now the flags correctly reflect whether the libraries are actually available at runtime, and all dependent symbols are properly imported when the libraries are present.

Fixes #635

### DIFF
--- a/examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py
+++ b/examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py
@@ -7,8 +7,20 @@ from functools import wraps
 import time
 from core.common.log import LOGGER
 
-MATPLOTLIB_AVAILABLE = False
-OPEN3D_AVAILABLE = False
+try:
+    import matplotlib
+    import matplotlib.pyplot as plt
+    from matplotlib.patches import Ellipse
+    from matplotlib.collections import PatchCollection
+    MATPLOTLIB_AVAILABLE = True
+except ImportError:
+    MATPLOTLIB_AVAILABLE = False
+
+try:
+    import open3d as o3d
+    OPEN3D_AVAILABLE = True
+except ImportError:
+    OPEN3D_AVAILABLE = False
 
 
 def timeit(func):


akshita317:
The diagnosis is right and the fix is idiomatic — `MATPLOTLIB_AVAILABLE` and `OPEN3D_AVAILABLE` are hard-coded to `False`, so every visualisation branch in `pose-estimation-llio` is unreachable no matter what is installed, and `try/except ImportError` is the correct way to restore it.

Two notes.

**1. Two other open PRs make the identical change:** **#636** (@iron-prog) and **#849** (yours, where it is bundled into an unrelated `core/` validation fix). All three add byte-identical blocks to this same file. This PR is the cleanest of the three — one file, one concern, accurate title — so it seems the natural one to keep; the other two copies will conflict once any of them lands.

**2. `open3d` is not declared anywhere in the repository.** `grep -rn "open3d" --include=requirements.txt .` returns nothing, and it is absent from `examples/industrialEI/pose-estimation-llio/requirements.txt`. So after this fix `MATPLOTLIB_AVAILABLE` will correctly become `True` (matplotlib is in the root `requirements.txt`), but `OPEN3D_AVAILABLE` will still be `False` in any standard install — the flag will now be *accurate* rather than *always wrong*, which is an improvement, but the point-cloud visualisation stays unreachable until `open3d` is declared as an optional dependency.

Adding it to the example's `requirements.txt`, or documenting it as optional in the README, would complete the fix.

PR #648, opened 2026-08-04
fix(metrics): resolve NoneType crashes, transfer ratio math error, and metric lookup guardrails

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Fixes several systemic defects in the system metrics evaluation subsystem (`core/testcasecontroller/metrics/metrics.py`):

1. **Defensive Validation for Missing System Metric Data**:
   Added `system_metric_info` and metric-specific payload checks across built-in metric functions (`samples_transfer_ratio_func`, `task_avg_acc_func`, `forget_rate_func`, `bwt_func`, `fwt_func`, `matrix_func`). When metric data is missing, empty, or `None`, the functions now gracefully return `np.nan` (or `{}` for matrix metrics) instead of crashing with `TypeError: 'NoneType' object is not iterable / subscriptable` or `KeyError`.

2. **Accurate Transfer Ratio Calculation**:
   Fixed mathematical ratio calculation in `samples_transfer_ratio_func()`:
   - Removed artificial `+ 1` denominator offset (`transfer_num / (inference_num + 1)`) to compute the exact ratio `transfer_num / inference_num`.
   - Added explicit zero-check handling when `inference_num == 0`.

3. **Consistent Tuple Unpacking & Metric Lookup Guardrails**:
   - Updated `compute()` to consistently return a 3-tuple (`[], np.nan, np.nan`) when matrix inputs are invalid, preventing 2-tuple vs 3-tuple unpacking crashes.
   - Updated `get_metric_func()` to check for function existence in `sys.modules` and raise a clear `ValueError` when an unknown metric name is specified without an external plugin `url`.

4. **Unit Tests**:
   Added `tests/test_metrics.py` covering system metric functions (`samples_transfer_ratio_func`, `task_avg_acc_func`, `forget_rate_func`, `bwt_func`, `fwt_func`, `matrix_func`), missing data handling, zero denominators, and metric lookup error handling (`5 passed`).

**Which issue(s) this PR fixes**:

Fixes #647

**Special notes for your reviewer**:

NONE

**Does this PR introduce a user-facing change?**:

```release-note
Fix system metrics evaluation subsystem to defensively handle missing metric data, ensure exact sample transfer ratio calculation, and prevent NoneType crashes.
```

PR #664, opened 2026-08-05
fix(core): add null guards for testenv/rank in BenchmarkingJob and validate visualization methods

### LINKED ISSUE #662: [Bug] Missing null-guards for testenv/rank in BenchmarkingJob, and/or validation bug in core config classes, and opaque AttributeError in get_visualization_func()
**What happened**:

Three closely related validation gaps in `core/cmd/obj/benchmarkingjob.py` allow a benchmarking job to pass configuration parsing and crash at runtime with opaque `AttributeError` or `RuntimeError` messages rather than failing fast with a clear `ValueError` at startup:

**1. `BenchmarkingJob._check_fields()` does not validate that `test_env` and `rank` were parsed**

`self.test_env` and `self.rank` are both initialised to `None` (lines 48, 47). If the YAML config omits the `testenv` or `rank` keys entirely, `_check_fields()` never verifies they are non-`None`. `run()` then calls `self.test_env.prepare()` (line 89) and `self.rank.save(...)` (line 97), crashing with:

```
AttributeError: 'NoneType' object has no attribute 'prepare'
AttributeError: 'NoneType' object has no attribute 'save'
```

These errors are wrapped in a generic `RuntimeError` from `benchmarking.py`, giving the user no indication that the YAML is missing a required section.

**2. `BenchmarkingJob._check_fields()` uses `and` instead of `or`, allowing wrong-typed values to bypass validation**

```python
if not self.name and not isinstance(self.name, str):  # line 54
if not self.test_object and not isinstance(self.test_object, dict):  # line 61
```

Because Python's `and` short-circuits, a non-empty value of the wrong type (e.g. `name: 42`, `test_object: "bad"`) passes these checks silently — `not 42` is `False` so the whole condition is `False` regardless of the `isinstance` check. The same `and`/`or` pattern appears in `core/storymanager/rank/rank.py` (tracked in #659) and `core/testcasecontroller/algorithm/algorithm.py` (lines 133, 136), indicating a systemic copy-paste defect across all core config classes.

**3. `get_visualization_func()` raises an opaque `AttributeError` for unrecognised visualization methods**

```python
# core/storymanager/visualization/visualization.py line 48
def get_visualization_func(mode):
    return getattr(sys.modules[__name__], mode)
```

When `method` in the YAML is misspelled or unsupported (e.g. `method: plot_table` instead of `print_table`), this raises:
```
AttributeError: module 'core.storymanager.visualization.visualization' has no attribute 'plot_table'
```
There is no guard, no listing of valid methods, and no `ValueError`. The error is surfaced only at `rank.plot()` time — long after configuration is complete.

1. If `testenv` or `rank` keys are absent from the YAML, `BenchmarkingJob._check_fields()` should raise a clear `ValueError` like `"benchmarkingjob's testenv must be provided"` before `run()` is ever called.
2. Type-validation guards should use `or` so that a non-empty, wrong-typed value is also rejected: `if not isinstance(self.name, str) or not self.name`.
3. `get_visualization_func(mode)` should validate `mode` against the set of known functions and raise a `ValueError` listing valid options when an unrecognised name is given.

**Reproduce #1 — missing `rank` key:**

```python
from core.cmd.obj.benchmarkingjob import BenchmarkingJob

# A config with a valid test_object but no 'rank' key
config = {
    "name": "test-job",
    "workspace": "/tmp/ws",
    "test_object": {"type": "algorithms", "algorithms": []},
    # 'rank' key intentionally omitted
}
job = BenchmarkingJob(config)
# No error here — _check_fields() passes.
# job.run() would crash with: AttributeError: 'NoneType' object has no attribute 'save'
print(job.rank)  # None
```

**Reproduce #2 — wrong-typed `name`:**

```python
from core.cmd.obj.benchmarkingjob import BenchmarkingJob

config = {
    "name": 42,          # should be a string, but passes validation
    "workspace": "/tmp/ws",
    "test_object": {"type": "algorithms", "algorithms": []},
}
job = BenchmarkingJob(config)
print(job.name)  # 42 — no ValueError was raised
```

**Reproduce #3 — bad visualization method:**

```python
from core.storymanager.visualization.visualization import get_visualization_func

# 'plot_table' is not a defined function
func = get_visualization_func("plot_table")
# AttributeError: module '...' has no attribute 'plot_table'
# No helpful error message listing valid options.
```

Affected files:
- `core/cmd/obj/benchmarkingjob.py` — missing `test_env`/`rank` null guards in `_check_fields()`; `and`/`or` validation bug on lines 54, 61
- `core/storymanager/visualization/visualization.py` — bare `getattr` in `get_visualization_func()` with no guard
- `core/testcasecontroller/algorithm/algorithm.py` — same `and`/`or` bug on lines 133, 136

A fix would involve:
1. Adding `if self.test_env is None` and `if self.rank is None` guards in `BenchmarkingJob._check_fields()`.
2. Replacing `not X and not isinstance(X, T)` with `not isinstance(X, T) or not X` across all affected `_check_fields()` methods.
3. Replacing the bare `getattr` in `get_visualization_func()` with an explicit allowlist check that raises `ValueError` with a listing of valid method names.
4. Adding unit tests covering each of these error paths.

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

This PR fixes three closely related validation gaps in the `core` benchmarking job orchestration and visualization modules:

1. **Missing `test_env` and `rank` Null Guards in `BenchmarkingJob`**:
   - `BenchmarkingJob._check_fields()` now explicitly verifies that `test_env` and `rank` were parsed from the YAML configuration. If missing, a clear `ValueError` is raised before `run()` can crash with unhandled `AttributeError` exceptions.

2. **Fixed `and`/`or` Evaluation Bug in `BenchmarkingJob` & `Algorithm` Validation**:
   - `_check_fields()` in `BenchmarkingJob` and `Algorithm` previously used `if not self.field and not isinstance(self.field, type):`, which allowed invalid non-empty values of incorrect types (e.g., integers for name strings) to bypass validation.
   - Refactored condition logic to `if not isinstance(self.field, str) or not self.field:` across both classes.

3. **Explicit Guard for `get_visualization_func()`**:
   - Replaced bare `getattr(sys.modules[__name__], mode)` in `visualization.py` with an explicit check against supported visualization methods (`print_table`, `draw_heatmap_picture`). Raises a clear `ValueError` detailing valid options when an invalid method name is supplied.

4. **Unit Tests**:
   - Added `tests/test_benchmarkingjob_validation.py` covering invalid field types, missing `testenv`/`rank` blocks, invalid algorithm names/paradigms, and unsupported visualization methods (14 tests passed).

Fixes #662

### DIFF
--- a/core/cmd/obj/benchmarkingjob.py
+++ b/core/cmd/obj/benchmarkingjob.py
@@ -51,27 +51,35 @@ class BenchmarkingJob:
         self._parse_config(config)
 
     def _check_fields(self):
-        if not self.name and not isinstance(self.name, str):
-            raise ValueError(f"benchmarkingjob's name({self.name}) must be provided"
+        if not isinstance(self.name, str) or not self.name:
+            raise ValueError(f"benchmarkingjob's name({self.name!r}) must be provided"
                              f" and be string type.")
 
         if not isinstance(self.workspace, str):
-            raise ValueError(f"benchmarkingjob's workspace({self.workspace}) must be string type.")
+            raise ValueError(f"benchmarkingjob's workspace({self.workspace!r}) must be string type.")
 
-        if not self.test_object and not isinstance(self.test_object, dict):
-            raise ValueError(f"benchmarkingjob's test_object({self.test_object})"
-                             f" must be dict type.")
+        if not isinstance(self.test_object, dict) or not self.test_object:
+            raise ValueError(f"benchmarkingjob's test_object({self.test_object!r})"
+                             f" must be a non-empty dict type.")
 
         test_object_types = [e.value for e in TestObjectType.__members__.values()]
         test_object_type = self.test_object.get("type")
         if test_object_type not in test_object_types:
             raise ValueError(
-                f"benchmarkingjob' test_object doesn't support the type({test_object_type}), "
+                f"benchmarkingjob' test_object doesn't support the type({test_object_type!r}), "
                 f"the following test object types can be selected: {test_object_types}.")
 
         if not self.test_object.get(test_object_type):
             raise ValueError(f"benchmarkingjob' test_object doesn't find"
-                             f" the field({test_object_type}).")
+                             f" the field({test_object_type!r}).")
+
+        if self.test_env is None:
+            raise ValueError("benchmarkingjob's testenv must be provided "
+                             "in the benchmarking config file.")
+
+        if self.rank is None:
+            raise ValueError("benchmarkingjob's rank must be provided "
+                             "in the benchmarking config file.")
 
     def run(self):
         """
--- a/core/storymanager/visualization/visualization.py
+++ b/core/storymanager/visualization/visualization.py
@@ -43,6 +43,13 @@ def draw_heatmap_picture(output, title, matrix):
     plt.savefig(output_dir)
     plt.show()
 
+_VISUALIZATION_FUNCS = {"print_table", "draw_heatmap_picture"}
+
 def get_visualization_func(mode):
     """ get visualization func """
+    if mode not in _VISUALIZATION_FUNCS:
+        raise ValueError(
+            f"unsupported visualization method: {mode!r}. "
+            f"Valid options are: {sorted(_VISUALIZATION_FUNCS)}"
+        )
     return getattr(sys.modules[__name__], mode)
--- a/core/testcasecontroller/algorithm/algorithm.py
+++ b/core/testcasecontroller/algorithm/algorithm.py
@@ -130,12 +130,12 @@ class Algorithm:
         return None
 
     def _check_fields(self):
-        if not self.name and not isinstance(self.name, str):
-            raise ValueError(f"algorithm name({self.name}) must be provided and be string type.")
+        if not isinstance(self.name, str) or not self.name:
+            raise ValueError(f"algorithm name({self.name!r}) must be provided and be string type.")
 
-        if not self.paradigm_type and not isinstance(self.paradigm_type, str):
+        if not isinstance(self.paradigm_type, str) or not self.paradigm_type:
             raise ValueError(
-                f"algorithm paradigm({self.paradigm_type}) must be provided and be string type.")
+                f"algorithm paradigm({self.paradigm_type!r}) must be provided and be string type.")
 
         paradigm_types = [e.value for e in ParadigmType.__members__.values()]
         if self.paradigm_type not in paradigm_types:
@@ -145,12 +145,30 @@ class Algorithm:
         if not isinstance(self.incremental_learning_data_setting, dict):
             raise ValueError(
                 f"algorithm incremental_learning_data_setting"
-                f"({self.incremental_learning_data_setting} must be dictionary type.")
+                f"({self.incremental_learning_data_setting}) must be dictionary type.")
 
         if not isinstance(self.lifelong_learning_data_setting, dict):
             raise ValueError(
                 f"algorithm lifelong_learning_data_setting"
-                f"({self.lifelong_learning_data_setting} must be dictionary type.")
+                f"({self.lifelong_learning_data_setting}) must be dictionary type.")
+
+        if not isinstance(self.fl_data_setting, dict):
+            raise ValueError(
+                f"algorithm fl_data_setting({self.fl_data_setting}) must be dictionary type.")
+
+        data_partition = self.fl_data_setting.get("data_partition")
+        if data_partition is not None and data_partition not in ("iid", "non-iid"):
+            raise ValueError(
+                f"algorithm fl_data_setting data_partition({data_partition!r}) must be 'iid' or 'non-iid'.")
+
+        non_iid_ratio = self.fl_data_setting.get("non_iid_ratio")
+        if non_iid_ratio is not None and (
+            isinstance(non_iid_ratio, bool)
+            or not isinstance(non_iid_ratio, (int, float))
+            or not 0 < non_iid_ratio <= 1.0
+        ):
+            raise ValueError(
+                f"algorithm fl_data_setting non_iid_ratio({non_iid_ratio!r}) must be a number in (0, 1].")
 
         if not isinstance(self.initial_model_url, str):
             raise ValueError(
--- a/tests/test_benchmarkingjob_validation.py
+++ b/tests/test_benchmarkingjob_validation.py
@@ -1,0 +1,195 @@
+# Copyright 2022 The KubeEdge Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for BenchmarkingJob validation, Algorithm validation, and get_visualization_func."""
+
+import sys
+import os
+import unittest
+
+# Ensure the repo root is on sys.path for imports
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+def _minimal_job_config(**overrides):
+    """Return a minimal BenchmarkingJob config dict, applying any overrides."""
+    cfg = {
+        "name": "test-job",
+        "workspace": "./workspace",
+        "test_object": {"type": "algorithms", "algorithms": [{"name": "a", "url": "/nonexistent.yaml"}]},
+    }
+    cfg.update(overrides)
+    return cfg
+
+
+# ---------------------------------------------------------------------------
+# BenchmarkingJob validation tests
+# ---------------------------------------------------------------------------
+
+class TestBenchmarkingJobValidation(unittest.TestCase):
+    """Unit tests for BenchmarkingJob._check_fields()."""
+
+    def _make_job(self, config):
+        """Import lazily to avoid top-level import side-effects."""
+        from core.cmd.obj.benchmarkingjob import BenchmarkingJob
+        return BenchmarkingJob(config)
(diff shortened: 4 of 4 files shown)


Arijit429:
**Code review — from LFX 2026 Term 3 pre-test Discussion #862**

**Recommendation: Minor Revision**

New finding not in any existing review: `metrics.py`'s `samples_transfer_ratio_func` is rewritten mid-PR, inverting the ratio direction and dropping a `+1` denominator smoothing term:

```python
# Before
return round(float(transfer_num) / (inference_num + 1), 4)

# After
return round(float(transfer_num) / float(inference_num), 4)
```

This changes the numeric output of a published system metric — historical leaderboard entries computed under the old formula stop being comparable after this merges. The change lives in its own commit (`bab0273`) but isn't explained in the PR description — worth a one-line justification there even though the commit history already separates it, which makes the split-into-its-own-PR recommendation straightforward.

The null guards and visualization allowlist fix are sound and ready to merge independently.

bhuvan-somisetty:
Thanks for reviewing @Arijit429! The \metrics.py\ commit was unintentionally included from an earlier branch and belongs to a separate PR (#648). I've rebased this branch on latest \main\ and dropped that commit so this PR remains strictly focused on the \BenchmarkingJob\, \Algorithm\, and visualization validation fixes for #662. All 14 tests pass cleanly.

Arijit429:
Thanks @bhuvan-somisetty for the quick turnaround on the rebase — confirming the `metrics.py` commit is gone and the PR is now correctly scoped to the BenchmarkingJob/Algorithm/visualization fixes. The separation makes the change history much cleaner for anyone bisecting later. 

Minor remaining note: with the formula change removed, the test coverage ratio looks even stronger now — 226 lines of tests for a tightly-scoped set of validation guards is solid. Ready for maintainer review from my side.

bhuvan-somisetty:
Thanks @Arijit429, appreciate the thorough review and feedback!

hira299:
### 2.3 PR #664: fix(core): add null-guards in benchmarkingjob and validation in config

**Author:** bhuvan (fixing Issue #662)
**Scope:** `core/cmd/obj/benchmarkingjob.py`, `core/testcasecontroller/algorithm/algorithm.py`
**Recommendation:** Minor Revision Required

**Buggy Guards in benchmarkingjob.py (Lines 54 and 61)**

```python
# core/cmd/obj/benchmarkingjob.py, lines 53-63 (CURRENT/BUGGY)
def _check_fields(self):
    if not self.name and not isinstance(self.name, str):         # L54: BUG
        raise ValueError(...)

    if not isinstance(self.workspace, str):                       # L58: CORRECT
        raise ValueError(...)

    if not self.test_object and not isinstance(self.test_object, dict):  # L61: BUG
        raise ValueError(...)
```

Line 54: `not self.name and not isinstance(self.name, str)` evaluates to `True` only if `name` is falsy AND not a string. Since `self.name` initializes to `""` (an empty string, a `str`), `not isinstance("", str)` is `False`. The guard never fires, even when `name` is actually empty and invalid.

**Buggy Guards in algorithm.py (Lines 133 and 136)**

```python
# core/testcasecontroller/algorithm/algorithm.py, lines 132-138 (CURRENT/BUGGY)
def _check_fields(self):
    if not self.name and not isinstance(self.name, str):            # L133: BUG
        raise ValueError(...)

    if not self.paradigm_type and not isinstance(self.paradigm_type, str):  # L136: BUG
        raise ValueError(...)
```

**Proposed Fix Pattern (All Four Sites)**

```diff
# benchmarkingjob.py L54 and algorithm.py L133
-    if not self.name and not isinstance(self.name, str):
+    if not self.name or not isinstance(self.name, str):

# benchmarkingjob.py L61
-    if not self.test_object and not isinstance(self.test_object, dict):
+    if not self.test_object or not isinstance(self.test_object, dict):

# algorithm.py L136
-    if not self.paradigm_type and not isinstance(self.paradigm_type, str):
+    if not self.paradigm_type or not isinstance(self.paradigm_type, str):
```

**Revision Requirements**

1. **Missing `fl_data_setting` type check:** The `fl_data_setting` dict in `algorithm.py` (lines 70-76) has no `isinstance(self.fl_data_setting, dict)` guard in `_check_fields()`. Add a check.

2. **Missing `fl_data_setting` sub-key validation:** The values `data_partition` (must be `"iid"` or `"non-iid"`) and `non_iid_ratio` (must be in `(0, 1]`) are never validated. A caller can set `non_iid_ratio: 0` and receive silently empty client partitions.

3. **No tests included:** At minimum, one test verifying that a `BenchmarkingJob` with an empty `name` field raises `ValueError` with a clear message.

bhuvan-somisetty:
Thanks for the thorough review and feedback @hira299! Updated the PR to:

1. Ensure all `name`, `workspace`, `test_object`, and `paradigm_type` guards strictly validate types and non-empty values.
2. Added `fl_data_setting` dictionary type validation in `Algorithm._check_fields()`, along with sub-key checks for `data_partition` (`'iid'` / `'non-iid'`) and `non_iid_ratio` (must be in `(0, 1]`).
3. Expanded unit tests under `tests/test_benchmarkingjob_validation.py` to cover all field guards and `fl_data_setting` validation — all 17 tests pass cleanly.

hira299:
> Thanks for the thorough review and feedback @hira299! Updated the PR to:
> 
> 1. Ensure all `name`, `workspace`, `test_object`, and `paradigm_type` guards strictly validate types and non-empty values.
> 2. Added `fl_data_setting` dictionary type validation in `Algorithm._check_fields()`, along with sub-key checks for `data_partition` (`'iid'` / `'non-iid'`) and `non_iid_ratio` (must be in `(0, 1]`).
> 3. Expanded unit tests under `tests/test_benchmarkingjob_validation.py` to cover all field guards and `fl_data_setting` validation — all 17 tests pass cleanly.

Thanks @bhuvan-somisetty confirming the update addresses what I flagged: `fl_data_setting` dict-type validation, `data_partition`/`non_iid_ratio` range checks, and the expanded test coverage in `tests/test_benchmarkingjob_validation.py` (17 tests) all look correct on re-check. Also worth noting for other reviewers: this PR now also fixes the third component of #662 the unguarded `get_visualization_func()`  via the new `_VISUALIZATION_FUNCS` allow-list. Recommendation: Accept as-is.

bhuvan-somisetty:
Thank you @hira299 for the thorough review, testing, and approval! Really appreciate the helpful feedback throughout.

Muhammad-Hashir-Code:
**Review — LFX 2026 Term 3 pre-test** (full analysis: https://github.com/kubeedge/ianvs/discussions/887)

Verified the `_check_fields()` fix directly re-ran the old vs. new validation logic against the same inputs:
```
input           old (buggy)                    this PR
''              silently accepted (bug)        raises (correct)
42               silently accepted (bug)        raises (correct)
None            raises                         raises
'valid_name'    silently accepted              silently accepted
```
Confirms this closes the `and`/`or` logic gap correctly in both `benchmarkingjob.py` and `algorithm.py` the empty-string and wrong-type cases that previously passed validation silently now correctly raise.

**One scope note, not a defect in what's here:** the same `and`/`or` pattern also exists in `core/storymanager/rank/rank.py`'s `Rank._check_fields()` (already tracked separately in #460) and isn't touched by this PR. Not something this PR needs to cover, just flagging so it doesn't read as "the pattern is fully gone repo-wide" once this merges.

`get_visualization_func`'s new allowlist is a good addition too clean, testable, and the test suite covers the rejection path (`test_arbitrary_attr_does_not_leak`) which is exactly the kind of case that's easy to skip.

MukandKrishna (commented):
**Recommendation: minor revision.** The new `non_iid_ratio` guard has one type-boundary gap: it accepts the YAML boolean `true` as a valid numeric ratio.

Python defines `bool` as a subclass of `int`, so this condition treats `True` as the number `1`:

```python
isinstance(True, (int, float))  # True
0 < True <= 1.0                # True
```

I executed `_check_fields()` from PR head `a59e2dba7e21969b5828f8f9d2a6fc615da4240b` with the same otherwise-valid configuration and obtained:

```text
True (bool): ACCEPTED
False (bool): ValueError: algorithm fl_data_setting non_iid_ratio(False) must be a number in (0, 1].
0.5 (float): ACCEPTED
'0.5' (str): ValueError: algorithm fl_data_setting non_iid_ratio('0.5') must be a number in (0, 1].
```

This is reachable from ordinary YAML because `non_iid_ratio: true` is loaded as Python `True`. It then silently behaves as ratio `1`, despite the validation message requiring a number.

Please explicitly reject `bool` before the existing `(int, float)` and range checks. The focused test matrix should include both `True` and `False`; `True` is the important regression case because it currently passes both predicates, while `False` is already rejected by the lower bound.

Verification boundary: I executed the exact PR implementation of `Algorithm._check_fields()` with minimal import stubs; no paradigm runtime behavior was replaced or exercised. The full federated-learning benchmark was not run.

bhuvan-somisetty:
Thanks for the catch on the boolean type boundary @MukandKrishna! Updated `Algorithm._check_fields()` to explicitly reject `bool` instances for `non_iid_ratio` before numeric comparison, and added `True` and `False` to the regression test matrix in `tests/test_benchmarkingjob_validation.py`. All 17 tests pass cleanly.

MukandKrishna:
> Thanks for the catch on the boolean type boundary @MukandKrishna! Updated `Algorithm._check_fields()` to explicitly reject `bool` instances for `non_iid_ratio` before numeric comparison, and added `True` and `False` to the regression test matrix in `tests/test_benchmarkingjob_validation.py`. All 17 tests pass cleanly.

Thanks for the quick update. I independently reran the same focused boundary probe against the new PR head `30a757f147d30c8197b00fcc408e07200319a748`.

```text
previous head a59e2dba:
True (bool): ACCEPTED

updated head 30a757f1:
True (bool): ValueError
False (bool): ValueError
0.5 (float): ACCEPTED
'0.5' (str): ValueError
```

I also confirmed that the implementation now explicitly excludes `bool` before the numeric/range checks and that the regression matrix includes both `True` and `False`. This resolves the finding from my review.

Verification boundary: I reran the exact `Algorithm._check_fields()` behavior from the updated PR file with the same minimal import stubs used for the original reproduction. I inspected the updated regression test, but did not independently run the complete 17-test suite.

bhuvan-somisetty:
Thank you @MukandKrishna for the prompt re-check and independent verification! Really appreciate the helpful feedback throughout.

PR #670, opened 2026-08-06
fix(dataset): correctly use np.isin for non-IID partitioning

### LINKED ISSUE #669: [Bug] `partition_data` non-IID branch uses incorrect `np.where` comparison with a list
**Describe the bug**
In `core/testenvmanager/dataset/utils.py`, the `partition_data()` function's non-IID branch compares `y_data` (a numpy array) against `current_class` (a Python list) using `np.where(y_data == current_class)`. This comparison is incorrect — it performs element-wise comparison with broadcasting, which will either crash with a shape mismatch or return unexpected results. The intent is to filter samples belonging to any class in `current_class`, which requires `np.isin(y_data, current_class)`.

**To Reproduce**
Call `partition_data(datasets, client_number=2, data_partition="non-iid")` where `datasets[1]` has more elements than the number of sampled classes.

**Expected behavior**
`np.isin(y_data, current_class)` should be used to correctly select all samples belonging to any of the selected classes.

**Location**
- File: `core/testenvmanager/dataset/utils.py`, Line 91

**Proposed Fix**
```diff
-            indices = np.where(y_data == current_class)[0]
+            indices = np.where(np.isin(y_data, current_class))[0]

Hiteshsai007:
Fixes #669 by replacing \
p.where(y_data == current_class)\ with \
p.where(np.isin(y_data, current_class))\ to properly handle list comparison.

### DIFF
--- a/core/testenvmanager/dataset/utils.py
+++ b/core/testenvmanager/dataset/utils.py
@@ -88,7 +88,7 @@ def partition_data(datasets, client_number, data_partition="iid", non_iid_ratio=
             sample_number = int(class_num * non_iid_ratio)
             current_class = random.sample(range(class_num), sample_number)
             LOGGER.info(f"for client{i} the class is {current_class}")
-            indices = np.where(y_data == current_class)[0]
+            indices = np.where(np.isin(y_data, current_class))[0]
             data.append([x_data[indices], y_data[indices]])
     else:
         raise ValueError("paritiion_methods must be 'iid' or 'non-iid'")


hira299:
**Review:** Verified diff: this is a minimal, correct one-line fix (`np.where(y_data == current_class)` -> `np.where(np.isin(y_data, current_class))`), confirmed by independent reproduction (buggy line raises `ValueError: operands could not be broadcast together with shapes (8,) (2,)`; fixed line returns the correct index set). No regression risk to the `iid` path (untouched, uses `np.arange`-based slicing).

**Note for maintainers:** PR #624 and PR #808 both submit the identical one-line fix independently (opened 2026-08-01 and 2026-08-15 respectively). I recommend triaging and closing the other two as duplicates once one merges.

**Residual items for follow-up (non-blocking):** Dedicated unit tests for `partition_data()` under `core/testenvmanager/`, a guard for when `non_iid_ratio` produces an empty `current_class` (which can cause division-by-zero downstream), and a type guard for scalar vs. list input.

Hiteshsai007:
Thanks for independently verifying the diff and confirming that the one-line change fixes the broadcasting failure without affecting the iid path.

I agree that PRs #624 and #808 are duplicate implementations of the same fix. I recommend keeping this PR narrowly scoped to Issue #669 and allowing maintainers to select one canonical implementation before closing the duplicates.

The dedicated partition_data tests, empty current_class guard, and scalar/list validation are good follow-up items. I would keep those separate from this minimal correction so the current PR remains easy to review and merge.

PR #672, opened 2026-08-06
fix: replace duplicate benchmarking.py with wrapper

### LINKED ISSUE #671: [Bug] Duplicate benchmarking.py at repository root is an exact copy of core/cmd/benchmarking.py
The file `benchmarking.py` at the repository root is a byte-for-byte duplicate of `core/cmd/benchmarking.py`. This creates maintenance confusion — changes to one may not be reflected in the other. The `setup.py` entry point already points to `core.cmd.benchmarking:main`, making the root-level copy redundant.

**Expected behavior**
Either remove the root-level `benchmarking.py` or have it simply import and delegate to `core.cmd.benchmarking.main()`.

**Location**
- File: `benchmarking.py` (root)
- File: `core/cmd/benchmarking.py`

**Proposed Fix**
Option A — Delete `benchmarking.py` from the root.
Option B — Replace its contents with:
```python
from core.cmd.benchmarking import main

if __name__ == '__main__':
    main()

Hiteshsai007:
Resolves #671 by replacing the root `benchmarking.py` with a wrapper that imports and executes `core.cmd.benchmarking.main()`.

This prevents duplicate code maintenance issues (DRY principle) while maintaining the existing entrypoint for users who clone the repository and execute the script directly instead of installing via pip.


Raghavbhola:
This PR is a good improvement. It replaces the root benchmarking.py with a wrapper that calls core.cmd.benchmarking.main(), which follows the DRY principle and avoids duplicate code maintenance.  

Strengths:
- Clean wrapper approach, reduces redundancy.
- Maintains backward compatibility for users running the script directly.
- Aligns with best practices by centralizing logic in core.

Possible Improvements:
- Adding a simple test to ensure the wrapper correctly calls core.cmd.benchmarking.main() would help catch regressions.
- Commit message could briefly mention the impact on multiple examples for clarity.

Overall, this PR goes beyond example-local fixes and improves Ianvs Core maintainability. I recommend **minor revision** (add test coverage), otherwise it is ready to merge.

Yahdel:
This PR fixes the duplicate root `benchmarking.py` file by making it call the main Core command instead. That is a good change: future command fixes only need to be made in one place.

I tested it with benchmark-only packages blocked. By itself, this PR sends the root script to the Core command, but the Core command still tries to load benchmark code too early. Because of that, version and help can still fail.

PR #811 fixes that Core problem. With #672 and #811 together, these four commands worked without loading unrelated benchmark packages:

| Command path | Version | Help |
| --- | --- | --- |
| `python benchmarking.py` | Works | Works |
| `python -m core.cmd.benchmarking` | Works | Works |

I suggest merging this PR together with #811, then adding a small test for version and help from both command paths.

I tested the command startup only. I did not run the full TAB or MOT17 benchmarks. Full context: Discussion #961.

PR #674, opened 2026-08-06
fix(deps): add missing colorlog to requirements.txt

### LINKED ISSUE #673: [Bug] colorlog is imported by core/common/log.py but is not listed in requirements.txt
The `Logger` class in `core/common/log.py` imports `colorlog` at line 18, but `colorlog` is not listed in `requirements.txt`. This causes an `ImportError` on a fresh install when `colorlog` is not already present.

**To Reproduce**
1. Create a fresh virtualenv.
2. `pip install -r requirements.txt`
3. `python -c "from core.common.log import LOGGER"` → `ModuleNotFoundError: No module named 'colorlog'`

**Expected behavior**
`colorlog` should be added to `requirements.txt` (and ideally to `setup.py`'s `install_requires`).

**Proposed Fix**
```diff
 prettytable~=2.5.0 # BSD
 scikit-learn
 numpy
 pandas
 tqdm
 matplotlib
 onnx
+colorlog

### PR DESCRIPTION
**Pull request by Hiteshsai007**
Resolves #673 by adding `colorlog` to `requirements.txt`. 

The `Logger` class in `core/common/log.py` explicitly imports `colorlog`, but it was missing from the dependencies list, causing `ModuleNotFoundError` on fresh installations.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ pandas
 tqdm
 matplotlib
 onnx
+colorlog
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,9 @@
 prettytable~=2.5.0 # BSD
+colorlog # MIT
+PyYAML # MIT
 scikit-learn
 numpy
 pandas
 tqdm
 matplotlib
 onnx
-colorlog

### CODE REVIEW

**Comment by kavaljeetsingh-dev** [comment]
Tested this directly. This fixes the `colorlog` import in `core/common/log.py`, but `core/common/utils.py` also has an undeclared `yaml` import in the same category, which this PR doesn't cover:

```
$ python3 -c "import core.common.utils"
ModuleNotFoundError: No module named 'yaml'
```

#791 and #468 (also open, targeting the same issue) both include the PyYAML line as well.

sheikhayaan:
There are now five open PRs for this one line: this one, #793, #791, #682 and #468. @kavaljeetsingh-dev has already covered the yaml half above, so I won't repeat it. What's still unresolved is which of the five should actually merge, so here's the tiebreak on the two that fix colorlog cleanly.

Placement and hygiene are the only real difference. #674 appends colorlog after onnx, and onnx is the line with no trailing newline — which is why this PR reads +2 −1 rather than +1: it rewrites the onnx line to add the newline it then consumes. The diff still shows \ No newline at end of file on the new side, so the file ends up with the same defect one line lower. #793 inserts at line 2 with a # MIT comment, matching prettytable~=2.5.0 # BSD, and leaves the end of the file untouched.

This PR has the stronger claim on two counts, though. It is three weeks old against #793's one week, and it is the only one of the five that names its issue, Resolves #673, so merging it closes the bug automatically. #793 is linked to #742 but doesn't close anything.

So the honest summary: #674 has priority and issue linkage, #793 has the tidier edit and now carries PyYAML too. Whichever is chosen, the other four should be closed as duplicates in the same pass, because five open PRs for a two-line dependency fix is itself now a maintenance cost.

Recommendation: reject as duplicate, in favour of #793 — but narrowly, and only because #793 now covers both undeclared imports while this one covers colorlog alone. If @Hiteshsai007 adds the PyYAML line and the trailing newline, I'd switch and merge this one instead, since it is older and closes #673. Either way this was independently and correctly diagnosed first, which is worth recording given how small the diff looks.

One thing none of the five addresses, which I've written up on #793 instead of repeating here: setup.py never passes install_requires to setup(), so pip install ianvs installs no dependencies at all regardless of which of these merges.

Hiteshsai007:
Thanks for the detailed comparison. I have updated #674 to include the missing PyYAML dependency alongside colorlog, moved the entries to the dependency section near the top, and fixed the missing trailing newline in requirements.txt.

This keeps the original Resolves #673 linkage while covering both undeclared runtime imports identified during review. Please reconsider #674 as the merge candidate; if the maintainers prefer #793, I am happy for this PR to be closed as a duplicate.

Hiteshsai007:
Thanks for testing this directly and catching the undeclared yaml import in core/common/utils.py.

You are right that adding only colorlog leaves a clean installation failing at the next import. I will update this PR to cover both colorlog and PyYAML, while keeping the scope focused on the dependency gap described by Issue #673.

I will also verify the final requirements.txt has a trailing newline and rerun both import checks from a clean environment. I am intentionally keeping the separate setup.py/install_requires concern outside this PR so the change remains limited to the colorlog/PyYAML dependency and newline repair.

ALbahrani5:
Directly resolves the colorlog/yaml failure chain I independently reproduced on a fresh Windows/pip install (ModuleNotFoundError: colorlog at core/common/log.py:18, then yaml at core/common/utils.py:24 after fixing colorlog alone). This still does not address Issue #681 (setup.py never passes install_requires to setup()), so pip install ianvs installs nothing regardless of this PR merging — that would need a follow-up fix in the same pass. Full analysis: [ https://github.com/kubeedge/ianvs/discussions/971 ]

PR #680, opened 2026-08-06
fix(metrics): harden system metrics evaluation and fix transfer ratio calculation

### LINKED ISSUE #647: [Bug] Unhandled NoneType crashes, mathematical calculation distortion, and missing validation in system metrics subsystem
**What happened**:

The system metrics evaluation subsystem in `core/testcasecontroller/metrics/metrics.py` contains several systemic defects related to missing data handling, mathematical calculation inaccuracies, and unhandled metric function resolution failures:

1. **Unhandled `NoneType` and Missing Key Crashes Across System Metrics**:
   - `samples_transfer_ratio_func()` executes `for inference_data, transfer_data in info:` on `system_metric_info.get(SystemMetricType.SAMPLES_TRANSFER_RATIO.value)`. When `info` is `None` (e.g., if the algorithm paradigm does not return sample transfer metrics), it crashes with `TypeError: 'NoneType' object is not iterable`.
   - `task_avg_acc_func()` attempts `info["accuracy"]` on `system_metric_info.get(SystemMetricType.TASK_AVG_ACC.value)`. When `info` is `None` or missing `"accuracy"`, it crashes with `TypeError` or `KeyError`.
   - `forget_rate_func()` executes `np.mean(info)` on `system_metric_info.get(SystemMetricType.FORGET_RATE.value)`. When `info` is `None` or an empty list `[]`, it crashes with `TypeError` or returns `nan` without proper type handling.
   - `bwt_func()`, `fwt_func()`, and `matrix_func()` access `info["all"]` or `info.keys()` on `system_metric_info.get(SystemMetricType.MATRIX.value)`. When `info` is `None`, they crash with `TypeError: 'NoneType' object is not subscriptable` or `AttributeError`.

2. **Mathematical Distortion in `samples_transfer_ratio_func`**:
   - Line 48 calculates `return round(float(transfer_num) / (inference_num + 1), 4)`. Adding an arbitrary `+ 1` to `inference_num` distorts the true ratio calculation. When `inference_num == 0`, it hides division-by-zero with an inaccurate `transfer_num / 1.0` result instead of returning `0.0` or `np.nan`.

3. **Unhandled `AttributeError` for Unsupported/Typoed Metric Names**:
   - In `get_metric_func()`, `getattr(sys.modules[__name__], str.lower(name) + "_func")` raises an unhandled `AttributeError` when an unknown or custom metric name is supplied without a plugin `url`, crashing the benchmarking run without listing supported metric names.

1. All system metric evaluation functions in `metrics.py` should check whether `system_metric_info` and metric-specific data (`info`) are valid and non-`None` before performing operations, returning `np.nan` or `None` gracefully if metric data is missing.
2. `samples_transfer_ratio_func` should compute the exact transfer ratio (`transfer_num / inference_num`) and handle `inference_num == 0` explicitly without artificial denominator offsets.
3. `get_metric_func()` should validate requested metric names against supported built-in metrics and raise descriptive `ValueError` exceptions when an unsupported metric name is configured.

1. Configure a `testenv.yaml` requesting `samples_transfer_ratio` or `task_avg_acc` metrics with an algorithm paradigm that returns empty or `None` system metric info.
2. Run the benchmarking job. Observe the resulting stack trace:

```python
from core.testcasecontroller.metrics import samples_transfer_ratio_func, task_avg_acc_func, forget_rate_func

# Crashes with TypeError: 'NoneType' object is not iterable
samples_transfer_ratio_func({})

# Crashes with TypeError: 'NoneType' object is not subscriptable
task_avg_acc_func({})

# Crashes with TypeError or nan formatting error
forget_rate_func({"forget_rate": None})
```

- **Affected Components**:
  - `core/testcasecontroller/metrics/metrics.py`
  - `core/testcasecontroller/testcase/testcase.py`

- **Proposed Solution Plan**:
  1. **Defensive Data Checks**: Update `samples_transfer_ratio_func`, `task_avg_acc_func`, `forget_rate_func`, `bwt_func`, `fwt_func`, and `matrix_func` to validate `info` presence and structure before processing, returning `np.nan` when data is missing.
  2. **Accurate Math Calculation**: Remove `+ 1` from `samples_transfer_ratio_func` denominator and add explicit zero check (`inference_num == 0`).
  3. **Descriptive Exception Handling**: Update `get_metric_func()` to validate built-in metric names and raise a clear `ValueError` listing available metrics when an unknown metric name is requested.
  4. **Unit Tests**: Add unit tests under `tests/test_metrics.py` covering system metric calculation edge cases, missing data handling, zero denominators, and metric function lookups.

priyam99:
Fixes #647

Adds defensive checks across system metric evaluation functions to prevent NoneType and missing key crashes, corrects transfer ratio math, and provides descriptive error handling for unknown metrics in get_metric_func.


priyam99:
Two things here that could be separated if you'd prefer: defensive guards across the system metric functions (missing keys / None values), and a correction to the transfer ratio calculation itself. Also made `get_metric_func` raise a descriptive error for unknown metric names instead of a bare KeyError.

The transfer ratio change is the one worth a close look, since it changes reported numbers rather than just preventing a crash.

/assign @moorezheng

Muhammad-Hashir-Code:
**Review — LFX 2026 Term 3 pre-test cross-PR comparison** (full analysis: https://github.com/kubeedge/ianvs/discussions/887)

Reviewed this alongside #648, which also targets #647.

**Confirmed working, directly executed:** unlike the alternative approach in #648, this version's `if not flag or length <= 1:` guard correctly handles a single-row matrix without crashing:
```python
matrix = [[]]  # length==1
compute('all', matrix)  # (None, nan, nan) no crash
```
I tested this specifically because #648's equivalent guard doesn't check `length`, and does crash with `ZeroDivisionError` on the same input so this PR closes a gap the other doesn't.

**Two smaller things worth a look, not blockers:**
1. `samples_transfer_ratio_func` returns `0.0` unconditionally when `inference_num == 0`. Since #648 instead returns `NaN` unless `transfer_num` is also `0`, this is a genuine semantic difference between the two PRs for the same edge case — might be worth deciding intentionally rather than by whichever merges first.
2. This PR (like #648) is scoped to preventing crashes in the metric-computation functions themselves. Neither adds any handling downstream in `core/storymanager/rank/rank.py`, which currently has zero NaN-awareness in its `sort_values()`-based leaderboard sort so a `NaN` returned here (correctly, per this PR's own fix) still silently sorts to last place in the leaderboard as if it were a valid low score, with no distinguishing marker. Full writeup of that mechanism is in the linked Discussion if useful context for scoping a follow-up.

Nice fix on the `+1` denominator removal too that's a real correctness bug independent of the crash-safety work.

rajdhruvsingh:
Credit to @Muhammad-Hashir-Code for the existing review, which directly executed this PR's guard against a single-row matrix and confirmed it correctly avoids a crash, and separately noted that rank.py's leaderboard sort has no NaN-awareness, so a NaN returned here currently sorts silently to last place with no distinguishing marker.

One addition on the guard's own internal consistency: the empty-matrix branch here returns None for the matrix component while the two accompanying scores are computed as NaN — a split convention (None for one field, NaN for the others) inside the same return tuple. Recommend picking one sentinel value across all three return positions rather than mixing None and NaN for what is conceptually the same "no data" case — it'll make downstream handling (including whatever addresses @Muhammad-Hashir-Code's rank.py point) easier to write consistently.

Issue #413, opened 2026-05-04
[Dependencies] Add missing colorlog and PyYAML to requirements.txt


AswaniSahoo:
## Problem

The root `requirements.txt` is missing two packages that are imported unconditionally in `core/common/`:

| Package | Import Location | Import Statement |
|---------|----------------|------------------|
| `colorlog` | `core/common/log.py` (line 18) | `import colorlog` |
| `PyYAML` | `core/common/utils.py` (line 24) | `import yaml` |

Both imports are at module level with no `try/except` guard. Following the [install guide](https://github.com/kubeedge/ianvs/blob/main/docs/guides/how-to-install-ianvs.md), a user runs `pip install -r requirements.txt` and then `pip install -e .`. Since `colorlog` and `PyYAML` are absent from `requirements.txt`, any subsequent `ianvs` CLI invocation crashes with `ModuleNotFoundError` unless the user happens to have these packages installed from another source.

## Additional Cleanup

The existing entries in `requirements.txt` were unsorted. This change also alphabetically sorts all entries for consistency and easier maintenance.

## Changes

- Add `colorlog` to `requirements.txt`
- Add `PyYAML` to `requirements.txt`
- Sort entries alphabetically

PR #682, opened 2026-08-06
fix: declare runtime package dependencies

### LINKED ISSUE #681: Fix: Packaging metadata does not include runtime dependencies from `requirements.txt`
# Packaging metadata does not include runtime dependencies from `requirements.txt`

## Summary

`setup.py` already reads dependencies from `requirements.txt`, but the parsed dependencies are never passed to `setup()` through `install_requires`. As a result, the generated package metadata does not declare Ianvs' runtime dependencies, which can lead to missing dependency errors after installation.

---

## What happened

Installing Ianvs as a Python package (for example, using `pip install .` or `pip install -e .`) does not automatically install all runtime dependencies required by the project.

Although `setup.py` parses `requirements.txt` via `_infos.basic_dependencies`, the value is never passed to the `install_requires` argument of `setup()`. Consequently, the package metadata does not list Ianvs' runtime dependencies.

This can cause the installed CLI to fail immediately with import errors such as:

```text
```

---

Installing Ianvs as a Python package should automatically install all runtime dependencies declared in `requirements.txt`, allowing the `ianvs` CLI to run without additional manual package installation.

---

## How to reproduce

1. Clone the repository.

```bash
cd ianvs
```

2. Install Ianvs as a package.

```bash
pip install -e .
```

(or)

```bash
pip install .
```

3. Run the CLI.

```bash
ianvs
```

4. Observe that startup may fail with an error similar to:

```text
```

---

## Root cause

`setup.py` already contains logic to read dependencies from `requirements.txt`.

For example:

```python
@property
def basic_dependencies(self):
    return self._read_requirements(self._requirements)
```

However, these dependencies are never passed into the package metadata via:

```python
install_requires=...
```

As a result:

- `requirements.txt` exists.
- `setup.py` successfully parses it.
- The parsed dependencies are discarded.
- The generated package metadata does not advertise the runtime requirements.

---

## Additional observations

During inspection, two runtime dependencies were also found to be required by the source code:

- `colorlog`
  - imported by `core/common/log.py`
- `PyYAML`
  - provides the `yaml` module imported by `core/common/utils.py`

If these packages are absent from the declared runtime dependencies, users may encounter import errors even after successfully installing Ianvs.

---

## Suggested fix

Use the already existing dependency parser by passing the parsed requirements into `setup()`:

```python
install_requires=_infos.basic_dependencies,
```

Also ensure that `requirements.txt` contains all runtime dependencies required by the project (including packages such as `colorlog` and `PyYAML` if they are intended runtime requirements).

---

## Impact

This issue affects package installations performed through standard Python packaging workflows, including:

- `pip install .`
- `pip install -e .`
- installation from source
- any future distribution built from the package metadata

Without proper dependency metadata, installations may succeed while the application fails immediately at runtime because required packages were never installed.

---

## Environment

- OS: macOS
- Python: 3.x
- Installation method: `pip install -e .`

iron-prog:
**What type of PR is this?**

/kind bug

This PR fixes the package dependency metadata for Ianvs by exposing the runtime dependencies parsed from `requirements.txt` through `install_requires` in `setup.py`.

It also adds the missing runtime dependencies `colorlog` and `PyYAML` to `requirements.txt`. These packages are imported by Ianvs core modules (`core/common/log.py` and `core/common/utils.py`) but were not declared as runtime dependencies, which can result in missing dependency errors when Ianvs is installed as a package.

Fixes #681


akshita317:
This addresses the cause behind a group of separately-reported dependency issues, and I think that is worth making explicit for reviewers.

`setup.py` on `main` computes the dependency list and then never uses it:

```python
@property
def basic_dependencies(self):
    return self._read_requirements(self._requirements)   # computed
```

`install_requires` appears nowhere in the `setup()` call, so the built package declares **zero** dependencies. `pip install ianvs` therefore installs core with nothing behind it, and adding entries to `requirements.txt` alone (as #674, #468, #791, #793 each do) only fixes the from-source path.

That makes this PR the upstream fix for #673, #413, #685 and #489, which are all reported as separate missing-dependency bugs.

The one-line `install_requires=_infos.basic_dependencies` is the right minimal change, and `_read_requirements` already filters comment lines, so the `# BSD` / `# MIT` annotations in `requirements.txt` will not leak into the metadata.

**One caution worth flagging before merge:** once `install_requires` is live, `requirements.txt` becomes a hard install contract for every consumer. Several entries are currently unpinned (`scikit-learn`, `numpy`, `pandas`, `tqdm`, `matplotlib`, `onnx`), which is harmless as documentation but starts participating in dependency resolution — including `onnx`, which is a heavy dependency that not every user of core needs. It may be worth pairing this with either upper bounds or an `extras_require` split, so that making the metadata authoritative does not simultaneously make the install heavier and less constrained.

hibaa23:

The `install_requires` wiring in `setup.py` is correct and addresses a real gap — confirmed `basic_dependencies` was computed but never passed into `setup()` on `main`. This is a genuine root-cause fix, not a symptom patch.

Two notes:
- Small artifact: the `setup.py` diff introduces a Windows CRLF line ending on the new `install_requires` line, in a file that's otherwise LF — worth a cleanup pass before merge to avoid a mixed-line-ending diff.
- This PR's `requirements.txt` additions (`colorlog`, `PyYAML`) overlap a five-way duplication already flagged elsewhere in this pre-test cycle (credit: a separate Discussion submission's Bonus review, PR #791 analysis) — not reclaiming that finding here. Worth adding on top: even after this PR, `sedna` has no `requirements.txt` entry and structurally can't get one under the current distribution model (no legitimate PyPI release exists — see #489). This PR makes every *other* dependency properly declared, which only sharpens the point that Sedna remains the one exception with no working install mechanism at all.

hibaa23:
Correction to my comment above: I verified the CRLF claim was wrong. `git show pr-682:setup.py` (reads the actual stored blob content, bypassing terminal/checkout line-ending conversion) shows 0 CRLF occurrences in the file — the `^M` I saw was a `git diff` terminal-rendering artifact, not a real character in the file. Retracting that point; the rest of the review stands unchanged.

kavaljeetsingh-dev:
Reviewed this as the deeper fix for #681: `setup.py`'s `basic_dependencies` property already parses `requirements.txt` correctly, but the parsed value was never passed to `install_requires=` in the `setup()` call. This PR's own commit fixes exactly that:

```
+    install_requires=_infos.basic_dependencies,
```

Confirmed this is the correct, minimal fix by isolating the PR's own commit from the base branch. One caution: this branch is currently 480 commits behind `main` (`git rev-list --count pr-682..main`), so landing it will need a rebase rather than a straightforward merge. The fix itself looks right; the staleness is the main blocker.

MukandKrishna (commented):
**Recommendation: changes requested.** The new `install_requires` wiring works when a wheel is built directly from the repository checkout, but the dependency contract is lost when the package follows the source-distribution path.

I reproduced this against PR head `2eb7067281a6ebbe63abec2830edf546a94731a9`:

```text
python -m pip wheel . --no-deps --no-build-isolation ...
Successfully built ianvs

Requires-Dist: prettytable~=2.5.0
Requires-Dist: colorlog
Requires-Dist: PyYAML
Requires-Dist: scikit-learn
Requires-Dist: numpy
Requires-Dist: pandas
Requires-Dist: tqdm
Requires-Dist: matplotlib
Requires-Dist: onnx
```

However, `python setup.py sdist` creates `ianvs-0.3.0.tar.gz` without either `requirements.txt` or `OWNERS`. Because `setup.py` reads both files at build time and silently falls back when they are absent, rebuilding a wheel from that sdist succeeds with different metadata:

```text
python -m pip wheel ianvs-0.3.0.tar.gz --no-deps --no-build-isolation ...
Successfully built ianvs

Name: ianvs
Version: 0.3.0
Author: ianvs
Maintainer: ianvs
```

The sdist-derived wheel contains **zero `Requires-Dist` fields**, so installation from the sdist recreates the undeclared-runtime-dependency bug this PR is intended to fix. It also changes the author and maintainer metadata because `OWNERS` is unavailable.

Please either include every setup-time input in the sdist (for example through `MANIFEST.in`) or move this metadata to a packaging configuration that does not depend on repository-only files. An artifact-level test should build the sdist, build a wheel from that sdist, and assert that its `Requires-Dist` set matches the wheel built directly from the checkout.

Verification boundary: I built and inspected both wheel paths locally from the exact PR head. I used `--no-deps`, so no runtime dependencies were installed and the Ianvs runtime itself was not executed.

suhaan-24:
Wiring `install_requires` is a real improvement — as it stands, `pip install ianvs` declares no
runtime dependencies at all, so a fresh install resolves nothing and fails on first import. Reading
them from `requirements.txt` through `InstallPrepare` keeps one source of truth, which is the right
shape.

One gap that I think undercuts the goal.

## The resulting manifest omits `sedna`

After this PR, `requirements.txt` reads:

```
prettytable~=2.5.0
colorlog
PyYAML
scikit-learn
numpy
pandas
tqdm
matplotlib
onnx
```

`sedna` is not there, and it is not in `setup.py` either. Ianvs Core imports it directly — for
example `core/testenvmanager/dataset/dataset.py` begins with

```python
from sedna.datasources import (
    ...
)
```

so with this PR applied, `pip install ianvs` will now install nine dependencies and still fail at
import time on the one the framework cannot start without.

It is currently shipped as a vendored wheel:

```
$ ls resources/third_party/sedna*.whl
resources/third_party/sedna-0.6.0.1-py3-none-any.whl
$ grep -n 'sedna' requirements.txt setup.py
(no matches in either file)
```

I appreciate that is probably deliberate — the wheel is not on PyPI under that name, and there is
an open issue (#489) about a name collision with an unrelated `sedna` package on PyPI, so a naive
`sedna>=0.6` line would be actively harmful. But that makes it worth stating explicitly rather than
leaving implicit: if `install_requires` is now the declared contract, a reader will reasonably infer
the list is complete.

### Options worth considering

- Leave `install_requires` as is but add a comment in `setup.py` recording that `sedna` is
  intentionally excluded and must be installed from `resources/third_party/`, so the omission is
  documented rather than looking like an oversight.
- Or declare it via a direct-reference / extras entry so the requirement is at least machine-visible.

## Smaller point on pinning

`colorlog` and `PyYAML` are added without bounds. Since `install_requires` now propagates these to
every install, a lower bound on each would make resolution reproducible. `prettytable` already
carries `~=2.5.0`, so the file is currently mixed in style.

## Verification boundary

The post-PR manifest and the absence of `sedna` were read directly from this branch and from
`upstream/main`. The `dataset.py` import is quoted from `upstream/main`. I have **not** run
`pip install` from this branch, so the failure described above is derived from the import graph
rather than observed.

shadow2022-coder:

I built head `2eb7067281a6ebbe63abec2830edf546a94731a9` as a wheel on macOS arm64 / Python 3.11.9. The artifact correctly contains nine `Requires-Dist` entries, including `onnx`, but none for Sedna. Installing the wheel and every declared dependency succeeded; the installed CLI did not:

```text
$ ianvs -v
...
from sedna.datasources import (
ModuleNotFoundError: No module named 'sedna'
exit 1
```

Installing the repository's vendored `sedna-0.6.0.1` wheel into that same environment changed the command to `0.3.0`, exit 0. This isolates the missing first-party dependency contract: the PR makes optional ONNX mandatory for every package install while the same eager startup graph still reaches an undeclared KubeEdge Sedna.

I also independently rebuilt every artifact path from this head. The checkout-built wheel has all nine `Requires-Dist` fields. The generated sdist omits `requirements.txt`, and the wheel rebuilt from that sdist has zero `Requires-Dist` fields. This confirms the existing thread finding: the same commit publishes incompatible install metadata depending on whether consumers receive a wheel or sdist.

Please keep the `install_requires` wiring, but revise the packaging design before merge: define a canonical, unambiguous KubeEdge Sedna distribution; split minimal bootstrap requirements from optional paradigm dependencies; include every setup-time input in the sdist; and test normalized direct-wheel/sdist-wheel metadata plus a clean installed `ianvs -v`. Do not add bare `sedna` to `install_requires`, because that name resolves to an unrelated PyPI project.

Cross-Example impact is repository-wide because this metadata governs every Ianvs install. Current branch comparison against `main` at `37a9c60` is 9 commits behind/1 ahead, so I am not using the older “480 commits behind” claim as a blocker.

Verification boundary: checkout wheel, sdist, wheel-from-sdist, metadata comparison, dependency installation, failing CLI startup, and manual-vendored-Sedna recovery were executed. I reproduced rather than originated the sdist finding. No full Example, dataset/model path, GPU path, or repaired-artifact test was executed.

observer04 (commented):
**Recommendation: major revision.** Connecting `basic_dependencies` to `install_requires` is the correct source-level fix, but the released contract is still path-dependent.

@MukandKrishna's formal Review and @shadow2022-coder's later comment already published the artifact finding. My independent run only confirms it: the direct wheel has nine `Requires-Dist` fields, while the sdist omits `requirements.txt`/`OWNERS` and its derived wheel has zero.

The same source revision therefore creates semantically different wheels. Other existing reviewers have also shown that KubeEdge Sedna is absent from the published dependency set and that adding the ambiguous public `sedna` name would be unsafe.

Please include or relocate every metadata input so it survives the sdist, add a CI comparison between direct-wheel and sdist-derived-wheel metadata, and run an installed-CLI smoke check after installing the canonical vendored Sedna artifact. The merged source validator’s editable `--no-deps` path is useful but does not exercise this release path.

Verification boundary: wheel/sdist construction and metadata comparison were executed locally. I did not publish the artifacts or run a full benchmark.

PR #683, opened 2026-08-06
Remove self referential ianvs dependency

### LINKED ISSUE #644: Self-referential ianvs dependency in requirements.txt breaks fresh installs (GovDoc2Poster, imagenet/multiedge_inference_bench)
**What happened**:
`pip install -r requirements.txt` fails immediately for two Ianvs examples because their `requirements.txt` lists `ianvs` itself as a pinned pip dependency:
- `examples/GovDoc2Poster/singletask_learning_bench/requirements.txt:5` → `ianvs>=0.1.0`
- `examples/imagenet/multiedge_inference_bench/requirements.txt:27` → `ianvs>=0.3.0`

`ianvs` is not published on PyPI (confirmed via `https://pypi.org/pypi/ianvs/json` → 404) — it's installed from source via `python setup.py install`, per the project's own Quick Start guide. Listing it as a version-pinned pip requirement causes the install to fail outright for any new user following the documented setup for either example.

`pip install -r requirements.txt` should succeed for both examples, resolving only genuine third-party PyPI packages, since `ianvs` itself is installed separately via `python setup.py install` before any example's own requirements are installed.

```bash
$ pip install "ianvs>=0.1.0"
ERROR: Could not find a version that satisfies the requirement ianvs>=0.1.0 (from versions: none)
ERROR: No matching distribution found for ianvs>=0.1.0
```
Same result running the full `pip install -r requirements.txt` against either of the two files listed above.

This is a repeated pattern across two independently-authored examples, not a one-off typo — likely a templated or copy-pasted requirements header that assumed `ianvs` was pip-installable. I checked related open issues (#132, #514, #535) and found none reporting this specific self-referential dependency pattern.

Note: `sedna>=0.1.0` in the GovDoc2Poster file is fine and should stay — Sedna is genuinely published on PyPI, unlike Ianvs.

Fix is a one-line deletion in each file:

```diff
--- a/examples/GovDoc2Poster/singletask_learning_bench/requirements.txt
+++ b/examples/GovDoc2Poster/singletask_learning_bench/requirements.txt
@@ -2,7 +2,6 @@
 # 新版政府代理依赖包

 # Core dependencies
-ianvs>=0.1.0
 sedna>=0.1.0

 # Document processing (轻量级方案)
```

```diff
--- a/examples/imagenet/multiedge_inference_bench/requirements.txt
+++ b/examples/imagenet/multiedge_inference_bench/requirements.txt
@@ -24,7 +24,6 @@ h11==0.14.0
 h5py==3.8.0
 huggingface-hub==0.16.4
 humanfriendly==10.0
-ianvs>=0.3.0
 idna==3.7
 imageio==2.31.2
 importlib-metadata==6.7.0
```

I'm happy to open a PR with this fix if it's welcome.

Arijit429:
/kind bug

`pip install -r requirements.txt` fails for two examples because their `requirements.txt` lists `ianvs` itself as a pinned pip dependency:

- `examples/GovDoc2Poster/singletask_learning_bench/requirements.txt` → `ianvs>=0.1.0`
- `examples/imagenet/multiedge_inference_bench/requirements.txt` → `ianvs>=0.3.0`

`ianvs` is not published on PyPI — it is installed from source via `python setup.py install` per the Quick Start guide. This self-referential dependency breaks the install for any new user following the documented setup.

This PR removes the two offending lines. `sedna>=0.1.0` in GovDoc2Poster is intentionally kept — Sedna is genuinely published on PyPI.

Fixes #644

### DIFF
--- a/examples/GovDoc2Poster/singletask_learning_bench/requirements.txt
+++ b/examples/GovDoc2Poster/singletask_learning_bench/requirements.txt
@@ -2,7 +2,6 @@
 # 新版政府代理依赖包
 
 # Core dependencies
-ianvs>=0.1.0
 sedna>=0.1.0
 
 # Document processing (轻量级方案)
--- a/examples/imagenet/multiedge_inference_bench/requirements.txt
+++ b/examples/imagenet/multiedge_inference_bench/requirements.txt
@@ -24,7 +24,6 @@ h11==0.14.0
 h5py==3.8.0
 huggingface-hub==0.16.4
 humanfriendly==10.0
-ianvs>=0.3.0
 idna==3.7
 imageio==2.31.2
 importlib-metadata==6.7.0
--- a/examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py
+++ b/examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py
@@ -7,8 +7,20 @@ from functools import wraps
 import time
 from core.common.log import LOGGER
 
-MATPLOTLIB_AVAILABLE = False
-OPEN3D_AVAILABLE = False
+try:
+    import matplotlib
+    import matplotlib.pyplot as plt
+    from matplotlib.patches import Ellipse
+    from matplotlib.collections import PatchCollection
+    MATPLOTLIB_AVAILABLE = True
+except ImportError:
+    MATPLOTLIB_AVAILABLE = False
+
+try:
+    import open3d as o3d
+    OPEN3D_AVAILABLE = True
+except ImportError:
+    OPEN3D_AVAILABLE = False
 
 
 def timeit(func):


kavaljeetsingh-dev:
Tested this directly:
```
$ git diff main <this PR> -- examples/GovDoc2Poster/singletask_learning_bench/requirements.txt examples/imagenet/multiedge_inference_bench/requirements.txt
-ianvs>=0.1.0   (GovDoc2Poster)
-ianvs>=0.3.0   (imagenet/multiedge_inference_bench)
```
Both self-referential lines removed cleanly in both affected examples. Confirmed `ianvs` is not published to PyPI, so this line only ever caused `pip install` to fail — removing it carries no regression risk.

Note: #687 (also open) proposes a byte-for-byte identical fix to this one, opened one day later.

sheikhayaan:
The dependency removal here is correct, and I think it matters more than the title suggests, so it's worth spelling out why. Separately, this PR carries a change beyond its title that I'd split out.

Two example requirement files declare the project as their own dependency: examples/GovDoc2Poster/singletask_learning_bench/requirements.txt:5 has ianvs>=0.1.0, and examples/imagenet/multiedge_inference_bench/requirements.txt:27 has ianvs>=0.3.0. setup.py names this distribution ianvs at version 0.3.0, so the reference is genuinely self-referential.

Why that's worse than cosmetic: packages=find_packages(...) publishes a top-level package named core. So either the name doesn't resolve and the install fails outright, or it does resolve and drops a generic core/ into site-packages that can shadow (or be shadowed by) the working tree's own core/, depending on sys.path order. Given that Core mutates sys.path during example loading, in py2dict and load_module, "depending on sys.path order" isn't a hypothetical here. Removing the self-reference is right either way; the core namespace exposure is worth a separate issue.

On the undeclared scope, and a correction to the comparison above. @kavaljeetsingh-dev's diff was scoped to the two requirements.txt paths, so within that scope "byte-for-byte identical to #687" is right. Across the whole PR it isn't: this one is +14 −4 over three files and #687 is −2 over two. The third file is examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py, where hardcoded MATPLOTLIB_AVAILABLE = False and OPEN3D_AVAILABLE = False become real try/except ImportError probes. That's a behaviour change in an unrelated example, switching on plotting paths that were previously dead code. It may well be correct and I'm not arguing it's wrong; it just doesn't belong in a dependency-removal PR, and the description doesn't mention it even though the commit subject does.

The convenient part is that it's already isolated in its own commit, bf79838, separate from 82fffe1 which does the dependency removal. So splitting costs a cherry-pick, not a rewrite.

Recommendation: minor revision. Split the utils.py import-guard change into its own PR with its own justification. The requirements.txt half is correct and should merge.

MukandKrishna (commented):
**Recommendation: major revision.** Removing the two self-referential `ianvs` requirements is correct, but GovDoc2Poster is still not safely installable because this PR intentionally retains `sedna>=0.1.0` and describes Sedna as a valid PyPI dependency.

I independently reproduced this in a virtual environment:

```text
python -m pip install "sedna>=0.1.0"
Collecting sedna>=0.1.0
Using cached Sedna-0.1.2-py3-none-any.whl
Successfully installed sedna-0.1.2
```

That wheel is the unrelated computational-mathematics project named `Sedna`. Its modules are `convert`, `maths`, `physics`, `prime`, `system`, and `voice`; it contains neither `sedna.common` nor `sedna.datasources`. Importing the contract used by this repository fails:

```text
from sedna.common.class_factory import ClassFactory
ModuleNotFoundError: No module named 'sedna.common'
```

GovDoc2Poster directly imports `sedna.common.class_factory` in both `testenv/acc.py` and `testalgorithms/gen/basemodel.py`. Consequently, `pip install -r requirements.txt` can move from the visible `ianvs` resolution failure to a successful installation of the wrong distribution, followed by a less obvious runtime import failure.

Please remove `sedna>=0.1.0` from this Example requirement file and use the repository's validated KubeEdge Sedna installation path (currently `resources/third_party/sedna-0.6.0.1-py3-none-any.whl`), or replace it with an explicit trusted artifact reference. A clean-environment gate should run `from sedna.common.class_factory import ClassFactory` after dependency installation.

Verification boundary: isolated dependency installation, wheel-content inspection, and the exact `sedna.common` import failure were executed against PR head `82fffe1e7c9d52a8d242cc68429fdf95f3c71c52`; the full GovDoc2Poster model pipeline was not run.

Arijit429:
> **Recommendation: major revision.** Removing the two self-referential `ianvs` requirements is correct, but GovDoc2Poster is still not safely installable because this PR intentionally retains `sedna>=0.1.0` and describes Sedna as a valid PyPI dependency.
> 
> I independently reproduced this in a virtual environment:
> 
> ```
> python -m pip install "sedna>=0.1.0"
> Collecting sedna>=0.1.0
> Using cached Sedna-0.1.2-py3-none-any.whl
> Successfully installed sedna-0.1.2
> ```
> 
> That wheel is the unrelated computational-mathematics project named `Sedna`. Its modules are `convert`, `maths`, `physics`, `prime`, `system`, and `voice`; it contains neither `sedna.common` nor `sedna.datasources`. Importing the contract used by this repository fails:
> 
> ```
> from sedna.common.class_factory import ClassFactory
> ModuleNotFoundError: No module named 'sedna.common'
> ```
> 
> GovDoc2Poster directly imports `sedna.common.class_factory` in both `testenv/acc.py` and `testalgorithms/gen/basemodel.py`. Consequently, `pip install -r requirements.txt` can move from the visible `ianvs` resolution failure to a successful installation of the wrong distribution, followed by a less obvious runtime import failure.
> 
> Please remove `sedna>=0.1.0` from this Example requirement file and use the repository's validated KubeEdge Sedna installation path (currently `resources/third_party/sedna-0.6.0.1-py3-none-any.whl`), or replace it with an explicit trusted artifact reference. A clean-environment gate should run `from sedna.common.class_factory import ClassFactory` after dependency installation.
> 
> Verification boundary: isolated dependency installation, wheel-content inspection, and the exact `sedna.common` import failure were executed against PR head `82fffe1e7c9d52a8d242cc68429fdf95f3c71c52`; the full GovDoc2Poster model pipeline was not run.

Good catch — confirmed. The `sedna>=0.1.0` line resolved to the unrelated computational-mathematics Sedna package on PyPI, not KubeEdge Sedna which is installed from `resources/third_party/sedna-0.6.0.1-py3-none-any.whl`. Removed in the latest commit. Thanks for the detailed reproduction.

suhaan-24:
The stated change is correct. `ianvs>=0.1.0` and `ianvs>=0.3.0` in an example's own
`requirements.txt` make the example depend on the framework that ships it, which either resolves to
a PyPI package that is not this repository or forces a reinstall over the local editable install.
Removing both lines is right, and it lines up with open issue #644.

## The diff contains a second, unrelated change

Alongside the two `requirements.txt` edits, this PR rewrites
`examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py`:

```python
-MATPLOTLIB_AVAILABLE = False
-OPEN3D_AVAILABLE = False
+try:
+    import matplotlib
+    ...
+    MATPLOTLIB_AVAILABLE = True
+except ImportError:
+    MATPLOTLIB_AVAILABLE = False
+
+try:
+    import open3d as o3d
+    OPEN3D_AVAILABLE = True
+except ImportError:
+    OPEN3D_AVAILABLE = False
```

That is a good fix — the flags were hardcoded `False`, so visualisation was permanently disabled
regardless of whether the libraries were present (open issue #635). But it is not a self-referential
dependency, and the PR title does not mention it, so a reviewer scanning titles will not know this
PR carries it.

## The same hunk also appears in #851

PR #851 (*"ci: add GitHub Actions workflow for example lint and requirements"*) modifies the same
file with a **byte-identical** hunk. Diffing both branches against their merge bases produces no
difference for that file.

Whichever merges second will conflict there. Since both PRs are yours, it would be worth deciding
which one owns the visualisation fix and dropping it from the other — or splitting it into its own
PR referencing #635, which would let it be reviewed on its own merits and land independently of
both the dependency cleanup and the CI workflow.

## Verification boundary

The removed `ianvs` lines, the `utils.py` hunk, and the byte-identical overlap with #851 were all
computed by diffing each branch against its merge base with `upstream/main`. I have **not** run any
of the three affected examples (`GovDoc2Poster`, `imagenet`, `industrialEI`), so the correctness of
the visualisation fix is assessed by reading rather than by exercising the plotting path.

Shikhar-Kesharwani:
## Task 2 Review — PR #683

### Recommendation: minor revision

The dependency-removal portion of this PR is technically correct and directly addresses the self-referential `ianvs` entries reported by #644.

I independently compared the PR revision with its parent and confirmed that:

- `ianvs>=0.1.0` is removed from `examples/GovDoc2Poster/singletask_learning_bench/requirements.txt`;
- `ianvs>=0.3.0` is removed from `examples/imagenet/multiedge_inference_bench/requirements.txt`.

This is the correct ownership boundary for the specific defect: an example-local requirements file should not attempt to install the Ianvs framework itself as an ordinary PyPI dependency when the example is being run from the Ianvs source environment.

I reviewed the complete current PR rather than only the dependency-removal commit. The existing discussion has already identified the unrelated `utils.py` change and the `sedna>=0.1.0` issue, so I am not claiming either of those as new findings.

My additional review concern is the installation contract exposed by this fix. Removing `ianvs` is correct, but the PR by itself does not establish a repository-wide rule describing which dependencies are owned by Ianvs core and which are owned by an individual example.

This distinction matters because the two failure modes examined in Task 1 are opposite:

1. an example can incorrectly declare the framework itself as an example dependency; or
2. an example can omit a dependency that its execution path requires because it assumes the root Ianvs environment will provide it.

Fixing only the individual requirement entries therefore prevents the immediate failure but does not by itself prevent the same dependency-boundary problem from recurring in other examples.

I recommend keeping the two `ianvs` removals, while documenting the dependency ownership/install model separately and enforcing that model through shared validation. This would make the fix more durable without moving the framework dependency into every example requirements file.

### Verification

I compared the parent and PR revisions and confirmed the two self-referential `ianvs` entries are removed.

I also reproduced the original `ianvs` dependency-resolution failure during the Task 1 investigation. My local pip installation attempt was additionally constrained by Ubuntu's PEP 668 `externally-managed-environment` restriction, so I do not claim a complete fresh installation or end-to-end benchmark execution.

The recommendation is based on the actual PR diff, repository dependency declarations, the existing PR discussion, and targeted reproduction.

Shikhar-Kesharwani:

The evidence in this PR supports **Dependency / Packaging** as the appropriate repair boundary for the self-referential `ianvs` declarations.

The defect is in the Example requirement files:

- `examples/GovDoc2Poster/singletask_learning_bench/requirements.txt`
- `examples/imagenet/multiedge_inference_bench/requirements.txt`

The affected files incorrectly **declare Ianvs itself as an external dependency**. The appropriate repair is therefore to remove those self-referential declarations rather than modify Ianvs Core runtime code.

Keeping this repair at the Dependency / Packaging boundary limits regression risk because the change affects the installation contract of the affected Examples without changing shared Core behavior.

A shared validation layer can subsequently prevent recurrence by checking the defined dependency ownership rules, but CI should enforce the contract rather than replace this concrete dependency repair.

I independently reproduced the dependency-resolution failure for `ianvs>=0.1.0`; I did not execute the complete GovDoc2Poster or ImageNet benchmark workflows, so no end-to-end restoration is claimed.

Arijit429:
The utils.py commit (bf79838) is already merged into main — it's not part of this PR's diff. Running `git diff main --name-only` on this branch shows only the two requirements.txt files. No conflict risk from this PR's side.

PR #686, opened 2026-08-06
[PHASE IV] fix(TAB): correct dataset file extension from .jsonl to .json to match README

### LINKED ISSUE #684: [PHASE IV] TAB: dataset filename mismatch (data.json vs data.jsonl) blocks benchmark execution
**What happened:**

While setting up `examples/TAB/cloud_edge_collaborative_inference_bench` from a fresh environment, following the README's dataset setup step exactly results in `ianvs` crashing instantly on execution. This is a hard blocker — not a warning, an actual fatal crash before any inference starts.

Environment: Ubuntu 24.04 (WSL2 on Windows), Ianvs main branch, tested Aug 2026.

`test_env/test_env.yaml` hardcodes the dataset path as `./data.jsonl`:
```
testenv:
  dataset:
    train_data: "./data.jsonl"
    test_data: "./data.jsonl"
```
But the README instructs users to name the file `data.json`. Since `data.jsonl` is never created, `ianvs` aborts on launch with:
```
ValueError: dataset file(./data.jsonl) is not a local file and not a absolute path.
RuntimeError: testenv config file(...) is not supported, error: dataset file(./data.jsonl) is not a local file and not a absolute path.
```

Following the documented setup steps exactly should produce a runnable benchmark, with the dataset filename referenced in `test_env.yaml` matching the filename the README instructs users to create.

**How to reproduce it (as minimally and precisely as possible):**

```
cd ~/ianvs
python setup.py install
```
(Succeeds — `ianvs -v` returns `0.3.0`)

```
cd examples/TAB/cloud_edge_collaborative_inference_bench
# README instructs: download the TAB dataset,
# rename echr_train.json -> data.json
```

```
cd ~/ianvs
ianvs -f examples/TAB/cloud_edge_collaborative_inference_bench/benchmarkingjob.yaml
```

Observe the `ValueError`/`RuntimeError` above — execution aborts before inference begins.

**Anything else we need to know:**

Suggested fix — either of the following resolves it, happy to open a PR for whichever the maintainers prefer:
1. Update `test_env.yaml`'s `train_data` / `test_data` paths from `./data.jsonl` to `./data.json`, to match the README, or
2. Update the README's rename instruction from `data.json` to `data.jsonl`, to match `test_env.yaml`

Also found and confirmed 3 smaller issues in this same example while investigating it (broken install path in the README, duplicate metric declarations in `test_env.yaml`, and an undeclared `prettytable` core dependency) — will file/PR these separately.

Found while investigating `TAB` for LFX Mentorship Fall 2026 (Ianvs Example Restoration, Phase IV). I'll be opening the PR fix for this myself.

hayyanbaig:
**What type of PR is this?**

/kind bug

Fixes the dataset filename/extension mismatch in the TAB (`cloud_edge_collaborative_inference_bench`) example that blocks benchmark execution when following the official README.

`test_env/test_env.yaml` previously hardcoded the dataset path as `./data.jsonl`, but the README instructs users to name the file `data.json`. This mismatch caused `ianvs` to crash immediately on launch with:
`ValueError: dataset file(./data.jsonl) is not a local file and not a absolute path.`
`RuntimeError: testenv config file(...) is not supported, error: dataset file(./data.jsonl) is not a local file and not a absolute path.`
This PR updates `test_env.yaml`'s `train_data` and `test_data` paths to `./data.json`, matching the filename the README already instructs users to create — no README changes needed.

Verified locally: reproduced the original crash, applied the fix, and confirmed the `data.jsonl`-related error no longer occurs — execution now correctly proceeds past this dataset field to later stages.

Fixes #684


mwemmz:
Confirmed via diff: `test_env.yaml`'s dataset reference is changed from `.jsonl` to `.json`, aligning it with what the README documents as the expected format. This is a minimal, correctly-scoped 2-line fix.

Wasn't able to verify against the actual distributed dataset file itself (not available in my environment) to confirm the file really is `.json` and not `.jsonl` — the fix is consistent with the README, which is a reasonable source of truth, but if the README itself is what's out of date (rather than the config), this would need to go the other direction instead. Worth a quick confirmation from whoever has access to the canonical dataset artifact before merge.

PR #687, opened 2026-08-06
fix: remove self-referential ianvs dependency from requirements.txt

### LINKED ISSUE #644: Self-referential ianvs dependency in requirements.txt breaks fresh installs (GovDoc2Poster, imagenet/multiedge_inference_bench)

Muhammad-Hashir-Code:
**What type of PR is this?**

/kind bug

Removes the self-referential `ianvs>=X` line from two requirements.txt files.
`ianvs` is not published on PyPI (confirmed 404 at https://pypi.org/pypi/ianvs/json)
— it's installed from source via `python setup.py install`, so listing it as a
pinned pip dependency breaks `pip install -r requirements.txt` for anyone
following the documented setup for either example:

- examples/GovDoc2Poster/singletask_learning_bench/requirements.txt
- examples/imagenet/multiedge_inference_bench/requirements.txt

Testing: verified `pip install "ianvs>=0.1.0"` fails with the exact error
described in the linked issue, and confirmed the fix by re-running
`pip install -r requirements.txt` against both files with the line removed.
I did not run either example fully end-to-end with all heavy ML dependencies
installed.

`sedna>=0.1.0` in the GovDoc2Poster file is untouched — Sedna is genuinely
published on PyPI, unlike ianvs.

Fixes #644


kavaljeetsingh-dev:
Compared this directly against #683 (also open, same target):
```
$ git diff <pr-683> <pr-687> -- examples/imagenet/multiedge_inference_bench/requirements.txt examples/GovDoc2Poster/singletask_learning_bench/requirements.txt
(no output — byte-for-byte identical fix)
```
Both remove the same self-referential `ianvs` line from the same two files. This PR was opened one day after #683 (2026-08-07 vs 2026-08-06), with no cross-reference between the two. Recommend picking one and closing the other as a duplicate rather than merging both.

sheikhayaan:
This overlaps #683, which removes the same two lines. @kavaljeetsingh-dev's comparison above recommends picking one and closing the other without saying which, so here's a reason to decide on, since a tie broken by timestamp would pick the wrong one here.

The two are identical inside the two requirements.txt files. They are not identical as PRs: this one is −2 across two files and does only what its title says, while #683 is +14 −4 across three, because it also rewrites utils.py in the unrelated industrialEI example to turn hardcoded MATPLOTLIB_AVAILABLE = False and OPEN3D_AVAILABLE = False into real try/except ImportError probes. I've written that up on #683. It's a scoped git diff on the two requirements paths that makes the pair look identical.

Two smaller things in this PR's favour. The PyPI check is stated as a specific verifiable fact, a 404 from https://pypi.org/pypi/ianvs/json, and not as an assertion. And the description draws its own verification boundary: "I did not run either example fully end-to-end with all heavy ML dependencies installed." That's the right way to report a partial verification, and it's rarer than it should be.

Recommendation: accepted as it is, and it's the one I'd merge of the pair. It's precisely scoped to the stated problem, whereas #683 bundles an undeclared behaviour change alongside it. Merging this frees #683 to be resubmitted as just the utils.py improvement, which is cleaner history for both. Worth noting the do-not-merge/invalid-commit-message label was raised and cleared by force-push, so nothing is outstanding.

I'll flag that this is a judgment call and not an obvious answer. #683 is a day earlier and the larger contribution, and rewarding the narrower PR is arguable. My reasoning is that reviewability of a two-line dependency fix shouldn't be coupled to a behaviour change in a different example, and that the split is cheap because #683's two halves are already separate commits.

TanishChahal:
Analyzing this dependency loop for the LFX pre-test. While removing the self-referential `ianvs>=X` tag prevents the immediate PyPI 404, it leaves the examples without a defined link to the core CLI. The architectural fix isn't just deleting the line, but replacing it with a relative editable install (`-e ../../..`) so local examples inherit the workspace context natively without hitting public registries. I've mapped this out in my dependency graph design here: https://github.com/kubeedge/ianvs/discussions/911.

Shikhar-Kesharwani:
## Task 4 — Target-specific restoration-path finding

PR #687 is the preferred dependency-only restoration candidate for the self-referential Ianvs dependency issue affecting GovDoc2Poster and ImageNet.

The restoration path should treat #687 as an alternative to #683, not as a sequential step with #683. Both address the same two self-referential declarations:

- `examples/GovDoc2Poster/singletask_learning_bench/requirements.txt` → `ianvs>=0.1.0`
- `examples/imagenet/multiedge_inference_bench/requirements.txt` → `ianvs>=0.3.0`

#687 is the narrower candidate because its change set contains only these dependency removals. The additional `utils.py` change present in #683 is already represented in `upstream/main`, so applying both would be redundant.

On the restoration critical path, #687 can be handled independently of the TAB dependency repair in #688; both can proceed as parallel Example-level repairs before the common validation stage.

Verification gate: after applying the selected dependency repair, confirm that neither affected requirements file contains a self-referential `ianvs` declaration, then run the existing Example dependency/validation infrastructure in a disposable environment.

If this gate fails, revert the selected dependency repair and return to the last validated state before continuing with the common validation path.

This validation has not been executed against a repaired state in this analysis, so no post-repair pass or end-to-end Example restoration is claimed.

shadow2022-coder:

The two `ianvs` deletions at head `b27b0972536b2590b16698ebfb09d351409e7178` are correct, but GovDoc2Poster still has a blocking first-party dependency error in the same manifest. The PR description deliberately retains `sedna>=0.1.0` on the premise that Sedna is genuinely published.

I resolved that exact retained requirement in an isolated Python 3.11 environment. Pip selected `Sedna-0.1.2`; its metadata says “Sedna is a python package for computational mathematics,” and its wheel contains `convert.py`, `maths.py`, `physics.py`, `prime.py`, `system.py`, and `voice.py`. Runtime verification:

```text
from sedna.datasources import BaseDataSource
ModuleNotFoundError: No module named 'sedna.datasources'
exit 1
```

Thus #687 removes the first resolver blocker but makes GovDoc2Poster silently accept the wrong package. ImageNet does not carry this Sedna line, so the two Examples have different post-merge outcomes.

Separate clean Python 3.11 resolver dry-runs confirm that split. GovDoc2Poster's full overlay resolves 120 packages but selects the unrelated Sedna 0.1.2. ImageNet does not complete its as-provided plan: pip clones the pinned ByteTrack/yolox commit, then the isolated editable-build metadata step imports missing `torch` and fails. These are dependency/bootstrap checks only, not Example runs.

Please retain the Ianvs deletions but remove/replace the misleading Sedna registry declaration and point to one canonical KubeEdge Sedna install mechanism. This must coordinate with #682: merging both unchanged still leaves the package metadata without KubeEdge Sedna. The #683 duplication is already covered in this thread and is not my finding.

Verification boundary: exact PR diff, both clean overlay dry-runs, downloaded Sedna wheel metadata/content, and failed `sedna.datasources` import were executed. The full dependency plans were not installed and the GovDoc2Poster/ImageNet benchmarks were not executed.

KiranUdayakumar:
Reviewed as part of a submitted Example Restoration pre-test analysis (full write-up: https://github.com/kubeedge/ianvs/discussions/963). This is treated as the critical PR in that submission, since it touches packaging across two separate example directories rather than a single example.

**What this PR does well:** precisely scoped two-line removal, correctly diagnosed and honestly verified at the packaging level (author explicitly discloses no full end-to-end run was performed). Agree with sheikhayaan's case for merging this over #683.

**Verdict: major revision**, upgraded from an initial minor-revision read after accounting for shadow2022-coder's finding in this thread: the `sedna>=0.1.0` line in GovDoc2Poster's `requirements.txt`, deliberately left untouched here, resolves to an unrelated PyPI package. Independently re-verified for this submission by querying PyPI's JSON API directly: the live `sedna` package (v0.1.2) is owned by an unrelated maintainer and described in its own metadata as "a python package for computational mathematics," with modules named convert, maths, physics, prime, system, and voice, nothing related to data sourcing.

Recommend: land this fix, but explicitly link to #683 as the duplicate counterpart, and separately address the `sedna` line, coordinated with any other in-flight packaging PR touching this file.

PR #688, opened 2026-08-06
[PHASE IV] fix(TAB): add missing prettytable dependency to example requirements.txt

### LINKED ISSUE #685: [PHASE IV] TAB: ianvs -v fails with ModuleNotFoundError for prettytable if only example requirements are installed
**What happened:**

Running `ianvs -v` after installing only the TAB (`cloud_edge_collaborative_inference_bench`) example's own `requirements.txt` — without also installing Ianvs core's root-level `requirements.txt` — crashes with:

```
File ".../core/storymanager/visualization/visualization.py", line 20, in <module>
    from prettytable import from_csv
ModuleNotFoundError: No module named 'prettytable'
```

`prettytable` is required by Ianvs core's `core.storymanager.visualization` module, but it is not listed anywhere in the TAB example's own `requirements.txt` — it's only present in the repo root's `requirements.txt`. A user who installs Sedna and the example-specific dependencies (as the README's example-setup steps suggest) but skips or misses the root-level install ends up with a broken `ianvs` CLI, with no indication from the example's own docs that this root dependency is required.

`ianvs -v` should either work correctly after following the example's documented install steps, or the README should explicitly and clearly call out that the repo root's `requirements.txt` must be installed first, since core CLI functionality depends on it.

**How to reproduce it (as minimally and precisely as possible):**

```
cd ~/ianvs
python setup.py install
```
(Skip installing the root-level `requirements.txt` at this point)

```
ianvs -v
```

Result:
```
ModuleNotFoundError: No module named 'prettytable'
```

Installing the root `requirements.txt` afterward resolves it:
```
cd ~/ianvs
ianvs -v
```
(Now succeeds, returns `0.3.0`)

**Anything else we need to know:**

Suggested fix — either of the following:
1. Add `prettytable` (and any other genuinely core-required packages) to the TAB example's own `requirements.txt` as a safety net, or
2. Make the README's install order explicit and unambiguous: root `requirements.txt` must be installed before any example-specific setup, since core CLI functions depend on root-level packages

Found while investigating `TAB` for LFX Mentorship Fall 2026 (Ianvs Example Restoration, Phase IV), as part of the same investigation that surfaced #684 .

hayyanbaig:
**What type of PR is this?**

/kind bug

`ianvs` core's visualization module (`core.storymanager.visualization`) depends on `prettytable`, but the TAB (`cloud_edge_collaborative_inference_bench`) example's own `requirements.txt` never listed it. A user who installs only this example's dependencies — without also installing the repo root's `requirements.txt` — ends up with a broken `ianvs` CLI:

`ModuleNotFoundError: No module named 'prettytable' `

This PR adds `prettytable` to the example's `requirements.txt` as a safety net so the example's own dependency list is self-sufficient for running `ianvs`.

Verified locally: installed only this example's `requirements.txt` in a clean environment, confirmed `ianvs -v` failed with the `ModuleNotFoundError` beforehand and succeeds after adding the dependency.

Fixes #685


whozahm3d:
### Supplementary Code Review — Leaderboard Render Protection & Dependency Lock

**Recommendation:** **Minor Revision**  
**Review Depth:** **Analysis**

#### Review Findings:
1. **Leaderboard Render Crash:** Missing `prettytable` in `examples/TAB/cloud_edge_collaborative_inference_bench/requirements.txt` causes benchmark jobs to run training/inference successfully but crash at the final visualization phase when rendering leaderboards (`ModuleNotFoundError: No module named 'prettytable'`).
2. **Dependency Drift Risk:** Adding raw `prettytable` without explicit version bounds risks downstream breaking changes on future PyPI releases. 
3. **Remedy:** Lock the dependency strictly to `prettytable==3.10.0` to match the root repository package locks and guarantee layout reproducibility.

#### Verification Boundary:
* **Manifest & Resolution Check:** Confirmed that `prettytable==3.10.0` matches root locks.

Shikhar-Kesharwani:
## Task 2 Review — PR #688

### Recommendation: minor revision

This PR directly addresses Issue #685 by adding `prettytable` to:

`examples/TAB/cloud_edge_collaborative_inference_bench/requirements.txt`

The targeted change is technically effective. The PR description states that the TAB example's own requirements previously omitted `prettytable`, even though the Ianvs core visualization module imports it, and that installing only the example requirements reproduced the `ModuleNotFoundError`.

I independently reproduced the same dependency boundary failure locally:

`ModuleNotFoundError: No module named 'prettytable'`

I then installed `prettytable==2.5.0` in an isolated environment and verified:

`prettytable import succeeded`

This provides direct evidence that the dependency added by the PR resolves the specific import failure.

### Root problem versus symptom

The immediate symptom is the missing `prettytable` import. The deeper issue is the undefined dependency contract between Ianvs core requirements and example-local requirements.

The PR makes the TAB requirements more self-sufficient, which is useful if example requirements are intended to support running an example independently. However, `prettytable` is also a dependency of Ianvs core. Duplicating a core dependency in an example can therefore create two possible sources of version ownership unless the project explicitly defines which layer controls the dependency.

This matters beyond TAB because the same dependency-boundary ambiguity can produce either of the Task 1 failure modes: an example can omit a dependency that its execution path needs, or different examples can independently declare and version the same core dependency.

### Existing discussion versus additional review

I am not claiming the missing `prettytable` defect itself as a new finding because it is the purpose of this PR and is already documented in the PR discussion.

I am also not claiming the general unbounded-version concern as a new finding because an existing review has already raised dependency drift and version locking.

My additional review concern is therefore the ownership/install contract: the project should explicitly state whether an example's `requirements.txt` is intended to be independently installable or whether it is intentionally layered on top of the Ianvs root requirements.

Without that rule, adding `prettytable` to TAB fixes this particular example but does not prevent the same class of missing-core-dependency problem from appearing in other examples.

### Cross-example and regression impact

The immediate code change is isolated to TAB and is low-risk for the reported import failure.

The broader risk is inconsistency across examples. If some examples duplicate Ianvs core dependencies while others rely on the root environment, future dependency upgrades can leave examples with different versions of the same core package.

A shared dependency policy and validation mechanism would therefore be preferable to solving each omission independently.

### Verification


I inspected the parent and PR revisions and confirmed that `prettytable` is added to the TAB example requirements.

I independently reproduced the missing import and then verified successful import after installing `prettytable==2.5.0` in an isolated environment.

The attached recording shows the successful isolated-environment installation and import verification.

I did not execute the complete TAB benchmark workflow, so no end-to-end benchmark success is claimed.


https://github.com/user-attachments/assets/30fb0d3e-c6d6-40b5-b2f5-77a0e698f5f7



### Conclusion

The PR is a valid targeted fix for #685 and should be retained. I recommend **minor revision** to clarify the dependency ownership/install contract so that the fix establishes a consistent model for examples rather than becoming another isolated dependency declaration.

Specifically, the revision should document whether Example-local `requirements.txt` files are expected to be independently installable or are expected to be installed together with the Ianvs root requirements, and the validation policy should enforce that chosen dependency-ownership model consistently across Examples.

Shikhar-Kesharwani:

The evidence in this PR supports **Dependency / Packaging** as the appropriate repair boundary for the TAB dependency failure.

The immediate problem is that `prettytable` is required by the Ianvs Core visualization path, while the TAB Example requirements originally did not **declare** it. The targeted requirements change therefore addresses the installation dependency boundary rather than requiring a modification to Ianvs Core visualization code.

A Core-level change would introduce a broader regression surface because the visualization module is shared by multiple Examples. Keeping the immediate repair in the TAB dependency declaration limits the runtime regression surface while allowing the repository to define a consistent Core-versus-Example dependency ownership policy separately.

CI / Validation should then act as the enforcement boundary by checking the defined dependency contract across Examples rather than moving this Example-specific repair into Core.

I independently reproduced the missing `prettytable` import and verified successful import after installing the dependency. I did not execute the complete TAB benchmark workflow, so no end-to-end benchmark success is claimed.

Shikhar-Kesharwani:
## Task 4 — Target-specific restoration-path finding

PR #688 is the independent restoration candidate for the missing `prettytable` dependency in the TAB Example.

The current `upstream/main` requirements for `examples/TAB/cloud_edge_collaborative_inference_bench/requirements.txt` do not declare `prettytable`, while #688 adds the missing dependency. This repair is independent of the GovDoc2Poster/ImageNet self-referential `ianvs` dependency repair addressed by #687, so the two repairs can proceed in parallel rather than as sequential steps.

On the restoration critical path, #688 should be completed as the TAB-specific dependency repair before the common Example validation stage.

Verification gate: after applying the repair, confirm that the TAB requirements explicitly declare `prettytable`, then run the existing Example dependency/validation infrastructure in a disposable environment.

If the dependency or validation gate fails, revert the #688 requirements change and return to the last validated state before continuing with the common validation path.

This validation has not been executed against a repaired TAB state in this analysis, so no post-repair pass or end-to-end TAB restoration is claimed.

shadow2022-coder:

Adding `prettytable` resolves the first import in #685, but it does not establish the claimed self-sufficient TAB environment. At head `1f639fbd2bfe54b57190630c87fe0bf383242253`, I built an isolated Python 3.11 bootstrap containing `prettytable`, every non-ONNX Core requirement, and the repository's correct Sedna wheel. With ONNX absent:

```text
$ python -m core.cmd.benchmarking -v
...
from .multiedge_inference import MultiedgeInference
...
import onnx
ModuleNotFoundError: No module named 'onnx'
exit 1
```

This happens before argument processing because Core imports the complete paradigm registry at startup. TAB's exact requirements do resolve in a clean Python 3.11 dry-run (94-package plan including `prettytable`), so the added line is technically effective at its narrow resolver goal.

In a **separate diagnostic prototype**, not in this PR, I moved ONNX's import to `_partition()`, its only use site. The focused matrix then produced:

```text
A. no ONNX, `ianvs -v`: 0.3.0, exit 0
B. no ONNX, `_partition()`: actionable missing-dependency RuntimeError, exit 2
C. ONNX 1.22.0: minimal model split into two valid submodels with expected map, exit 0
```

This demonstrates that the shared Core seam is viable for the generated minimal graph; real Example partition workflows still require testing. It does not make PR #688 itself a successful runtime execution.

`prettytable` is directly imported by Core visualization and already belongs to root requirements. A revised #682 should install it for every Ianvs consumer. Duplicating it, unbounded, in TAB creates a second version owner and would become redundant under that package contract. The file also still contains `matplotlib>=3.7.0` twice; this patch does not deduplicate it.

Cross-Example impact: the eager registry is shared by all CLI invocations, while this patch only changes TAB. The added line is relevant rather than harmful/unnecessary, so rejection would be too strong; however, it cannot merge as a complete restoration fix.

Please revise this against an explicit ownership contract. Preferred: let corrected Ianvs packaging own `prettytable`, keep TAB as an Example-specific overlay, remove the duplicate `matplotlib` line, and coordinate the lazy-ONNX Core companion fix. If maintainers instead require independently installable Example manifests, retain a bounded `prettytable` entry and enforce that policy consistently across Examples.

Verification boundary: TAB resolver dry-run, exact PR-as-provided CLI failure, and the separate three-case lazy-ONNX prototype (including generated minimal-model partition) were executed. TAB config construction later stops on absent `data.jsonl`. No full TAB dataset/model/GPU benchmark or real Example-model partition was executed. No video is claimed because the PR-as-provided runtime failed; only the separate prototype succeeded.

Yahdel:
This is a useful small fix for TAB. Adding `prettytable` to the example's requirements means someone who installs only that example is less likely to hit a missing-package error.

However, it only fixes one part of the problem. I tested the command with several benchmark-only imports blocked. Adding `prettytable` does not stop `ianvs -v` or `ianvs --help` from loading the full benchmark code before reading the command. A different missing package can still make a simple version or help command fail.

So I see #688 as a reasonable dependency fix, but not the full command-startup fix. PR #811 is needed to delay benchmark loading until a real benchmark command is used. With #811 plus #672, version and help work from both command paths even when the benchmark-only imports are unavailable.

Suggestion:

1. keep this PR as the TAB dependency fix;
2. merge the command-startup fix separately;
3. add a test that installs TAB requirements only and checks both the TAB workload and the basic command help/version behavior.

I tested command startup and import behavior, not a full TAB benchmark run. Full context: Discussion #961.

PR #689, opened 2026-08-06
[PHASE IV] fix(phys_scene_gen): migrate langchain to langchain-core/langchain-openai for pydantic v2 compatibility

### LINKED ISSUE #666: [Phase IV] phys_scene_gen: fastapi/pydantic version conflict breaks ianvs after following README install steps
## Overview

While setting up `examples/phys_scene_gen/singletask_learning_bench` from a fresh environment, following the README's install steps exactly and in order results in `ianvs` itself becoming completely unusable. This is a hard blocker — not a warning, an actual crash on `ianvs -v`.

## Environment
- OS: Ubuntu 24.04 (WSL2 on Windows)
- Python: 3.10.18 (conda env, as specified in the README)
- Ianvs: main branch, tested Aug 2026

## Reproduction Steps

**1. Create environment (per README):**
```
conda create --name phys_scene_gen python=3.10.18
conda activate phys_scene_gen
```

**2. Install Ianvs core:**
```
pip install -e .
```
Succeeds. `ianvs -v` works correctly at this point.

**3. Install example-specific dependencies:**
```
cd examples/phys_scene_gen/singletask_learning_bench
```
This pulls in `langchain==0.0.171`, which forces `pydantic==1.10.22`.

Pip warns but does not block:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed.
fastapi 0.141.1 requires pydantic>=2.9.0, but you have pydantic 1.10.22 which is incompatible.
onnx 1.22.0 requires typing_extensions>=4.15.0, but you have typing-extensions 4.14.1 which is incompatible.
```

**4. Run `ianvs -v` again:**
```
ImportError: cannot import name 'IncEx' from 'pydantic.main'
```

Full traceback (trimmed to relevant frames):
```
File ".../fastapi/types.py", line 7, in <module>
    from pydantic.main import IncEx as IncEx
ImportError: cannot import name 'IncEx' from 'pydantic.main'
```

Root cause: `fastapi` (used by Ianvs core via `sedna.service.server.inference`) requires `pydantic>=2.9.0` for the `IncEx` type. This example's `requirements.txt` silently downgrades `pydantic` to `1.10.22` via `langchain==0.0.171`, breaking `ianvs` entirely — not just this example.

## Hard Blocker
- `ianvs` command becomes non-functional system-wide after following this example's install steps, due to the `pydantic` downgrade

## Secondary Issues (non-fatal, but worth fixing)
- `onnx 1.22.0` requires `typing_extensions>=4.15.0`, example environment ends up with `4.14.1`
- The README's own steps conflict with each other on `numpy`:
  - Step: `pip install --extra-index-url https://ai2thor-pypi.allenai.org ai2thor==...` installs `numpy 2.2.6`
  - Next README step: `pip install numpy==1.24.3` downgrades numpy again
  - But `opencv-python` (a transitive dependency pulled in during ai2thor install) requires `numpy>=2`, so the final pinned state leaves `opencv-python`'s requirement unmet

## Suggested Fix
Pin a `pydantic`/`langchain` version in this example's `requirements.txt` that's compatible with Ianvs core's `fastapi` requirement (`pydantic>=2.9.0`), or isolate this example into its own venv separate from Ianvs core's shared environment.

## Related
Also opened a small doc fix for an incorrect install path in the same README: https://github.com/kubeedge/ianvs/pull/665

Found while investigating `phys_scene_gen` for LFX Mentorship Fall 2026 (Ianvs Example Restoration, Phase IV). Happy to help test a fix or take this on directly if pointed in the right direction.

hayyanbaig:
**What type of PR is this?**

/kind bug

Fixes issue #666 in full — both the hard blocker and both secondary compatibility issues found while investigating the `phys_scene_gen` (`singletask_learning_bench`) example for LFX Mentorship Fall 2026 (Ianvs Example Restoration, Phase IV).

**The hard blocker:**

This example's `requirements.txt` previously pinned `langchain==0.0.171`, which force-downgrades `pydantic` to `1.10.22`. This directly conflicts with Ianvs core's `fastapi` requirement (`pydantic>=2.9.0`), and following this example's documented install steps broke `ianvs` **system-wide**, not just for this example:

`ImportError: cannot import name 'IncEx' from 'pydantic.main' `

Any user following the README top-to-bottom would end up with a fully non-functional `ianvs` CLI.

**Fix for the hard blocker:**
- Replaced `langchain==0.0.171` with the modern split packages `langchain-core==0.3.29` and `langchain-openai==0.2.14`, both of which support `pydantic>=2`
- Removed the now-unnecessary `pydantic==1.10.22` pin, letting it correctly resolve to Ianvs core's required `pydantic>=2.9.0`
- Bumped `openai==0.27.6` (the old pre-1.0 SDK) to `openai==1.58.1`, required by the new `langchain-openai`
- Removed `openapi-schema-pydantic==1.2.4`, a legacy pydantic-v1-only helper no longer used by anything in this dependency chain
- Adjusted `packaging==25.0` down to `packaging==24.2` to satisfy `langchain-core`'s dependency floor
- Migrated all 10 files under `testalgorithms/ai2holodeck/generation/` off the old langchain 0.0.x top-level import style (`from langchain import PromptTemplate, OpenAI`), which no longer exists in modern langchain, to the current split-package imports (`from langchain_core.prompts import PromptTemplate`, `from langchain_openai import OpenAI`)

**The two secondary issues (also fixed):**
- `onnx 1.22.0` required `typing_extensions>=4.15.0`, but the environment resolved to `4.14.1`. Fixed by bumping the pin to `typing_extensions>=4.15.0`.
- The README's own install steps conflicted with each other on `numpy`: installing `ai2thor` pulls in `numpy 2.2.6`, but a later documented step manually downgrades to `numpy==1.24.3` — leaving `opencv-python`'s transitive requirement of `numpy>=2` unmet. Fixed by bumping the pin to `numpy>=2.2.6`, removing the need for the conflicting manual downgrade.

**Also fixed in passing:** `requirements.txt` itself was encoded as UTF-16LE with CRLF line endings and a UTF-8 BOM — invisible when viewed normally, but silently broke standard Unix text tooling (`sed`, `grep`) run against it. Converted to plain UTF-8 with LF line endings, consistent with the rest of the repo.

Fixes #666

**Testing done**:

Verified locally end-to-end:
- Reproduced the original `ImportError: cannot import name 'IncEx'` crash from a clean environment
- Applied all dependency and import fixes described above
- Confirmed `ianvs -v` now returns `0.3.0` cleanly with no crash or warnings
- Confirmed all 10 migrated files import successfully with no errors
- Did not run the full end-to-end benchmark (requires an OpenAI API key and large Holodeck/GAPartNet dataset downloads — out of scope for this environment-level fix). This PR resolves the blocker preventing `ianvs` from running at all; the benchmark's actual generation pipeline was not exercised.

Related: also opened a small doc fix for an incorrect install path in the same README (#665).


observer04 (commented):
**Recommendation: major revision.** The LangChain/Pydantic migration is moving in the right direction, but the dependency file also introduces a separate runtime regression.

The manifest keeps `torch==1.13.1` and changes NumPy from `==1.24.3` to `>=2.2.6`. I reproduced the first reached array bridge with Python 3.10.20 and `torch==1.13.1+cpu` held constant:

```text
numpy 1.24.3 -> torch.from_numpy PASS, exit 0
numpy 2.2.6  -> _ARRAY_API not found; RuntimeError: Numpy is not available, exit 1
numpy 1.26.4 -> torch.from_numpy PASS, exit 0
```

This PR’s code reaches that interface in `objaverse_retriever.py:77,82` while converting the CLIP/SBERT feature arrays. Resolution and imports can succeed before this call fails.

There is also a reviewability problem: the current manifest is UTF-16, so git and GitHub show its PR change as a binary file with zero line additions or deletions.

Please either retain Torch 1.13.1 and constrain NumPy below 2 (`1.26.4` is a tested point), or update Torch/Torchvision/NumPy as one tested matrix. I would also re-encode the manifest as UTF-8 in a content-neutral commit so the version changes receive a normal line review.

Verification boundary: the three-point NumPy/Torch bridge and source reachability were executed on CPU. I did not run the Objaverse/Holodeck assets, GPU path, or full benchmark.

Mahendr99ar:
Cross-posting from my LFX 2026 Term 3 pretest analysis (full review: discussion #984).

Recommendation: minor revision. This correctly migrates off langchain==0.0.171 (which forced pydantic==1.10.22, conflicting with fastapi's pydantic>=2.9.0 floor — see #666), moving to langchain-core/langchain-openai instead. Right direction for the reported instance.

One open question I haven't verified: whether langchain-core/langchain-openai preserve the exact prompting/chain behavior phys_scene_gen's benchmark depends on — I only checked the dependency-resolution side, not runtime behavior.

Broader context: this fixes one instance of a pattern also seen in #735 (protobuf/onnx) — full analysis in discussion #984.

Issue #157, opened 2024-10-17
Pylint Errors Due to Compatibility Issues with Pylint3.9


XueSongTap:
**Description:**

Currently, this repository runs CI checks using `pylint` versions 3.7, 3.8, and 3.9.

In `pylint` 3.9, a new check `too-many-positional-arguments` was introduced. This check is triggering warnings in `core/testenvmanager/dataset/dataset.py` at several locations:

```bash
core/testenvmanager/dataset/dataset.py:119:4: R0917: Too many positional arguments (8/5) (too-many-positional-arguments)
core/testenvmanager/dataset/dataset.py:206:4: R0917: Too many positional arguments (6/5) (too-many-positional-arguments)
core/testenvmanager/dataset/dataset.py:213:4: R0917: Too many positional arguments (7/5) (too-many-positional-arguments)
core/testenvmanager/dataset/dataset.py:246:4: R0917: Too many positional arguments (7/5) (too-many-positional-arguments)
core/testenvmanager/dataset/dataset.py:285:4: R0917: Too many positional arguments (7/5) (too-many-positional-arguments)
core/testenvmanager/dataset/dataset.py:329:4: R0917: Too many positional arguments (7/5) (too-many-positional-arguments)
core/testenvmanager/dataset/dataset.py:368:4: R0917: Too many positional arguments (6/5) (too-many-positional-arguments)
```

**Long-term solution:**

Refactor the affected function definitions to reduce the number of positional arguments.

**Short-term solution:**

Add the following directive to disable the `too-many-positional-arguments` check:

```python
# pylint: disable=too-many-positional-arguments
```

**Problem:**

Adding this directive causes `pylint` versions 3.7 and 3.8 to raise an error because they do not recognize the `too-many-positional-arguments` message:

```
core/testenvmanager/dataset/dataset.py:119:0: W0012: Unknown option value for 'disable', expected a valid pylint message and got 'too-many-positional-arguments' (unknown-option-value)
```

**Question:**

I'm not sure how to resolve this issue in the short term. One possible solution is to keep only `pylint` 3.9 in the CI pipeline, as Python 3.9 is backward compatible with 3.7, and `pylint` 3.9 provides stricter checks. Would this be an acceptable approach, or is there a better way to handle this situation?

FuryMartin:
## 原因分析
[...]

PR #694, opened 2026-08-07
build: update CI to test Python 3.9-3.12 and drop EOL versions

### LINKED ISSUE #693: [Enhancement] CI tests only Python 3.7–3.9; should include 3.10+ and drop EOL versions
The CI workflow (`.github/workflows/main.yaml`) tests against Python `3.7`, `3.8`, and `3.9`. All three versions have reached end-of-life:
- Python 3.7 EOL: June 2023
- Python 3.8 EOL: October 2024
- Python 3.9 EOL: October 2025

Meanwhile, `setup.py` declares `python_requires=">=3.6"`, which is even more outdated.

**Proposed Changes**
1. Update CI matrix to `["3.9", "3.10", "3.11", "3.12"]` (or at least add 3.10+).
2. Update `setup.py` `python_requires` to `">=3.9"`.
3. Update `setup.py` assertion to `assert sys.version_info >= (3, 9)`.
4. Update GitHub Actions from `actions/checkout@v3` / `actions/setup-python@v3` to `@v4`.

Hiteshsai007:
Resolves #693 by updating the CI and project configurations to test against actively supported Python versions and drop those that are End-of-Life (EOL).

Changes included:
1. Updated `.github/workflows/main.yaml` matrix to test Python `3.9`, `3.10`, `3.11`, and `3.12`.
2. Updated `setup.py` `python_requires` to `>=3.9`.
3. Updated `setup.py` runtime assertion to `sys.version_info >= (3, 9)`.
4. Upgraded GitHub Actions `actions/checkout` and `actions/setup-python` to `v4`.
5. Simplified the `pylint` step in CI since all tested Python versions are now >=3.9, which all require the `--max-positional-arguments=10` workaround (originally introduced for 3.9 in #157).

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -15,11 +15,11 @@ jobs:
     name: pylint
     strategy:
       matrix:
-        python-version: [ "3.7", "3.8", "3.9" ]
+        python-version: [ "3.9", "3.10", "3.11", "3.12" ]
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
@@ -31,11 +31,7 @@ jobs:
           python -m pip install ${{github.workspace}}/resources/third_party/*
           python -m pip install -r ${{github.workspace}}/requirements.txt
       - name: Analysing code of core with pylint
-        # `--max-positional-arguments=10` is set for Python 3.9 to avoid `R0917: too-many-positional-arguments`.
+        # `--max-positional-arguments=10` is set for Python >= 3.9 to avoid `R0917: too-many-positional-arguments`.
         # See details at https://github.com/kubeedge/ianvs/issues/157
         run: |
-          if [ "${{ matrix.python-version }}" = "3.9" ]; then
-            pylint --max-positional-arguments=10 '${{github.workspace}}/core'
-          else
-            pylint '${{github.workspace}}/core'
-          fi
+          pylint --max-positional-arguments=10 '${{github.workspace}}/core'
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ import os
 
 from setuptools import setup, find_packages
 
-assert sys.version_info >= (3, 6), "Sorry, Python < 3.6 is not supported."
+assert sys.version_info >= (3, 9), "Sorry, Python < 3.9 is not supported."
 
 
 class InstallPrepare:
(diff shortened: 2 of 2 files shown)


Yash4616:
Reviewed against `main` at `37a9c60`. Conflicts against `main` as it stands.

The `setup.py` half of this PR is the part I would keep:

```diff
-assert sys.version_info >= (3, 6), "Sorry, Python < 3.6 is not supported."
+assert sys.version_info >= (3, 9), "Sorry, Python < 3.9 is not supported."
-    python_requires=">=3.6",
+    python_requires=">=3.9",
```

`setup.py` is not one of the files renamed in `7f7776a`, so this applies to a live target unaffected by the workflow question, and it closes a real gap: the declared floor and the tested matrix currently disagree.

Two problems in the workflow half.

The `setup-python` bump is to v4, while `main` currently uses `actions/setup-python@v7`. That line would move the workflow backwards.

More substantive, this removes the version conditional around the pylint flag:

```diff
-          if [ "${{ matrix.python-version }}" = "3.9" ]; then
-            pylint --max-positional-arguments=10 '${{github.workspace}}/core'
-          else
-            pylint '${{github.workspace}}/core'
-          fi
+          pylint --max-positional-arguments=10 '${{github.workspace}}/core'
```

`--max-positional-arguments` is only accepted by newer pylint, and the install step does not pin a pylint version. On a 3.9 to 3.12 matrix this is probably safe today, but if resolution ever produces an older pylint on any matrix entry, the job fails on argument parsing rather than on code quality, and the error will not obviously point here. The conditional being removed exists to avoid exactly that. Pinning pylint in the install step would make the removal safe.

Recommendation: minor revision. Keep the `setup.py` change, drop the action bump, and pin pylint or keep the conditional.

observer04 (commented):
**Recommendation: minor revision.** The two `setup.py` changes are useful and apply cleanly to current main:

```text
python_requires -> >=3.9
default_version -> remains 0.3.0
setup.py conflict markers -> 0
```

The PR as a whole still conflicts in the renamed pylint workflow. Its branch pins `actions/checkout` and `actions/setup-python` at v2–v4, while current main is uniformly on v7. The existing review on this thread also correctly notes the unconditional, unpinned pylint flag concern.

That existing review already owns the rename analysis, the v4-versus-v7 regression, the pylint conditional, and the recommendation to keep the `setup.py` half. I credit those conclusions rather than presenting them as new. My added evidence is the controlled merge comparison below.

Please rebase, retain the current v7 Actions, and resolve/justify the pylint command while keeping the focused `setup.py` change. This is a small conflict resolution rather than a redesign, which is why I classify it as minor.

For comparison, I ran the same merge with #623: #623 conflicts across `setup.py`, whereas this PR’s two-line version auto-merges and preserves the current project version. The difference is the whole-file line-ending rewrite, not the intended Python policy.

Verification boundary: both merges were executed against current main. The 3.9–3.12 CI matrix itself was not run.

Issue #695, opened 2026-08-07
[Bug] RoboDK Palletizing example fails with the published Kaggle dataset because the dataset index files contain hardcoded absolute paths


SurbhiD404:
**Body:**
```markdown
**What happened**:

The RoboDK Palletizing Dataset published on Kaggle
(https://www.kaggle.com/datasets/kubeedgeianvs/the-robodk-palletizing-dataset)
— the dataset currently referenced by this example's README as the
official download source — ships with `train_index.txt` and
`test_index.txt` containing hardcoded absolute paths from the original
contributor's environment, for example:

```
/root/ianvs/project/data/dataset/RoboDK_Palletizing_Dataset/images/train/snapshot_20250903_105641.png /root/ianvs/project/data/dataset/RoboDK_Palletizing_Dataset/labels/train/snapshot_20250903_105641.txt
```

Because these entries are absolute paths pointing to the original
contributor's environment, the downloaded dataset cannot be used directly
on another machine without manually modifying the index files.

Training fails. In my testing (Ultralytics 8.4.115), the error was:
```
ValueError: train: No labels found in \root\ianvs\project\data\dataset\RoboDK_Palletizing_Dataset\labels\train.cache
```
The exact wording may differ slightly across Ultralytics versions, but the
underlying cause is the same: the loader attempts to open image/label
files under `/root/ianvs/project/...`, a path that does not exist on the
local machine.

**Root cause**: The published dataset archive contains absolute paths
embedded directly in `train_index.txt` and `test_index.txt`. Ianvs already
supports resolving relative paths via `_process_txt_index_file()` in
`core/testenvmanager/dataset/dataset.py`. However, because the entries are
already absolute, they are treated as valid paths and are not rewritten
relative to the dataset's actual location. Therefore the issue originates
from the published dataset's contents, rather than from the ianvs
dataset-loading implementation.

**What you expected to happen**:

The dataset's index files should use relative paths (e.g.
`images/train/snapshot_....png labels/train/snapshot_....txt`), consistent
with how `_process_txt_index_file()` already supports relative-path
resolution for other examples in the repository. With relative paths,
ianvs correctly resolves each entry relative to the index file's own
directory, making the dataset portable across machines and operating
systems with no manual editing required.

**How to reproduce it (as minimally and precisely as possible)**:

1. Follow the example's README to install ianvs and download the dataset
   via the Kaggle API:
   ```bash
[...]

PR #696, opened 2026-08-07
Fix RoboDK Palletizing example: broken config paths, dataset portability, and weight-discovery bug

SurbhiD404:
## What this PR does

Restores the `RoboDK Palletizing` singletask_learning_bench example to a
working state. Previously the example failed immediately due to stale
configuration paths, and even after fixing those, failed again due to the
published dataset's absolute-path index files and a post-training
weight-discovery bug.

## Changes

- **`benchmarkingjob.yaml`, `algorithm.yaml`, `testenv.yaml`**: replaced
  hardcoded, non-existent paths
  (`/root/ianvs/project/ianvs-0.3.0/examples/Palletizing/...`) with correct
  repository-relative paths matching the actual folder name
  (`examples/RoboDK Palletizing/singletask_learning_bench/...`).

- **`scripts/convert_dataset_indices.py`** (new): idempotent utility that
  converts the Kaggle-published dataset's absolute index paths to relative
  ones, so ianvs's existing path-resolution logic can locate the dataset
  regardless of where the user places it or which OS they're on.

- **`testalgorithms/basemodel.py`**: improved post-training weight
  discovery by checking both `self.workspace` and Ultralytics' nested
  output directory (`runs/detect/<workspace-name>`), allowing the example
  to locate the trained weights across the observed output layouts.

- **`readme.md`**: documented the dataset-conversion step, and fixed the
  README's own unzip command to match the actual downloaded filename
  (`the-robodk-palletizing-dataset.zip`, not `RoboDK_Palletizing_Dataset.zip`).

## Verification

Ran the full benchmark end-to-end multiple times on a clean Windows checkout:
1. Downloaded the dataset via the Kaggle API as documented.
2. Ran `convert_dataset_indices.py` to portabilize the index files.
3. Ran `ianvs -f "examples/RoboDK Palletizing/singletask_learning_bench/benchmarkingjob.yaml"`.
4. Confirmed successful training, evaluation, and a populated results table
   (e.g. mAP50 ≈ 0.90 at 2 epochs — reduced epoch count used only for fast
   local verification; `algorithm.yaml` ships with the original 30-epoch
   setting).

## Scope notes

- The dataset's absolute-path packaging issue is tracked separately in
  #695, since the root cause is in the published Kaggle dataset, not this
  repository. `convert_dataset_indices.py` is a compatibility utility for
  that dataset, not a framework change.

## Related

Related to #695.

### DIFF
--- a/examples/RoboDK Palletizing/singletask_learning_bench/benchmarkingjob.yaml
+++ b/examples/RoboDK Palletizing/singletask_learning_bench/benchmarkingjob.yaml
@@ -2,13 +2,13 @@ benchmarkingjob:
   name: "palletizing_bench"
   workspace: "./workspace/palletizing_bench"
 
-  testenv: "/root/ianvs/project/ianvs-0.3.0/examples/Palletizing/singletask_learning_bench/testenv/testenv.yaml"
+  testenv: "./examples/RoboDK Palletizing/singletask_learning_bench/testenv/testenv.yaml"
 
   test_object:
     type: "algorithms"
     algorithms:
       - name: "YOLOv8n"
-        url: "/root/ianvs/project/ianvs-0.3.0/examples/Palletizing/singletask_learning_bench/testalgorithms/algorithm.yaml"
+        url: "./examples/RoboDK Palletizing/singletask_learning_bench/testalgorithms/algorithm.yaml"
 
   rank:
     sort_by:
--- a/examples/RoboDK Palletizing/singletask_learning_bench/readme.md
+++ b/examples/RoboDK Palletizing/singletask_learning_bench/readme.md
@@ -183,8 +183,33 @@ python -m pip install kaggle
 kaggle datasets download kubeedgeianvs/the-robodk-palletizing-dataset
 
 # Unzip the dataset
-unzip RoboDK_Palletizing_Dataset.zip
+unzip the-robodk-palletizing-dataset.zip
 ```
+**Important — dataset path portability:**
+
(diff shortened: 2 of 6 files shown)


MooreZheng (commented):
1. the convert step is not necessary; need to remove the code and redundant steps in the readme file
2. testenv.yaml from absolute path to related path 
3. add the video

diegodevxd:
The reviewer already asks to remove the dataset-index conversion. Independently, the PR combines three recovery concerns with different ownership: checked-in configuration paths, external Kaggle index portability, and Ultralytics output-layout/weight discovery. A single end-to-end claim cannot isolate which of those changes fixed a future regression.

Please separate configuration closure from model-output discovery and give each a focused test/verification step. Dataset conversion should stay out unless the project decides to own that external dataset format; the config patch can then be compared cleanly with the competing RoboDK path-only proposals.

ROGUESA1LOR:
Beyond the config-path and dataset-conversion concerns already raised: even
once this PR's fixes land, the underlying bug class — absolute,
machine-specific dataset paths in testenv.yaml — is not caught by any
existing CI check (verified against PR #835's validator: reports 0 broken
paths for a synthetic absolute dataset path). The identical pattern exists
independently in Cloud_Robotics (#816), government_rag, and smart_coding.
Recommend pairing this Example-local fix with a shared CI/validator-layer
check for the 8 dataset config fields so future Examples can't reintroduce
the same issue. Full analysis:
https://github.com/kubeedge/ianvs/discussions/953

PR #697, opened 2026-08-07
fix(examples): restore curb-detection lifelong learning benchmark and update strategy registry

### LINKED ISSUE #254: [PRE-TASK] : Critical Framework Integration Bugs in examples/cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection Example
***

# **Critical Framework Integration Issues in Cityscapes-Synthia Curb Detection Example**

### **Selected Example**
**Example:** `examples/cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/`  
**Paradigm:** Lifelong Learning with RFNet architecture  
**Purpose:** Semantic segmentation for autonomous driving curb detection using real and synthetic data

### **Problem Description**
This flagship lifelong learning example suffers from multiple critical framework integration issues that make it completely non-functional for new users. Through systematic debugging, I identified several layers of problems:

#### **1. Missing Knowledge Management Registry (Critical)**
```bash
ValueError: can't find class type knowledge_management class name UpdateStrategyDefault in class registry
```
**Root Cause:** The Sedna framework's lifelong learning pipeline is hardcoded to look for `UpdateStrategyDefault` class, but this class is never registered in the ClassFactory registry.

#### **2. TxtDataParse Object Interface Mismatch**
```python
TypeError: 'TxtDataParse' object is not subscriptable
# Error occurs at: disp_path = self.disparities[self.split][index].rstrip()
```
**Root Cause:** Dataset loader assumes `TxtDataParse` objects can be indexed like arrays, but they don't implement `__getitem__`.

#### **3. Dataset Interface Inconsistency**
```bash
CityscapesSegmentation.__init__() got an unexpected keyword argument 'data'
```
**Root Cause:** Ianvs framework passes `data` parameter to dataset constructor, but the dataset class doesn't accept this parameter.

#### **4. Permission Issues**
```bash
PermissionError: [Errno 13] Permission denied: '/var/lib/sedna'
```
**Root Cause:** Sedna tries to create system directories without proper permissions.

### **Impact Justification**
- **High Severity:** Complete example failure affects new contributors and researchers
- **Documentation Gap:** Official docs suggest simple execution, but reality involves complex debugging
- **Community Impact:** Users report similar issues across multiple examples
- **Research Hindrance:** Prevents evaluation of lifelong learning capabilities in edge AI scenarios

### **Debug Process Evidence**

Image

Image

***

### **Primary Contributions to KubeEdge Ianvs Community**

1. **Framework Integration Fixes**
   - Implement proper knowledge management class registration
   - Fix dataset interface compatibility issues
   - Resolve permission and directory structure problems

2. **Enhanced Example Robustness**
   - Create fallback mechanisms for different paradigm types
   - Implement proper error handling and user feedback
   - Add configuration validation

3. **Documentation & Usability**
   - Provide clear setup instructions with dependency management
   - Create troubleshooting guide for common issues
   - Update example documentation to match actual requirements

4. **CI/Testing Infrastructure**
   - Implement automated testing for this example
   - Create dependency compatibility checks
   - Add integration tests for framework components

***

### **Expected Users**
- **Primary:** AI researchers working on lifelong learning and edge AI
- **Secondary:** New contributors exploring Ianvs capabilities  
- **Tertiary:** Students and developers learning distributed AI benchmarking

### **Project Scope**
- **Immediate:** Fix the curb-detection example to be fully functional
- **Medium-term:** Create reusable solutions for similar examples
- **Long-term:** Establish patterns for robust example development

### **Uniqueness vs Existing Issues**
Unlike other reported issues that focus on single configuration problems or missing files, this addresses **systematic framework integration issues** that affect multiple layers:
- **Framework Level:** Sedna-Ianvs integration problems
- **Example Level:** Dataset and algorithm compatibility
- **System Level:** Permission and directory structure
- **Documentation Level:** Accuracy and completeness gaps

This is a **comprehensive analysis** rather than isolated bug reports, providing both deep technical understanding and practical solutions.

***

### **4.1 Architecture: Affected Ianvs Modules**

The proposed revisions will target these core modules:

```
ianvs/
├── core/testcasecontroller/algorithm/paradigm/lifelong_learning/
│   └── lifelong_learning.py (knowledge management integration)
├── examples/cityscapes-synthia/.../testalgorithms/rfnet/
│   ├── basemodel.py (dataset interface compatibility)
│   └── RFNet/dataloaders/datasets/cityscapes.py (TxtDataParse handling)
├── core/cmd/
│   └── benchmarking.py (error handling and validation)
└── docs/
    └── examples/ (documentation updates)
```

**External Dependencies:**
- Sedna framework (knowledge management classes)
- Dataset parsing utilities (TxtDataParse interface)

#### **Core Framework Modifications (Minimal)**
**Justification:** Preserve framework stability while enabling functionality

1. **Knowledge Management Registry Fix**
   ```python
   # Add to lifelong_learning.py
   def register_default_knowledge_management():
       if not ClassFactory.is_registered(ClassType.KM, 'UpdateStrategyDefault'):
           # Register minimal implementation or provide clear error message
   ```

2. **Enhanced Error Handling**
   - Add descriptive error messages for missing classes
   - Provide fallback mechanisms for optional components
   - Implement graceful degradation

#### **Example-Level Modifications (Primary Focus)**
**Justification:** Most issues can be resolved at example level without core changes

1. **Dataset Interface Compatibility**
   ```python
   # Enhanced CityscapesSegmentation.__init__()
   def __init__(self, args, root, split="train", transform=None, data=None, **kwargs):
       # Handle both direct file access and Ianvs data parameter
   ```

2. **TxtDataParse Wrapper**
   ```python
   # Create robust data access layer
   def safe_index_access(txt_parse_obj, index):
       # Multiple fallback strategies for data access
   ```

3. **Configuration Validation**
   - Pre-flight checks for required files and permissions
   - Clear setup instructions and dependency verification
   - Alternative paradigm configurations (incremental learning fallback)

#### **System Integration**
1. **Permission Handling**
   - Use user-space directories as default
   - Provide clear instructions for system-wide installation
   - Environment variable configuration options

2. **Dependency Management**
   - Version compatibility matrix
   - Automated dependency checking
   - Clear error messages for version conflicts

#### **Testing & Validation**
1. **Integration Tests**
   - End-to-end example execution
   - Multiple Python/dependency versions
   - Different system configurations

2. **Documentation Tests**
   - Verify all instructions work on clean systems
   - Test setup procedures step-by-step
   - Validate troubleshooting guides

### **Implementation Priority**
1. **Phase 1:** Fix critical blocking issues (knowledge management, dataset interface)
2. **Phase 2:** Improve robustness and error handling
3. **Phase 3:** Documentation and testing infrastructure
4. **Phase 4:** Generalize solutions for other examples

**Expected Outcome:** Transform a completely broken example into a reliable, well-documented reference implementation that demonstrates Ianvs capabilities effectively.

***

Image

Image

Image

***

ahsan-zamil:
**What type of PR is this?**

/kind bug

This PR restores the `cityscapes-synthia` curb-detection lifelong learning benchmark.

- **Registry Fix**: Registers `UpdateStrategyDefault` dynamically in `lifelong_learning.py` under the `'knowledge_management'` category to prevent ClassFactory registry lookup crashes.
- **Path Fixes**: Corrects double-slashes and relative config folder paths in `benchmarkingjob.yaml`, `rfnet_algorithm.yaml`, and `testenv.yaml`. Converts absolute dataset paths to relative paths.
- **Dataset Initialization**: Updates all dataset class constructors to accept `data=None` and adds robust list/tuple/TxtDataParse parsing to prevent un-subscriptable object type errors.
- **Dependency**: Appends missing `watchdog` library to requirements.

Fixes #254


suhaan-24:
Reviewed against `sedna 0.6.0.1` / Python 3.9.6 / macOS Apple M4. Static review — I have **not**
executed this branch; everything below is diff and code-path evidence, and I have flagged it as
such rather than claiming a run.

## 1. The title does not match the diff

The PR is titled *"restore curb-detection lifelong learning benchmark"*, but the diff contains no
file under `examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/`:

```
$ git diff --name-only $(git merge-base upstream/main HEAD) HEAD | grep 'lifelong_learning_bench/curb-detection'
(no output)
```

The curb-detection files it does touch are under
`examples/cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/`, which is a
different Example, and the lifelong-learning edits land on
`examples/cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/`.

Worth retitling — as it stands, someone looking for the curb-detection lifelong-learning fix will
find this PR by name and merge something that does not touch it.

## 2. Six dependencies are added to the root `requirements.txt` with no version bounds

```diff
+colorlog
+PyYAML
+tenacity
+pydantic
+fastapi
+watchdog
```

Since this is the **root** requirements file, it changes the resolved dependency set for every
installation of Ianvs, not only for the Examples in this PR.

### `fastapi` + `pydantic` is the one I would hold the PR on

Issue #666 reports that `fastapi` — reached from Ianvs core via
`sedna.service.server.inference` — requires `pydantic>=2.9.0` for the `IncEx` type, and that any
install path which downgrades `pydantic` breaks the `ianvs` command outright:

```
File ".../fastapi/types.py", line 7, in <module>
    from pydantic.main import IncEx as IncEx
ImportError: cannot import name 'IncEx' from 'pydantic.main'
```

In #666 that is triggered by one Example's own requirements pulling an old `langchain` that forces
`pydantic==1.10.22`. This PR takes the same fragile pair and puts it in the base install for
everyone, with neither side pinned. My environment currently resolves `fastapi 0.128.8` with
`pydantic 2.13.4`, which is fine — but nothing in this diff makes that the outcome, and #666
documents what happens when the resolution goes the other way.

If these dependencies are genuinely needed at the core level, they would be much safer with lower
bounds at minimum (`pydantic>=2.9`, and a `fastapi` floor matching it).

### `watchdog` — smaller version of the same issue

sedna's own wheel also declares it unbounded:

```
$ python -c "import importlib.metadata as m; print([r for r in m.requires('sedna') if 'watchdog' in r])"
['watchdog']
```

and sedna passes `recursive` positionally:

```python
self.unseen_sample_observer.schedule(
    UnseenSampleUploadingHandler(), self.local_unseen_save_url, True)
```

On the installed version that is fine —
`schedule(self, event_handler, path, recursive=False)` accepts three positional arguments. The
concern is that an unbounded requirement does not hold that guarantee for future installs, and
this line is the one that would break.

## 3. The registry alias resolves the lookup but not the failure

```python
if not ClassFactory.is_exists(ClassType.KM, "UpdateStrategyDefault"):
    ClassFactory.register_cls(UpdateStrategyByFinetune, ClassType.KM, alias="UpdateStrategyDefault")
```

The diagnosis is correct — sedna defaults `task_update_decision` to
`{"method": "UpdateStrategyDefault"}`, its `__all__` advertises that name, and the name is not
registered:

```
__all__ advertises            : ('UpdateStrategyDefault',)
UpdateStrategyDefault registered   : False
UpdateStrategyByFinetune registered: True
```

But the class being aliased cannot execute for these Examples. `UpdateStrategyByFinetune.__call__`
does:

```python
task = tasks[[i for i, task in enumerate(tasks) if
              task.meta_attr == self.attribute[0]]][0]
```

which indexes a list with a list and raises
`TypeError: list indices must be integers or slices, not list` whenever nothing matches. It also
filters samples on a hardcoded city name (`kwargs.get("city", "berlin")`). So this change turns
an unresolvable-name error into a `TypeError` one stage later.

## 4. Scope and overlap

38 files across four unrelated Examples — `cityscapes-synthia`,
`cloud-edge-collaborative-inference-for-llm`, `federated-llm/fedllm-peft`, `robot` — plus two Core
files and the root requirements. Two consequences:

- **19 of those files are also changed by #657**, by the same author, with divergent content.
  The two need reconciling before either merges; whichever lands second will conflict.
- `core/testenvmanager/dataset/dataset.py` is touched by four other open PRs. This PR rewrites the
  parsing path there, which is a different strategy from the deprecation-warning approach taken
  elsewhere. Worth a maintainer decision on which strategy the loader should adopt before either
  is merged.

Splitting this into per-concern PRs (dependency bound / Core dataset parsing / registry / each
Example) would make each piece reviewable and let the safe parts land quickly.

## Verification boundary

**This branch was not executed.** Findings above come from the diff, from the installed sedna
source, and from runtime registry introspection on my environment (Python 3.9.6, sedna 0.6.0.1,
fastapi 0.128.8, pydantic 2.13.4, watchdog 3.0.0).

The `IncEx` traceback is quoted from issue #666, not reproduced by me — I have not constructed the
incompatible `pydantic` resolution locally. What I verified directly is that this diff adds
`fastapi` and `pydantic` to the root requirements with no bounds, and that #666 is open and
describes the failure that combination permits. Likewise the `watchdog` point is a risk argument
from the absent upper bound, not an observed failure on this branch.

Laiiiiba:
### Review Major revision

I reviewed the changes with particular attention to the interaction between the restored example logic and the shared/core changes.

My main concern is a correctness issue in the current implementation: `my_inference()` / `my_evaluate()` are called unconditionally before the subsequent `hasattr()` check that is intended to guard those methods.

This means the guard does not actually protect the call site. If the target object does not provide the corresponding method, execution can fail before `hasattr()` is reached. The check needs to occur before the method invocation, rather than after it.

I also think the PR's scope deserves reconsideration because it combines core/dependency changes with restoration work across multiple examples. Separating unrelated changes would make the behavior easier to validate and reduce the risk of coupling an example restoration to shared dependency changes.

Major revision

I would recommend restructuring the affected call path so the capability check precedes invocation, and considering whether the unrelated scope can be separated into smaller changes.

barakahabimbola:
Reviewing this as part of LFX 2026 Term 3 pretest (full review in Discussion #964). This PR is Core/shared-layer, not Example-local — 38 files across 4 unrelated Examples plus 2 Core files and the root requirements.txt. Flagging two concerns: (1) 19 of these files are also touched by PR #657 (same author, divergent content) — whichever merges second will conflict on those files. (2) This adds fastapi+pydantic to root requirements with no version bounds, which reproduces rather than fixes the unbounded-dependency risk described in #666. Recommend major revision: pin version bounds and reconcile the #657 overlap before merge. (Static/traced review, not executed.)

Mahendr99ar:
Reviewing as part of my LFX 2026 Term 3 pretest (full review: discussion #984).

Recommendation: major revision. Adding fastapi and pydantic to the root requirements.txt with no version bounds reproduces rather than fixes the unbounded-dependency risk in #666.

Adding to prior reviews: I connected this PR's pattern to #735 as well (protobuf/onnx conflicting with root onnx==1.22.0's floor) — a completely different package pair showing the same failure shape. This isn't just a risk specific to the fastapi/pydantic pair already flagged here; it's a general pattern of adding Core-shared dependencies without bounds, independently demonstrated twice. Full writeup in discussion #984.

(Static/diff review — I haven't executed this branch.)

phantom-712:
The Cityscapes-Synthia path/schema changes are useful, but this PR should not
be accepted as one indivisible restoration unit. The current commit series
contains changes for unrelated metrics, cloud-edge/federated examples, Core
lifelong-learning behavior, Windows/backend handling, and the Cityscapes
configuration. That makes it difficult to answer which behavior a reviewer is
approving and makes a regression in one area expensive to revert.

For the target Example, acceptance should require a narrowly scoped commit or follow-up
that proves:

1. every edited benchmarking-job, testenv, and algorithm reference resolves in
   the current repository layout;
2. the parser receives supported dataset keys and a dummy file-backed fixture
   reaches `Dataset.process_dataset()`;
3. the dynamic update-strategy registration is tested with the exact Sedna
   artifact, including the required parameters of the selected implementation;
4. root dependency additions have explicit version policy and an import smoke
   test; and
5. the semantic-segmentation path is not represented by a title or file set
   belonging to a different Cityscapes curb-detection Example.

The registry alias is only the first step if the aliased implementation later
indexes a task list with a list-valued index or relies on a hardcoded city
attribute. The integration test must exercise the next constructor/call path,
not merely assert that the class name is present in the registry.

Recommendation: major revision. Split Example configuration, Core dataset
parsing/registry, dependency policy, and unrelated Example work into
reviewable units; keep the Cityscapes path/schema repair once its file-backed
static test is attached. The branch itself was not executed in this review;
static and minimal-fixture evidence must not be described as a complete
benchmark pass.

PR #702, opened 2026-08-07
fix(utils): preserve keyword-only parameters in parse_kwargs helper

### LINKED ISSUE #597: [Bug] parse_kwargs() in core/common/utils.py filters out valid keyword-only parameters

priyam99:
Fixes #597

Updates parse_kwargs in core/common/utils.py to inspect both positional args and keyword-only args from inspect.getfullargspec(), preventing keyword-only arguments from being silently discarded during parameter filtering.


priyam99:
This one's small and self-contained: `parse_kwargs` was only inspecting positional args from `getfullargspec`, so keyword-only parameters were being silently dropped when filtering kwargs. Added a unit test covering the keyword-only case.

Happy to adjust the approach if you'd prefer the filtering handled differently.

/assign @moorezheng

henryng15 (commented):
Reviewed at head `68b5c28` against `37a9c60`.

The change is correct and the added unit test is welcome. Two things the PR could not
have known about itself.

**1. #598 is an older, open PR making the same fix to the same function for the same
issue (#597).** Neither PR references the other. #598 additionally handles `varkw`
declared under a name other than `kwargs`, which this one leaves to the existing
`need_kw.varkw == 'kwargs'` check. On coverage alone #598 is the broader of the two.

**2. The function has no callers in this repository.** At `37a9c60`, an AST scan of
`core/` and `examples/` finds `core/common/utils.py:46` as the only definition and
**zero** invocations and zero imports. The bundled Sedna wheel ships and calls its own
`parse_kwargs` (`sedna/backend/base.py:47`, 14 call sites) and does not import Ianvs's.
So the keyword-only drop this fixes cannot currently affect any Example.

The defect itself is real — running the current helper gives
`parse_kwargs(kwonly, threshold=0.9) -> {}` — so neither PR is wrong. It is a question of
which one lands and whether the helper should exist at all.

Reproduce: `python3 tools/probe_parse_kwargs_dead.py ianvs` —

**Recommendation: reject as redundant, not as incorrect.** I would close this in favour
of #598 and move the unit test there — the test is the part of this PR #598 lacks, and it
should not be lost.

PR #704, opened 2026-08-07
fix: restore PIPL example integration

### LINKED ISSUE #703: PIPL example is not runnable due to invalid dependency and broken Ianvs integration
## What happened

The `examples/PIPL/edge-cloud_collaborative_learning_bench` example is currently not runnable with Ianvs because multiple integration issues prevent it from being installed and executed successfully.

I identified the following problems:

1. **`pip install -r requirements.txt` fails**
   - `membership-inference-attacks>=0.1.0` does not exist on PyPI.
   - The package is not imported anywhere in the example. Membership inference attack logic is already implemented locally in `testenv/privacy_metrics.py`, making this dependency unnecessary.

2. **`benchmarkingjob.yaml` uses paths that do not resolve when following the documented workflow**
   - `testenv: "./testenv/testenv.yaml"`
   - `url: "./test_algorithms/algorithm.yaml"`
   - Ianvs resolves configuration paths relative to the current working directory, while the example README instructs users to execute the benchmark from the repository root. As a result, these configuration files cannot be located.

3. **`algorithm.yaml` does not follow the Ianvs algorithm schema**
   - The configuration does not define the required `paradigm_type` or `modules` fields expected by `Algorithm._parse_config()`.
   - Consequently, the algorithm cannot be instantiated correctly.

4. **The implementation class is registered under unused aliases**
   - `privacy_preserving_llm.py` registers one implementation as:
     - `PIPLPrivacyDatasetProcessor`
     - `PrivacyPreservingEdgeModel`
     - `PrivacyPreservingCloudModel`
   - However, the implementation exposes the `train`, `predict`, and `evaluate` interface used by the `singletasklearning` `basemodel` rather than a `jointinference` edge/cloud implementation, and none of these aliases are referenced by the example configuration.

5. **`testenv.yaml` uses a metrics schema that does not match the schema used by other Ianvs examples**
   - The metrics are organized into nested `utility`, `privacy`, `performance`, and `compliance` sections instead of the flat `{name, url}` format used by the framework.
   - The shipped metric implementations are also not registered with `ClassFactory`.
   - This appears to require a larger design discussion because `PrivacyPreservingLLM.predict()` returns structured output rather than the standard prediction format consumed by existing metrics.

---

The PIPL example should install successfully using the provided `requirements.txt`, load its benchmark configuration correctly, instantiate the configured algorithm, and run using the documented Ianvs command without requiring manual modifications.

---

### 1. Clone the repository

```bash
cd ianvs
```

### 2. Install the example dependencies

```bash
cd examples/PIPL/edge-cloud_collaborative_learning_bench
```

Observed:

```text
ERROR: Could not find a version that satisfies the requirement membership-inference-attacks>=0.1.0
ERROR: No matching distribution found for membership-inference-attacks>=0.1.0
```

### 3. Run the benchmark from the repository root

```bash
cd ../../..
ianvs -f examples/PIPL/edge-cloud_collaborative_learning_bench/benchmarkingjob.yaml
```

The benchmark configuration cannot be loaded correctly because the referenced configuration paths are resolved relative to the current working directory instead of the configuration file location.

### 4. Inspect the algorithm configuration

`algorithm.yaml` does not follow the schema expected by `Algorithm._parse_config()`, preventing Ianvs from constructing the configured modules.

---

## Anything else we need to know?

I investigated the example and prepared local fixes for the first four issues listed above.

Issue #5 was intentionally left unchanged because it appears to require maintainer guidance on the expected metrics interface for structured prediction outputs.

**Environment**

- macOS
- Apple Silicon (arm64)
- Python 3.10
- Miniconda

iron-prog:
**What type of PR is this?**

/kind bug

This PR restores the PIPL example so it can be installed and loaded correctly by the Ianvs framework.

The changes include:

- Remove the invalid `membership-inference-attacks` dependency that prevents `pip install -r requirements.txt` from succeeding.
- Fix configuration paths in `benchmarkingjob.yaml` so the benchmark loads correctly when executed from the repository root.
- Update `algorithm.yaml` to follow the Ianvs algorithm schema by defining the required `paradigm_type` and `modules` configuration.
- Register the implementation using the `singletasklearning` `basemodel` interface expected by the framework instead of unused aliases.

Fixes #703


iron-prog:
During verification I confirmed that these changes resolve the initial integration problems:

- `pip install -r requirements.txt` no longer fails because of the invalid dependency.
- The benchmark now successfully loads `benchmarkingjob.yaml`, `testenv.yaml`, and the algorithm configuration.
- The algorithm is instantiated using the expected `singletasklearning` `basemodel` interface.

Running:

```bash
ianvs -f examples/PIPL/edge-cloud_collaborative_learning_bench/benchmarkingjob.yaml
```

now progresses further into initialization and reaches dataset validation.

Execution currently stops because the example references:

```
./data/chnsenticorp_lite/
```

but those dataset files are not present in the repository, and I could not find any download instructions for them. Since this appears to be independent of the framework integration fixes in this PR, I have intentionally left it out of scope.

PR #713, opened 2026-08-08
Fix bdd curb detection paths

### LINKED ISSUE #712: [Bug] BDD lifelong-learning curb-detection example uses stale config paths
**What happened**:
BDD curb-detection example has broken paths in the yaml files.

It should use the files inside the actual `curb-detection` folder.

**How to reproduce it**:
Run `ianvs -f examples/bdd/lifelong_learning_bench/curb-detection/benchmarkingjob.yaml`.

The config points to `examples/bdd/lifelong_learning_bench/...`, but the files are under `examples/bdd/lifelong_learning_bench/curb-detection/...`.

ishansurdi:
**What type of PR is this?**
/kind bug

Fixes broken BDD curb-detection config paths and old dataset keys.

Fixes #712

### DIFF
--- a/examples/bdd/lifelong_learning_bench/curb-detection/benchmarkingjob.yaml
+++ b/examples/bdd/lifelong_learning_bench/curb-detection/benchmarkingjob.yaml
@@ -19,7 +19,7 @@ benchmarkingjob:
       - name: "yolo_lifelong_learning_five_model_750data"
         # the url address of test algorithm configuration file; string type;
         # the file format supports yaml/yml
-        url: "./examples/bdd/lifelong_learning_bench/testalgorithms/yolo/yolo_algorithm.yaml"
+        url: "./examples/bdd/lifelong_learning_bench/curb-detection/testalgorithms/yolo/yolo_algorithm.yaml"
 
   # the configuration of ranking leaderboard
   rank:
--- a/examples/bdd/lifelong_learning_bench/curb-detection/benchmarkingjob.yaml
+++ b/examples/bdd/lifelong_learning_bench/curb-detection/benchmarkingjob.yaml
@@ -6,7 +6,7 @@ benchmarkingjob:
 
   # the url address of test environment configuration file; string type;
   # the file format supports yaml/yml;
-  testenv: "./examples/bdd/lifelong_learning_bench/testenv/testenv.yaml"
+  testenv: "./examples/bdd/lifelong_learning_bench/curb-detection/testenv/testenv.yaml"
 
   # the configuration of test object
   test_object:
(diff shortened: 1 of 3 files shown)


suhaan-24:
The rename here is correct — `core/testenvmanager/dataset/dataset.py` treats `train_url`/`test_url`
as values it computes inside `process_dataset()`, not as config inputs, so the previous keys fell
through to:

```python
else:
    raise NotImplementedError('not one of train_index/train_data/train_data_info')
```

Two observations specific to this Example.

## 1. There is a commented-out line that will confuse the next reader

`examples/bdd/lifelong_learning_bench/curb-detection/testenv/testenv.yaml` carries, above the live
keys, a commented pair pointing at a different dataset entirely:

```yaml
    # train_url: "dataset/curb-detection/train_data/index.txt"
    train_url: "dataset/bdd/bdd_compress_train_label.txt"
```

Since this PR is already touching these lines, it would be worth removing the commented pair rather
than renaming only the live one — otherwise the file ends up with a `train_index:` key next to a
stale `# train_url:` comment referencing a different Example's dataset layout, which is more
confusing than either state alone.

## 2. The label-file format differs from the index-file format

The BDD values here are `bdd_compress_train_label.txt`, whereas the sibling curb-detection Example
uses `train_data/index.txt` with lines of the form
`./images/_leftImg8bit.png ./images/_gtFine_labelTrainIds.png`.

`train_index` routes to `_process_index_file()`, which expects the two-column index form and
rewrites relative paths to absolute. If the BDD label file is a different shape, `train_index` may
not be the correct target key — `train_data` routes to `_process_data_file()` instead. Worth
confirming which loader the BDD file is meant for before settling on `train_index`, since both keys
are accepted and only one will parse correctly.

I was not able to check this myself: BDD100K is distributed via Baidu Pan links I cannot resolve,
so I could not inspect the label file's contents or run the Example.

## Note on recurrence

Core discards unrecognised dataset keys silently, so a config written against the old contract
fails with `NotImplementedError` rather than anything naming the offending key. That is worth being
aware of when reviewing this kind of rename — the failure message does not point at the key that
caused it.

## Verification boundary

**Not executed.** The `NotImplementedError` path is quoted from `upstream/main` source and was
observed on a different Example (`cityscapes-synthia/.../curb-detection`), not on BDD. The
commented-line observation is from reading `upstream/main`. The format concern in point 2 is a
question, not a confirmed defect — I could not obtain the dataset to check.

rajdhruvsingh:
Credit to @suhaan-24 for the existing review confirming the rename is correct and flagging the commented-line cleanup and label-format concerns.

One addition: I'd flag that the key migration here is fully complete, not partial — the active train_index:/test_index: keys are correctly set, with only the already-commented alternative paths left commented (which matches what @suhaan-24 already noted).

On @suhaan-24's open question about whether train_index is even the right key for this dataset's label format: I wasn't able to resolve it either — BDD100K's distribution links aren't accessible to check the actual file shape. Flagging it as still open rather than assuming it's resolved, since if the label file isn't in two-column index form, train_data would be the correct key instead and train_index would silently mis-parse rather than crash — a harder failure to catch than the one this PR fixes.

Separate from the key migration: the metric block also swaps accuracy→map, repointing to a real, already-registered function (alias="map", confirmed in the target file) — functional, just undocumented in the PR description. Worth a line in the description since it changes what this Example measures.

PR #715, opened 2026-08-08
fix(core): optimize leaderboard generation to avoid O(n^2) DataFrame insertions

### LINKED ISSUE #633: Quadratic O(n^2) leaderboard construction in Rank._get_all due to row-by-row DataFrame.loc insertion
**What happened**:

`Rank._get_all()` in `core/storymanager/rank/rank.py` builds the leaderboard result table by inserting each test case's row into a live `pandas.DataFrame` one at a time via `all_df.loc[i] = row_data` inside a `for` loop. Each such label-based `.loc[i] = ...` assignment on an existing `DataFrame` triggers an internal reindex/copy of the whole frame, so the total cost of building the leaderboard table is `O(n^2)` in the number of test cases rather than `O(n)`.

Since Ianvs is designed to run leaderboards/benchmarks across many algorithm/module/hyperparameter combinations, this scales poorly exactly where Ianvs is used the most: large benchmark matrices.

Leaderboard/rank table construction time should scale roughly linearly with the number of test cases, e.g. by collecting row dicts in a plain Python list and constructing the `DataFrame` once after the loop.

The relevant code is `Rank._get_all` in `core/storymanager/rank/rank.py` (around lines 147-172):

```python
all_df = pd.DataFrame(columns=self.all_df_header)
for i, test_case in enumerate(test_cases):
    ...
    all_df.loc[i] = row_data
```

This exact pattern can be reproduced as a standalone micro-benchmark:

```python
import pandas as pd, time
N = 2000
cols = [f"c{i}" for i in range(15)]

t0 = time.time()
df = pd.DataFrame(columns=cols)
for i in range(N):
    df.loc[i] = {c: i for c in cols}
print("loc-based row insertion:", time.time() - t0)

t0 = time.time()
rows = [{c: i for c in cols} for i in range(N)]
df2 = pd.DataFrame(rows, columns=cols)
print("list-of-dicts + single DataFrame():", time.time() - t0)
```

Observed output on this environment (Python 3, pandas 3.0.5):

```
loc-based row insertion: 0.515s
list-of-dicts + single DataFrame(): 0.003s
```

That is roughly a 170x slowdown for only 2000 rows, and the gap widens superlinearly as the row count grows.

This is distinct from the earlier, closed pandas-deprecation issue (#134), which addressed removal of deprecated pandas APIs (`pd.np`, `DataFrame.append`, chained assignment warnings) in this same method. The `.loc[i] = row_data` row-by-row insertion pattern itself remains and causes a real performance regression, not just a deprecation warning.

Suggested direction: accumulate `row_data` dicts into a plain Python list during the loop, then build `all_df = pd.DataFrame(rows, columns=self.all_df_header)` once after the loop, matching the fast branch of the benchmark above. Leaderboard output (`all_rank.csv` / `selected_rank.csv`) should remain equivalent to current output for existing examples.

ahsan-zamil:
**What type of PR is this?**

/kind bug

Optimizes the leaderboard DataFrame generation performance from quadratic $O(n^2)$ to linear $O(n)$ by collecting rows in a list of dicts first, rather than doing row-by-row `.loc[i] = ...` DataFrame insertions in a loop.

This results in a ~100x speedup for large leaderboards (e.g. 2000 rows builds in 0.04s instead of >0.5s).

Fixes #633


suhaan-24:
Straightforward improvement — `all_df.loc[i] = row_data` reallocates the frame on every assignment,
so building the leaderboard was quadratic in the number of test cases. Collecting rows and
constructing once is the standard fix.

Because `_get_all` feeds the CSV that gets published as the benchmark result, I checked whether the
change alters output rather than assuming equivalence. It does not — comparing both constructions
on a representative row:

| Check | `.loc` (before) | bulk (after) |
|---|---|---|
| dtypes | `rank int64, algorithm object, accuracy float64, samples_transfer_ratio float64` | **identical** |
| CSV row | `1,rfnet,0.2116751104900657,0.4979` | **identical** |
| key present in `row_data` but not in header | dropped | dropped |
| key in header but missing from `row_data` | `NaN` | `NaN` |
| empty `test_cases` | shape `(0, 4)` | shape `(0, 4)` |

Float repr is preserved to full precision, and the empty case still produces a correctly-shaped
frame for `_concat_existing_data` to consume. So no behaviour change to flag.

## One nit

```python
-        for i, test_case in enumerate(test_cases):
+        for _, test_case in enumerate(test_cases):
```

The index is no longer used, so `enumerate` is now vestigial:

```python
        for test_case in test_cases:
```

reads more clearly and avoids the throwaway binding.

## One question

Do you have a sense of the realistic upper bound on `len(test_cases)`? A benchmarking job expands
to one test case per algorithm-module combination, so a large hyperparameter sweep could plausibly
reach the hundreds — in which case this is a worthwhile win. If the practical ceiling is a handful,
the change is still strictly better and harmless, just less consequential than the title suggests.
Either way it is worth landing; asking mainly because it would be useful to know whether
`_concat_existing_data` deserves the same treatment.

## Verification boundary

The equivalence table above was **executed** against the pandas installed in this environment
(Python 3.9.6). I have **not** run this branch inside a full benchmarking job, so the check covers
`_get_all`'s construction semantics in isolation rather than end-to-end leaderboard output.

PR #721, opened 2026-08-08
[PHASE IV] fix(GovDoc2Poster): add diagnostics for silently-dropped results in average_scores calculation

### LINKED ISSUE #720: [PHASE IV][GovDoc2Poster]: Silent Result Exclusion in average_scores Calculation (Missing 'evaluation' Key)
### What happened

During end-to-end execution testing of the GovDoc2Poster example benchmark, a silent data-loss issue was discovered in the results aggregation logic in `basemodel.py`.

In `_format_results()`, results are filtered before computing `average_scores`:

```python
evaluation_scores = [r.get('evaluation', {}) for r in results if 'evaluation' in r]
```

Any result dict missing the `'evaluation'` key (e.g. because evaluation failed upstream, or the document was skipped) is silently excluded from the average score calculation. There is no log message, no count, and no indication anywhere in the output that some results were excluded from scoring — `total_processed` still reports the full count, while `average_scores` quietly reflects only a subset of results.

`_format_results()` should log a warning indicating how many results were excluded from `average_scores` due to a missing `'evaluation'` key, and the returned `average_scores` dict should include explicit `evaluated_count` / `skipped_count` fields, so downstream consumers of benchmark output can tell how many results actually contributed to the reported score.

Run the following reproduction script inside `examples/GovDoc2Poster/singletask_learning_bench`:

```python
import sys
sys.path.insert(0, 'testalgorithms/gen')
from basemodel import NewGovernmentPosterAgent

agent = NewGovernmentPosterAgent(evaluate_only=True)

fake_results = [
    {'evaluation': {'score': 8.0}, 'processing_time': 1.2},
    {'evaluation': {'score': 6.0}, 'processing_time': 1.1},
    {'error': 'parse failed', 'processing_time': 0.5},  # missing 'evaluation' entirely
]

formatted = agent._format_results(fake_results)
print('total_processed:', formatted['total_processed'])
print('average_scores:', formatted.get('average_scores'))
```

Result:

```
total_processed: 3
average_scores: {'avg_score': 7.0}
```

The third result is silently excluded — no warning, no count of how many were dropped.

Suggested code fix: in `_format_results()`, log a warning when `results_missing_evaluation` is non-empty, and add `evaluated_count`/`skipped_count` fields to the returned `average_scores` dict.

Found while investigating GovDoc2Poster for LFX Mentorship Fall 2026 (Ianvs Example Restoration, Phase IV).

hayyanbaig:
### What type of PR is this?

/kind bug

While investigating the GovDoc2Poster example benchmark for LFX Mentorship Fall 2026 (Ianvs Example Restoration, Phase IV), I found and fixed a silent data-loss issue in the results aggregation logic in `basemodel.py` (#720):

**Add Diagnostics for Silently-Dropped Results (basemodel.py)**
`_format_results()` filtered out any result missing the `'evaluation'` key before computing `average_scores`, with no log message and no indication anywhere in the returned output that results had been excluded. `total_processed` still reported the full result count, while `average_scores` silently reflected only a subset — making it impossible to tell from the output alone whether a low average score meant poor model performance or dropped results. Added a warning log that fires whenever one or more results are excluded, and added explicit `evaluated_count` / `skipped_count` fields to the returned `average_scores` dict so downstream consumers can see exactly how many results contributed to the reported score.

Fixes #720

### Testing done:

Reproduction Script: Executed a minimal reproduction script isolating `_format_results()` with a mix of results containing and missing the `'evaluation'` key; confirmed the warning log fires correctly and `evaluated_count`/`skipped_count` are populated as expected.
End-to-End Execution: Ran `ianvs -f examples/GovDoc2Poster/singletask_learning_bench/benchmarkingjob.yaml` locally; verified the pipeline completes through to evaluation metric generation with no regressions (final score: 0.76).


blackdragoon26:
Hi,

The missing-key diagnostics are useful, but `evaluation=None` is counted as an evaluated result and then crashes on `.get()`. Please accept only mapping-shaped evaluations, classify null/non-mapping entries separately, and add a mixed valid/null/missing test. I would also keep coverage metadata separate from the metric-score mapping if downstream code assumes every entry is a metric.

Links: [Issue #720](https://github.com/kubeedge/ianvs/issues/720) · [#721 aggregation path](https://github.com/kubeedge/ianvs/blob/5b3b9978c5fe37324eb0e03b132ed5a79560e949/examples/GovDoc2Poster/singletask_learning_bench/testalgorithms/gen/basemodel.py#L872-L892)

PR #543, opened 2026-06-13
fix: replace hardcoded absolute paths with root-relative paths in RoboDK Palletizing example

Fahmid-Arman:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it:**
The `RoboDK Palletizing` example previously contained hardcoded absolute paths tied to the original author's local environment (`/root/ianvs/project/ianvs-0.3.0/examples/Palletizing/...`). Furthermore, the directory name was incorrectly referenced as `Palletizing` instead of `RoboDK Palletizing`. This caused immediate failures for any new user attempting to run the benchmark.

Changes implemented:
- **Configuration Remediation:** Updated `benchmarkingjob.yaml`, `testenv.yaml`, and `algorithm.yaml` to utilize portable, root-relative paths (`./examples/RoboDK Palletizing/...`).
- **Directory Name Correction:** Fixed the incorrect `Palletizing` directory references in all configuration files to match the actual repository structure (`RoboDK Palletizing`).
- **Documentation Overhaul:** Updated the `readme.md` execution commands to run from the repository root, ensuring the relative paths resolve correctly. The `ianvs -f` command is now properly quoted to handle the space in the directory name.
- **Dataset Path Alignment:** Updated the dataset download instructions in `readme.md` to map directly to the new expected relative path (`./examples/RoboDK Palletizing/singletask_learning_bench/dataset/RoboDK_Palletizing_Dataset/`).

**Which issue(s) this PR fixes:**
This is a proactive fix for an unreported bug discovered during repository-wide path auditing.

PR #724, opened 2026-08-08
fix: replace hard-coded paths in RoboDK Palletizing example

### LINKED ISSUE #723: RoboDK Palletizing example uses hard-coded absolute paths in benchmark configuration
## What happened

`examples/RoboDK Palletizing/singletask_learning_bench` contains hard-coded absolute paths in multiple configuration files. These paths point to a specific local environment:

- `/root/ianvs/project/ianvs-0.3.0/...`
- `/root/ianvs/project/data/...`

Some of these paths also reference the old directory name (`Palletizing`) instead of the current example directory (`RoboDK Palletizing`).

The affected files are:

1. `benchmarkingjob.yaml`
   - `testenv`
   - algorithm `url`

2. `testalgorithms/algorithm.yaml`
   - `basemodel` module `url`

3. `testenv/testenv.yaml`
   - `train_index`
   - `test_index`
   - metric `url`s

Because these paths are machine-specific, the example cannot be used from a normal repository checkout without manually editing the configuration.

This example also does not appear in the runnable examples tracking issue (#194), so this problem may not have been reported previously.

---

The example should use repository-relative paths so it can be executed from a normal Ianvs checkout without modifying configuration files.

---

## How to reproduce it

Clone the repository and inspect the example configuration:

```bash
cd ianvs

grep -R "/root/ianvs/project" -n "examples/RoboDK Palletizing"
```

The command shows multiple hard-coded absolute paths embedded in the example configuration.

---

## Verified fix

I replaced the hard-coded absolute paths with repository-relative paths that match Ianvs's current path resolution behavior.

After making these changes:

- All benchmark, algorithm, and metric configuration paths resolve correctly.
- `basemodel.py` loads successfully using Ianvs's existing `utils.load_module()`.
- Running

```bash
ianvs -f "examples/RoboDK Palletizing/singletask_learning_bench/benchmarkingjob.yaml"
```

now progresses through benchmark configuration parsing, algorithm loading, and module initialization.

Execution then stops only because the referenced dataset files are not present locally, which is independent of the hard-coded path issue.

---

## Environment

- macOS
- Apple Silicon (arm64)
- Python 3.10
- Miniconda

iron-prog:
**What type of PR is this?**

/kind bug

This PR replaces machine-specific absolute paths in the RoboDK Palletizing example with repository-relative paths.

The affected configuration files previously referenced paths such as:

- `/root/ianvs/project/ianvs-0.3.0/...`
- `/root/ianvs/project/data/...`

These paths only exist in the original development environment, preventing the example from being used from a normal Ianvs repository checkout.

This PR updates:

- `benchmarkingjob.yaml`
  - `testenv`
  - algorithm `url`
- `testalgorithms/algorithm.yaml`
  - `basemodel` module `url`
- `testenv/testenv.yaml`
  - dataset index paths
  - metric module paths

to use repository-relative paths that match Ianvs's existing path resolution behavior.

**Verification**

- Verified all updated configuration and module paths resolve correctly.
- Verified `basemodel.py` loads successfully through Ianvs's `utils.load_module()`.
- Verified `ianvs -f examples/RoboDK\ Palletizing/singletask_learning_bench/benchmarkingjob.yaml` progresses through benchmark configuration parsing, algorithm loading, and module initialization.

Execution now stops only because the referenced dataset files are not present locally, which is independent of the hard-coded path issue addressed in this PR.

Fixes #723


ijayhub:
Reviewed this as part of a cross-example investigation into #723, #514, and #535 (full analysis: #906).

This PR correctly replaces all hard-coded paths and stale directory references identified in #723. I checked every changed line against the current repository structure and confirmed each new relative path resolves to a real file.

One note: the train_index and test_index paths now point inside the example directory instead of the old /project/data/... location. This improves portability, but the dataset files aren't present there yet, so the example may still stop at the dataset stage. That's separate from the path fix in this PR.

Recommendation: Accept as is.

Prachi194agrawal:
Confirmed the underlying defect  `benchmarkingjob.yaml` lines 5 and 11 on
`main` point at `examples/Palletizing/...`, which doesn't exist. The fix here
is correct for what it targets.

Flagging for triage: three other open PRs address the same lines - #543
(2026-06-13, earliest, also updates the README dataset instructions), #696
(broadest scope, also covers dataset portability and weight discovery), and
#783. All four are individually `mergeable: true`, so GitHub reports no
conflict - but they rewrite the same lines with different path conventions,
which means merging two in sequence would silently overwrite one approach
with another.

#696 looks like the better single landing point, since it covers the adjacent
portability issues in the same pass rather than leaving them for a follow-up.

**Recommendation: superseded by #696** -no defect in this change itself.

Environment: Ubuntu 24.04.3, Python 3.12.3, commit `37a9c60`

PRANABKUMARNEOGI:
Hi everyone. I was reviewing this as part of a broader architectural analysis, and @Prachi194agrawal's triage note about the conflicting path conventions in #543, #696, and #783 hits on exactly the framework-level issue I wanted to highlight.

The fact that four separate authors submitted four completely different file-system traversals to fix the same `benchmarkingjob.yaml` defect perfectly illustrates a "Portability Gap" in Ianvs Core. 

Currently, the central orchestrator does not automatically inject a unified, OS-agnostic workspace directory into the Example's constructor. As a result, authors are forced to independently guess and reinvent path-traversal logic (`os.path` vs `pathlib`, relative vs absolute). This lack of a standardized data-ingestion contract is exactly what leads to these triage collisions and cross-OS inconsistency risks.

**Recommendation:**
I completely agree with Prachi's recommendation to supersede this PR in favor of #696 to capture the weight discovery fixes. 

However, to prevent this exact triage bottleneck from happening on future examples, we should open an architectural tracking issue to implement a centralized `WorkspaceManager` in Core. If the orchestrator resolves paths and passes them down, Example authors won't have to write manual OS-level path logic at all.

ANISHA-RAWAT:
Reviewed as part of a cross-example analysis (#514, #535 — full writeup: 
https://github.com/kubeedge/ianvs/discussions/967). The diff correctly 
fixes all 7 hardcoded-path occurrences in algorithm.yaml, testenv.yaml, 
and benchmarkingjob.yaml. However, readme.md in the same directory still 
has 7 unfixed occurrences of the same /root/ianvs/project/... paths — 
a contributor following the README will still hit the original error. 
Recommendation: minor revision — update readme.md in the same PR.

Liyuchen0118 (commented):
**Recommendation**

Major revision.

**Tested head**

`62a66ca106adf4cb547524d03d3561b79b8764ee`

**Validation method**

Focused source-level assertion harness run against the tested PR head. Scope: reference existence only; not an Ianvs runtime test. The harness checked repository-root, README-documented, and temporary-directory contexts. Harness exit: `0`.

**Observed result**

- Repository-root context: five code/config references exist; both dataset indexes are absent.
- README-documented invocation context: all seven references resolve false; the referenced directory is absent in the tested PR snapshot.
- `/tmp` context: all seven references resolve false.
- An earlier local runtime attempt was blocked before execution by missing `pandas`. Subsequent Linux verification installed the dependency and confirmed the GPU stack; the remaining full-runtime blockers are absent dataset indexes and other Example-specific external resources. This is an environment/provisioning boundary, not an Example runtime verdict.

**Impact**

The patch removes contributor-specific and obsolete references, which is necessary. The evidence shows an unclear or narrow supported-CWD contract. It does not establish that arbitrary-CWD execution is required.

**Requested change**

1. Document the supported path base and invocation directory.
2. Add regression tests for every documented invocation context.
3. Add a deterministic Kaggle provisioning command, output location, and post-setup existence check for both indexes.

**Prior discussion**

https://github.com/kubeedge/ianvs/pull/724#issuecomment-5409255240

**Additional evidence**

The README documents `ianvs -f benchmarkingjob.yaml` from the Example directory, while the replacement values retain repository-root prefixes.

PR #726, opened 2026-08-08
fix: correct dashscope requirement specifier

### LINKED ISSUE #718: bug: requirement.txt has invalid pip syntax (dashscope=1.24.5) and aborts the entire install
**What happened**:

Line 6 of `examples/Cloud_Robotics/cloud-edge-collaborative-inference_bench/perception-reasoning/requirement.txt` reads `dashscope=1.24.5`, using a single `=` where pip requires `==`. pip rejects the file outright:

```
$ pip install -r requirement.txt --dry-run
ERROR: Invalid requirement: 'dashscope=1.24.5': Expected semicolon (after name with no version specifier) or end
    dashscope=1.24.5
             ^ (from line 6 of .../requirement.txt)
Hint: = is not a valid operator. Did you mean == ?
```

The dependency file installs cleanly.

```
pip install -r examples/Cloud_Robotics/cloud-edge-collaborative-inference_bench/perception-reasoning/requirement.txt --dry-run
```

Nothing in the file installs - pip aborts on the malformed line rather than skipping it, so this blocks environment setup entirely. Line 5 of the same file has a separate problem (`pytorch` instead of `torch`), filed separately.

Environment:
- Ianvs version: `38a5319` (main)
- Python version: 3.10.19
- OS: Ubuntu 24.04.3 LTS, Linux 7.0.0-28-generic x86_64

JAYATIAHUJA:
/kind bug

**What type of PR is this?**

Bug fix.

Fixes the invalid `dashscope` version specifier in the perception-reasoning requirements file. The single `=` prevents pip from reading the requirements file.

Fixes #718

**Testing**:

`python -m pip install --dry-run -r examples/Cloud_Robotics/cloud-edge-collaborative-inference_bench/perception-reasoning/requirement.txt`


JAYATIAHUJA:
Hi, this fixes the invalid dashscope version specifier from #718. I verified it with: python -m pip install --dry-run -r examples/Cloud_Robotics/cloud-edge-collaborative-inference_bench/perception-reasoning/requirement.txt. Please take a look when you have time.

observer04 (commented):
**Recommendation: accepted as it is.**

The `dashscope=1.24.5` → `dashscope==1.24.5` change is correct and is exactly what this PR sets out to do. pip treats a single `=` as an invalid requirement and aborts the whole file, so this one character is what makes the manifest readable at all. The author already demonstrated that failure, so I do not claim the finding.

Three further defects sit in the same six-line install path. All three predate this PR and survive it unchanged, so I am recording them as follow-up work on that directory rather than asking for them here:

1. The tracked file is `requirement.txt`, while the README instructs `pip install -r requirements.txt`.
2. The README uses `cloud-edge-collaborative-inference-bench`, while the real directory is `cloud-edge-collaborative-inference_bench`.
3. The file declares `pytorch`. Executed in a clean Python 3.10 environment, `pytorch==1.0.2` exits 1 during its build, reporting that the package named for PyTorch is `torch`.

Verification boundary: the specifier behaviour and the `pytorch` build failure were executed locally on CPU. I did not run the Example, and no dataset, model, GPU path, or API credential was involved.

PR #667, opened 2026-08-06
fix(core): enhance benchmark execution resiliency, result retention, and module isolation

priyam99:
Fixes #610

Fixes infinite output directory loops in testcase.py, retains completed testcase results on partial suite failures in testcasecontroller.py, fixes rank validation/heatmap matrix guards in rank.py, and implements isolated dynamic module loading in utils.py.

PR #729, opened 2026-08-08
fix: restore sys.path on exception in py2dict and load_module

### LINKED ISSUE #728: [Bug] sys.path left permanently modified on exception in core/common/utils.py (py2dict and load_module)
**What happened**:
`core/common/utils.py` contains two functions — `py2dict()` and `load_module()` — that both temporarily
insert a directory into `sys.path` to import a module, then remove it afterward. In both cases, the
removal (`sys.path.pop(0)`) only runs on the success path; if the import raises an exception, `sys.path`
is left permanently modified for the rest of the process's lifetime.

`py2dict()` (lines 63-68):
```python
sys.path.insert(0, config_dir)
mod = import_module(module_name)
sys.path.pop(0)
```
If `import_module(module_name)` raises, `sys.path.pop(0)` never executes.

`load_module()` (lines 98-102):
```python
sys.path.insert(0, module_path)
try:
    importlib.import_module(module_name)
    sys.path.pop(0)
except Exception as err:
    raise RuntimeError(f"load module(url={url}) failed, error: {err}") from err
```
Here the `try/except` structure makes it look handled, but `sys.path.pop(0)` is still only called
*inside* the `try` block before the exception is raised — the `except` branch re-raises without ever
popping, so the leak happens on every failed import.

`sys.path` should be restored to its prior state regardless of whether the import succeeds or fails —
i.e. the cleanup should happen in a `finally` block.

```python
import sys
from core.common.utils import load_module

before = list(sys.path)
try:
    load_module("/some/path/to/a_module_that_does_not_exist.py")
except RuntimeError:
    pass

assert sys.path == before, f"sys.path was mutated and never restored: {sys.path[0]} still present"
```
This assertion fails on current `main` — `sys.path[0]` still contains the inserted directory after the
exception propagates.

Both functions are used during config/module loading (`py2dict` for `.py` config files, `load_module`
for dynamically loading algorithm modules), which in a long-running process (e.g. repeated
`BenchmarkingJob` runs in the same interpreter, or a server-style deployment) means a single failed
config/module load silently and permanently pollutes `sys.path` for every subsequent import in that
process — a subtle source of hard-to-diagnose import shadowing or `ModuleNotFoundError`s down the line
that won't obviously correlate with the original failure.

Muhammad-Hashir-Code:
Fixes a `sys.path` leak in `core/common/utils.py` affecting two functions, `py2dict()` and
`load_module()`. Both temporarily insert a directory into `sys.path` to perform a dynamic import, then
remove it but the removal only ran on the success path. If the import raised an exception, the
inserted directory was left permanently in `sys.path` for the remaining lifetime of the process, which
in a long-running or repeated-run process can silently pollute subsequent, unrelated imports.

This PR moves the `sys.path.pop(0)` call into a `finally` block in both functions so cleanup always
runs, whether the import succeeds or fails.

Testing: added a reproduction check confirming `sys.path` is left mutated after a failed
`load_module()` call on unpatched code, and confirming it is correctly restored after this fix. Ran
against a fresh clone of current `main`.

Fixes #728


Omkeswani27:
Tested this branch directly (`git fetch origin pull/729/head:pr-729`):

load_module leaked? False
py2dict leaked? False

Confirms this correctly fixes #728. Note: cross-referenced against #759 and
#806, which independently implement the same fix — full duplication analysis
and recommendation (minor revision):

Yash4616:
The shape here is right: `try/finally` applied to both `load_module()` and `py2dict()`, which is the smallest change that covers the failure path in this file. Executed, `sys.path` is restored correctly after a failed import.

The cleanup itself has an edge case. `finally: sys.path.pop(0)` removes whatever is at index 0 when it runs, which is not necessarily what this function inserted. If the module being imported calls `sys.path.insert(0, ...)` during its own import, index 0 is then the module's entry, so `pop(0)` removes that and leaves the inserted path behind.

This is reachable rather than theoretical: `examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/basemodel.py:29` does exactly that and is loaded through `load_module`. The insert is guarded by `if current_dir not in sys.path` on line 28, which passes on a first load, so the guard does not prevent this case.

Snapshotting `sys.path` before the insert and restoring it in the `finally`, as #806 does, closes that case while keeping the `try/finally` structure this PR establishes.

**Recommendation: major revision.** Correct structure, and the fix for the index-0 case is a few lines.

Aryansingh-ai:

I checked out PR #729 and tested the modified `core/common/utils.py` directly.

Checked-out commit:

```text
1e38a7e fix: restore sys.path on exception in py2dict and load_module

I forced load_module() to fail while recording sys.path before and after the call.

Observed result:

BEFORE_LEN: 10
AFTER_LEN: 10
UNCHANGED: True

The import itself failed with the expected missing-dependency error, but the temporary path insertion was removed correctly.

I also tested the py2dict() failure path with a deliberately missing dependency:

ERROR: ModuleNotFoundError
MESSAGE: No module named 'definitely_missing_package_ianvs_test'
BEFORE_LEN: 10
AFTER_LEN: 10
PATH_UNCHANGED: True
LEAKED_DIR: False

So the PR successfully addresses the failure-path state-leakage property identified in #728.

2. Basename collision — FAIL

I then tested the separate #557 property using two files with the same basename:

one/plugin.py → VALUE = "one"
two/plugin.py → VALUE = "two"

Sequential loading produced:

AFTER FIRST: one
AFTER SECOND: one
EXPECTED SECOND: two

Therefore, #729 does not address the module-name collision.

This is an important scope distinction:

#729
    ↓
fixes failure-path sys.path cleanup

but

#729
    ↓
does not change basename-based module identity
    ↓
#557 remains reproducible

I would accept the PR for the specific cleanup defect, subject to minor revision/clarification.

The implementation is appropriately narrow and the tested failure path is fixed.

However, I recommend explicitly scoping the PR as a #728 cleanup fix rather than a complete dynamic-module-loading fix, because the independent #557 collision remains reproducible after applying this patch.

I would also recommend adding/retaining a regression test that asserts sys.path is unchanged when both py2dict() and load_module() fail during import.

**Comment by sheikhayaan** [comment]
The change is correct as far as it goes: on current code an exception inside `import_module`
leaks the inserted entry, and moving the `pop` into `finally` fixes that.

The problem is that `pop(0)` is positional. @Yash4616 reached that conclusion before me and it
is the right conclusion, so I won't re-argue it, and Copilot raised the same shape on #806
earlier still. One correction to the evidence, though, because it changes which examples are
actually exposed.

**The cited instance doesn't reproduce.**
`examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/basemodel.py:29` is
offered as a reachable case, on the grounds that the guard on line 28 "passes on a first load".
It doesn't pass. I ran the real `core.common.utils.load_module` against a fixture of exactly
that shape:

```
mod_single.py   Core inserted: /tmp/fx/single
      LEAKED after pop(0): NONE
```

`current_dir` on line 27 is `os.path.dirname(os.path.abspath(__file__))`. `load_module` computes
its own insert as `os.path.split(url)[0]`, which for that file is the identical directory. So by
the time line 28 runs Core has already placed that path in `sys.path`, the `not in` test is
False, and the module inserts nothing whatsoever. `pop(0)` then removes precisely Core's own
entry. On a first load the guard is not failing to prevent the case; it is the reason there is
no case to prevent.

Worth getting right, because that single-directory idiom is the one nearly every `basemodel.py`
in `examples/` uses, and it is the safe one. There's a practical risk too: a maintainer who opens
that file to check the concern will find no leak, and could reasonably conclude the concern is
unfounded. It isn't. The conclusion is correct and only the example is wrong.

**What does reproduce is a module that inserts a second, *different* directory.** Same harness,
shape lifted from a real example and traced below:

```
mod_pair.py     Core inserted: /tmp/fx/pair/sub
      LEAKED after pop(0): ['/tmp/fx/pair/sub']
```

**It's quantified.** 12 example `.py` files manipulate `sys.path`, 8 of them with `insert(0,
...)` and the other 4 with `append`:

```
$ grep -rl "sys\.path\.\(insert\|append\)" examples/ | wc -l
12
$ grep -rl "sys\.path\.insert" examples/ | wc -l
8
```

The 8 `insert(0)` files are the ones that matter, because `append` can't displace index 0. They
cluster in four places: `cloud-edge-speculative-decoding-benchmark` (4 files),
`industrialEI/pose-estimation-llio/.../basemodel.py`,
`phys_scene_gen/.../phys_scene_generator.py`, and `TAB/...` (2 files). Three of those four are
either `basemodel.py` or a module Core loads directly, so they sit on the `load_module` path,
not in test helpers.

**It fires on the success path, which is where `finally` stops helping.** Take
`examples/cloud-edge-speculative-decoding-benchmark/testalgorithms/speculative-decoding/algorithms/ar/drafter.py`,
which manipulates `sys.path` at module top level — that is, *during* `import_module`:

```python
MODULE_DIR  = os.path.dirname(os.path.abspath(__file__))
SPECDEC_DIR = os.path.dirname(os.path.dirname(MODULE_DIR))
for path in (MODULE_DIR, SPECDEC_DIR):
    if path not in sys.path:
        sys.path.insert(0, path)
```

Sequence for `load_module(".../algorithms/ar/drafter.py")`:

1. Core inserts `.../algorithms/ar` at position 0.
2. `import_module("drafter")` runs the block above. `MODULE_DIR` is `.../algorithms/ar`, which
   is **already present, because Core just inserted it**, so the `if path not in sys.path`
   guard skips it. `SPECDEC_DIR` isn't present, so it goes in at 0.
3. `sys.path` is now `[SPECDEC_DIR, .../algorithms/ar, ...]`.
4. `pop(0)` removes `SPECDEC_DIR`, the example's own entry, and Core's entry stays behind
   permanently.

Step 2 is the non-obvious part: Core's insert is what causes the module's own idempotence guard
to skip its own directory, which is what leaves a foreign path at index 0. The two mutations
conspire. And because the imports that need `SPECDEC_DIR` (`common.runtime`, `common.modeling`,
`common.config`) all resolve *before* the pop, the immediate load succeeds. The damage is
latent, and resurfaces later as an import failure or a stale path in an unrelated module. No
exception is involved anywhere, so a `finally` can't catch it.

**The risk I'd actually raise is merge order.** This is the smallest and cleanest of the three
diffs, which makes it the most likely to get merged. Merging it closes the issue while leaving
the misidentification in place. The cluster already contains the better fix in #806. It'd be a
shame to merge past it.

Suggestions:

- Adopt snapshot/restore for both `py2dict` and `load_module`, as #806 now does with
  `sys.path[:] = old_sys_path`. A few more lines, and the positional assumption disappears.
- Whichever PR lands, the regression fixture that matters is a module inserting a **different**
  directory at index 0 during its own import, asserting `sys.path` is unchanged afterwards.
  #806 ships four tests across success and failure paths; I haven't read them, so I can't say
  whether that particular shape is among them. Worth confirming, because it's the shape that
  separates the three fixes from each other.
- Separately, `examples/yaoba/.../demo.py` and `TTA_strategy.py` append absolute
  `/home/wjj/...` paths (#515), which will interact with any `sys.path` hygiene work here.

**Where I could be wrong, and why I don't think I am.** The obvious counter is that `pop(0)` is
safe because Core inserted at 0 and Python is single-threaded, so nothing can get in between:
the entry it inserted really is still at index 0 when the pop runs. That holds for every case
except the one where the imported module inserts a directory Core did **not** insert, and the
`drafter.py` trace above is exactly that case: the mutation happens inside `import_module`, in
the window between Core's insert and Core's pop, so single-threadedness offers no protection at
all. The distinction is the whole point of the fixture pair above. Modules that re-insert their
own directory are protected by Core's insert; modules that reach for a parent or sibling
directory are not. What lifts it from theoretical to reachable is the grep: 12 example files
touch `sys.path`, 8 of them with `insert(0, ...)`. This particular file also sits on the load
path of a shipped example, not in a test fixture. I also considered the weaker objection, that
the leak is harmless because the imports resolve anyway. They do resolve, and that's precisely
what makes it dangerous: the load succeeds, so nothing surfaces at the point of the bug, and the
corrupted `sys.path` is inherited by every later example in the same process.

**On deciding the cluster deliberately, instead of letting merge order decide it.** Merging any
one of
#729, #759 or #806 will close the underlying issue, and #729 is the smallest and cleanest
diff, which makes it the most likely to be picked — so the likely outcome is that the cluster
closes on the fix that doesn't address the positional assumption. There's a second-order hazard
too: #729 and #759 both modify the same region of `core/common/utils.py`, so merging one forces
a rebase of the other, and a mechanical rebase could easily produce a `finally` block that pops
the wrong entry without any reviewer noticing. My suggestion is to decide the cluster as a unit
and close two of the three deliberately.

**Recommendation: reject**, not because the change is wrong, but because it shouldn't be the
cluster's fix. It's a strict improvement over current `main` and I'd have no objection to it in
isolation; in context, merging it closes the issue while leaving the positional assumption in
place. My preference is #806, extended with a regression fixture that calls `sys.path.insert(0,
...)` at import time, with #729 and #759 closed with credit to their authors. Three people
independently found a real Core bug and the diagnosis in each is sound. If maintainers would
rather use this PR as the base, then it needs the snapshot/restore change and the fixture,
which makes it a **major revision** rather than the one-line move it currently is.

suhaan-24:
The intent here is right, and the `load_module` half in particular fixes a real leak — the original
popped **inside** `try`, before `except`, so any import failure left `sys.path` permanently modified
for the rest of the process.

One concern with the mechanism.

## `pop(0)` removes whatever is at index 0, not necessarily what was inserted

```python
    sys.path.insert(0, module_path)
    try:
        importlib.import_module(module_name)
    except Exception as err:
        raise RuntimeError(...) from err
    finally:
        sys.path.pop(0)
```

If the module being imported inserts its own entry at index 0 during import, the `finally` block
removes **that** entry and leaves this function's `module_path` behind — reintroducing the leak the
PR is closing, and additionally corrupting the imported module's path setup.

That is not hypothetical in this repository. Eight call sites across four example families insert at
index 0 at import time:

```
examples/TAB/.../cloud_model.py:24                    sys.path.insert(0, bench_root)
examples/TAB/.../test_env/privacy_metrics.py:14       sys.path.insert(0, bench_root)
examples/cloud-edge-speculative-decoding-benchmark/.../ar/drafter.py:11      sys.path.insert(0, path)
examples/cloud-edge-speculative-decoding-benchmark/.../ar/verifier.py:11     sys.path.insert(0, path)
examples/cloud-edge-speculative-decoding-benchmark/.../block/drafter.py:16   sys.path.insert(0, path)
examples/cloud-edge-speculative-decoding-benchmark/.../block/verifier.py:17  sys.path.insert(0, path)
examples/industrialEI/pose-estimation-llio/.../llio_fusion/basemodel.py:29   sys.path.insert(0, current_dir)
examples/phys_scene_gen/.../ai2holodeck/phys_scene_generator.py:25           sys.path.insert(0, parent_dir)
```

The four speculative-decoding call sites insert inside a loop over candidate paths, so more than one
entry can be added by a single import.

### Suggestion

Snapshot and restore instead of popping by position:

```python
    saved = sys.path[:]
    sys.path.insert(0, module_path)
    try:
        importlib.import_module(module_name)
    except Exception as err:
        raise RuntimeError(f"load module(url={url}) failed, error: {err}") from err
    finally:
        sys.path[:] = saved
```

This costs one list copy per load and makes no assumption about what the imported module does. The
same applies to the `py2dict` hunk.

## Coordination note

Six other open PRs currently modify `sys.path` handling in this same file — #552, #558, #611,
#667, #759 and #806 — against two open issues (#728, #805). Worth checking which of these is
intended to land, since several rewrite the same few lines. For what it is worth, #806 already
takes the snapshot approach described above, and #558 is solving a different problem (module-name
collision via hashed module names) that would interact with any change here.

## Verification boundary

The four `sys.path.insert(0, ...)` call sites were located with `git grep` over `upstream/main` and
are quoted verbatim. The duplication list was computed by diffing each PR against its merge base.
I have **not** reproduced the index-shift end to end — no example currently on the lifelong-learning
path triggers it — so this is a latent risk rather than an observed failure, and I would not treat
it as a merge blocker.

PR #734, opened 2026-08-09
[PHASE IV] fix(RoboDK Palletizing): replace hardcoded image dimensions in map50/map90 metric calculation with dynamic image resolution

### LINKED ISSUE #730: [PHASE IV][RoboDK Palletizing]: Silent map50/map90 Score Corruption via Hardcoded Metric Dimensions
### What happened

While investigating the RoboDK Palletizing example (`examples/RoboDK Palletizing/singletask_learning_bench/`), I found a silent math bug in the evaluation metric scripts `testenv/map50.py` and `testenv/map90.py`:

1. **Hardcoded image dimensions during coordinate conversion**
   Inside `read_label_file()`, normalized YOLO bounding box coordinates (`x_center`, `y_center`, `w`, `h`) are converted back to absolute pixel coordinates for IoU matching against ground truth using hardcoded values:
   ```python
   img_width = 640
   img_height = 480
   ```
   If a dataset uses any other resolution or aspect ratio (like 800x600), `read_label_file()` completely ignores it and scales the normalized values using 640 and 480 anyway. On an 800x600 frame, this creates a systematic **0.8x spatial scaling error**, shifting every ground-truth box out of place.

2. **Silent metric corruption (no exceptions)**
   Because the IoU fails to match the shifted ground-truth boxes with the correct predictions, `map50()` and `map90()` return garbage scores or `0.0` even when the model's predictions are completely right—and Python doesn't throw a single error.

3. **Internal resolution mismatch**
   To make matters worse, `testalgorithms/basemodel.py` defaults to square `640x640` training and inference, while the metric files assume non-square `640x480`. The example literally disagrees with itself.

---

* `read_label_file()` should dynamically check the image's actual resolution (via PIL or OpenCV) and scale the normalized YOLO boxes using true dimensions.
* The model wrapper and evaluation scripts should agree on the image size.

---

1. Check the hardcoded lines in the metric script:
   ```bash
   grep -n "img_width =\|img_height =" "examples/RoboDK Palletizing/singletask_learning_bench/testenv/map50.py"
   ```
   Output:
   ```text
   img_width = 640
   img_height = 480
   ```

2. Run a synthetic test with an 800x600 image and a centered YOLO label (`0 0.5 0.5 0.25 0.25`):
   * Correct box for 800x600: `[300, 225, 500, 375]`
   * What `read_label_file()` actually outputs: `[240, 180, 400, 300]`

---

Verified with a synthetic test harness. Feeding a prediction scaled for 800x600 got a `0.0` mAP50 because of the coordinate shift, but feeding boxes manually pre-shifted to the buggy 640x480 scale scored `0.995`. Path matching works fine; the score corruption is 100% caused by the hardcoded dimensions.

*Found while investigating RoboDK Palletizing for LFX Mentorship Fall 2026 (Ianvs Example Restoration, Phase IV).*

hayyanbaig:
### What type of PR is this?

/kind bug

While investigating the RoboDK Palletizing example for LFX Mentorship Fall 2026 (Ianvs Example Restoration, Phase IV), I found and fixed a silent scoring-corruption bug in the evaluation metric scripts (#730):

**Replace Hardcoded Image Dimensions with Dynamic Resolution (map50.py, map90.py)**
Both `map50.py` and `map90.py` hardcoded `img_width=640, img_height=480` inside `read_label_file()` and applied these values unconditionally when converting normalized YOLO bounding box coordinates to absolute pixel coordinates. If a dataset's actual images are any other resolution, every ground-truth box is silently shifted to the wrong location, and `map50()`/`map90()` return corrupted or near-zero scores with no exception ever raised. Fixed both files to dynamically read each image's real dimensions via PIL at evaluation time, falling back to the historical 640x480 default only if the image cannot be found or read.

**Clarify Training Resolution vs. Raw Dataset Resolution (basemodel.py)**
The example's model wrapper defaults to a square 640x640 `image_size` (YOLO's internal training/inference resize target), which is unrelated to the dataset's actual raw image dimensions but could easily be misread as the same thing given the metric files now,removed hardcoded 640x480 assumption. Added a clarifying log line at initialization explaining that YOLO's resize target is independent of the dataset's raw image size, which the metric scripts now handle dynamically.

Fixes #730

### Testing done:

Reproduction Script: Built a synthetic 800x600 test image with a centered YOLO-format label (`0 0.5 0.5 0.25 0.25`). Prior to the fix, feeding a correctly-scaled prediction (`[300, 225, 500, 375]`) into `map50()` returned `0.0` due to the coordinate mismatch caused by the hardcoded 640x480 assumption. After the fix, the same test scores `0.995` on both `map50()` and `map90()`, confirming ground-truth coordinates are now computed against the image's actual dimensions.


Hiteshsai007:
### Request changes — undeclared Pillow dependency and unsafe image-dimension fallback

I checked commit `646b085829fa09d83a4f0c2730102aaaaf50cf9a`. The changed Python files compile successfully, and the dynamic-resolution idea is useful.

There are two merge-blocking concerns:

1. Both `map50.py` and `map90.py` add:

```python
from PIL import Image

PR #738, opened 2026-08-10
fix(government): restore government_rag example benchmark

### LINKED ISSUE #362: [Bug Report] government_rag Example: 7 Execution-Blocking Bugs Prevent Any User From Running This Benchmark

ahsan-zamil:
**What type of PR is this?**

/kind bug

Restores the `government_rag` example benchmark by addressing the 7 execution-blocking bugs:
- **Paths & Imports**: Dynamically resolves the import path for `gov_rag` in `basemodel.py` and standardizes hardcoded absolute developer paths to relative paths in YAMLs.
- **Dynamic Config**: Retrieves model credentials and embedding configurations dynamically from environment variables (e.g. `DEEPSEEK_API_KEY`, `RAG_MODEL_URL`) with fallback defaults.
- **Location Resolution**: Dynamically reads the folders in the dataset directory to populate the `self.all_locations` RAG locations variable.
- **Device Fallback**: Implements dynamic GPU/CPU fallback to run smoothly on CPU-only machines.
- **API deprecations**: Guards deprecated LangChain (`invoke` fallback) and ChromaDB (`persist()` fallback check) API calls.

Fixes #362


christinaliyevav-coder:
I looked at this PR as an alternative restoration of the same `government_rag` example reviewed in PR #536.

Using environment variables here is a useful improvement, but I found that missing variables still fall back to placeholder credential strings:

```python
api_key = os.getenv("DEEPSEEK_API_KEY") or "<DeepSeek API Key>"
token = os.getenv("SILICONFLOW_API_KEY") or "<token>"
client_id = os.getenv("QIANFAN_ACCESS_KEY") or "[应用API Key]"
client_secret = os.getenv("QIANFAN_SECRET_KEY") or "[应用Secret Key]"
```

I reproduced these assignments directly from the PR source with all four environment variables removed. The result was:

```text
api_key: <DeepSeek API Key>
token: <token>
client_id: [应用API Key]
client_secret: [应用Secret Key]
```

So a missing prerequisite is not rejected as a configuration error. It is converted into a non-empty placeholder value that can continue into the backend request code.

This is especially relevant because Issue #362 already describes the desired direction as reading credentials from environment variables and reporting a clear error when they are missing. The environment-variable part is implemented here, but the placeholder fallback preserves the original failure pattern when configuration is absent.

I also compared this implementation with PR #536. The two PRs use different credential contracts (`require_env(...)` versus placeholder fallback), and a direct merge simulation produced content conflicts in both `basemodel.py` and `gov_rag.py`.

I would keep the environment-variable configuration, remove the placeholder fallbacks, and fail explicitly when credentials required by the selected backend are missing.

PR #739, opened 2026-08-10
fix(examples): remove protobuf<=3.20.3 pin conflicting with core onnx

### LINKED ISSUE #735: examples/industrialEI/pose-estimation-llio: requirements.txt pins protobuf<=3.20.3, conflicting with core onnx requirement
/kind bug

### What happened

`examples/industrialEI/pose-estimation-llio/requirements.txt` pins `protobuf<=3.20.3`.

Core ianvs installs `onnx` via the root `requirements.txt`, and onnx 1.22.0 declares `protobuf>=4.25.1`. Following step 2 of the example's README therefore downgrades protobuf below the floor onnx declares, and pip reports the conflict:

```
Installing collected packages: protobuf
  Attempting uninstall: protobuf
    Found existing installation: protobuf 7.35.1
    Uninstalling protobuf-7.35.1:
      Successfully uninstalled protobuf-7.35.1
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
onnx 1.22.0 requires protobuf>=4.25.1, but you have protobuf 3.20.3 which is incompatible.
Successfully installed protobuf-3.20.3
```

Installing an example's requirements should not violate a constraint declared by a core dependency.

### How to reproduce it

```bash
# core install, per docs/guides/how-to-install-ianvs.md
pip install -e .

# example install, per examples/industrialEI/pose-estimation-llio/README.md step 2
pip install -r examples/industrialEI/pose-estimation-llio/requirements.txt
```

### Scope

To be precise about impact: `import onnx` still succeeds under protobuf 3.20.3, so I have not observed a runtime failure — only the declared incompatibility reported by pip. Code paths that exercise ONNX serialization may behave differently; I have not tested those.

### Environment

- OS: Ubuntu 24.04.2 LTS (GitHub Codespaces)
- Python: 3.12.1
- ianvs: 0.3.0, installed from source via `pip install -e .`
- onnx: 1.22.0

### Possible directions

Two options, depending on why the pin was added:

1. **Drop the pin.** The README already documents `export PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python`, which addresses the protobuf 4.x descriptor incompatibility that a `<=3.20.3` pin is usually there to avoid. If that is the reason here, the pin is redundant with the documented workaround.
2. **Keep the pin and document the constraint.** If the pin is genuinely required, this example cannot share an environment with core `onnx`, and the README should say so — or `onnx` could be made optional, which is what #450 proposes for `multiedge_inference.py`.

Happy to open a PR once maintainers indicate a preference.

### Additionally

`requirements.txt:6` refers to a script that does not exist in the repository:

```bash
$ find examples/industrialEI -name "run_ianvs.sh"

$ grep -rn "run_ianvs" examples/ docs/
examples/industrialEI/pose-estimation-llio/requirements.txt:6:# Or use the provided run_ianvs.sh script
```

Either the script should be added or the reference removed. Happy to fold this into the same PR.

Related: #450, #635

harshaaaaw:
Fixes #735

## Problem

`examples/industrialEI/pose-estimation-llio/requirements.txt` pins `protobuf<=3.20.3`, but core ianvs installs `onnx` (via the root `requirements.txt`), and onnx >= 1.22.0 declares `protobuf>=4.25.1`. Following the example's install steps therefore downgrades protobuf below onnx's floor and pip reports the conflict.

## Fix

The pin is redundant: the example README already documents `export PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python` as the workaround for the protobuf 4.x descriptor incompatibility. This PR:

- Drops the `<=3.20.3` pin so the example shares the same environment as core ianvs
- Documents the compatibility note in `requirements.txt`
- Removes the reference to the non-existent `run_ianvs.sh` script in the requirements comment

## Verification

- `pip install -r examples/industrialEI/pose-estimation-llio/requirements.txt` no longer downgrades protobuf below onnx's floor
- No remaining `run_ianvs` references in `examples/` or `docs/`
- `import onnx` succeeds under protobuf >= 4.25.1

Related: #450


subham-hq:
Thanks for picking this up. One thing I'd flag for the maintainers: neither of us can run the example end-to-end without the KITTI dataset, so the removal is untested against the failure the pin was originally guarding against.

The README's PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python should cover the protobuf 4.x descriptor incompatibility, but that's inference rather than something I've verified. Worth someone with the dataset confirming before merge.

31groot:
Dependency-boundary correctness

**Recommendation: minor revision.**

The proposed direction is correct: removing the Example-local `protobuf<=3.20.3` constraint avoids forcing the shared environment below the protobuf floor required by Core's `onnx` dependency.

The PR description also reports successful dependency installation and `import onnx` verification.

The remaining risk is verification/merge hygiene rather than the core dependency change.

1. The current commit message is rejected by the repository bot because it contains an `@`/`#` keyword sequence. This should be corrected before merge.
2. The full LLIO example path has not been executed because the KITTI dataset is unavailable. Therefore, the verification should be stated precisely: dependency installation and `import onnx` are verified, but the complete Example benchmark remains unverified.
3. The removal of the protobuf upper bound is preferable to replacing it with another Example-local upper bound, because the latter could recreate the same Core/Example dependency conflict.

Overall, I consider the dependency-direction change technically sound, subject to the commit-message correction and the stated verification limitation.

henryng15 (commented):
Reviewed at head `900ae72` against `37a9c60`. The direction is right — a
`protobuf<=3.20.3` ceiling in an Example does conflict with the `onnx` that core Ianvs
installs. Two points before merge.

**1. The added comment states a constraint the change does not encode.**

```text
# Note: must stay compatible with onnx (installed by core ianvs), which requires
# protobuf>=4.25.1.
protobuf
```

The requirement line is now completely unconstrained, so a resolver may still select
protobuf 3.x — the situation this PR exists to prevent. If `>=4.25.1` is the correct
floor, it belongs in the requirement rather than in a comment above it:

```text
protobuf>=4.25.1
```

That is a one-line change and it makes the PR self-enforcing.

**2. The floor is derived from an unpinned dependency.** Root `requirements.txt:7` lists
`onnx` with no version specifier at all, so which protobuf version onnx actually demands
is itself free to move. Documenting a floor read off today's onnx is fragile. Either
constrain `onnx` too, or phrase the comment as "onnx currently requires >= 4.25.1 as of
<version>" so a future reader knows it needs rechecking.

**Confirmed correct:** removing the `# Or use the provided run_ianvs.sh script` line is
right — no `run_ianvs.sh` exists anywhere under `examples/industrialEI/`.

For context on why the encoded-vs-documented distinction matters here: across all Example
requirements files in the repository, 302 of 536 requirement lines carry no version
constraint at all. Comments describing intended bounds are common; enforced bounds are
not.

**Recommendation: minor revision** — change `protobuf` to `protobuf>=4.25.1`.

Mahendr99ar:
Cross-posting from my LFX 2026 Term 3 pretest analysis (full review: discussion #984).

Recommendation: minor revision. Straightforward, correctly-scoped fix — removes the protobuf<=3.20.3 pin that conflicted with onnx 1.22.0's protobuf>=4.25.1 floor. I independently confirmed the conflict via pip's dry-run resolver targeting this issue's exact Python 3.12/manylinux environment, matching the pip output in #735 exactly.

Same caveat as similar fixes elsewhere: this resolves the specific instance but not the underlying gap that let it happen (no check between Example and Core requirements) — see discussion #984 for the broader pattern, also connected to #666.

PR #741, opened 2026-08-10
fix: enforce the non-empty test split invariant in FCI evaluation

### LINKED ISSUE #740: Empty test splits in FCI evaluation yield a nan metric, then ZeroDivisionError
**What happened**:

`FederatedClassIncrementalLearning._split_test_dataset` computes `step = all_data // split_time`. When the test dataset holds fewer samples than `incremental_rounds`, `step` is `0` and every split but the last comes back empty, with no error and no log:

```
5 test samples split into 10 rounds -> split sizes [0, 0, 0, 0, 0, 0, 0, 0, 0, 5]
```

`evaluation` then fails in two stages. The accuracy loop never executes, so `old_class_acc_list` is empty and `np.mean([])` writes `nan` into `system_metric_info["task_avg_acc"]["accuracy"]`. That is silent, since `nan` is a valid float and every comparison against it is `False`. On the following round the forget rate divides by `len(old_class_acc_list)`:

```python
current_forget_rate = (
    max_acc_sum / len(old_class_acc_list) if incremental_round > 0 else 0.0
)
```

The ternary checks `incremental_round`, but the divisor is `len(old_class_acc_list)`, so the run stops with `ZeroDivisionError` one round after the metric was already corrupted, and two layers away from the split that caused it.

A test dataset that cannot supply one sample per incremental round should be rejected where the splits are created, with a message saying what to change. Failing that, an empty split should at worst produce a logged `0.0`; it should never place `nan` in the reported metrics.

```python
from core.testcasecontroller.algorithm.paradigm.federated_learning.\
    federated_class_incremental_learning import FederatedClassIncrementalLearning as FCIL


class StubDataset:
    test_url = "unused"

    @staticmethod
    def load_data(*_args, **_kwargs):
        class Data:
            x = list(range(5))
            y = list(range(5))
        return Data()


paradigm = object.__new__(FCIL)
paradigm.dataset = StubDataset()
print([len(s["x"]) for s in paradigm._split_test_dataset(10)])
# [0, 0, 0, 0, 0, 0, 0, 0, 0, 5]

paradigm = object.__new__(FCIL)
paradigm.accuracy_func = lambda y_true, y_pred: 1.0
paradigm.accuracy_per_round = []
paradigm.system_metric_info = {"task_avg_acc": {}}
paradigm.get_global_model = lambda: None
empty_split = [{"x": [], "y": []}]

print(paradigm.evaluation(empty_split, 0))
# 0.0, and the log shows "task avg acc: nan"
print(paradigm.evaluation(empty_split, 1))
# ZeroDivisionError: division by zero
```

The stubs stand in for the dataset and the paradigm job so the defect can be shown without a full benchmark run; the code paths exercised are the real ones.

This is latent rather than observed. The shipped examples do not reach it, since cifar100 provides 10000 test samples against `incremental_rounds` of 2 (`fci_ssl/fedavg`), 10 (`fci_ssl/glfc`) and 50 (`fci_ssl/fed_ci_match_v2`), so `step` is never `0`. It becomes reachable as soon as the benchmark is pointed at a smaller or custom dataset, which is the common case when trying the paradigm out for the first time.

Worth noting that the current guard is correct while every split is non-empty: `incremental_round > 0` and a check on `len(old_class_acc_list)` are equivalent in that case, because with `incremental_round == 0` the inner loop never runs and the numerator is `0` as well. The defect is the missing invariant, not the expression itself.

/kind bug

Eason09053360:
**What type of PR is this?**

/kind bug

`FederatedClassIncrementalLearning.evaluation` assumes every test split holds at least one sample, and nothing enforces that assumption.

`_split_test_dataset` computes `step = all_data // split_time`, which is `0` when the test dataset has fewer samples than `incremental_rounds`. Every split but the last is then empty, with no error and no log:

```
5 test samples split into 10 rounds -> split sizes [0, 0, 0, 0, 0, 0, 0, 0, 0, 5]
```

While the assumption holds, the existing code is correct: the forget rate guards on `incremental_round > 0` rather than on its divisor `len(old_class_acc_list)`, but the two are equivalent as long as the accuracy list is non-empty. Once a split is empty they diverge, and the failure arrives in two stages:

1. `np.mean([])` returns `nan`, which is written into `system_metric_info["task_avg_acc"]["accuracy"]`. This is silent: `nan` is a valid float and every comparison against it is `False`, so it reaches the report unnoticed.
2. The next round evaluates `max_acc_sum / len(old_class_acc_list)` with `incremental_round > 0`, and raises `ZeroDivisionError`.

**This is a latent defect, not an observed one.** The shipped examples do not reach it: cifar100 provides 10000 test samples against `incremental_rounds` of 2 (`fci_ssl/fedavg`), 10 (`fci_ssl/glfc`) and 50 (`fci_ssl/fed_ci_match_v2`), so `step` is never `0`. It becomes reachable when the benchmark is pointed at a smaller or custom dataset, where the first symptom is a `nan` metric one round *before* the crash, two layers away from the split that caused it.

Changes:

- `_split_test_dataset` rejects a test dataset smaller than `incremental_rounds` with an actionable message, so a broken invariant is reported where it originates instead of surfacing later as a metric. The style follows the existing `ValueError` in `federated_learning.py`.
- `evaluation` guards the task average accuracy and the forget rate on `old_class_acc_list` itself, the value that can actually be zero, and logs a warning when a round evaluates no sample. This stays a warning rather than an error, since the upstream check already covers the main path and a single empty split should not abort a whole benchmark run.

Round 0 behaviour is unchanged. With `incremental_round == 0` the inner loop `for j in range(incremental_round)` never runs, so `max_acc_sum` is `0` and `0 / len(...)` yields the same `0.0` the old ternary returned. The new guard is a strict superset of the old one.

Deliberately not changed: the last split still absorbs the remainder, so 105 samples over 10 rounds give the last round 15. That is intentional so no test sample is dropped, and changing it would move existing benchmark numbers.

<details>
<summary>Reproduce (before / after)</summary>

```python
from core.testcasecontroller.algorithm.paradigm.federated_learning.\
    federated_class_incremental_learning import FederatedClassIncrementalLearning as FCIL


class StubDataset:
    test_url = "unused"

    @staticmethod
    def load_data(*_args, **_kwargs):
        class Data:
            x = list(range(5))
            y = list(range(5))
        return Data()


paradigm = object.__new__(FCIL)
paradigm.dataset = StubDataset()
print([len(s["x"]) for s in paradigm._split_test_dataset(10)])
# before: [0, 0, 0, 0, 0, 0, 0, 0, 0, 5]
# after:  ValueError: the test dataset only has 5 samples, which is less than
#         the number of incremental rounds (10). ...

paradigm = object.__new__(FCIL)
paradigm.accuracy_func = lambda y_true, y_pred: 1.0
paradigm.accuracy_per_round = []
paradigm.system_metric_info = {"task_avg_acc": {}}
paradigm.get_global_model = lambda: None
empty_split = [{"x": [], "y": []}]

print(paradigm.evaluation(empty_split, 0))
# before: 0.0, but task_avg_acc is silently set to nan
print(paradigm.evaluation(empty_split, 1))
# before: ZeroDivisionError: division by zero
# after:  0.0, with "no test sample is evaluated in round 1" logged
```

</details>

Tested with the snippet above, before and after the change. Normal splitting is untouched: `5 samples / 5 rounds -> [1, 1, 1, 1, 1]` and `5 samples / 2 rounds -> [2, 3]`, so the new check only rejects configurations that could not have produced a meaningful evaluation. `pylint --max-positional-arguments=10 core`, the CI invocation for Python 3.9, rates the file 10.00/10.

Fixes #740


AdityaKumarSethia:
Reproduced this fix against the real patched file (via `git worktree`, since this file uses a relative import and can't be loaded standalone):

```bash
git worktree add /tmp/ianvs-pr741 pr-741
cd /tmp/ianvs-pr741
python3 -c "
import sys, tempfile, os
sys.path.insert(0, '.')
from core.testcasecontroller.algorithm.paradigm.federated_learning.federated_class_incremental_learning import FederatedClassIncrementalLearning

tmp_dir = tempfile.mkdtemp()
test_file = os.path.join(tmp_dir, 'test.txt')
with open(test_file, 'w') as f:
    for i in range(5):
        f.write(f'img_{i}.jpg {i % 2}\n')

fci = object.__new__(FederatedClassIncrementalLearning)
ds = object.__new__(Dataset)
ds.test_url = test_file
fci.dataset = ds

result = fci._split_test_dataset(split_time=10)
print(result)
"
```
Confirmed working: raises a clear `ValueError` naming the exact sample/round mismatch ("the test dataset only has 5 samples, which is less than the number of incremental rounds (10)"), instead of silently producing 9 empty splits like unpatched `main` does. This is the more defensible failure mode of the two possible fixes here, loud and immediate versus silent corruption discovered only after a full run.

This bug is one instance of a broader pattern (unguarded division by a count that can legitimately be zero) recurring across 3 unrelated subsystems, also #591 and #462. Full cross-issue analysis in [Discussion #873](https://github.com/kubeedge/ianvs/discussions/873).

AdityaKumarSethia:
Follow-up to my review above. While auditing sibling methods in `dataset.py` for the same unguarded-division pattern (part of a broader cross-issue review, see [Discussion #873](https://github.com/kubeedge/ianvs/discussions/873)), found that this PR's guard doesn't cover one adjacent edge case.

`_split_test_dataset`'s guard is `if all_data < split_time: raise ValueError(...)`. This correctly catches the case originally reported in #740 (`all_data` smaller than `split_time`). It does not catch `split_time=0`, since `all_data < 0` is false for any real sample count, so the guard is silently bypassed and the original `step = all_data // split_time` still executes.

Reproduced against the real patched file, same method as my original review:

```python
# same setup as above, real patched file via git worktree
result = fci._split_test_dataset(split_time=0)
```
Output:
```
ZeroDivisionError: integer division or modulo by zero
```
Context on reachability: `split_time` here is `self.incremental_rounds`, a plain kwarg (`kwargs.get("incremental_rounds", 1)`) with no validation preventing `0`. Checked all 21 configs in the repo that set `incremental_rounds` explicitly, none currently use `0`, so this isn't actively breaking anything today. But it's one config edit away, and worth closing while this PR is still open, either an `elif split_time <= 0: raise ValueError(...)` alongside the existing check, or folding into `all_data <= 0 or split_time <= 0` as a single condition.

Separately, and probably more relevant to how this gets fixed: `_split_dataset` (a sibling method in `federated_learning.py`, called just before `_split_test_dataset` in `run()`) routes to `Dataset._splitting_more_times`, which has the exact same `split_time=0` gap, confirmed by direct execution against the real file. In the current call order, that one fails first, so today `_split_test_dataset`'s gap is masked, but both are real and both are the same root pattern. Not asking for a fix here specifically, since it may make more sense as a shared guard at the `Dataset` layer rather than patched twice, just flagging so it's visible while reviewing.

PR #744, opened 2026-08-10
Feat/ci validate example configs

### LINKED ISSUE #743: Add CI check to catch broken example config paths before merge

Add a lightweight CI validator that checks whether the configuration paths referenced by `examples/*/benchmarkingjob.yaml` actually exist, following the `testenv → algorithm → module` reference chain.

Integrate it into GitHub Actions so new broken references fail CI, while a baseline file allows existing issues to remain without blocking the build.

**Why is this needed**:

Many Ianvs examples currently contain broken, stale, or developer-specific paths. Static analysis found around 40 such references across the examples. These failures are often discovered only when someone tries to run an example.

A lightweight CI check would catch new path regressions before merge without requiring GPU, PyTorch, Sedna, or other heavy runtime dependencies.

iron-prog:
**What type of PR is this?**

/kind feature

Adds a lightweight CI validator for Ianvs example configuration paths.

The validator checks the `benchmarkingjob.yaml → testenv → algorithm → module` reference chain and reports broken or missing paths before they reach users. It does not require GPU, PyTorch, Sedna, or other heavy runtime dependencies.

A baseline is used for existing broken references, so CI only fails when a new regression is introduced.

Adds CI validation for broken example configuration paths as proposed in the related issue.

Fixes #743


diegodevxd:
The validator's YAML-error behavior and its explicit dataset key list are useful, but merging it as a new workflow would create a third path-validation authority alongside merged #771 and open #835. Different baseline formats and severities will make one CI job pass while another reports the same example differently.

Please extract/reuse this chain traversal in #771's existing validator instead of adding `scripts/validate_example_configs.py` as parallel policy. Keep baseline entries traceable to an issue and removable, and test malformed YAML plus a missing checked-in module separately from absent external datasets.

akshita317:
This is a useful check and the baseline design is the right call — grandfathering existing breakage lets the validator land without requiring a repository-wide fix first, which is what usually stalls this kind of PR.

Two findings.

**1. The baseline already contains at least one entry that no longer reproduces.**

```
examples/cifar100/fci_ssl/fedavg/benchmarkingjob.yaml: INVALID YAML - while scanning
a simple key ... line 17, column 5 could not find expected ':' ... line 18, column 7
```

I hit this same error while testing and traced it: the file contained `algorithms:conda` — a stray token in the YAML. That has since been fixed on `main`, and the file now parses cleanly. A baseline entry that no longer reproduces will silently absorb a genuine future regression in that file.

Two suggestions: regenerate the baseline at merge time, and consider failing the check when a baseline entry *stops* reproducing, so the file cannot rot. That converts the baseline from a static allowlist into a burn-down list.

**2. Coverage note.** The validator checks that referenced files exist, which catches the stale-path class well. It does not check whether configuration *keys* are consumed, and there is a failure mode where every path is valid and the example still cannot run: `Dataset._parse_config` (`core/testenvmanager/dataset/dataset.py:71-73`) assigns only keys already declared on the instance, so an undeclared key is discarded whatever its value.

`examples/yaoba/**/testenv.yaml` declares `image_folder_url`, `known_dataset_url` and `unknown_dataset_url` with resolvable paths, and all three are dropped — while `core/.../singletask_learning_active_boost.py:49,61,62` reads them. That config passes this validator and still raises `AttributeError` at runtime.

A complementary schema check is short and derives its invariant from Core alone:

```python
# every attribute Core reads as self.dataset.<attr> must be declared in Dataset.__init__
import re, pathlib
src = open("core/testenvmanager/dataset/dataset.py", encoding="utf-8").read()
declared = set(re.findall(r"self\.(\w+)\s*:\s*\w+\s*=", src))
reads = set()
for p in pathlib.Path("core").rglob("*.py"):
    reads |= set(re.findall(r"self\.dataset\.(\w+)",
                            p.read_text(encoding="utf-8", errors="ignore")))
missing = sorted(reads - declared - {"load_data", "split_dataset", "process_dataset"})
assert not missing, f"core reads undeclared Dataset attributes: {missing}"
```

On current `main` this reports `['image_folder_url', 'known_dataset_url', 'unknown_dataset_url', 'val_url']`. Might be a natural addition to this workflow.

Also worth coordinating: **PR #835** adds a very similar validator (`scripts/check_example_paths.py` + `scripts/known_broken_paths.txt`) and modifies the same `.github/workflows/main.yaml`, so the two will conflict.

Yash4616:

The baseline design is the right shape for landing a validator against an existing backlog: grandfather what is already broken, fail only on new regressions. That part I would keep.

The problem is what the baseline matches on. Entries are stored as fully rendered message strings, including resolved paths:

```
examples/RoboDK Palletizing/singletask_learning_bench/benchmarkingjob.yaml: algorithm 'YOLOv8n'
  config does not exist -> /root/ianvs/project/ianvs-0.3.0/examples/Palletizing/...
```

That makes the baseline a snapshot of message text rather than of underlying facts. Consider a PR that partially fixes one of these entries, for example correcting the `testenv` path but not yet the algorithm path. The rendered message changes, the baseline entry no longer matches, and the checker reports it as a new regression and fails the build.

The mechanism intended to let the validator land without a repository-wide cleanup would then fire against precisely the PRs doing that cleanup, which is the opposite of the intent.

Keying each entry on the config file plus the offending field, rather than on the full rendered message, would keep the grandfathering behaviour while surviving partial fixes.

Recommendation: minor revision. Keep the baseline, change what it keys on.

ROGUESA1LOR:
Adjacent but distinct finding from @akshita317's undeclared-key discard
issue above: even for the 8 dataset fields that ARE declared and consumed
(train_url, test_url, train_index, test_index, train_data, test_data,
train_data_info, test_data_info), this validator's check_testenv() treats a
wrong VALUE — specifically an absolute, machine-specific path — as a
non-blocking dataset_notes entry rather than an error. Reproduced against
RoboDK Palletizing (#781) and Cloud_Robotics (#816), both of which declare
these fields correctly but with hardcoded developer-specific paths
(/root/ianvs/..., /home/aryan/...). Recommend distinguishing "absolute
machine-specific path" (should error) from "missing externally-downloaded
dataset" (should stay non-blocking) — full analysis:
https://github.com/kubeedge/ianvs/discussions/953

shrav784:
Thanks for the thorough discussion above — @akshita317's declared-key finding and @Yash4616's baseline-keying issue are both well-evidenced.

One more angle, verified by actually running the script locally (main, PR #771 already merged): the Palletizing entry that's already in the baseline as a clean "config does not exist" message doesn't actually get reached that way when I run it —

    $ python3 scripts/validate_example_configs.py
    Found 44 benchmarkingjob config(s) under examples/
      File ".../validate_example_configs.py", line 74, in check_benchmarkingjob
        if not testenv_path.exists():
    PermissionError: [Errno 13] Permission denied: '/root/ianvs/project/ianvs-0.3.0/examples/Palletizing/singletask_learning_bench/testenv/testenv.yaml'

`Path.exists()` raises `PermissionError` rather than returning `False` when the resolved path sits under a directory the current user can't read (`/root/...`). The script has no `try/except` around this call, so it crashes with an unhandled exception before ever reaching the baseline-comparison logic being discussed above. This means the "only fail on new regressions" design @diegodevxd and @Yash4616 are evaluating may not actually be reachable for this specific entry in practice — worth checking whether this reproduces in CI or is local-environment-specific.

Also reproduced the identical crash, same file, same line pattern, in #835's script — happy to share full details if useful, or see the write-up here: [Discussion #973 link]

Liyuchen0118 (commented):
**Recommendation**

Major revision. Supersede the standalone implementation as an independent validation authority, while preserving non-duplicative structured traversal or baseline-expiry behavior in the validator already present on main.

**Tested head**

PR: `03f8611003421efd5d0a067b6ce0622fa1d4f10d`

Main: `37a9c60a9747af0cfe3170f84249bc349c56e8d5`

**Validation method**

The PR validator was run from its checkout with `scripts/validate_example_configs.py --baseline scripts/example_config_baseline.txt`; a separate focused fixture compared the PR behavior with main. Exit: `0`. Output: 40 reproduced baseline problems, 45 absent dataset notes, no new structural problems. The fixture harness also exited `0`.

**Observed result**

For an inactive commented value:

```yaml
# dataset_processor:
#   url: examples/nonexistent/processor.py
```

PR #744 returns no problems. Main's raw-text validator reports `ERROR` for line 6.

**Impact**

Blocking reference checks should operate on active structured values. Exit `0` for the full validator is not itself incorrect because the baseline intentionally allows known problems. The risks are dual authority, stale debt, and unowned allowlist entries.

**Requested change**

1. Move useful structured active-value traversal into the current validator.
2. Keep one CI authority.
3. If text debt remains, require an owner, reason, and expiry condition for each entry.
4. Detect stale baseline entries when a problem stops reproducing.

**Prior discussion**

- https://github.com/kubeedge/ianvs/pull/744#issuecomment-5381275765
- https://github.com/kubeedge/ianvs/pull/744#issuecomment-5381815928

**Additional evidence**

The focused fixture isolates the structured-versus-raw parsing difference without treating the full baseline's exit semantics as a defect.

SurbhiD404:
Findings from diff review and independent execution on this branch (03f8611):

1. PermissionError crash is platform-dependent, not universal. Ran scripts/validate_example_configs.py --baseline scripts/example_config_baseline.txt on Windows (exit 0, no crash -RoboDK Palletizing's /root/... paths reported as ordinary "file does not exist" entries) and on WSL/Ubuntu (PermissionError: [Errno 13] at line 74, testenv_path.exists(), on RoboDK Palletizing's testenv.yaml). This refines the earlier crash report — the "only fail on new regressions" gate is unreachable on Linux CI runners specifically, not on all platforms.

2. Duplicates #771's merged static_validator.py - a second parallel validator with different traversal/baseline format creates two definitions of "valid."

3. Merge conflict with #835 (both edit .github/workflows/main.yaml) blocks this from merging as-is.

4. Baseline keyed on rendered message strings will false-positive as "new regression" on any PR that partially fixes an existing entry.

Recommendation: major revision - fix the OS-dependent crash, consolidate into #771's validator, resolve the #835 conflict, switch baseline key to stable (file, field) identity.

SurbhiD404:
Repair Boundary Finding: The fix for the root cause (detection decoupled from gating) is consolidation into #771's static_validator.py, not adding a fourth parallel workflow. Before any consolidation, this PR's own bugs must be fixed:
1. PermissionError crash on /root/... absolute paths - self-verified this session as platform-dependent (crashes on WSL/Linux, not on Windows), making the "only fail on new regressions" gate structurally unreachable for Class A Examples on Linux CI runners specifically.
2. Malformed YAML false-success - must be a blocking ERROR.
3. Baseline keyed on rendered message strings - must use stable (file, field) identity.
4. Duplicates #771's merged validator.
5. Merge conflict with #835 on .github/workflows/main.yaml.

Restoration Path note: This crash must be fixed before any consolidation or baseline regeneration can safely proceed - first blocking step in the restoration path.

Issue #348, opened 2022-02-11
[Feature] Simulator for Edge-cloud Collaborative AI: a Lifelong Learning example


MooreZheng:
**Background**:
For most distributed AI scheme developers, it is often complicated and a heavy burden to establish and deploy a large-scale cloud-edge collaborative system. Currently, ianvs provide a algorithm tester in one node, and accuracy-wise metrics can be tested through test datasets. However, the system-wise metrics then become impossible to measure in a real-world manner for large-scale nodes and out-of-the box testing frequently. 

This issue then aims to provide an industrial-grade distributed collaborative system simulation for a large-scale deployment of an edge-cloud collaborative AI system. To show how it works, this issue selects an edge-cloud collaborative lifelong learning example for testing. 

**What contents are to be added/modified**:
In the test case controller of Ianvs, it is expected to add a simulation controller. It will provide Ianvs with a worker-in-worker system. Then even if one is using a signle machine, simulated workers can be generated to simulated multi-node systems. 

The modules in the simulation controller could be as follows:
- A Simulation Environment Administrator is used to
1. Parse the system config to establish the simulation
2. Check the host environment, e.g., check if the host has all requirements met, and whether memory > 4GB
3. Build the simulation environment
4. Create and deploy the modules needed in the simulation environment
5. Close and delete the simulation environment
- The Simulation Job Administrator is critical to manage the simulation job, and provides the following functions:
1. Build the images of algorithms to be tested, e.g., docker image
2. Generate the YAML file of the simulation job
3. Deploy and delete the simulation job with workers
4. List-watch the results of the simulation job with workers

==============
Appendix on Feb 11, 2022
The KubeEdge SIG AI is chartered to facilitate Edge AI applications with KubeEdge. An overview of SIG AI activities can be found in this [charter](https://github.com/kubeedge/community/blob/master/sig-ai/charter.md). 

KubeEdge-Sedna has released the first [edge-cloud collaborative lifelong learning](https://github.com/kubeedge/sedna/releases/tag/v0.3.0) on June 2021, together with [a hands-on example](https://sedna.readthedocs.io/en/latest/examples/lifelong_learning/atcii/README.html) and [a free playground on Katacoda](https://www.katacoda.com/kubeedge-sedna/scenarios/lifelong-learning-example). The sceme learn coming tasks from the edge based on previously learned tasks on the cloud knowledgebase. 

**Why is this project needed**:
[...]

PR #745, opened 2026-08-11
fix(simulation): teardown environment after benchmarking

vishnudathks:
**What type of PR is this?**

/kind bug

`BenchmarkingJob.run()` calls `build_simulation_enviroment()` when a `simulation:`
block is configured, but never calls the matching `destory_simulation_enviroment()`
— on success or failure. Every simulation-mode run leaks its kind cluster (and
cloud/edge worker containers).

This PR wraps the post-build steps of `run()` in `try/finally`, so teardown always
fires once a build has actually completed, on both success and exception paths,
without attempting to tear down a cluster that was never built. It also makes
`destory_simulation_enviroment()` log success/failure instead of silently
discarding the subprocess return code, and fixes a copy-pasted docstring.

Scoped narrowly on purpose:
- Doesn't touch the host-environment checks — already covered by #656.
- Doesn't touch config validation in `simulation.py` — already covered by #698.
- Doesn't touch the `shlex.quote()` injection fix — already covered by #620.
- Doesn't rename `destory_simulation_enviroment` — already planned separately
  as part of the Stage 1 restoration work described in #526.

Related to #348 — this fixes one specific lifecycle bug within that broader
simulation-restoration effort, not marking as `Fixes` since it doesn't close
the whole feature request on its own.

**Testing**:

- Added `tests/test_benchmarkingjob_simulation_teardown.py`: 4 mocked regression
  tests covering teardown after a clean run, teardown after testcases raise
  (asserting the original exception still propagates through `finally`), no
  teardown when no `simulation:` block is configured, and no teardown attempted
  when build itself fails.
- Confirmed the same 4 tests fail with `AttributeError` against unpatched `main`.
- `pylint` on both changed files → 10.00/10, matching baseline.
- Not verified against a real docker+kind cluster — no such environment was
  available while preparing this.


vishnudathks:
assign @MooreZheng 
assign @Poorunga

PRANABKUMARNEOGI:
Hi there! Thanks for putting this fix together. Wrapping the `run()` execution in a `try/finally` block to guarantee `destory_simulation_enviroment()` fires is a crucial architectural fix. 

I am reviewing this as part of a broader analysis on framework execution resiliency (specifically looking at cascading failures like those in Issue #423), and your fix directly addresses one of the most severe consequences of unhandled exceptions.

Before this PR, if a benchmark encountered a fatal downstream error (e.g., an Example-layer `FileNotFoundError` due to a bad config path), the orchestrator would crash and skip the teardown. This meant orphaned `kind` clusters and edge worker containers were left running. Consequently, the next benchmark in a queue would fail not because of its own code, but because of **Resource Starvation** (ports being in use, system RAM/VRAM being held hostage by the leaked containers). 

By enforcing the `finally` block, you've ensured that the simulation infrastructure is sandboxed and cleared even on exception paths. 

**Recommendation:**
This PR is excellently scoped and should be merged to stop the container leakage! 

As a follow-up for the Stage 1 restoration work you mentioned (or #348), we should ensure that destroying the `kind` cluster also successfully triggers a hardware-level release on the host machine (e.g., explicitly freeing the CUDA VRAM allocated to those edge containers) so that GPU resources are completely reset for the next benchmark.

Great work tracking down this lifecycle bug!

31groot:
### Request changes — partial simulation builds can still leak resources

The new `try/finally` teardown correctly covers failures that occur *after* `build_simulation_enviroment()` returns, but there is still a lifecycle gap when provisioning itself partially succeeds and then raises.

The current control flow is:

```python
simulation_built = False

if self.simulation is not None:
    build_simulation_enviroment(self.simulation)
    simulation_built = True
```

followed by:

```python
finally:
    if simulation_built:
        destory_simulation_enviroment(self.simulation)
```

Because `simulation_built` is assigned only after the complete build function returns, any exception raised inside the provisioning step leaves it `False` and skips teardown.

I reproduced this against the exact PR head `7e038e7` with an isolated harness that simulates a partially-created environment followed by an installer failure:

```text
SIMULATING: environment partially created, installer fails afterward
run raised: installer failed after cluster creation
build calls: 1
destroy calls: 0
```

So the PR now handles:

```text
build succeeds -> testcase failure -> teardown
```

but still leaves:

```text
partial build -> build raises -> teardown skipped
```

This matters because `build_simulation_enviroment()` performs environment provisioning through a shell installer. A non-zero installer exit can occur after external resources have already been created, so "build raised" does not necessarily mean "nothing exists to clean up."

The new regression test `test_teardown_skipped_when_build_itself_fails` currently codifies the opposite assumption: it requires destroy not to run whenever the build function raises. That is safe only if the builder has an all-or-nothing guarantee, which is not established by the current implementation.

I recommend arming cleanup before entering the provisioning phase (or otherwise returning an explicit provisioning state), and making teardown safe when the cluster was never created. The key invariant should be:

> once simulation provisioning may have started, cleanup is attempted on every exit path.

**Verification boundary:** reproduced the partial-build control-flow failure against the exact PR head with mocked provisioning; no real Docker/Kind/Sedna environment was provisioned.

PR #752, opened 2026-08-11
 [PHASE IV] fix(llm-agent): implement save()/load() persistence contract and harden related edge cases

### LINKED ISSUE #751: [PHASE IV][llm-agent]: Model Persistence Contract Broken (save()/load() No-Ops) and Related Latent Defects in llm_agent BaseModel
Title:

## What happened

Filed as part of ongoing work toward the CNCF-KubeEdge Ianvs "Comprehensive Example Restoration: Phase IV" effort (LFX Mentorship Fall 2026, Term 3). `examples/llm-agent/singletask_learning_bench` has been substantially restored since Issue #364 was filed (paths, dataset, and LoRA config are all now correct and working), but one problem #364 flagged is still present and unaddressed: `save()`/`load()` in `testalgorithms/basemodel.py` are non-functional stubs.

**1. `save()`/`load()` silently break the framework's model-persistence contract**

```python
def load(self, model_url, **kwargs):
    if model_url:
        logging.info("load model url: ",model_url)

def save(self, model_path = None):
    pass
```

The core paradigm controller (`core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning.py`) expects a real, usable contract here:

```python
trained_model_path = job.save(train_output_dir)   # always None, since save() is `pass`
...
job.load(trained_model)                            # called with None, does nothing
```

A single benchmarking run doesn't visibly break, because the trained model stays on `self.model` in memory and `predict()` reuses it directly. But `train_config.json` explicitly configures `output_dir` and `save_strategy: "epoch"`, showing clear intent to persist a usable checkpoint — yet the framework-facing method that's supposed to report that path back does nothing, and `load()` has no logic to reassign `self.model` even if given a real path. Any future workflow depending on reloading a previously-trained model (skip-retraining, comparing against an `initial_model_url`, a resumed run) would silently do nothing rather than fail loudly.

## What you expected

`save()` should actually persist the trained (LoRA-adapted) model to `model_path` and return a real, loadable path. `load()` should actually reload weights into `self.model` from a given path.

## How to reproduce

No dataset or GPU needed — reading `testalgorithms/basemodel.py` directly confirms both methods are no-ops independent of any run:

```bash
grep -n "def save\|def load" examples/llm-agent/singletask_learning_bench/testalgorithms/basemodel.py
```

## Anything else

A few related, smaller findings surfaced during this same investigation — verified but confirmed **not currently reachable/live** under normal usage, included here for completeness rather than as separate issues:

- `load()`'s own logging call is malformed: `logging.info("load model url: ", model_url)` has no `%s` placeholder, so it would raise `TypeError` if ever invoked with a truthy path. Currently unreachable only because `save()` never returns one — fixed as part of the same patch below.
- `__init__` will crash with `TypeError: expected str, bytes or os.PathLike object, not NoneType` if `BaseModel()` is ever instantiated without `config`/`train_config` kwargs. Not reachable through the normal `ianvs -f` pipeline (which always supplies real paths via `test_algorithm.yaml`), but worth a defensive guard.
- `preprocess()` truncates `input_ids`/`labels` from the end when a sample exceeds `MAX_LENGTH`. Reproduced with synthetic data: if the prompt portion alone exceeds `MAX_LENGTH`, the response gets truncated away entirely, leaving every label as `-100` (fully masked, zero training signal) with no error. Confirmed via direct inspection of the shipped dataset (`dataset/activity_classification.jsonl`, 10 samples, max 17 words) that this does **not** currently fire — flagged for awareness if the dataset ever grows.
- `config.json`'s `"half_model": true` key is never read anywhere in `basemodel.py` — dead configuration with no effect.

Issue #364 (Feb 2026) originally flagged `save()`/`load()` among several other problems; those other problems (path mismatches, missing dataset/config directories, missing LoRA `r` parameter, wrong pipeline type) have since been resolved in the current codebase. This issue narrows scope to the one part of #364 still accurate today, with fresh verification.

hayyanbaig:
## What type of PR

/kind bug

## What this PR does

Implements a real `save()`/`load()` contract for `LLM_agent`'s `BaseModel`, and hardens three related latent issues found during the same investigation, as part of ongoing Ianvs "Comprehensive Example Restoration: Phase IV" work (LFX Mentorship Fall 2026).

```python
# save() now actually persists the trained (LoRA-adapted) model and returns a real path
def save(self, model_path=None):
    if model_path is None:
        return None
    self.model.save_pretrained(model_path)
    self.tokenizer.save_pretrained(model_path)
    return model_path

# load() now actually reloads weights from a given path, and its log call is fixed
def load(self, model_url, **kwargs):
    if model_url:
        logging.info("load model url: %s", model_url)
        self.model = PeftModel.from_pretrained(self.model, model_url)
```

Also:
- `__init__` now raises a clear `ValueError` (naming the missing key) instead of an opaque `TypeError` if `config`/`train_config` kwargs are omitted, rather than crashing inside `open()`.
- `preprocess()` now guards against fully-masked label sequences (logs a warning and skips the sample rather than silently training on zero signal) if a future dataset ever produces a prompt longer than `MAX_LENGTH`.
- Removed the unused `"half_model"` key from `config/config.json` (never read anywhere in `basemodel.py`).

## Which issues it fixes

Fixes #751 

## Testing done

- Verified `save()` writes real adapter/tokenizer files to `model_path` and returns a non-`None` path, tested against a small local training run on the shipped `activity_classification.jsonl` dataset (10 samples).
- Verified `load()` given the path returned by `save()` correctly reloads the adapter into `self.model`, confirmed via `predict()` producing consistent output before/after a save→load round trip.
- Confirmed the `__init__` guard raises a clear, actionable error instead of the previous opaque `TypeError` when `config`/`train_config` are omitted.
- Confirmed the `preprocess()` guard does not change behavior on the current dataset (no sample currently exceeds `MAX_LENGTH`), verified via the same word-count check used during investigation (max 17 words per sample, well under the truncation threshold).
- Full end-to-end run: `ianvs -f examples/llm-agent/singletask_learning_bench/benchmarkingjob.yaml` completes and produces the same ROUGE scores as before this change, confirming no regression.


zjr060424-lab:
Thanks for taking this issue. I tested the exact same-instance path used by `SingleTaskLearning`: `train()` leaves `self.model` as a `PeftModel`, `save()` returns the adapter directory, and `_inference()` then calls `load()` on that same job instance.

With a tiny local GPT-2 model (no model download), PEFT 0.20.0, Transformers 5.6.2, and Torch 2.6.0, this line:

```python
PeftModel.from_pretrained(existing_peft_model, adapter_path)
```

created a nested `PeftModel` and emitted both a multiple-adapters warning and missing-adapter-key warnings. Reloading from a clean base model, or calling `unload()` first, avoided the nesting:

```python
base_model = self.model.unload() if isinstance(self.model, PeftModel) else self.model
self.model = PeftModel.from_pretrained(base_model, model_url)
```

The other point I noticed is that the new `preprocess()` branch does not actually skip the fully masked sample as the PR description says; it logs and still returns the sample. That warning is also suppressed by the existing module-level `logging.disable(logging.WARNING)`. So either the description should be narrowed, or the train loop needs explicit skip/error handling and logging needs to remain visible.

These are focused observations from the current diff; the core persistence fix is still the right direction.

mwemmz:
Reviewed the diff to `testalgorithms/basemodel.py` and `config.json`. The added `save()`/`load()` methods correctly implement the persistence contract the base class expects — this is the only PR I reviewed in this pretest that adds missing functionality rather than fixing a config/path/validation defect, so it's a useful point of contrast on that front.

One scoping question: `config.json` has 19 lines changed for what the commit describes as a persistence fix. Skimming the diff, several of those lines don't appear directly related to `save()`/`load()` (they look like unrelated formatting/value changes). Might be worth splitting the persistence-contract implementation from any incidental config cleanup into separate commits (or confirming in the PR description that the extra changes are intentional) so a reviewer can evaluate the core fix independently of the config tidy-up.

Shikhar-Kesharwani:
**Review depth: Analysis**

The `save()`/`load()` implementation correctly addresses the previously missing persistence behavior, but there is a remaining asymmetry in the persistence contract that is worth addressing before considering the example fully restorable.

`save()` now persists both the LoRA-adapted model and the tokenizer:

```python
self.model.save_pretrained(model_path)
self.tokenizer.save_pretrained(model_path)
```

However, `load()` only restores the model adapter:

```python
self.model = PeftModel.from_pretrained(self.model, model_url)
```

It does not reload the tokenizer from the same persisted directory.

This means the new persistence path is not fully symmetric: artifacts explicitly written by `save()` are not all restored by `load()`. A future workflow that changes or replaces the in-memory tokenizer between training and reload could therefore produce a model/tokenizer mismatch even though the checkpoint directory itself contains the tokenizer artifacts.

This is separate from the nested-`PeftModel` issue already identified in the existing review: that finding concerns how the adapter is attached to the base model, whereas this finding concerns completeness of the persisted model state.

I recommend either reloading the tokenizer from `model_url` inside `load()` or explicitly documenting that tokenizer state is intentionally managed outside the persistence contract. A regression test should save a model, reconstruct the inference state from the saved directory, and verify that both model and tokenizer are restored from that directory rather than relying on the pre-existing in-memory tokenizer.

**Evidence:** the PR's implementation explicitly saves tokenizer artifacts in `save()` but only calls `PeftModel.from_pretrained()` in `load()`. The existing review discussion focuses on nested adapter loading, preprocessing behavior, and configuration scope, rather than tokenizer restoration.

**Verification boundary:** this is a static code-path finding based on the submitted `save()`/`load()` implementation. I have not claimed an end-to-end tokenizer-mismatch reproduction.

PR #753, opened 2026-08-11
fix: create output directory before saving heatmap

### LINKED ISSUE #628: [Bug] draw_heatmap_picture() crashes with FileNotFoundError because output/ subdirectory is never created

xaemonn:
## What

`draw_heatmap_picture()` passed a path under `<output>/output/` directly to
`plt.savefig()`, which does not create parent directories. On a fresh run
directory this raised `FileNotFoundError` and aborted the benchmark before
any results were rendered.

## Changes

- Create the target directory with `os.makedirs(..., exist_ok=True)` before saving
- Build the path with `os.path.join` instead of an embedded `/`, so it behaves
  correctly on Windows as well as Linux
- `plt.close()` after saving, so repeated calls don't leak figures

## Testing

- Reproduced the crash against a temporary directory with no `output/`
  subdirectory and confirmed `FileNotFoundError`
- Confirmed the patched version writes `<output>/output/<title>-heatmap.png`

Fixes #628

### DIFF
--- a/core/storymanager/visualization/visualization.py
+++ b/core/storymanager/visualization/visualization.py
@@ -39,9 +39,12 @@ def draw_heatmap_picture(output, title, matrix):
     plt.ylabel('task', fontsize=15)
     plt.title(title, fontsize=15)
     plt.colorbar(format='%.2f')
-    output_dir = os.path.join(output, f"output/{title}-heatmap.png")
-    plt.savefig(output_dir)
+    output_dir = os.path.join(output, "output")
+    os.makedirs(output_dir, exist_ok=True)
+    output_file = os.path.join(output_dir, f"{title}-heatmap.png")
+    plt.savefig(output_file)
     plt.show()
+    plt.close()
 
 def get_visualization_func(mode):
     """ get visualization func """


31groot:
### Request changes : `title` can produce invalid output paths

**Review depth: Analysis**

The PR correctly creates the `output/` directory, but `title` is still used directly when constructing the output filename:

```python
output_dir = os.path.join(output, "output")
os.makedirs(output_dir, exist_ok=True)
output_file = os.path.join(output_dir, f"{title}-heatmap.png")
```

This creates two path-handling issues:

1. A title such as `task/round` produces `<output>/output/task/round-heatmap.png`, but the intermediate `task/` directory is not created, resulting in `FileNotFoundError`.
2. An absolute title such as `/tmp/escaped` causes `os.path.join()` to discard the intended output directory, producing `/tmp/escaped-heatmap.png`.

I reproduced both behaviors locally with the submitted path construction.

**Recommendation:** ensure `title` is converted to a filesystem-safe filename before joining it with the output directory. Please also add regression tests for normal, path-containing, and absolute titles, verifying that the resulting file always remains under `output/output`.

**Verification boundary:** Reproduced both path-handling behaviors locally; the full visualization pipeline was not executed.

haneul-24:
This PR correctly fixes the problem by creating the `output/` directory before saving the heatmap.

Using `os.makedirs(output_dir, exist_ok=True)` also works when the directory already exists.

The added `plt.close()` is useful because it closes the figure after it is shown.

**Recommendation: Accepted as it is.**

PR #755, opened 2026-08-12
fix(core): guard NoneType matrix in rank visualization and zero division in lifelong learning

Raj-vardhan01:
### What type of PR is this?
/kind bug

This PR resolves two runtime failure modes in the `ianvs` core engine:
1. **Rank Visualization Crash**: When executing benchmarking jobs that do not yield a matrix output (such as single-task learning paradigms), configuring `save_mode: "selected_and_all_and_picture"` caused `Rank._draw_pictures()` to crash with `AttributeError: 'NoneType' object has no attribute 'keys'` because `test_result.get("Matrix")` returned `None`.
2. **Lifelong Learning ZeroDivisionError**: When `tasks_detail` is empty during evaluation rounds in `lifelong_learning.py`, `i` remained `0`, causing lines 129, 151, 227, and 250 to raise `ZeroDivisionError: float division by zero`.

Fixes # (N/A)

- **`core/storymanager/rank/rank.py`**: Added check `if not matrix or not isinstance(matrix, dict): continue` before calling `matrix.keys()`.
- **`core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py`**: Added zero-division guards (`task_avg_score['accuracy']/i if i > 0 else 0.0`) across all 4 accuracy calculation loops.
- **`core/storymanager/rank/test_rank.py`**: Added regression unit test verifying `_draw_pictures` handles missing/None matrix objects gracefully.

### Does this PR introduce a user-facing change?:
```release-note
NONE

### CODE REVIEW

**Comment by AdityaKumarSethia** [comment]
Unrelated to this PR's stated scope (the `rank.py` `NoneType` guard, which looks correct), flagging one thing about the other half of this PR's diff.

This PR also patches the same 4 `lifelong_learning.py` division sites (`task_avg_score['accuracy']/i` at lines 129, 151, 227, 250) as two other currently open PRs, #588 and #496. Confirmed via real `git merge`:

```bash
git checkout -b test-755-vs-496
git merge pr-496 --no-edit   # clean
git merge pr-755 --no-edit
# Auto-merging lifelong_learning.py
# CONFLICT (content): Merge conflict in core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py
```
Three independent PRs on the identical lines. Recommend maintainers pick one (my read: #496 is the broadest, since it also fixes a related crash in `metrics.py` that neither this PR nor #588 touches) and close the other two as duplicates rather than resolving the conflict three separate times as each lands.

Full cross-PR analysis in [Discussion #873](https://github.com/kubeedge/ianvs/discussions/873).

Muhammad-Hashir-Code:
##  Review PR #755 (fix(core): guard NoneType matrix in rank visualization and zero division in lifelong learning)

**Depth level: Analysis** (verified via direct execution against both the pre- and post-patch logic).

**Background:** this PR was already reviewed in Discussion #873 (AdityaKumarSethia), but only for its `lifelong_learning.py` component specifically, proving it conflicts with #496/#588 on the same division-by-zero lines. That review explicitly describes the `rank.py` change as "an unrelated NoneType guard" and doesn't examine it further. This review covers exactly that unexamined piece, connecting it to a different root problem (this candidate's own Task 1 analysis, `rank.py`'s handling of malformed/missing metric data) rather than the division-by-zero cluster.

**What the `rank.py` change does, verified via real diff:**
```diff
             matrix = test_result.get("Matrix")
+            if not matrix or not isinstance(matrix, dict):
+                continue
             for key in matrix.keys():
                 draw_heatmap_picture(out_put, key, matrix[key])
```

**Finding — this fixes a real, separate crash, verified via direct execution of both versions:**
```python
test_result = {'f1_score': 0.87}  # a normal scalar-metric result, no Matrix key
# Before this PR:
matrix.keys()  # AttributeError: 'NoneType' object has no attribute 'keys'
# After this PR:
# guard triggers, function returns cleanly, no crash
```
This occurs in `Rank._draw_pictures()`, called when `save_mode` is configured as `"selected_and_all_and_picture"`. Any benchmarking job using that save mode against an algorithm that reports standard scalar metrics (accuracy, F1, etc.) rather than a confusion-matrix-style result crashes at the final reporting stage after evaluation has already completed successfully, meaning a user loses access to already-computed results due to a crash in an unrelated visualization step.

**A completeness gap, not raised by the existing conflict-focused review:** the fix `continue`s silently no log line, no warning, nothing indicating a picture was skipped because the expected `Matrix` data wasn't present. For a paradigm the user configured expecting a heatmap output, this could look like the picture-generation step simply didn't run, with no indication of why. A one-line `LOGGER.warning(...)` before the `continue` would close this gap with negligible cost, consistent with the pattern favored elsewhere in this codebase's more careful validation code (e.g. `benchmarkingjob.py`'s `_check_fields()`, which raises descriptive `ValueError`s rather than failing silently).

**Why this matters for the PR as a whole:** the `rank.py` fix and the `lifelong_learning.py` fix, bundled in one PR, address two structurally unrelated bugs (a missing-data crash in visualization vs. an unguarded division in accuracy computation) worth flagging for a maintainer considering whether to split the PR, independent of the division-conflict issue #873 already raised, since the `rank.py` half has no merge-conflict risk and could land immediately regardless of how the `lifelong_learning.py` half's three-way duplication gets resolved.

**Uniqueness:** the `rank.py` component's correctness has not been independently verified anywhere in this PR's thread or in #873's review, which explicitly scoped itself away from this file.

**Verification Boundary:** both the crash reproduction and the fix confirmation are directly executed against the real pre- and post-patch logic extracted from the actual branches, not simulated or inferred from the diff alone.

phantom-712:
The rank guard is a valid robustness improvement: calling `.keys()` on a
missing or `None` Matrix should not destroy scalar-metric results during the
final reporting phase. The current `continue` behavior should, however, emit a
clear warning when a configured picture cannot be produced. Otherwise a user
can mistake "no heatmap was generated" for successful visualization.

The task-average half needs a separate semantic decision. When `i == 0`,
returning `0.0` prevents the arithmetic exception but publishes a value that
is indistinguishable from a genuine zero accuracy. "No task was scored" and
"scored tasks achieved zero" are different states. The implementation should
either propagate an explicit not-measured value supported by the leaderboard
or document why zero is the required sentinel.

The regression matrix should include:

- no Matrix key;
- `Matrix=None`;
- a non-dictionary Matrix value;
- a valid Matrix that still produces a picture; and
- empty and non-empty `tasks_detail` through each of the four accuracy loops.

Those tests should be attached to the corresponding Core paths rather than
using the rank test as evidence for the unrelated lifelong-learning division
change. Recommendation: minor revision. Keep the two fixes if the project
prefers one PR, or split them for review clarity; add the warning/metric policy
and direct empty-task coverage. The local evidence for the working-tree
equivalent is fixture-level only and does not retroactively prove the PR branch
ran end-to-end. The reconstructed composition reached final metric
handoff and rank-table generation after three training states and two updates;
that confirms the surrounding reporting path can complete, but it does not
replace direct tests for this PR's Matrix and empty-task branches.

The reporting path was also checked after a five-round official-resource Robot
run in the repaired composed tree. Ianvs returned code 0 and wrote the scalar
rank row and the five-column all-task matrix. The final accuracy was
`0.002232849365268865`, task-average accuracy was `0.002`, and both BWT and
FWT were `0.0`. This confirms that the surrounding rank output can carry a
real-resource result, but it does not prove that PR #755 passed unchanged or
that the single-candidate score is a comparative quality result.

Issue #707, opened 2026-08-07
fix: cifar100 testenv.yaml uses deprecated train_url/test_url keys


rajdhruvsingh:
All `testenv.yaml` files under `examples/cifar100/` use `train_url`
and `test_url` as dataset keys, but the current ianvs `Dataset` API
expects `train_index` and `test_index`.

This causes a silent misconfiguration where the dataset is never
loaded correctly.

**Affected files:**
- `fci_ssl/fedavg/testenv/testenv.yaml`
- `fci_ssl/glfc/testenv/testenv.yaml`
- `fci_ssl/fed_ci_match/testenv/testenv.yaml`
- `fci_ssl/fed_ci_match_v2/testenv/testenv.yaml`
- `federated_class_incremental_learning/fedavg/testenv/testenv.yaml`
- `federated_learning/fedavg/testenv/testenv.yaml`

**Fix:** Rename `train_url` → `train_index` and `test_url` → `test_index`.

subham-hq:
Reproduced on Ubuntu 24.04 / Python 3.12.1, ianvs 0.3.0 installed from source.

Running:

```bash
ianvs -f examples/cifar100/fci_ssl/glfc/benchmarkingjob.yaml
```

fails at `core/testenvmanager/dataset/dataset.py:165`:

```
Traceback (most recent call last):
  File "/workspaces/ianvs/core/testenvmanager/testenv/testenv.py", line 78, in prepare
    self.dataset.process_dataset()
  File "/workspaces/ianvs/core/testenvmanager/dataset/dataset.py", line 165, in process_dataset
    raise NotImplementedError('not one of train_index/train_data/train_data_info')
NotImplementedError: not one of train_index/train_data/train_data_info
```

Worth noting the drift isn't confined to `examples/cifar100/`:

```bash
grep -rn "train_index\|train_url" examples/ --include="*.yaml"
```

returns roughly 22 configs still using `train_url` against 5 already migrated to `train_index` — including MOT17, bdd, robot, imagenet, cityscapes, cityscapes-synthia and yaoba.

Happy to help with the wider sweep if that's useful.

PR #758, opened 2026-08-12
fix(core): warn on deprecated and unknown dataset config fields

subham-hq:
/kind bug

`Dataset._parse_config` silently discards any config key that is not already an attribute on the instance:

```python
def _parse_config(self, config):
        if attr in self.__dict__:
            self.__dict__[attr] = value
    self._check_fields()
```

A typo (`trian_index`) or a deprecated key is dropped without any indication. The failure then surfaces much later, in `process_dataset`, as:

```
NotImplementedError: not one of train_index/train_data/train_data_info
```

— an error that names three keys the user never wrote, and says nothing about the key they did write. This is the mechanism behind the "silent misconfiguration where the dataset is never loaded correctly" described in #707.

`train_url` / `test_url` are a specific case worth calling out. They are declared as attributes and are therefore accepted from config:

```python
self.train_url: str = ""
self.test_url: str = ""
```

but they are only ever assigned as *outputs* inside `process_dataset`, never read as inputs:

```
$ grep -n "self.train_url\|self.test_url" core/testenvmanager/dataset/dataset.py
45:        self.train_url: str = ""
46:        self.test_url: str = ""
158:            self.train_url = self._process_index_file(self.train_index)
160:            self.train_url = self._process_data_file(self.train_data)
162:            self.train_url = self._process_data_info_file(self.train_data_info)
168:            self.test_url = self._process_index_file(self.test_index)
170:            self.test_url = self._process_data_file(self.test_data)
172:            self.test_url = self._process_data_info_file(self.test_data_info)
```

So a config that sets `train_url:` is accepted, stored, and then ignored — which is why the resulting error is so hard to trace back to the config.

This PR logs a warning in both cases: for the two deprecated fields, and for any key the loader does not recognise.

**Behaviour is otherwise unchanged.** Nothing that works today stops working — this is warning-only and backward compatible. Whether unrecognised keys should eventually raise rather than warn seems like a maintainer decision; happy to follow up if that's preferred.

**Before / after**

Running `examples/cifar100/fci_ssl/glfc/benchmarkingjob.yaml` (which uses the deprecated keys):

Before — straight to the `NotImplementedError`, with no indication which key was at fault.

After:

```
[WARNING] - dataset field `train_url` is not read by the dataset loader; use train_index, train_data or train_data_info instead.
[WARNING] - dataset field `test_url` is not read by the dataset loader; use test_index, test_data or test_data_info instead.
```

Verified that an unaffected example (`examples/cloud-edge-speculative-decoding-benchmark`) produces no new warnings and fails exactly as before.

**Note on an existing config field this surfaces:**

The unknown-field warning also fires on `name:`, which appears under `dataset:` in 8 example configs — cifar100 (×5), PIPL, and federated-llm — e.g.:

```yaml
dataset:
  name: 'cifar100'
```

`Dataset` has no `name` attribute, and `TestEnv` only defines `name` inside `model_eval.model_metric`, so this key is currently dropped in every one of those configs. I've left the warning in rather than special-casing the field, since suppressing it would hide exactly the kind of silent drop this PR is meant to surface. Happy to add `name` as a supported field, or update the example configs, if maintainers would prefer either.

**Verification:**

```
$ pylint core/testenvmanager/dataset/dataset.py
```

**Environment:**

- OS: Ubuntu 24.04.2 LTS (GitHub Codespaces)
- Python: 3.12.1
- ianvs: 0.3.0, installed from source via `pip install -e .`

Related: #589, #707, #318


akshita317:
The `else:` branch here is a real improvement — warning on unrecognised dataset keys would surface a class of silent misconfiguration that is currently invisible. 14 of the 48 example `testenv*.yaml` files carry keys that `_parse_config` discards without a word, so this would have practical value immediately.

One finding on the `_DEPRECATED_FIELDS` half.

**`train_url` and `test_url` are not deprecated inputs; they are outputs Core writes.**

`core/testenvmanager/dataset/dataset.py:157-175`:

```python
if self.train_index:
    self.train_url = self._process_index_file(self.train_index)     # L158
```

`process_dataset()` computes and assigns `self.train_url` / `self.test_url`, and paradigms read them afterwards — `singletask_learning.py:122`, `lifelong_learning.py:141,155,240,254,296,430`, `incremental_learning.py:108,227`, `multiedge_inference.py:83,85,93`, `federated_learning.py:120,144,145`, `federated_class_incremental_learning.py:176,189`, `joint_inference.py:78`.

The warning text — "is not read by the dataset loader" — is therefore accurate about the *loader* but misleading overall: the attribute is populated by Core a moment later, so a user who follows the advice and then inspects the object will find it set.

**Suggestion.** Modelling these as a third category rather than as deprecated inputs would be more precise and would also address #604 at its cause:

```python
_OUTPUT_FIELDS = {"train_url", "test_url"}    # written by process_dataset()
```

with a warning that names the correct input key (`train_index` / `train_data` / `train_data_info`) when a config supplies one. Twenty of the 48 example configs supply `train_url`/`test_url` and *none* of them supplies any of the three input keys, so they all currently stop at L165 with `NotImplementedError` — a warning naming the right key would turn that into an actionable message.

**One compatibility note:** PR #789 also modifies this method and adds `train_url`/`test_url` to a `_PATH_FIELDS` tuple for `os.path.expanduser`. The two branches merge without conflict, but the combined result places the `continue` above the expansion, so those two entries in `_PATH_FIELDS` become unreachable:

```python
if attr in self._DEPRECATED_FIELDS:
    LOGGER.warning(...)
    continue                                   # <-- exits first
if attr in self.__dict__:
    if attr in self._PATH_FIELDS and isinstance(value, str):
        value = os.path.expanduser(value)      # train_url/test_url never reach here
```

Worth coordinating with @purunjay-bhardwaj on a single model of these fields before either lands.

**Minor:** with 14 configs carrying unknown keys (PIPL alone declares 12 under `dataset:`), the `else:` warning will be fairly noisy on first merge. An allowlist, or downgrading to debug for keys known to be intentional, might help adoption.

Nikhil-VS1811:
Cross-referencing from #742: this issue's proposed warning-on-unrecognized-keys would need to distinguish "dead config" (Class 2 — e.g. calib_dir, poses_dir in industrialEI, zero consumers found) from "intentional metadata" (Class 3 — e.g. name, description in cifar100/PIPL, also zero consumers but clearly documentation-style) to avoid false-positive warning noise on ~10 configs. Full analysis: https://github.com/kubeedge/ianvs/issues/230#issuecomment-5387796858

akshita317:
@Nikhil-VS1811 agreed on the Class 2 / Class 3 distinction — that is the part that decides whether the warning is useful or just noise.

Small heads-up: the link in your comment is missing the repo path, so it 404s. It renders as `https://github.com//issues/230#...` — this is the working one:

https://github.com/kubeedge/ianvs/issues/230#issuecomment-5387796858

Flagging because the pre-test notes say links that do not open score zero, so worth fixing wherever else you have pasted it.

suhaan-24:
Static review. This is the smallest change in the area and, in my reading, the one placed at the
right layer.

## Why the loader is the right place

Configs carrying these two key names are not confined to one Example. Counting key-form matches at
`upstream/main`:

```
$ git grep -lE '^[[:space:]]*(train|test)_url[[:space:]]*:' upstream/main -- 'examples/**/*.yaml' | wc -l
26
```

Those 27 files span `bdd`, `cifar100` (6), `cityscapes`, `cityscapes-synthia` (6),
`Cloud_Robotics`, `cloud_VLA_finetune`, `imagenet`, `MOT17` (2), `robot` (5),
`robot-cityscapes-synthia`, and `yaoba` (2) — multiple paradigms, not just lifelong learning.

The alternative being pursued in other open PRs is to rename the keys per Example. That repairs
today's state but leaves nothing preventing the 28th Example from being written against the old
names, because the loader currently ignores them silently. Putting the knowledge in
`_parse_config` means it is stated once.

## The failure this makes legible

Without this change, a config using `train_url` produces:

```
NotImplementedError: not one of train_index/train_data/train_data_info
  core/testenvmanager/dataset/dataset.py in process_dataset
RuntimeError: prepare dataset failed, error: not one of train_index/train_data/train_data_info.
```

which names the three keys that *are* accepted but not the one that was supplied. A warning that
names the offending key closes that gap.

On the `_DEPRECATED_FIELDS` naming — worth noting `process_dataset()` *assigns* `self.train_url` /
`self.test_url` and paradigms read them afterwards, so these are attributes Core writes rather than
inputs it once accepted. Modelling them as outputs, and warning that a supplied value will be
overwritten, would describe the actual contract more accurately than "deprecated".

## Two suggestions

1. **Warn-and-continue still fails, just later.** A run with `train_url` will emit the warning and
   then hit the same `NotImplementedError` a few lines down. If the intent is a migration path,
   mapping `train_url` → `train_index` with the warning would make old configs work; if the intent
   is strictness, raising immediately with the deprecation message would be clearer than warning
   and then failing with a less specific error. The current middle position gives a better message
   but the same outcome.

2. **The unknown-field warning may be noisy.** The `else` branch warns on *any* key not in
   `self.__dict__`. Configs that carry intentional extra fields will now log a warning per field
   per run. Consider restricting the warning to the deprecated set, or gating the unknown-field
   case behind a stricter validation mode.

## Overlap to be aware of

`core/testenvmanager/dataset/dataset.py` is currently touched by four other open PRs. This one is
additive and small enough to rebase easily, but the strategy question — whether the loader should
warn, migrate, or reject — is worth settling once rather than per PR.

## Verification boundary

**Not executed.** The `NotImplementedError` trace above is from running the current `main` loader
against an Example configured with `train_url`, not from running this branch. The 27-file count is
a `git grep` over `HEAD`.

suhaan-24:
Following the discussion above on how these fields should be modelled — I measured the corpus this
change would apply to, and I think the numbers argue for going one step further than warning.

## The corpus, measured at `upstream/main`

```
$ git grep -l '' upstream/main -- 'examples/**/testenv*.yaml' | wc -l
48
$ git grep -lE '^[[:space:]]*(train|test)_url[[:space:]]*:' upstream/main -- 'examples/**/testenv*.yaml' | wc -l
26
$ git grep -lE '^[[:space:]]*(train|test)_(index|data|data_info)[[:space:]]*:' upstream/main -- 'examples/**/testenv*.yaml' | wc -l
22
```

Key-form matches only, so commented-out lines and path strings containing `train_data/` are
excluded. The 26 span eleven Example families: `bdd`, `cifar100`, `cityscapes`,
`cityscapes-synthia`, `Cloud_Robotics`, `cloud_VLA_finetune`, `imagenet`, `MOT17`, `robot`,
`robot-cityscapes-synthia`, `yaoba`.

**The number I did not expect: zero configs declare both forms.**

```
$ comm -12 <(git grep -lE '...(train|test)_url...' | sort) \
           <(git grep -lE '...(train|test)_(index|data|data_info)...' | sort) | wc -l
0
```

That is a clean generational split — each config was written against whichever contract was
current, and none was ever migrated. It also means a mapping shim has no ambiguity to resolve.

## Suggestion: migrate rather than warn

As written, a config supplying `train_url` gets a clearer message and then still stops at
`NotImplementedError`. All 26 files remain broken until each is edited by hand. Since no config
declares both keys, the loader can simply accept the legacy name:

```python
_LEGACY_INPUT_MAP = {"train_url": "train_index", "test_url": "test_index"}

    if attr in self._LEGACY_INPUT_MAP:
        target = self._LEGACY_INPUT_MAP[attr]
        if not config.get(target):
            LOGGER.warning(
                "dataset field `%s` is deprecated as an input; treating it as `%s`. "
                "Update the config — this shim will be removed.", attr, target)
            self.__dict__[target] = value
        continue
    ...
```

Behaviour change: the 26 configs proceed instead of aborting, and log a deprecation warning naming
the correct key. The 22 configs already using `train_index` never reach the branch, so they are
unaffected — which matters, because those 22 are the real regression surface for any change to this
method.

This also composes with the output-field point raised above rather than competing with it: the
legacy *input* name gets mapped on the way in, and `train_url` remains what
`process_dataset()` assigns on the way out. Those are two different roles for the same string, and
separating them is what makes both behaviours describable.

## The recurrence half

Whichever modelling is chosen, the loader currently discards unrecognised keys silently, so nothing
stops a new Example being written against the old contract. A shim fixes the existing corpus; it
does not prevent the 27th. A CI check that fails on any `examples/**/testenv*.yaml` introducing
`train_url:`/`test_url:` would cover that half, and it is small enough to land alongside this
(related: #743).

## Verification boundary

The counts above are `git grep` over `upstream/main` and are reproducible. I have **not** executed
this branch. The behaviour of the proposed shim is a design sketch, not a tested patch — the only
part I have run is the underlying failure it addresses: on
`cityscapes-synthia/lifelong_learning_bench/curb-detection`, renaming the two keys is what moves
execution past `test_env.prepare()`, after which the benchmark completes and reports accuracy
within 0.00062 of the value published in that Example's README.

azaynul10:
Three things from a static pass, in order of how much they affect the decisions already
being discussed above.

## 1. The corpus counts above are one commit stale

@suhaan-24's migration shim is sized against 26 legacy and 22 current configs. Both
numbers were accurate until 2026-08-16 and are not accurate at `37a9c60`:

```
                        42fd607^     HEAD (37a9c60)
old  train_url/test_url       26                 20
new  *_index/_data/_data_info 22                 28
```

The boundary is `42fd607` "fix+docs: cifar100 federated learning", authored 2026-08-08,
landed on main 2026-08-16 via #711 (merge `2bfa710`), which migrated six cifar100
configs from the legacy keys to `train_index`/`test_index`. That is the entire
difference: 26 - 6 = 20, 22 + 6 = 28. A `git fetch upstream` should reproduce this.

Reproduce, both scopes, from a clean checkout:

```bash
git grep -lE '^[[:space:]]*(train_url|test_url)[[:space:]]*:' HEAD -- examples \
  | grep -E 'testenv.*\.yaml$' | wc -l          # 20
git grep -lE '^[[:space:]]*(train|test)_(index|data_info|data)[[:space:]]*:' HEAD -- examples \
  | grep -E 'testenv.*\.yaml$' | wc -l          # 28
```

**The design conclusion survives, and this is worth saying plainly.** 20 + 28 = 48,
which is still every `testenv*.yaml` in the tree, so the generational split is still
clean and no config declares both forms. The shim still has no ambiguity to resolve.
What changes is the sizing: the backlog a shim would clear is 20 files rather than 26,
and the regression surface for any change to this method is 28 files rather than 22,
which is 27% larger than stated. Since the case for the shim rests on the ratio between
those two numbers, both moved in the direction that weakens it.

## 2. The `continue` is inert, which resolves the #789 coordination question

@akshita317 is right that merging this with #789 puts the `continue` above the
`expanduser` call and makes the `train_url`/`test_url` entries in `_PATH_FIELDS`
unreachable. Those two entries are already dead at HEAD, before either PR, so the
conflict is real but has no behavioural consequence and neither PR needs to wait for the
other.

A config-supplied `train_url` cannot be observed by anything, on any path:

1. `_parse_config` (`dataset.py:70-75`) is the only place it can land.
2. `_check_fields` (`dataset.py:56-68`) validates the six `*_index`/`*_data`/`*_data_info`
   attributes and never reads `train_url` or `test_url`.
3. `process_dataset()` has exactly one call site, `testenv.py:78`, unconditional inside
   `prepare()`.
4. `process_dataset()` assigns both attributes on every path that returns
   (`dataset.py:158,160,162` and `:168,170,172`) or raises (`:165,175`), and a raise
   aborts the run as `RuntimeError` at `testenv.py:79-80`.
5. Every reader of `dataset.train_url`/`test_url` is in a paradigm or in
   `testcase.py:106`, all of which run after `prepare()`.

So the value assigned at step 1 is overwritten at step 4 before any reader at step 5
sees it. Dropping the assignment, as the `continue` does, changes nothing, and adding
`expanduser` to it, as #789 does, also changes nothing.

This is why I would not describe the `continue` as load-bearing. The warning is the
whole value of that branch.

## 3. This is one of six identical sites

The silent discard is not specific to `Dataset`. Of seven `_parse_config`
implementations under `core/`, six drop unrecognised keys with no diagnostic:

```
core/cmd/obj/benchmarkingjob.py:110            if k in self.__dict__:
core/storymanager/rank/rank.py:56              if attribute in self.__dict__:
core/testcasecontroller/algorithm/algorithm.py:165   if k in self.__dict__:
core/testcasecontroller/algorithm/module/module.py:153  if k in self.__dict__:
core/testcasecontroller/simulation/simulation.py:50  if attribute in self.__dict__:
core/testenvmanager/dataset/dataset.py:72      if attr in self.__dict__:   <- this PR
```

The seventh, `TestEnv._parse_config` (`testenv.py:70-71`), does the inverse and accepts
anything it does not recognise:

```python
else:
    self.__dict__[k] = v
```

That asymmetry sharpens the `name:` observation in the PR body. Under `dataset:` the key
is dropped, and under `testenv:` the same key is accepted onto the instance, so its fate
depends on which block it appears in and neither outcome is reported.

The practical point for this PR: @suhaan-24 suggests settling the warn-migrate-reject
question once rather than per PR, across the five PRs touching `dataset.py`. The census
suggests the same argument applies per *class*, and that whatever is settled here has
five more sites to be applied to. `Rank._parse_config:56` is the one I have looked at
closely, where the discarded key is `save_mode` and the effect is a leaderboard silently
built from defaults rather than from the config, so the consequences of the idiom are
not limited to datasets.

## On the `_DEPRECATED_FIELDS` naming

Both readings offered above are correct, for different periods. The field changed role,
and `git log -S` on `core/testenvmanager/dataset/dataset.py` dates it, which may be
useful for choosing the wording.

**`train_url` was a genuine input for two years.**

`d439ff8` (2022-06-20) introduced it, validated it, and read it:

```python
self._check_dataset_url(self.train_url)                      # user's value validated
self.train_url = self._process_index_file(self.train_url)    # user's value read
```

`179d39d` (2024-07-06) introduced `train_index` and moved the read:

```python
self.train_url = self._process_index_file(self.train_index)   # source is now train_index
```

So `_DEPRECATED_FIELDS` describes 2022-06-20 to 2024-07-06 accurately, and "output
written by `process_dataset`" describes 2024-07-06 to today accurately. A comment naming
the date would keep both facts available to the next reader.

Reproduce:

```bash
git log -S "self._process_index_file(self.train_url)"  --oneline -- core/testenvmanager/dataset/dataset.py
git log -S "self._process_index_file(self.train_index)" --oneline -- core/testenvmanager/dataset/dataset.py
git show d439ff8:core/testenvmanager/dataset/dataset.py | grep -n "train_url"
```

**The contract then flip-flopped, which is why the examples are inconsistent.**

Classifying every historical version of `dataset.py` by which attribute
`process_dataset` reads:

| commit | date | `process_dataset` reads |
|---|---|---|
| `d439ff8` | 2022-06-20 | `train_url` |
| `179d39d` | **2024-07-06** | `train_index` |
| `7bf6b1b` | 2024-07-17 | `train_index` |
| `b118931` | **2024-07-30** | `train_url` |
| `fb95a9c` | 2024-08-03 | `train_index` |
| `37a9c60` | HEAD | `train_index` |

`b118931`'s first parent is already on the old form, so this was not a deliberate revert:
that branch forked before 2024-07-06 and was merged without rebasing, rewriting
`dataset.py` (+294/-145) from a stale base. The same commit added
`examples/cifar100/fci_ssl/{fedavg,fed_ci_match,fed_ci_match_v2,glfc}` and two
`federated_*` examples, whose `testenv.yaml` files use `train_url`, consistent with the
core it shipped alongside. `fb95a9c` restored `train_index` four days later, but only in
Core. The six configs were left on the superseded key and stayed that way until `42fd607`
(landed 2026-08-16 via #711) migrated them, just over two years after the contract they
were written against had been replaced. That two-year gap between a Core contract change
and the config migration that completes it is the interval this PR's warning is meant to
shorten.

**Why the `else:` branch earns its place.**

Classifying every `examples/**/*.yaml` that declares a dataset key, by style and by the
date the file entered the tree. Note the scope is all YAML rather than `testenv*.yaml`,
which is why the new-style total is 29 here and 28 in §1:

| style | added before 2024-07-06 | added on/after |
|---|---|---|
| `train_url`/`test_url` | 18 | **2** |
| `train_index`/`train_data`/`train_data_info` (and `test_*`) | 0 | 29 |

Two things follow. No file predating 2024-07-06 uses an input key, which independently
confirms that date as the schema boundary. And two files adopted `train_url` *after* it
stopped being read: `examples/imagenet/multiedge_inference_bench` (2024-09-26) and
`examples/Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation` (2025-01-14,
roughly six months after). No file uses both styles.

The inflow is small but not zero, and the most recent instance is six months past the
change, which is the part relevant to this PR: nothing in the loader signals that a
supplied key is not consulted, so the only feedback a contributor gets is a
`NotImplementedError` naming three keys they did not write. Migrating the remaining 20
files clears the backlog; the `else:` branch is what makes the next one self-diagnosing.

Reproduce (PowerShell, from the repo root):

```powershell
Get-ChildItem examples -Recurse -Include *.yaml,*.yml -File | ForEach-Object {
  $t = Get-Content $_.FullName -Raw
  $s = if ($t -match '(?m)^\s*(train_url|test_url)\s*:') { 'old' }
       elseif ($t -match '(?m)^\s*(train|test)_(index|data|data_info)\s*:') { 'new' }
  if ($s) { "$s $(git log --diff-filter=A --format=%ad --date=short -1 -- $_.FullName)" }
} | Group-Object { ($_ -split ' ')[0] } | Select-Object Name, Count
```

**Suggestion, narrow.** Keep the two-category split this PR introduces, but name the
category for what it is at HEAD and let the message carry the history:

```python
_OUTPUT_ONLY_FIELDS = {
    "train_url": "train_index, train_data or train_data_info",
    "test_url":  "test_index, test_data or test_data_info",
}
# Accepted as inputs until 179d39d (2024-07-06); process_dataset has computed
# them from the *_index / *_data / *_data_info keys ever since.
```

The `continue` can stay or go without changing behaviour, per §2, so it is worth keeping
only as a statement of intent that the attribute is not an input. If it stays, a short
comment saying that `process_dataset` overwrites it anyway would stop the next reader
from concluding, as I did on a first pass, that it guards against a stale value reaching
a consumer.

The corpus counts at both revisions, the `42fd607` boundary,
the `_parse_config` census, the history and the style/date table are executed against
`37a9c60` and reproducible with the commands above. §2 is read from the call graph, five
steps each grounded in a cited line, and is not an executed experiment: I have not
constructed a config that supplies both `train_url` and `train_index` to observe the
overwrite directly, and that is the check I would want before anyone acts on it. I have
not run this PR's branch, so the before/after warning output in the PR body is not
something I can independently confirm, and the affected examples need datasets I do not
have locally.

Nikhil-VS1811:
"Hi, I reviewed this PR for the LFX pre-test. While the warnings are great, I noticed a logical merge conflict with PR #789 regarding the continue statement skipping path expansion. Full details in my review:  https://github.com/kubeedge/ianvs/discussions/937

subham-hq:
Thanks @akshita317, @suhaan-24, @azaynul10 and @Nikhil-VS1811 — this is far more scrutiny than a `size/S` PR usually gets, and the analysis has been genuinely useful. Responding to what I think is settled, and flagging two questions I don't believe are mine to answer.

## The overwrite check

@azaynul10 — §2 of your analysis ends by noting that you had not constructed a config supplying both `train_url` and `train_index` to observe the overwrite directly, and that this was the check you would want before anyone acted on it. I ran it against `upstream/main`:

```
$ python -c "
import tempfile, pathlib
d = pathlib.Path(tempfile.mkdtemp())
(d/'idx.txt').write_text('a.jpg 0\n')
ds = Dataset({'train_url': '/sentinel/should-be-overwritten.txt',
              'train_index': str(d/'idx.txt'),
              'test_index': str(d/'idx.txt')})
print('after parse:  ', ds.train_url)
ds.process_dataset()
print('after process:', ds.train_url)
"
after parse:   /sentinel/should-be-overwritten.txt
after process: /tmp/tmpgvytsi9w/index.txt
```

The config-supplied value survives `_parse_config` and is replaced by `process_dataset()`. Your call-graph reasoning holds empirically.

Two consequences follow. The `continue` in this PR is inert — as you said, the warning is the whole value of that branch. And the `train_url`/`test_url` entries in #789's `_PATH_FIELDS` were already dead at HEAD, so the merge order between the two PRs has no behavioural effect. @Nikhil-VS1811 and @akshita317 — that should close out the coordination question you both raised; @purunjay-bhardwaj, flagging it here so #789 isn't blocked on this.

## Changed: modelled as output-only

@akshita317 raised this first, and @suhaan-24 and @azaynul10 both independently arrived at the same place, so I've made the change. `_DEPRECATED_FIELDS` is now `_OUTPUT_ONLY_FIELDS`, and the warning no longer says the field "is not read by the dataset loader" — accurate about the loader, but misleading about the object, since the attribute is populated moments later. Exactly the reason @akshita317 gave.

```python
# Accepted as genuine inputs until 179d39d (2024-07-06); process_dataset()
# has computed them from the *_index / *_data / *_data_info keys ever since.
_OUTPUT_ONLY_FIELDS = {
    "train_url": "train_index, train_data or train_data_info",
    "test_url": "test_index, test_data or test_data_info",
}
```

The warning now reads:

```
[WARNING] - dataset field `train_url` is computed by process_dataset() and any supplied value is overwritten; use train_index, train_data or train_data_info instead.
```

@azaynul10 — the `git log -S` history is what made this land for me. Both readings were correct for different periods, and dating the boundary in a comment keeps the fact available rather than choosing one name and losing the other. Adopted as you suggested.

Verified with CI's flags: `pylint --max-positional-arguments=10 core/testenvmanager/dataset/dataset.py` → 10.00/10.

## Deferred: two design questions

**Warn, migrate, or reject.** @suhaan-24 is right that warn-and-continue gives a better message and the same outcome — the run still stops at `NotImplementedError` a few lines later. The migration shim would change that, and the zero-overlap finding means it has no ambiguity to resolve, which is a genuinely strong argument. But @azaynul10's correction moves the sizing to 20 legacy against 28 current at HEAD, so the shim clears a smaller backlog against a 27% larger regression surface than the original case assumed. That trade-off seems like a maintainer call, and adopting it would turn this from a warning-only change into a behavioural one affecting 20 configs.

**Warning noise.** Raised by @akshita317, @Nikhil-VS1811 and @suhaan-24, and the proposed remedies differ — an allowlist, debug level for known-intentional keys, or @Nikhil-VS1811's Class 2 (dead config) versus Class 3 (intentional metadata) split. I don't think I should choose between those unilaterally. The Class 2/3 distinction in particular rests on judgements about individual example configs that maintainers are better placed to make than I am.

@MooreZheng @hsj576 — happy to implement either direction, or to keep this as the narrow warning-only change and let a shim land separately. A steer on the first question especially would help, since @suhaan-24 notes five open PRs touch this method and it seems better settled once than per PR.

## Worth separating out

@azaynul10's §3 census — six of seven `_parse_config` implementations under `core/` discard unrecognised keys silently, while `TestEnv` does the inverse and accepts anything — suggests whatever is decided here has five more sites to apply to. I confirmed the `TestEnv` half separately:

```
incremental_rounds  = 2      # default, validated and passed
incremental_round   = 10     # the typo, stored on the instance, never read
totally_made_up_key = xyz    # accepted
```

So a key's fate depends on which block it appears in, and neither outcome is reported. That seems worth its own scope rather than expanding this PR — @azaynul10 has already filed #938 for the `Rank` case.

PR #759, opened 2026-08-12
fix(core/common): guarantee sys.path cleanup in try-finally on module import failures

Raj-vardhan01:
### What type of PR is this?
/kind bug

In `core/common/utils.py`, dynamic module importing functions `load_module()` and `py2dict()` prepend module directories to `sys.path` via `sys.path.insert(0, path)`. 

Before this fix, `sys.path.pop(0)` was placed inside the `try` block after `import_module()`. If `import_module()` raised an exception (such as `SyntaxError`, `ImportError`, or `ModuleNotFoundError`), execution jumped directly to the `except` block, bypassing `sys.path.pop(0)`. This caused **`sys.path` pollution**, permanently leaking temporary configuration paths into `sys.path` during multi-test benchmark runs.

This PR wraps `import_module()` calls inside `try...finally` blocks to guarantee that `sys.path` is always cleaned up regardless of import success or failure.

Fixes # (N/A)

- **`core/common/utils.py`**: Placed `sys.path.pop(0)` inside `finally` blocks in `py2dict()` and `load_module()`.
- **`core/common/test_utils.py`**: Added unit tests verifying `sys.path` restoration when dynamic module imports raise errors.

### Does this PR introduce a user-facing change?:
```release-note
NONE


Omkeswani27:
Tested this branch directly (`git fetch origin pull/759/head:pr-759`):

load_module leaked? False
py2dict leaked? False

Confirms this correctly fixes the sys.path leak, with defensive index checks
that #729 lacks. Cross-referenced against #729 and #806, which independently
implement overlapping fixes — full duplication analysis and recommendation
(minor revision):

Yash4616:
The positional guard here, `if sys.path and sys.path[0] == module_path`, reads as an attempt to avoid removing the wrong entry, which is the right thing to be worried about. Executed, it inverts the outcome in the case it is guarding against.

If the module being imported inserts at index 0 during its own import, then when the `finally` runs, `sys.path[0]` is the module's entry rather than `module_path`. The guard does not match, nothing is removed, and both entries remain:

| | restores after a failed import | correct when the loaded module inserts at index 0 |
|---|---|---|
| this PR | PASS | FAIL, guard does not match, both entries leak |
| #729 | PASS | FAIL, removes the module's entry instead |
| #806 | PASS | PASS, snapshot and restore |

`examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/basemodel.py:29` triggers this, and is loaded through `load_module`. Its `if current_dir not in sys.path` guard on line 28 passes on a first load, so it does not prevent the case.

Snapshotting the whole list before the insert and restoring it in the `finally` removes the need for a guard at all, since it is correct regardless of what the imported module did to `sys.path`.

Separately, this PR does not declare an issue it closes, so nothing links or auto-closes if it merges. Given four other open PRs are changing the same function, a `Fixes #...` line would help whoever sequences them.

**Recommendation: major revision.** The guard needs replacing rather than adjusting.

Aryansingh-ai:

I checked out PR #759 and tested the modified `core/common/utils.py` directly.

Checked-out commit:

```text
7e9ff13 [PHASE IV] fix(core/common): guarantee sys.path cleanup in try-finally on module import failures

I forced load_module() to fail and compared sys.path before and after the call.

Observed result:

BEFORE_LEN: 10
AFTER_LEN: 10
UNCHANGED: True

The failing import did not leave the temporary module directory in sys.path.

I also tested the separate basename-collision property after applying this PR.

2. Basename collision — FAIL

Using two files with the same basename:

one/plugin.py → VALUE = "one"
two/plugin.py → VALUE = "two"

the result was:

AFTER FIRST: one
AFTER SECOND: one
EXPECTED SECOND: two

Therefore, #759 fixes the failure-path cleanup defect but does not address the independent module-identity collision reported in #557.

Comparison with #729

The implementation overlaps substantially with #729: both use try/finally cleanup around the existing dynamic import path.

My recommendation is therefore not to merge both independently without first deciding which cleanup implementation should be retained.

The tested cleanup behavior is correct and the PR is appropriately scoped for the sys.path failure-path defect.

I recommend:

Explicitly scope the PR to the cleanup defect.
Add/retain a regression test covering failed load_module() / py2dict() imports.
Coordinate with #729/#806 to avoid merging multiple substantially overlapping implementations of the same fix.

The separate #557 basename-collision defect remains unresolved by this PR.

rajdhruvsingh:
Credit to @Omkeswani27 (executed leak test confirming the fix works), @Yash4616 (executed counter-example below), and @Aryansingh-ai (confirms the failure-path fix works but a separate basename-collision defect, #557, remains out of scope).

Flagging @Yash4616's finding as worth resolving before merge, not just noting: the guard here — if sys.path and sys.path[0] == module_path — is a positional check, and it fails specifically when the module being imported inserts its own entry at sys.path[0] during its own import. When that happens, the check no longer matches, nothing gets removed, and both entries leak. This isn't hypothetical — examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/basemodel.py:29 does exactly this on import. A full-list snapshot-and-restore (capture sys.path before the insert, restore the whole list in finally) would be correct regardless of what the imported module does, since it doesn't depend on position at all. Would recommend that shape here rather than the positional check, given the guard is otherwise well-tested (the failure-path test coverage in this PR is a real strength worth keeping).

PR #760, opened 2026-08-12
docs: document GPU access for Docker quick start

ETS2K7:
**What type of PR is this?**

/kind documentation

The Docker setup for the cloud-edge LLM example starts the container without exposing the host GPU. The guide lists a GPU as a prerequisite and the example uses a vLLM edge model, so the documented command leaves the required GPU unavailable inside the container.

I added `--gpus all` to the command and documented the NVIDIA driver and Container Toolkit requirements. I made the same update in the quick-start guide and the example README so the two sets of instructions stay consistent.

The GPU option can still be omitted when someone only wants to inspect existing result files without running the benchmark.

Related to #230

**Validation**:

- `git diff --check`
- Sphinx HTML documentation build: `sphinx-build -M html docs docs/_build`


ETS2K7:
Superseded by #761 after renaming the branch.

PR #762, opened 2026-08-12
fix(cifar100): remove hardcoded developer paths

Aryansingh-ai:
**What type of PR is this?**

/kind bug

This PR fixes portability issues in the CIFAR-100 example caused by hardcoded developer-specific filesystem paths.

The CIFAR-100 utility and training worker previously referenced `/home/wyd/ianvs/project/...` directly, which does not exist on other developers' machines and prevents the example from being used without manually modifying the source code.

This change:
- Replaces hardcoded absolute paths with paths derived using `pathlib.Path`.
- Makes the CIFAR-100 data and index file locations repository-relative.
- Preserves the existing CIFAR-100 data processing behavior.
- Updates the federated learning training worker to use the portable training index path.

No issue was opened; this PR fixes the identified hardcoded developer-specific paths in the CIFAR-100 example.

**Testing**:

- `git diff --check` passes.
- Python syntax compilation was performed for the modified files.
- End-to-end execution of `utils.py` was not performed because TensorFlow was not installed in the local environment.


31groot:
### Request changes — the portability fix creates incompatible CIFAR-100 data paths

**Review depth: Analysis**

The PR correctly removes the developer-specific `/home/wyd/ianvs/project/...` paths, but the replacement `Path.parents[...]` calculations are inconsistent between the two modified files.

In `examples/cifar100/utils.py` the PR uses:

```python
PROJECT_ROOT = Path(__file__).resolve().parents[4]
DATA_DIR = PROJECT_ROOT / "data" / "cifar100"
```

while in `examples/cifar100/sedna_federated_learning/train_worker/train.py` it uses:

```python
PROJECT_ROOT = Path(__file__).resolve().parents[2]
DATA_DIR = PROJECT_ROOT / "data" / "cifar100"
```

These do not resolve to the same project root.

I reproduced the exact path expressions from the PR layout:

```text
utils.py PROJECT_ROOT : /home
utils.py DATA_DIR     : /home/data/cifar100
train.py PROJECT_ROOT : /home/parv/ianvs/examples/cifar100
train.py DATA_DIR     : /home/parv/ianvs/examples/cifar100/data/cifar100
PATH MISMATCH REPRODUCED
```

So `utils.py` and the federated training worker are targeting different directories.

This matters because they participate in the same CIFAR-100 data pipeline: the preprocessing code generates `cifar100_train.txt`, `cifar100_test.txt`, and the corresponding index files, while the training worker consumes the training index. With these two path calculations, the producer and consumer no longer agree on where those artifacts live.

The original hard-coded path consistently targeted:

```text
<repo>/data/cifar100
```

The replacement should preserve that producer/consumer contract rather than independently deriving different roots from each file's location.

**Recommendation:** derive the repository root from a stable anchor in both files, or centralize the CIFAR-100 data-path construction in one shared helper. Please also add a regression test that verifies the preprocessing output path is exactly the same path consumed by the training worker.

A useful invariant is:

```python
assert utils_data_dir == train_worker_data_dir
```

and ideally the test should exercise generation followed by consumption rather than only checking that the old hard-coded string has disappeared.

**Verification boundary:** I evaluated the exact `Path.parents[...]` expressions from the submitted PR structure and reproduced the producer/consumer path mismatch locally. I did not run the full TensorFlow/CIFAR-100 training pipeline.

PR #763, opened 2026-08-12
fix: evaluate LLIO poses against ground truth

ETS2K7:
/kind bug

The LLIO pose-estimation example's `position_error` and `orientation_error` metrics ignored ground truth after slicing it. They measured consecutive predicted motion instead of prediction error: a perfect 1 m translation could score 1 m error, and a perfect 90 degree rotation could score 90 degrees error.

This PR:

- returns each estimate with the ground-truth pose from the end of the same processed KITTI window;
- uses that same endpoint rule in both training and prediction;
- computes Euclidean position error and geodesic orientation error from aligned pose pairs;
- records sequence lengths and evaluates trajectory consistency within each KITTI sequence, never across unrelated sequence boundaries;
- handles static paths without divide-by-zero warnings; and
- validates pose shapes, finite values, homogeneous bottom rows, rotation matrices, and sequence-length coverage.

This restores the metric contract documented by the example and prevents plausible but incorrect benchmark scores.

Related to #230

**Validation**:

- `python3 -m unittest discover -s examples/industrialEI/pose-estimation-llio/testenv -p 'test_*.py' -v` (13 tests)
- `python3 -m compileall -q examples/industrialEI/pose-estimation-llio/testenv examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/basemodel.py`
- `git diff --check origin/main...HEAD`
- Real KITTI dataset end-to-end benchmark on macOS/Apple M4:
  - 7 training sequences (166 frames)
  - 8 test sequences (277 evaluated poses)
  - training, prediction, and all three metrics completed successfully


ETS2K7:
Hi @hsj576 @Poorunga, could you take a look when you get a chance? I added tests and also ran the example end to end with the KITTI data. Happy to fix anything you spot. Thanks!

PR #767, opened 2026-08-12
[PHASE IV] fix(core): honour use_gpu: false and apply gpu visibility before modu…

### LINKED ISSUE #765: [PHASE IV] use_gpu: false in testenv.yaml is a no-op, and use_gpu: true is applied after the base model is already constructed
**What happened**:

`use_gpu` is declared in `core/testenvmanager/testenv/testenv.py` and parsed from `testenv.yaml`, but it is read in exactly one place in the core (the matches under `examples/…/GAN/lpips/` are an unrelated constructor argument in a vendored third-party library):

```python
# core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning.py:55-56
if kwargs.get("use_gpu", True):
    os.environ["CUDA_VISIBLE_DEVICES"] = "0"
```

Two separate defects follow from those two lines.

**1. `use_gpu: false` does nothing at all.** There is no `else` branch, so a run configured for CPU never restricts device visibility — it silently inherits whatever `CUDA_VISIBLE_DEVICES` the host has, and the model takes the GPU anyway. `examples/llm-edge-benchmark-suite/single_task_bench_with_compression/testenv/testenv.yaml` ships with `use_gpu: false` and does not get the CPU-only run it asks for.

**2. Even `use_gpu: true` is applied too late to have its intended effect.** Line 50 of the same file calls `ParadigmBase.__init__`, which calls `_get_module_instances()` — and that imports *and instantiates* the user's `basemodel.py`. That is where TensorFlow/PyTorch get imported and where many models select their device. The environment variable is only assigned on line 56, six lines after the model already exists.

Additionally, `use_gpu` is honoured only by the `singletasklearning` paradigm. Every other paradigm ignores the setting entirely, even though it is declared on the shared `TestEnv`.

- `use_gpu: false` restricts the run to CPU.
- `use_gpu: true` takes effect before any algorithm module is imported.
- The setting behaves consistently across paradigms, since it lives on the shared `TestEnv`.

Any `singletasklearning` job works. Point a testenv at a base model that just reports what it sees:

```python
# testalgorithms/basemodel.py
import os
from sedna.common.class_factory import ClassType, ClassFactory

__all__ = ["ProbeModel"]

def _report(stage):
    print(f"[PROBE] {stage}: CUDA_VISIBLE_DEVICES={os.environ.get('CUDA_VISIBLE_DEVICES', '__unset__')!r}")

@ClassFactory.register(ClassType.GENERAL, alias="ProbeModel")
class ProbeModel:
    def __init__(self, **kwargs):
        _report("basemodel.__init__")

    def train(self, train_data, valid_data=None, **kwargs):
        _report("train")
        return None

    def save(self, model_path):
        os.makedirs(model_path, exist_ok=True)
        path = os.path.join(model_path, "model.txt")
        with open(path, "w", encoding="utf-8") as f:
            f.write("probe")
        return path

    def load(self, model_url=None):
        return model_url

    def predict(self, data, **kwargs):
        _report("predict")
        return [0 for _ in data]
```

Set an ambient value standing in for whatever devices the host exposes, and run the job. No actual GPU is needed — the probe only reads the variable, so the defect is observable on any machine:

```
export CUDA_VISIBLE_DEVICES=0,1,2,3
ianvs -f ./benchmarkingjob.yaml
```

Observed, on current `main`:

| stage | `use_gpu: true` | `use_gpu: false` |
|---|---|---|
| `basemodel.__init__` | `'0,1,2,3'` — not applied yet | `'0,1,2,3'` |
| `train` | `'0'` | `'0,1,2,3'` — never applied |
| `predict` | `'0'` | `'0,1,2,3'` |

The right-hand column should read `-1` (or equivalent) at every stage. The left-hand column should read `0` at every stage, including `__init__`.

The fix needs care around one trap. `TestEnv` declares `use_gpu = False` as its default, but the *effective* behaviour when a testenv omits the key has always been "do not touch `CUDA_VISIBLE_DEVICES`" — because the only consumer's fallback is `kwargs.get("use_gpu", True)`, defaulting to `True`. The declared default and the real behaviour contradict each other.

Only 2 of the 41 `testenv.yaml` files under `examples/` declare `use_gpu` at all. So honouring the declared `False` default literally would force every other example onto CPU — a silent, repo-wide regression. Any fix should treat "not configured" as a third state distinct from an explicit `false`.

I have a patch for this and will open a PR shortly.

vjuhhii:
**What type of PR is this?**

/kind bug

`use_gpu` is declared on `TestEnv` and parsed from `testenv.yaml`, but it was read in exactly one place — `SingleTaskLearning.__init__`:

```python
if kwargs.get("use_gpu", True):
    os.environ["CUDA_VISIBLE_DEVICES"] = "0"
```

That produced two defects:

1. **`use_gpu: false` was a silent no-op.** No `else` branch, so a CPU-configured run inherited the host's `CUDA_VISIBLE_DEVICES` and took the GPU anyway. `examples/llm-edge-benchmark-suite/single_task_bench_with_compression` ships `use_gpu: false` and never got its CPU-only run.
2. **`use_gpu: true` was applied too late.** `ParadigmBase.__init__` runs first (line 50) and calls `_get_module_instances()`, which imports and instantiates the user's `basemodel.py` — where AI frameworks get imported and pick their device. The assignment landed six lines later, after the model already existed.

`use_gpu` was also honoured only by `singletasklearning`, despite living on the shared `TestEnv`.

**The change**

Moved the policy into `ParadigmBase.__init__`, immediately before `_get_module_instances()`:

- fixes the ordering by construction — the variable is set before any algorithm module is imported;
- applies to every paradigm rather than only `singletasklearning`;
- `use_gpu: false` now sets `CUDA_VISIBLE_DEVICES=-1`.

`-1` rather than `""`, because an empty value is read back as unset by some frameworks, which then fall back to using every device. The repo still ships TF1-era examples, so the more portable value is worth preferring.

**Why `TestEnv.use_gpu` now defaults to `None` — the part worth reviewing closely**

`TestEnv` declared `use_gpu = False`, but the *effective* behaviour for a testenv omitting the key was "do not touch `CUDA_VISIBLE_DEVICES`", because the only consumer's fallback was `kwargs.get("use_gpu", True)`. Declared default and real behaviour contradicted each other.

Only 2 of ~45 examples declare `use_gpu`. Honouring the declared `False` literally would have forced every other example onto CPU — a silent, repo-wide regression. I hit exactly this with a first version of the patch and backed it out.

So the default is now `None`, meaning "not configured, inherit the environment", giving three states:

| testenv | `CUDA_VISIBLE_DEVICES` | changed? |
|---|---|---|
| key absent | untouched | no — identical to today |
| `use_gpu: false` | `-1` | **yes — this is the fix** |
| `use_gpu: true` | `0` | same value as today, applied earlier |

This is safe because `_parse_config` normalises any present key through `bool(v)`, so `None` can only ever mean "absent from the yaml".

**Testing**

Verified with a `singletasklearning` job whose base model prints `CUDA_VISIBLE_DEVICES` at `__init__`, `train` and `predict`, run with `CUDA_VISIBLE_DEVICES=0,1,2,3` in the environment. Before/after for all three testenv states:

| testenv | before | after |
|---|---|---|
| key absent | `0,1,2,3` at every stage | `0,1,2,3` at every stage — unchanged |
| `use_gpu: false` | `0,1,2,3` at every stage | `-1` at every stage |
| `use_gpu: true` | `0,1,2,3` at `__init__`, then `0` | `0` at every stage |

`pylint --max-positional-arguments=10 core` stays at 10.00/10.

**Not addressed here**, to keep this focused — happy to open follow-ups:

- `TestEnv._parse_config` uses `bool(v)`, so a quoted `use_gpu: "false"` evaluates to `True`.
- With `use_gpu: true`, an operator's existing `CUDA_VISIBLE_DEVICES=2` is still overwritten with `0`. Arguably it should be respected, but that is a behaviour change beyond this fix.

Fixes #765

### DIFF
--- a/core/testcasecontroller/algorithm/paradigm/base.py
+++ b/core/testcasecontroller/algorithm/paradigm/base.py
@@ -53,8 +53,26 @@ class ParadigmBase:
         self.dataset = kwargs.get("dataset")
         self.workspace = workspace
         self.system_metric_info = {}
+        # note: this must be applied before _get_module_instances(), which imports
+        # and instantiates the algorithm module, because AI frameworks read
+        # CUDA_VISIBLE_DEVICES when they initialize their device list.
+        self._set_gpu_visibility(kwargs.get("use_gpu"))
         self.module_instances = self._get_module_instances()
         os.environ["LOCAL_TEST"] = "TRUE"
+
+    @classmethod
+    def _set_gpu_visibility(cls, use_gpu):
+        if use_gpu is None:
+            # use_gpu is not configured in testenv: leave the gpu visibility of
+            # the environment as it is, which is the behaviour ianvs has always had.
+            return
+
+        if use_gpu:
+            os.environ["CUDA_VISIBLE_DEVICES"] = "0"
+        else:
+            # "-1" instead of "" because an empty value is read back as unset by
+            # some AI frameworks, which then fall back to using every device.
+            os.environ["CUDA_VISIBLE_DEVICES"] = "-1"
 
     def dataset_output_dir(self):
         """
--- a/core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning.py
+++ b/core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning.py
@@ -52,8 +52,6 @@ class SingleTaskLearning(ParadigmBase):
         self.mode = kwargs.get("mode")
         self.quantization_type = kwargs.get("quantization_type")
         self.llama_quantize_path = kwargs.get("llama_quantize_path")
-        if kwargs.get("use_gpu", True):
-            os.environ["CUDA_VISIBLE_DEVICES"] = "0"
 
     def run(self):
         """
--- a/core/testenvmanager/testenv/testenv.py
+++ b/core/testenvmanager/testenv/testenv.py
@@ -46,7 +46,9 @@ class TestEnv:
         self.round = 1
         self.client_number = 1
         self.dataset = None
-        self.use_gpu = False  # default false
+        # None means not configured, in which case ianvs does not touch the gpu
+        # visibility of the environment. Set it to false to force a cpu-only run.
+        self.use_gpu = None
         self._parse_config(config)
 
     def _check_fields(self):


mwemmz:
Reviewed the diff to `core/testcasecontroller/algorithm/paradigm/base.py` directly. The fix correctly:
- Distinguishes `use_gpu: true` / `use_gpu: false` / unset as three distinct states, rather than treating unset and `false` as equivalent (the bug in the linked report).
- Sequences `_set_gpu_visibility()` before `_get_module_instances()`, so `CUDA_VISIBLE_DEVICES` is set before any AI framework in the algorithm module reads it at import time — matches the root cause described.
- Uses `CUDA_VISIBLE_DEVICES = "-1"` rather than an empty string for the `false` case, correctly noting (in the PR's own comment) that some frameworks treat an empty value as unset.

One coverage gap: I don't see a test in this diff that exercises the `use_gpu: false` path on a host with multiple GPUs to confirm all devices are actually hidden (vs. just the first one, or vs. a framework-specific quirk in reading `-1`). Might be worth an explicit unit test asserting `os.environ["CUDA_VISIBLE_DEVICES"] == "-1"` after `ParadigmBase.__init__` with `use_gpu=False`, since this behavior was silently broken before and regression coverage would guard against it recurring.

Hiteshsai007:
The Core-side changes are correct and well-motivated:

- Moving the logic into `ParadigmBase.__init__` before `_get_module_instances()` fixes the ordering bug.
- Three-state handling (`None` / `true` / `false`) correctly preserves existing behaviour for the many examples that omit the key.
- Using `CUDA_VISIBLE_DEVICES="-1"` for the false case is the right portable choice.

**Recommendation: minor revision**

One gap remains: there is still no unit test that asserts the `use_gpu: false` path (especially on a host with multiple GPUs). A small test that checks `os.environ["CUDA_VISIBLE_DEVICES"] == "-1"` after `ParadigmBase.__init__(use_gpu=False)` would guard against regression of the original silent no-op.

Also note that while this PR correctly fixes the Core mechanism, it does not require Examples to actually honour `use_gpu`. Examples that hardcode device selection (e.g. `government_rag`) remain unaffected.

Hiteshsai007:
### Cross-example impact assessment [STATIC ANALYSIS]

Inspected 9 Examples for device-selection behaviour. None receive `use_gpu` as input. After this PR merges, Core sets no `CUDA_VISIBLE_DEVICES` under any condition (the removed lines were the only assignment). Examples that call `torch.cuda.is_available()` will reflect hardware state with no Core constraint.

The PR correctly fixes the ordering defect and adds the `False → "-1"` path. The remaining gap is that no mechanism requires Examples to consult Core's decision. Consider exposing a `get_device()` utility from `ParadigmBase` that Examples can call, using the `resolve_device()` pattern already present in `examples/cloud-edge-speculative-decoding-benchmark/.../runtime.py:519`.

vjuhhii:
Thanks for reviewing, and for asking the cross-example question — that is the right thing to ask about this change.

On the unit test: agreed, and it is now open as #890. I kept it separate rather than growing this one, so this diff stays at +21/−3 and easy to review.

It covers all three states, and each case asserts `CUDA_VISIBLE_DEVICES` *at the moment the algorithm module is instantiated* rather than after construction returns — asserting afterwards would pass against the pre-fix code and prove nothing. Against `main` five of the seven cases fail; with this PR applied all seven pass. The two that pass either way are the unset-key invariants, which 39 of the 41 `testenv.yaml` files depend on.

One correction on the second comment:

> After this PR merges, Core sets no `CUDA_VISIBLE_DEVICES` under any condition (the removed lines were the only assignment).

The assignment was moved, not removed. The lines deleted from `singletask_learning.py` are re-added in `base.py`, which is why the change applies to every paradigm rather than just one:

```python
# core/testcasecontroller/algorithm/paradigm/base.py
self._set_gpu_visibility(kwargs.get("use_gpu"))
self.module_instances = self._get_module_instances()

@classmethod
def _set_gpu_visibility(cls, use_gpu):
    if use_gpu is None:
        return
    if use_gpu:
        os.environ["CUDA_VISIBLE_DEVICES"] = "0"
    else:
        os.environ["CUDA_VISIBLE_DEVICES"] = "-1"
```

And `use_gpu` does reach `ParadigmBase.__init__`:

- `core/testcasecontroller/testcase/testcase.py:68` copies every attribute of `TestEnv` into `test_env_config`
- `testcase.py:72` passes it on as `**test_env_config`
- `SingleTaskLearning.__init__` forwards the same dict unchanged: `ParadigmBase.__init__(self, workspace, **kwargs)`

`TestEnv.use_gpu` is set in `core/testenvmanager/testenv/testenv.py`, so it is present in that dict. The before/after table in the PR description was produced by running this, not by reading the diff.

On the substantive point you raise, though, I agree, and I think it is the more interesting half: nothing requires Examples to consult Core's decision. There are 160 hardcoded `cuda` references across 73 files under `examples/`, against 2 of 41 `testenv.yaml` files that declare `use_gpu` at all. `resolve_device()` at `examples/cloud-edge-speculative-decoding-benchmark/testalgorithms/speculative-decoding/common/runtime.py:519` is a good illustration of what happens without a contract — an Example reinventing device resolution locally, unaware of Core.

I have opened #888 with a concrete design for that contract, including a finding that `torch.cuda.is_available()` is not a sufficient predicate either. Happy to continue the design discussion there so this PR can stay scoped to the ordering fix.

Hiteshsai007:
Thanks for the detailed clarification and for correcting the statement about the CUDA_VISIBLE_DEVICES assignment being removed.

The explanation confirms that the assignment was moved into ParadigmBase rather than deleted, which is the important distinction for cross-paradigm behavior. The constructor-order assertion in #890 is also the right test design because checking the environment only after construction could miss the original bug.

I will treat #890 as the regression-test companion to this PR and #888 as the broader follow-up for the cross-example device-selection contract. I agree that #767 should remain narrowly scoped to the Core ordering and three-state visibility behavior.

rahulnr107:
## Review — PR #767

### Summary

PR #767 addresses the `use_gpu` behavior reported in Issue #765.

The previous implementation handled GPU visibility inside the single-task learning path and did not reliably honor `use_gpu: false`. The PR moves the visibility handling into the shared algorithm base initialization path and applies it before algorithm modules are imported and instantiated.

### Code Assessment

The new flow is:

```text
Read use_gpu
    ↓
Set CUDA_VISIBLE_DEVICES
    ↓
Import / instantiate algorithm module
```

This is technically appropriate because device visibility may be read by AI frameworks during initialization.

The PR also distinguishes:

```text
use_gpu = None
    → leave the existing environment unchanged

use_gpu = True
    → expose GPU 0

use_gpu = False
    → set CUDA_VISIBLE_DEVICES = -1
```

This is clearer than treating an omitted configuration and an explicit CPU request as the same case.

### Does it address the reported problem?

Yes. The implementation directly changes both the timing of device-visibility configuration and the handling of an explicit `use_gpu: false`.

### Regression Risk

Because this is a Core-level change, the behavior affects multiple paradigms rather than a single Example.

The main regression cases that should be protected are:

1. `use_gpu` omitted;
2. `use_gpu: true`;
3. `use_gpu: false`.

The Core should preserve the existing environment when `use_gpu` is not configured while enforcing the requested visibility when it is explicitly configured.

### CI / Verification Observation

The PR currently passes the displayed DCO, CodeQL, CI, and build checks.

However, the displayed checks do not themselves demonstrate functional verification of the three `use_gpu` states or prove that device visibility is applied before model/framework initialization.

Therefore, CI success should not be treated as equivalent to functional GPU-behavior verification.

### Recommendation

**Minor revision**

The implementation is technically well aligned with the reported problem, but explicit regression tests for the `use_gpu` state matrix would make the Core-level behavior safer to merge.

The relevant cases are:

```text
None  → preserve existing GPU visibility
True  → expose GPU 0
False → hide GPU
```

A regression-test PR addressing these cases would provide stronger evidence for the behavioral contract.

### Verification Boundary

Source diff and GitHub CI results were inspected.

End-to-end GPU execution was not independently verified in this review environment.

vjuhhii:
Thanks for the review.

On the regression tests: those are #890, which you have since looked at, so I believe that request is already met. It covers the three states and asserts CUDA_VISIBLE_DEVICES at the moment the algorithm module is instantiated rather than after construction returns. If there is a case you would want covered beyond those, say so and I will add it there.

Agreed on the verification boundary — CI passing is not GPU verification. The tests deliberately assert the Core environment contract rather than framework behaviour. The hardware-dependent half is #888, which measures what actually happens on a device once visibility is set correctly.

Hiteshsai007:
## Follow-up: `CUDA_VISIBLE_DEVICES` is not restored across sequential TestCases

The three-state ordering fix here is correct for a single construction (verified: pre-`_get_module_instances()` set, `false → "-1"`, omitted leaves the environment untouched). One process-lifecycle case remains: TestCases run sequentially in one process, and the mask is written but never restored when a paradigm/TestCase scope ends.

I instantiated the real `ParadigmBase`/`SingleTaskLearning` twice in one interpreter (module loader stubbed so no example/torch import) and recorded the value present at each `_get_module_instances()` call, with the host environment starting at `CUDA_VISIBLE_DEVICES="0"`:

```text
case1 use_gpu=False  ->  case2 use_gpu omitted
  case1 load saw CUDA_VISIBLE_DEVICES = -1
  case2 load saw CUDA_VISIBLE_DEVICES = -1     # case2 should "leave inherited env untouched"
  SET writes = ['-1']

case1 use_gpu=True  ->  case2 use_gpu=False
  case1 load saw 0 ; case2 load saw -1 ; SET writes ['0','-1']   # case2 correctly masked
```

This also holds on the **exception path**: when case 1's module loader raises during construction (i.e. after the mask is applied), case 2 (omitted) still loads under `-1`:
```text
case1(false) result : raised (example-module failure during construction)
case2(omitted) result: constructed
LOAD observations: [('case1=false+RAISE', '-1'), ('case2=omitted', '-1')]
-> after a FAILING case1, case2 load saw '-1' instead of host '0'  => exception-path leak
```
Consistent with the source: across `core/` the only `CUDA_VISIBLE_DEVICES` write is `singletask_learning.py:56`, with no `pop`/`finally`/restore anywhere, so there is no cleanup on either the normal or the exception path. On `main`, `use_gpu:false` never writes `-1` (the no-op), so the leak appears only once this PR makes the mask effective — the write is correct but never scoped/restored.

Consequence: after a `use_gpu:false` TestCase, the process stays on `"-1"`, so a later TestCase that omits `use_gpu` — the state this PR defines as "preserve the inherited environment" — loads its module with GPUs hidden rather than at the host value. A CPU-only case can silently force CPU for every later case in the job.

This is independent of the per-construction ordering fix and doesn't block it, but it's worth handling either here or in the follow-up (#890 tests): snapshot `CUDA_VISIBLE_DEVICES` before applying the mask in `ParadigmBase` and restore it on scope exit (including the exception path), or scope visibility per worker instead of mutating the shared process environment. A cheap regression gate is two constructions in one process (`false` then omitted) asserting the second load observes the host value, not `-1`.

shubhika123:
# Pre-test Review — LFX Term 3 2026 Phase IV

## Technical Evaluation

This PR addresses **Issue #765** by relocating GPU visibility configuration from `singletask_learning.py` into a new `_set_gpu_visibility()` method on `ParadigmBase`, which is called **before** `self._get_module_instances()`.

### Key Strengths

1. **Correct execution order**

   `CUDA_VISIBLE_DEVICES` is now configured before `_get_module_instances()` loads the algorithm modules. This ensures that frameworks such as PyTorch and TensorFlow see the intended device visibility during initialization.

2. **Paradigm-wide scope**

   Moving the logic into `ParadigmBase` means that the device configuration is inherited by the paradigms built on top of it, including SingleTask, Lifelong, Federated, and Incremental Learning.

### Initial Verification Boundary

I initially verified the execution order through local code inspection. In the previous implementation, `ParadigmBase.__init__` called `_get_module_instances()` before `singletask_learning.py` applied the `CUDA_VISIBLE_DEVICES` change.

---

## Update — 2026-08-28: Additional Execution Evidence

### Executed Reproduction

I went beyond static inspection and built a small stub harness to execute the relevant `ParadigmBase` / `SingleTaskLearning` device-visibility logic from both the pre-fix (`main`) and post-fix (`pr-767`) branches.

The `sedna` dependency was mocked because it is not installed locally and is not required for this specific execution path.

### 1. Pre-fix (`main`) — `use_gpu=False`

The pre-fix implementation does not define `_set_gpu_visibility()`.

I reproduced the `use_gpu` handling present in `SingleTaskLearning.__init__`. When `use_gpu=False`, the condition is skipped because there is no corresponding `else` branch.

**Observed result:**

```text
No _set_gpu_visibility method found (pre-fix code)
CUDA_VISIBLE_DEVICES after use_gpu=False request: <UNSET>

**Comment by vjuhhii** [comment]
## End-to-end verification on a real benchmarking job

The behaviour table in the description was produced with a probe model. Today I ran a full
benchmarking job instead — `examples/llm_simple_qa` with `Qwen/Qwen2.5-0.5B-Instruct` on a
GTX 1080 — with `use_gpu: false` added to its `testenv.yaml`, once on `main` and once with this
patch applied. Everything else was identical.

I added a temporary print in `basemodel.py` to observe where the model actually landed. It is local
instrumentation only and is not part of this PR.

**On `main`:**

```
[PROBE] CUDA_VISIBLE_DEVICES='0'  model.device=cuda:0  model.dtype=torch.bfloat16
```

The job asked for a CPU-only run and the model was loaded onto the GPU. This is #765 reproduced
through the public entry point rather than through a probe.

**With this PR:**

```
[PROBE] CUDA_VISIBLE_DEVICES='-1'  model.device=cpu  model.dtype=torch.bfloat16
```

The setting is applied before the algorithm module is instantiated, so the framework never sees a
device.

Two things worth adding for reviewers:

**The job completes normally in both cases.** Both runs finished, evaluated the ten test rows and
printed the leaderboard. That was the property I most wanted to check before asking anyone to merge
a change in `core/`: it alters device visibility for every paradigm, and the risk worth ruling out
is that it breaks an ordinary run. It does not.

**Do not read anything into the accuracy figures** the two runs produced. I initially did, and I was
wrong: this example samples during generation and sets no seed, so its score varies between
identical runs. I filed that separately as #988. It does not affect what is being demonstrated here,
which is device placement, but it is the reason I am not presenting a before/after metric comparison
alongside the device output.

Regression coverage for the three states is in #890, and the wider capability question this exposed
is #888 and #957.

PR #768, opened 2026-08-12
fix(llio): parse KITTI index paths consistently

### LINKED ISSUE #766: [BUG] LLIO pose estimation uses OS separator for KITTI index paths
**What happened**:

In `examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/basemodel.py`, KITTI sequence paths read from `train_index.txt` and `test_index.txt` are split using `os.sep`.

The README defines the index-file format using `/`, for example:

`2011_09_26/2011_09_26_drive_0001_sync`

On Windows, `os.sep` is `\`, so the `/` delimiter is not split. As a result, the date and drive components cannot be extracted correctly.

The same parsing pattern occurs at lines 200, 243, 575, and 617.

KITTI sequence paths from the index files should be parsed consistently regardless of the operating system, with:

* `2011_09_26` extracted as the dataset/date name
* `2011_09_26_drive_0001_sync` parsed to obtain the drive identifier

On Windows, run:

```python
import os

p = "2011_09_26/2011_09_26_drive_0001_sync"

print("os.sep =", repr(os.sep))
print("old =", p.split(os.sep))
print("new =", p.split("/"))
```

Observed output:

```text
os.sep = '\\'
old = ['2011_09_26/2011_09_26_drive_0001_sync']
new = ['2011_09_26', '2011_09_26_drive_0001_sync']
```

The repository README specifies `/` as the delimiter in both `train_index.txt` and `test_index.txt`.

The issue is limited to parsing the fixed-format index-file entries. The affected code currently uses:

```python
path_parts = str(seq_path).split(os.sep)
```

A minimal fix is to parse the index-file format using `/` rather than the host operating system's filesystem separator.

The issue was reproduced on Windows with Python 3.x. No KITTI dataset download is required to reproduce the underlying parsing problem.

Aryansingh-ai:
**What type of PR is this?**

/kind bug

Fixes a Windows portability issue in the LLIO pose-estimation example.

KITTI sequence paths in `train_index.txt` and `test_index.txt` use `/` as their fixed delimiter, for example:

`2011_09_26/2011_09_26_drive_0001_sync`

The LLIO implementation was parsing these paths with `os.sep`. On Windows, `os.sep` is `\`, so the sequence path was not split correctly and the dataset/date and drive components could not be extracted.

This PR changes the four affected parsing locations to split using `/`, matching the documented KITTI index-file format.

The change is intentionally minimal and does not alter the KITTI processing logic.

Fixes #766


rahulnr107:
## Review — PR #768

### Summary

PR #768 addresses the KITTI index-path parsing problem reported in Issue #766.

The existing implementation uses:

```python
path_parts = str(seq_path).split(os.sep)
```

The KITTI index-file format uses `/` as its fixed separator. On Windows, `os.sep` is `\`, so the existing implementation does not split the index entry correctly.

This PR changes the four affected occurrences to:

```python
path_parts = str(seq_path).split('/')
```

### Assessment

The change directly addresses the root cause of the reported parsing failure rather than masking the resulting error.

Using the format-defined separator is appropriate here because the input is a fixed-format index file, not a host-native filesystem path.

### Recommendation

**Accepted as it is.**

The implementation is small, targeted, and consistent across the four identified parsing locations.

### Additional observation

I recommend adding a regression test for the fixed-format KITTI index representation.

For example, a test could verify that:

```text
2011_09_26/2011_09_26_drive_0001_sync
```

is parsed into:

```text
2011_09_26
2011_09_26_drive_0001_sync
```

on Windows as well as on Unix-like systems.

This would protect the format-specific `/` contract from being accidentally changed back to `os.sep` in the future.

### Verification

I reproduced the underlying separator behavior on Windows and inspected the PR diff to verify that all four affected `split(os.sep)` occurrences in the target file were changed to `split('/')`.

task2-pr768-reproduction png


Screenshot (66)
Screenshot (67)

shubhika123:
## Pre-test Review (LFX Term 3 2026 Phase IV)

**Recommendation: Accepted as it is**

### Technical Evaluation

Fixes **Issue #766** by replacing `os.sep` path splitting with fixed POSIX `/` splitting across four parsing locations in:

`examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/basemodel.py`

### Key Strengths

1. **Cross-Platform Compatibility**

   KITTI sequence files (`train_index.txt`, `test_index.txt`) explicitly use `/` as path separators. On Windows systems, where `os.sep` is `\`, splitting sequence strings using `os.sep` does not work correctly.

2. **Minimal & Non-Breaking**

   Directly targets the parsing logic without altering downstream odometry estimation routines.

### Verification Boundary

- Verified that the index path string `2011_09_26/2011_09_26_drive_0001_sync` is not correctly split when using `split('\\')` on Windows, confirming the portability issue addressed by the PR.

---

### Update (2026-08-28): Precise Failure Mode

I ran the actual parsing logic against a real KITTI-style path while simulating a Windows host (`os.sep = '\'`), rather than relying only on static inspection.

#### Reproduction

```python
seq_path = '2011_09_26/2011_09_26_drive_0001_sync'

haneul-24:
This PR fixes the problem by changing the KITTI path parsing from `split(os.sep)` to `split('/')`.

This matches the format used in the KITTI index files and makes the parsing work the same way on different operating systems.

The change is limited to the KITTI index path parsing in the LLIO Example.

**Recommendation: Accepted as it is.**

PR #769, opened 2026-08-12
fix: correct invalid module url in phys_scene_gen algorithmenv.yaml

### LINKED ISSUE #757: phys_scene_gen: algorithmenv.yaml module url is an invalid dotted-import path, not a file path
Image

**What happened**:

`examples/phys_scene_gen/singletask_learning_bench/testalgorithms/ai2holodeck/algorithmenv.yaml` uses a dotted Python import-style string for the `basemodel` module URL:

```yaml
url: "examples.phys_scene_gen.singletask_learning_bench.testalgorithms.ai2holodeck.phys_scene_generator"
```

Ianvs's `utils.load_module()` expects a module URL that can be resolved as a file path or importable module. This value cannot be resolved in the current repository layout, so the `phys_scene_generator` module fails to load.

The `basemodel` module should be referenced using the actual Python file path so Ianvs can load it successfully.

From the Ianvs repository root:

```python
from core.common import utils

utils.load_module(
    "examples.phys_scene_gen.singletask_learning_bench."
    "testalgorithms.ai2holodeck.phys_scene_generator"
)
```

This produces:

```text
RuntimeError: load module(url=...) failed,
error: No module named 'examples.phys_scene_gen'
```

The failure was reproduced on both Linux/Python 3.12 and macOS/Python 3.10.

The `basemodel` URL can be corrected to the actual repository-relative file path:

```yaml
url: "./examples/phys_scene_gen/singletask_learning_bench/testalgorithms/ai2holodeck/phys_scene_generator.py"
```

This matches the path format expected by Ianvs's existing module-loading mechanism.

iron-prog:
/kind bug

Fixes an invalid module `url` in `algorithmenv.yaml` - it was a dotted
Python-import path, which Ianvs's `load_module()` can't resolve (it
expects a real file path). One-line fix. Verified the failure and the fix.
Fixes #757


Nikhil-VS1811:
Hi, I reviewed this algorithmenv.yaml fix for the LFX pre-test to demonstrate how Example-local configuration rot extends beyond data definitions. Full review here:https://github.com/kubeedge/ianvs/discussions/937

Shikhar-Kesharwani:
**Review depth: Analysis**

The change correctly replaces the dotted Python import path with a filesystem path, which addresses the immediate `load_module()` resolution failure. However, there is a remaining portability issue in the new path form.

The replacement is:

`./examples/phys_scene_gen/singletask_learning_bench/testalgorithms/ai2holodeck/phys_scene_generator.py`

This is a relative filesystem path, so its resolution depends on the process's current working directory rather than on the location of `algorithmenv.yaml` or the Ianvs repository root. The PR verifies the command from a repository-root context, but that does not establish that the same configuration works when Ianvs is invoked from another directory, which is a realistic usage pattern for a CLI/framework configuration.

This matters because the original defect was a configuration-to-module resolution problem. Replacing an invalid dotted path with a relative path fixes the immediate format mismatch, but can leave the example environment-dependent if the launcher is executed outside the repository root.

I recommend resolving the module path relative to a stable configuration/repository base (or normalizing it through the same path-resolution mechanism used by other Ianvs configuration entries) rather than relying on the caller's current working directory. A regression test should invoke the example from both the repository root and a different working directory and confirm that the same module is loaded.

**Evidence:** the PR's diff changes the value from a dotted module identifier to a `./examples/...` filesystem path, and the PR's verification establishes the failure/fix for the module-loading format but does not test a non-root working directory.

**Verification boundary:** the module-path format change is verified by the PR author. The working-directory portability case is a static path-resolution finding; I have not claimed an executed reproduction from a non-root directory.

This is separate from the existing fix: the PR addresses **invalid path representation**, while this review identifies the remaining **path-base dependency**.

PR #772, opened 2026-08-12
fix: handle quasi identifiers in hard sample mining

31groot:
## What does this PR do?

Fixes a typo in the QUASI identifier check in the hard sample mining
algorithm.

The code was checking `dentifier_type` instead of `identifier_type`,
which caused QUASI sensitive entities to be skipped when building the
answer text.

## What was changed?

Changed:

`ent.get("dentifier_type")`

to:

`ent.get("identifier_type")`

## Validation

- `python -m py_compile` passed
- Python AST syntax check passed
- `git diff --check` passed
- Verified a QUASI entity is transformed correctly:
  `Alice lives in Paris` → `Alice lives in LOC`

### DIFF
--- a/examples/TAB/cloud_edge_collaborative_inference_bench/test_algorithms/hard_sample_mining/hard_sample_mining.py
+++ b/examples/TAB/cloud_edge_collaborative_inference_bench/test_algorithms/hard_sample_mining/hard_sample_mining.py
@@ -75,7 +75,7 @@ class OracleRouter:
         for ent in sensitive_entities or []:
             if not isinstance(ent, dict):
                 continue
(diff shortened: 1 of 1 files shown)


31groot:
assign @moorezheng

Yahdel:
The spelling correction looks right. Before this change, `dict.get("dentifier_type")` would return nothing instead of raising an error, so QUASI identifiers could be skipped quietly. That is risky because a normal run may look successful while handling this particular type of data incorrectly.

Could you add a small regression test that:

- checks that an entity with `identifier_type: QUASI` follows the intended transformation path;
- checks that an entity without that field keeps its current behavior; and
- checks the produced result, not only that the code runs without an error.

This would make the fix safer and prevent the same silent failure from returning later.

PR #773, opened 2026-08-12
fix: make inference cache keys deterministic

31groot:
## Problem

The EdgeModel and CloudModelAPI implementations used Python's built-in
`hash()` as part of their persistent cache keys.

Python string hashes are randomized between interpreter processes. As a
result, the same inference request can produce different cache keys after
a process restart, making persisted `cache.json` entries unusable.

This defeats the purpose of the persistent inference cache.

## Fix

Replace process-dependent `hash()` values with deterministic SHA-256 cache keys.

The EdgeModel cache key now includes:
- query
- model
- backend
- max_tokens
- temperature
- top_p
- repetition_penalty

The CloudModelAPI cache key now includes:
- input text
- privacy method
- model
- max_tokens
- temperature
- top_p
- presence_penalty
- frequency_penalty

This also ensures that different inference configurations do not incorrectly
reuse the same cached result.

## Validation

- Verified Python's built-in hash produces different values for the same
  query across independent processes.
- Verified the SHA-256 cache key remains identical across processes.
- Verified different EdgeModel generation settings produce different cache keys.
- `python -m py_compile` passes for both modified modules.
- `git diff --check` passes.

### DIFF
--- a/examples/TAB/cloud_edge_collaborative_inference_bench/test_algorithms/cloud_model/cloud_model.py
+++ b/examples/TAB/cloud_edge_collaborative_inference_bench/test_algorithms/cloud_model/cloud_model.py
@@ -1,3 +1,4 @@
+import hashlib
 # cloud_model.py
 import time
 import json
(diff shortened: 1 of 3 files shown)


Hiteshsai007:
### Request changes — CloudModelAPI cache key omits effective generation configuration

I independently checked the cache-key construction at commit `835f4457f387eff218541ca8d433b465dd960765`.

The new SHA-256 construction is deterministic, but the CloudModelAPI path does not include the effective generation configuration consistently.

`CloudModelAPI.inference()` builds the key from its local `kwargs`:

```python
kwargs.get("temperature", 0.7)

31groot:
> ### Request changes — CloudModelAPI cache key omits effective generation configuration
> 
> I independently checked the cache-key construction at commit `835f4457f387eff218541ca8d433b465dd960765`.
> 
> The new SHA-256 construction is deterministic, but the CloudModelAPI path does not include the effective generation configuration consistently.
> 
> `CloudModelAPI.inference()` builds the key from its local `kwargs`:
> 
> ```python
> kwargs.get("temperature", 0.7)
> ```

Thanks for catching this.

I traced the cache path through `CloudModel -> BaseLLM` and confirmed that the persistent cache is handled at the `BaseLLM` layer, where the effective generation settings are already stored.

I updated the cache key to include `max_tokens`, `temperature`, `top_p`, and `repetition_penalty`, and applied the same key construction consistently when loading, reading, and updating the persistent cache.

Verification performed locally:
- identical query + identical generation configuration -> same cache key;
- identical query + different temperature -> different cache key;
- a cache entry created for one configuration is not reused for another configuration;
- `python -m compileall` passes;
- `git diff --check` passes.

The fix is pushed in commit `69f723b`.

PR #774, opened 2026-08-12
[PHASE IV] fix(pcb-aoi): replace removed tensorflow.contrib.slim with tf-slim fallback (Issue #749)

### LINKED ISSUE #749: PCB-AoI fault_detection example fails: `tensorflow.contrib.slim` removed in TF2.x
### What happened

Running the single-task learning PCB-AoI fault_detection benchmark fails with:

### Where

`examples/pcb-aoi/singletask_learning_bench/fault_detection/testalgorithms/fpn/basemodel.py`, line 26:

```python
import tensorflow.contrib.slim as slim
```

### Root cause

The `tensorflow.contrib` namespace was removed entirely in TensorFlow 2.0 (2019) — it isn't deprecated, it no longer exists. Any environment installing `tensorflow` via pip today gets TF 2.x by default, so this import fails unconditionally on a fresh setup. `requirements.txt` only specifies `tensorflow` with no version pin, so there's no way to get a working install this way.

### Suggested fix

`tensorflow.contrib.slim` was spun off into a standalone community-maintained package, `tf-slim`, which preserves the same API surface:

```bash
pip install tf-slim
```

```python
# before
import tensorflow.contrib.slim as slim

# after
import tf_slim as slim
```

### Environment

- Python 3.12
- Windows
- Fresh venv, TensorFlow installed via `pip install tensorflow` (latest)

### Related

This looks like the same class of dependency-drift issue tracked in #170 and the broader restoration effort proposed in #329, just in a different file/example path that those didn't cover. Happy to open a PR with the fix if this looks right.

Raj-vardhan01:
Fixes #749 where running the PCB-AoI fault detection benchmark fails with  in TensorFlow 2.x environments.

 was removed in TensorFlow 2.0. This PR adds a  fallback importing  and adds  to , while preserving backward compatibility with TF 1.x environments.

### Code changes:
- : Replace direct  import with  fallback.
- : Replace direct  import with  fallback.
- : Add  dependency.
- : Add unit test suite for PCB-AoI dependency imports.

### Verification:
Ran unit tests in  (3 tests passed).


Raj-vardhan01:
Closing this PR to respect the issue assignment on #749.

PR #775, opened 2026-08-12
fix(core/storymanager/rank): handle string elements and column names in sort_by and guard empty metric list in _sort_all_df

Raj-vardhan01:
Fixes three defects in `Rank._sort_all_df` (`core/storymanager/rank/rank.py`) during leaderboard generation:
1. **String element support in `sort_by`**: Previously, string items in `sort_by` configurations (e.g. `["accuracy"]` or `["f1_score"]`) caused `next(iter("accuracy"))` to iterate individual characters and yield `'a'`, followed by an `AttributeError` when calling `.get()`. This PR supports both string and dictionary items in `sort_by`.
2. **Support sorting by non-metric DataFrame columns**: Allows sorting by standard DataFrame columns (such as `algorithm`, `time`, or `paradigm`) when specified in `sort_by`.
3. **Guard empty `sort_metric_list`**: Prevents pandas `ValueError` crash when `sort_metric_list` is empty `[]` by returning `all_df` unmodified.

### Code changes:
- `core/storymanager/rank/rank.py`: Update `_sort_all_df` implementation with type checks, column fallback, and empty list guard.
- `core/storymanager/rank/test_rank_sort.py`: Add unit test suite for `Rank._sort_all_df`.

### Verification:
Ran unit tests in `core/storymanager/rank/test_rank_sort.py` (4 tests passed).


Hiteshsai007:
## Major revision — the test module globally replaces NumPy and pandas in `sys.modules`

**Decision: request changes.** The production sorting changes are reasonable, but the added test module mutates interpreter-global import state during pytest collection.

At [`test_rank_sort.py` lines 19–21](https://github.com/kubeedge/ianvs/blob/6803326aed7876b47330a13949403d5ec3417f42/core/storymanager/rank/test_rank_sort.py#L19-L21), the module unconditionally executes:

```python
sys.modules['numpy'] = MagicMock()
sys.modules['pandas'] = MagicMock()
sys.modules['core.storymanager.visualization'] = MagicMock()
```

This is not scoped to a test or reverted afterward. Pytest imports test modules during collection, so every module imported later in the same process can receive mocked NumPy/pandas instead of the real libraries. The outcome becomes collection-order dependent; even if another test imported NumPy first, this assignment replaces the canonical `sys.modules` entry for subsequent imports.

I imported the PR's actual test module in a fresh process where real NumPy 2.4.6 was already loaded. The observed state was:

```text
before numpy: 2.4.6 module
after numpy is MagicMock: True
after pandas is MagicMock: True
visualization is MagicMock: True
```

The four tests also use a `MagicMock` dataframe, so they do not verify real pandas sorting behavior or catch mismatched `by`/`ascending` semantics accepted by the mock.

**Recommendation:** remove the global `sys.modules` assignments and run these focused tests with real NumPy/pandas, which are runtime dependencies of `Rank`. Construct a small real `pandas.DataFrame` and assert row order for string, dictionary, non-metric-column, and empty-filter cases. If visualization must be isolated, patch only that dependency with `unittest.mock.patch.dict(...)` inside a fixture/context manager and restore it immediately.

**Verification boundary:** I executed the PR's actual test module import and confirmed the three global replacements. I did not run the repository's complete pytest suite, so I have demonstrated the contamination mechanism, not a specific later test that fails under a particular collection order.

PR #776, opened 2026-08-13
fix(pcb-aoi): replace removed tensorflow.contrib.slim with tf-slim

### LINKED ISSUE #749: PCB-AoI fault_detection example fails: `tensorflow.contrib.slim` removed in TF2.x

ANISHA-RAWAT:
**What type of PR is this?**

this PR fixes a broken import in the PCB-AoI fault_detection example. The code imports 'tensorflow.contrib.slim' , but the 'tensorflow.contrib' namespace was removed entirely in TensorFlow 2.0 '(2019) - it isn't deprecated, it isn't deprecated, it no longer exists. Since 'requirements.txt' only specifies 'tensorflow' with no version pin, any fresh install via pip today pulls TF 2.x,causing this import to fail unconditionally.
The fix replaces the import with 'tf-slim',a standalone community-maintained package that preserves the same API surface as the original 'tensorflow.contrib.slim' module,so no other code changes are needed beyond the import line.
Tested locally on windows with python 3.12 by running the benchmarking job end-to-end:
'ianvs -f examples/pcb-aoi/singletask_learning_bench/fault_detection/benchmarkoing.job . yaml ` 
This specific import error resolved after the fix.
Note: further testing surfaced a separate, unrelated issue - the ` FPN_TensorFlow`  wheel bundled in `resources/algorithms/` pins ` tensorflow~=1.140`, which has no distribution camapatible with Python 3.12. This is a deeper compatability problem outside the scope of thisPR; I'll file it separately.
*Used AI-assisted guidance for environment setup and git workflow steps during this contribution; root cause diagnosis, the fix, and verification were done independently.*

Fixes #749


akshhkaushik (commented):
### Request changes — the new import is undeclared, and `.gitignore` is damaged

Replacing removed `tensorflow.contrib.slim` with `tf_slim` is the right source-level direction. The PR does not make a fresh environment able to import it. I searched the PR tree for `tf-slim`/`tf_slim`: the only occurrence is the new import. This example has no requirements file, and the clean test environment reports `tf_slim_installed: False`.

There is also an unrelated binary rewrite of the repository `.gitignore`. The original is LF text with no NUL bytes. The PR version converts it to CRLF and ends with mixed bytes:

```text
i\x00a\x00n\x00v\x00s\x00-\x00e\x00n\x00v\x00/\x00\r\x00\n\x00
```

Git consequently shows “Binary files differ.” That makes the last ignore pattern invalid and hides the actual one-line code review in a binary patch.

Recommendation: revert `.gitignore`, add an example dependency contract that installs a compatible TensorFlow + `tf-slim` pair, and run an import smoke test from that declared environment. The PR body mentions a requirements file, but none exists under this example on the submitted head.

Verification boundary: inspected the PR tree/bytes and dependency declarations; I did not install TensorFlow or run FPN training.

Nikhil-VS1811:
Hi, I reviewed this PR for the LFX pre-test Bonus Task to highlight ML framework obsolescence (TensorFlow 1.x to 2.x) as a distinct driver of Example decay. Full review here:https://github.com/kubeedge/ianvs/discussions/937

aryananand-04:
Took a look at the updated branch. The `.gitignore` conflict is sorted and `tf-slim==1.1.0` is declared now, which moved this from `DIRTY` to `MERGEABLE`. Two things still open though.

The bigger one: this import exists twice in the pcb-aoi tree, at line 26 of both files.

```
examples/pcb-aoi/singletask_learning_bench/fault_detection/testalgorithms/fpn/basemodel.py:26
examples/pcb-aoi/incremental_learning_bench/fault_detection/testalgorithms/fpn/basemodel.py:26
```

Both are `import tensorflow.contrib.slim as slim`, byte for byte. This PR fixes the singletask one and leaves the incremental one alone, so `incremental_learning_bench/fault_detection` still dies the same way on any TF 2.x install. I hit it on a clean venv with TensorFlow 2.21.0:

```
ModuleNotFoundError: No module named 'tensorflow.contrib'
```

`tensorflow.contrib` was dropped in TF 2.0, so there's no version of 2.x where this works. The same `import tf_slim as slim` swap you've already made would cover it.

Second thing, smaller. The new `requirements.txt` sits at `examples/pcb-aoi/requirements.txt`, two levels up from the job directories. Anything resolving dependencies relative to the folder holding `benchmarkingjob.yaml` won't see it there. I ran that kind of check against this branch and both job directories still come back missing:

```
::error:: Missing requirements.txt in examples/pcb-aoi/singletask_learning_bench/fault_detection
::error:: Missing requirements.txt in examples/pcb-aoi/incremental_learning_bench/fault_detection
```

The dependency itself is declared correctly, so this is about where the file lives rather than whether it exists. Still means per-example resolution doesn't pick up `tf-slim` for either job.

Minor revision from me. The import fix and the dependency declaration are both right, and adding the one-liner to the incremental copy plus moving the manifest would close out the whole example instead of half of it.

I ran this on macOS arm64 in a clean venv on TF 2.21.0, and only exercised the import path and manifest resolution. I didn't run the benchmark itself.

PR #777, opened 2026-08-13
fix(simulation): update stale pinned kind version in auto-install fal…

vishnudathks:
check_host_kind()'s auto-install fallback was pinned to kind v0.17.0 (Jan 2023), roughly 3.5 years and 15 minor releases stale. Extracts the version into a KIND_VERSION constant and bumps it to v0.32.0, the current stable release, confirmed against kind's actual release tags. Adds tests/test_simulation_system_admin_kind_version.py guarding that the install command tracks the constant rather than a re-hardcoded literal.

Which issue(s) this PR fixes
Automatically closes linked issue when PR is merged.
Usage: `Fixes #<issue number>`, or `Fixes (paste link of issue)`.
If PR is about `failing-tests or flakes`, please post the related issues/tests in a comment and do not use `Fixes


vishnudathks:
@jaypume

Shikhar-Kesharwani:
### Review — supplementary installation-integrity finding

The version update from Kind v0.17.0 to v0.32.0 is correct, and extracting the version into `KIND_VERSION` plus testing that the install command uses the constant is a useful regression guard.

One additional concern is that the new test only verifies the version string used to construct the installation command. It does not verify the integrity or identity of the binary that the auto-install path ultimately downloads and executes.

This matters because `KIND_VERSION` being pinned does not by itself make the installation artifact immutable: the release asset behind that version is still fetched remotely, and a modified or substituted artifact would satisfy the same version-based test while changing the executable that Ianvs installs and runs.

For an automatic host-tool installation path, I recommend pinning/verifying the expected release artifact as well as the version—for example by verifying a published SHA-256 checksum or signature before installing it. The regression test could then cover both properties: the command targets the intended version and the downloaded artifact is verified against the expected digest/signature.

This is separate from the stale-version problem fixed by this PR: the current change restores the correct version selection, while the remaining concern is that version selection and artifact integrity are two different guarantees.

PR #780, opened 2026-08-13
fix: preserve hyperparameter config during parsing

subham-hq:
/kind bug

`Module._parse_hyperparameters()` currently extracts each hyperparameter entry with `dict.popitem()`. Because each entry is a reference to a dictionary in the caller-provided configuration, this operation removes the entry from the original object.

The first parse still produces the expected Cartesian product because the removed key and value are retained locally. However, the source configuration is left containing empty dictionaries. As a result, the same valid configuration cannot be safely reused, and parsing it a second time raises `KeyError: 'popitem(): dictionary is empty'`.

This PR:

- reads each hyperparameter entry through `dict.items()` without mutating the source configuration;
- preserves the existing combination-generation behavior for valid configurations;
- adds a regression test that verifies both the generated combinations and preservation of the input configuration.

This change does not alter the configuration schema or public API.

None — discovered and reproduced directly.

**Testing**:

```console
$ python -m unittest discover -s tests -p 'test_module.py' -v
test_parsing_preserves_input_configuration (test_module.TestModuleHyperparameters.test_parsing_preserves_input_configuration)
Parsing should generate combinations without modifying its input. ... ok

----------------------------------------------------------------------
Ran 1 test in 0.001s

OK
```


akshhkaushik (commented):
### Request changes — non-mutating parse, but multi-key entries are silently truncated

The new test passes and confirms the important part of this change: `_parse_hyperparameters()` no longer mutates its input (`1 passed`).

`next(iter(ele.items()))` introduces an unchecked schema edge. A list entry containing two hyperparameter keys is accepted and only the first is used. I executed the patched `Module` with:

```python
{"first": {"values": [1]}, "second": {"values": [2]}}
```

Result:

```text
MULTI_KEY_RESULT: [{'first': 1}]
SECOND_KEY_PRESENT: False
```

This changes which key is lost compared with `popitem()`, but it still turns malformed configuration into a plausible, incomplete experiment. That is worse than a clear configuration error because the leaderboard can be produced without the requested hyperparameter.

Recommendation: before extraction, require `ele` to be a mapping with exactly one item and require the value to be a mapping. Add zero-key, multi-key, and non-mapping tests alongside the preservation test.

Verification boundary: ran the PR test and the real patched parser; did not run a benchmark sweep.

karyxx:
I pulled this branch and tested the new unit test suite locally as part of my LFX Mentorship Term 3 pre-test investigation.

### What I tested:
I ran `tests/test_module.py` against `Module._parse_hyperparameters()` to verify input preservation:

```python
import copy

config = {
    "type": "basemodel",
    "name": "DemoModel",
    "url": "",
    "hyperparameters": [
        {"learning_rate": {"values": [0.1, 0.01]}},
        {"batch_size": {"values": [16, 32]}},
    ],
}
config_before = copy.deepcopy(config)
module = Module(config)

# Assertions verified locally:
assert module.hyperparameters_list == [
    {"learning_rate": 0.1, "batch_size": 16},
    {"learning_rate": 0.1, "batch_size": 32},
    {"learning_rate": 0.01, "batch_size": 16},
    {"learning_rate": 0.01, "batch_size": 32},
]
assert config == config_before  # Input dict is preserved without in-place mutation!
```

---

### My findings and analysis:

1. **Eliminates Destructive In-Place Mutation:**  
   Replacing `ele.popitem()` with `hp_name, hp_config = next(iter(ele.items()))` is a great fix for the core bug. Previously, `popitem()` was mutating the user's config dictionary in-place, which caused `self.hyperparameters` to end up corrupted as `[{}]` and broke downstream state.

2. **The Multi-Key Dictionary Edge Case:**  
   While `next(iter(ele.items()))` is non-destructive, it assumes every element in `hyperparameters` contains exactly one key. If a user writes multiple parameters under a single YAML list item:
   ```yaml
   hyperparameters:
     - learning_rate: { values: [0.01] }
       batch_size: { values: [32] } # Multiple keys in one dictionary
   ```
   `next(iter(...))` extracts only the first key (`learning_rate`) and silently ignores `batch_size`.  
   Adding a quick length check before extraction (`if len(ele) != 1: raise ValueError(...)`) would turn this into an immediate, clear configuration error rather than silently omitting parameters.

3. **Solid Unit Test Coverage:**  
   Appreciate that this PR includes dedicated unit tests in `tests/test_module.py` asserting that `config == config_before_parsing`.

---

### Recommendation:
**minor revision (fixes without coding effort, e.g., docs, squash commits)**  
*Justification:* Resolves the in-place dictionary mutation cleanly with good test coverage. It just needs a minor 1-line check ensuring `len(ele) == 1` so that multi-key dictionary entries aren't silently truncated.

ishansurdi:
I tested the mutation fix directly:
 
```
before: [{'learning_rate': {'values': [0.1, 0.2]}}]
after:  [{'learning_rate': {'values': [0.1, 0.2]}}]
```
 
Confirmed — `main` leaves the same input as `[{}]` after parsing; this PR doesn't.
 
The multi-key silent-truncation edge case is already reported twice in this thread with working repros, so I won't repeat that test. What I'd add: I ran the *same* malformed input (one dict, two keys) through both `main` and this PR to see which key survives on each:
 
```
main (popitem):        keeps batch_size, drops learning_rate
this PR (next(iter)):  keeps learning_rate, drops batch_size
```
 
Both versions silently drop one key with no error, but which key survives is version-dependent - not just a residual gap, but a behavior change between `main` and this PR for the same malformed input. The `len(ele) != 1` guard already suggested here would close this the same way it closes the truncation case.

Nikhil-VS1811:
Hi, I reviewed this PR's hyperparameter parsing logic for the LFX pre-test Bonus Task, connecting it to similar architectural flaws in Dataset parsing. Full review here: https://github.com/kubeedge/ianvs/discussions/937

subham-hq:
@akshhkaushik @karyxx — thanks both. You independently landed on the same gap, and it's a real one, so I've implemented the check rather than argue scope.

**What changed:** before extraction, `_parse_hyperparameters()` now requires each entry to be a mapping with exactly one item, and requires that item's value to be a mapping. Each failure raises a `ValueError` naming the offending entry.

On the behaviour question — `popitem()` and `next(iter(...))` were both silently truncating multi-key entries, but you're right that they truncate differently. `popitem()` returns the *last* item of an insertion-ordered dict; `next(iter(...))` returns the *first*. So my original change altered which key survived without making the truncation visible. Turning it into an explicit error resolves both.

**Broader than the reported case.** Writing the tests you asked for surfaced three further shapes that were also reaching the extraction unguarded, each producing an unrelated crash rather than a configuration error. Against `54e5ec0`:

```
ERROR: test_empty_entry_raises
    hp_name, hp_config = next(iter(ele.items()))
StopIteration

ERROR: test_non_mapping_entry_raises
    hp_name, hp_config = next(iter(ele.items()))
AttributeError: 'str' object has no attribute 'items'

ERROR: test_non_mapping_hyperparameter_config_raises
    hp_values = hp_config.get("values")
AttributeError: 'list' object has no attribute 'get'

FAIL: test_multi_key_entry_raises
AssertionError: ValueError not raised

Ran 5 tests — FAILED (failures=1, errors=3)
```

With the validation applied:

```
Ran 5 tests in 0.001s
OK
```

A `StopIteration` or an `AttributeError` on `'str' object has no attribute 'items'` gives a user editing YAML nothing to work with. All four now fail at parse time with a message naming the entry.

@karyxx — thanks for running the parser against your own config and confirming the preservation assertion independently. That was useful to see.

PR #784, opened 2026-08-13
fix: preserve source JSONL during key normalization

### LINKED ISSUE #492: Issue summarization for example of federated-llm/fedllm-peft
LFX 2026 Term 2 pre-test, Task 1 (for #230).

Example: `examples/federated-llm/fedllm-peft/`. Paradigm: `federatedlearning` in LLM mode (`if_mode_llm: true`). It came in through design issue #198, proposal PR #226, and implementation PR #267 (by `wwhwhu`, merged 2025-10-29, merge commit `16fa1ac`).

### 1.1 What the example is

`federated-llm/fedllm-peft/` is the federated LLM fine-tuning benchmark. It's the only example in the repo that puts federated learning together with generative LLM fine-tuning, which is what makes it worth keeping alive. The idea: a handful of simulated clients each fine-tune a model with PEFT/LoRA on their own slice of a medical-QA dataset (`keivalya/MedQuad-MedicalQnADataset`), a server averages the LoRA weights with FedAvg or FedAvg-M, and the result is scored with ROUGE-1/2/L and BLEU-4. #307 tracks it as the "Federated LLM" feature.

The pieces: `algorithm/model.py` holds the `LLMFederatedModel` estimator, `algorithm/FedAvg-PEFT.py` and `FedAvgM-PEFT.py` are the aggregators, plus `algorithm.yaml`, `testenv/testenv.yaml`, and four metric files. One thing that matters for the rest of this issue: PR #267 didn't just add the example, it also edited four core files to wire in the LLM path - `core/.../federated_learning/federated_learning.py`, `sedna_federated_learning.py`, `core/testenvmanager/dataset/dataset.py`, and `dataset/utils.py`. So some of the problems here live in core, not just in the example folder.

### 1.2 Existing issues that touch this example

| Class | Issue | State | What it is |
|---|---|---|---|
| Origin / design | #198 LLM Federated Fine-tuning Algorithm Based on KubeEdge-Ianvs | closed | the design issue this example implements |
| Paradigm | #306 Implement Federated Learning Paradigm Support | closed | the `federatedlearning` controller this example extends |
| Paradigm | #111 Federated learning paradigm | closed | the earliest paradigm tracking issue |
| Dependency / env | #489 `pip install sedna` installs the wrong PyPI package | open | the documented install pulls an unrelated unit-conversion package; breaks this example's first `sedna.datasources` import |
| Dependency / env | #152 Pylint errors due to Sedna compatibility in CI | open | the Sedna-compat surface that also hits the FL core path in CI |
| Dependency / env | #132 Complete unclear dependencies for Ianvs | open | repo-wide undeclared dependencies, including the LLM stack this example needs |
| Related sibling | #97 Federated Incremental Learning for Label Scarcity | open | cifar100 `fci_ssl`, shares the `FL_AGG` aggregation contract |
| Related sibling | #252 / #454 federated_learning/fedavg and cifar100 pretest issues | open | cifar100 FL siblings, same paradigm, different example |
| Trackers | #307 Track of Implemented Features, #194 Track of Runnable Examples | open | #307 lists "Federated LLM" pointing to #198; #194 doesn't yet mark this example as runnable |

I've split these into origin (A), paradigm/core foundation (B), dependency/environment (C), related siblings (D, listed because they share code, not because they're bugs in this example), and trackers (E).

### #198 - LLM Federated Fine-tuning Algorithm (origin)
This is the design issue the whole example came from. The goal was an efficient federated fine-tuning setup for LLMs on Ianvs using PEFT, so places like hospitals can fine-tune on local data without shipping it anywhere. The hard part it had to solve: full-parameter federated learning on an LLM is far too expensive to move around, so the design introduced a LoRA path and an aggregation contract that works on tensors instead of Sedna's plain numeric FedAvg. Worth reading first because it explains why `model.py` is written so specifically around ChatGLM-6B, which is the root of the API-drift problems in section 3 (X-1, X-2). Implemented by #267, proposed in #226.

### #306 and #111 - Federated learning paradigm (foundation)
These set up the `federatedlearning` controller in `core/.../federated_learning/` that this example sits on top of. The controller has to juggle N client threads, a shared aggregator, and per-round weight exchange; LLM mode adds the `partition_llm_data` path on top. The reason this matters: anything broken in the controller (like the GPU fan-out I describe in L-3) hits this example and every other federated one at the same time. It's also shared with cifar100 `fci_ssl`, so any fix here needs to stay additive or it'll break that example too.

### #489 - `pip install sedna` grabs the wrong package (open, and the first thing that bites everyone)
On PyPI, `sedna` is a unit-conversion library (`convert`, `maths`, `physics`, and so on), not KubeEdge Sedna. It installs without any warning. What makes it nasty is that it fails silently: the install works, the import works, and then `from sedna.common.class_factory import ClassFactory` blows up somewhere completely unrelated, so you spend an hour looking in the wrong place. The real Sedna is the wheel bundled under `resources/`. This blocks the very first import of `fedllm-peft` and every other LLM example, which makes it the single most common thing a new contributor trips on. No fix merged yet; the answer is to document the bundled-wheel install and add a guard.

### #152 and #132 - Sedna compatibility and missing dependencies (open)
#152 is about Sedna-compatibility pylint failures in CI; #132 is the broader "we never declared half our dependencies" problem. Both are relevant here: the missing LLM stack I describe in E-2 is basically a specific case of #132, and #152's CI surface covers the same FL core that #267 changed.

### #97, #252, #454 - sibling federated examples (related, not this example)
Listing these for context, not counting them as bugs in `fedllm-peft`. They register `FL_AGG` aggregators against the same contract, so if the aggregation side gets touched (L-2), these are the examples to regression-test against.

### #307 and #194 - trackers
#307 marks "Federated LLM" as implemented and points at #198. #194 doesn't currently list `fedllm-peft` as runnable, and section 3 is basically the evidence for why, plus what it'd take to flip it.

## 3. What actually breaks when you run it

This section is from running the example myself rather than from existing issues - the only execution issue currently filed against `fedllm-peft` is the cross-cutting dependency one (#489). I set up a clean Python 3.10 environment, installed the bundled Sedna wheel instead of the PyPI one (because of #489), and ran:

```
ianvs -f examples/federated-llm/fedllm-peft/benchmarkingjob.yaml
```

It doesn't get anywhere. It dies before any federated or LLM code runs. Here's the actual traceback:

```
ValueError: dataset file(./project/data/fedllm/medicalQA/train_sub.jsonl) is not a local file and not a absolute path.
  core/testenvmanager/dataset/dataset.py:80
RuntimeError: testenv config file(.../testenv/testenv.yaml) is not supported
  core/cmd/obj/benchmarkingjob.py:123
RuntimeError: benchmarkingjob runs failed
  core/cmd/benchmarking.py:41
```

Below are the 15 problems that actually matter, grouped by where they live. There's also a tail of smaller stuff (print vs LOGGER, a dead `eval()`, `max_new_tokens=1024`, an off-by-one in a log f-string) that I've left out to keep this readable, and some of that overlaps the Gemini review anyway - I cover that in the Task 2 comment. Severity is P0 (won't launch), P1 (won't finish a round or gives invalid results), P2 (works but fragile/wrong), P3 (cosmetic). Every one of these came in with PR #267.

### Environment
| ID | Sev | Where | What happens | Why | What I'd do |
|----|-----|-------|--------------|-----|-------------|
| E-1 | P0 | `testenv/testenv.yaml:6,8` | the ValueError above | the dataset tree `./project/data/fedllm/medicalQA/` is hardcoded but #267 never creates it and the README never builds it | add a `scripts/prepare_medquad.py` that pulls the HF dataset, drops `qtype`, renames the columns to `question`/`answer`, sub-samples, and writes the train/test split |
| E-2 | P0 | root `requirements.txt` (7 lines) and `README.md:230` | `ModuleNotFoundError: No module named 'evaluate'` when metrics load | the root requirements only lists `prettytable, scikit-learn, numpy, pandas, tqdm, matplotlib, onnx`; the LLM deps show up only as a stray README line (`pip install peft transformers datasets rouge_score nltk`) that still misses `torch`, `evaluate`, `sentencepiece`, `sacrebleu`, `accelerate`, `protobuf`; there's no per-example requirements file at all | add a pinned `examples/federated-llm/fedllm-peft/requirements.txt` with the full stack including `evaluate` |
| E-3 | P1 | `README.md:219` | `python -m pip install third_party/*` doesn't resolve | the README went stale after the `resources/third_party/` rename in PR #406 | fix the path |

### Example code (`algorithm/model.py`)
| ID | Sev | Line | What happens | What I'd do |
|----|-----|------|--------------|-------------|
| X-1 | P0 | `:91` | loads the model with `AutoModelForSeq2SeqLM` on the `initial_model_url` branch, while the LoRA config at `:116` sets `task_type=TaskType.CAUSAL_LM` (the else branch at `:94` correctly uses `AutoModelForCausalLM`) | use `AutoModelForCausalLM` on both branches, or auto-detect via `AutoConfig` |
| X-2 | P0 | `:52` (assembly), reads `mask_token_id` at `:164` | the prompt is built as `[BOS, ids, MASK, ids, EOS]`, but `tokenizer.mask_token_id` is `None` on any non-ChatGLM tokenizer, which corrupts the labels and crashes inside `tokenizer.pad` | use `tokenizer.apply_chat_template`, with a fallback for non-GLM models |
| X-3 | P1 | `:103` | `self.model.half().float()` casts to fp16 and loses precision, then casts straight back to fp32 without recovering it | just `self.model.half()`, behind a flag |
| X-4 | P2 | `:158` | `AdamW(self.model.parameters())` hands the optimizer every parameter including the frozen base. AdamW allocates state lazily so it doesn't actually OOM, but it's fragile and misleading and would break if a future torch materialized state eagerly or if someone toggled `requires_grad` | pass `filter(lambda p: p.requires_grad, self.model.parameters())` |
| X-5 | P1 | `:196` (`self.save()`, `save_pretrained` at `:286`) | all four client threads write to the same `save_dir` at once, so whichever finishes last wins and the others are gone (this is the concurrency bug, separate from Gemini's note that the `save_dir` default is a hardcoded path) | suffix the dir with `client_{id}_round_{r}/` |
| X-6 | P1 | `:83` and the generation path | `self.device` can still be `cpu` on the global-model eval path even though the model is on GPU | resolve the device per call instead of caching it |

### Example config
| ID | Sev | Line | What happens | What I'd do |
|----|-----|------|--------------|-------------|
| C-1 | P1 | `testenv.yaml:36` | `gpu_num: 4` gives `CUDA error: invalid device ordinal` on a box with fewer than 4 GPUs | make it `gpu_num: auto` |
| C-2 | P1 | `algorithm.yaml:28-35` | the FedAvg-M block is commented out, so the FedAvg-M numbers in the README can't be reproduced | uncomment it or add a sibling yaml |
| C-3 | P1 | `algorithm.yaml:3-6` | `fl_data_setting` only has `train_ratio`, `splitting_method`, `label_data_ratio` - no `data_partition` or `non_iid_ratio`, so it's forced IID even though the whole hospital framing is non-IID | extend `partition_llm_data` to honor a `data_partition: non-iid` key |

### Ianvs core (cross-cutting, helps the sibling federated examples too)
| ID | Sev | Line | What happens | What I'd do |
|----|-----|------|--------------|-------------|
| L-1 | P1 | `core/testenvmanager/dataset/utils.py:154` (called at `federated_learning.py:109,337`) | `rename_keys_jsonl` does `os.replace(tmp, src)`, so it rewrites the user's actual dataset file in place | write a `.normalized.jsonl` next to the original and use that |
| L-2 | P2 (improvement) | sedna `aggregation.FedAvg` vs `algorithm/FedAvg-PEFT.py` | Sedna's built-in FedAvg only takes `List[List[float]]`, but PEFT weights are `Dict[str, Tensor]`, so #267 had to ship its own `FedAvg-PEFT.py`. It works, so this isn't a blocker, but it's duplicate code that shouldn't have to exist | upstream a `FedAvgDict` to Sedna and drop the per-example copy |
| L-3 | P1 | `federated_learning.py:298` plus `model.py:154` | `for gpu_idx in range(self.gpu_num)` starts one worker per GPU index and passes `device_id=gpu_idx` straight into `cuda:{gpu_idx}`. With `gpu_num: 4` on a box that only has 1-3 GPUs, the workers for the missing ordinals hit `CUDA error: invalid device ordinal`. The CPU path at `model.py:153` is guarded, so this is the 1-3-GPU case specifically, not pure CPU | read it as `min(requested, torch.cuda.device_count()) or 1` |

That's 15 that matter: 4 P0, 9 P1, 2 P2, plus the smaller tail I skipped. The P0 group is what stops it from launching (E-1 is the literal first crash above). L-1 and L-3 are core fixes that also help the cifar100 federated siblings, and L-2 is really an upstream Sedna improvement. All of it traces back to PR #267, which I go through in the Task 2 comment.

subham-hq:
**What type of PR is this?**

/kind bug
/kind api-change

`rename_keys_jsonl()` previously normalized a JSONL dataset by atomically
replacing the path supplied by the caller. In federated LLM mode, the helper is
used for both training and test data, so running a benchmark could permanently
rewrite the caller's source dataset from its original schema to `question` and
`answer` keys.

This PR changes the helper to preserve its input:

- If the source already uses `question` and `answer`, return the original path
  unchanged.
- Otherwise, write the normalized records to a sibling file such as
  `dataset.normalized.jsonl` and return that path.
- Update the federated-learning training and prediction paths to consume the
  path returned by the helper.
- Add regression tests covering both an already-normalized source and a source
  that requires normalization.

The helper's internal contract therefore changes from in-place mutation with
no return value to returning the path that callers should load. The two in-tree
call sites are updated in this PR.

**Root cause**:

The helper wrote normalized records to a temporary file and then called
`os.replace(tmp_path, src)`, replacing the input dataset. Its callers relied on
that side effect and ignored the return value.

**Validation**:

```text
python -m py_compile \
  core/testenvmanager/dataset/utils.py \
  core/testcasecontroller/algorithm/paradigm/federated_learning/federated_learning.py

python -m unittest discover -s tests -v
```

Result: 2 tests passed.

An end-to-end federated LLM benchmark was not run because it requires the
external model, dataset, and GPU environment. The focused tests exercise the
normalization behavior directly.

Related to #492 (finding L-1: in-place JSONL source mutation).

No dedicated issue; the behavior was reproduced directly and is already
documented in #492. This PR intentionally does not use `Fixes #492` because
that issue tracks multiple independent findings.


akshhkaushik (commented):
### Request changes — preserving the source still requires writing beside it

Both added tests pass, and returning the normalized path fixes the old in-place mutation. The destination remains coupled to the source directory:

```python
normalized_path = src.with_name(...)
tempfile.mkstemp(dir=src.parent)
```

That fails for a normal read-only dataset mount. I executed the real function with a readable JSONL file in a directory chmod'd `0555`:

```text
READ_ONLY_ERROR: PermissionError .../tmp...jsonl
```

There is a second lifecycle edge. If a later JSONL record is malformed, the temporary file created by `mkstemp()` is not removed:

```text
ERROR: JSONDecodeError
FILES_AFTER_ERROR: ['dataset.jsonl', 'tmpw3lx2uxj.jsonl']
```

Recommendation: take a caller-owned output/staging directory instead of writing beside input data, use a unique role-qualified filename, and remove the temporary file in `finally` on all failures. Add read-only-input, malformed-later-record, and repeated/concurrent-call tests.

Verification boundary: ran the PR tests and both filesystem/error cases; did not execute federated model training.

subham-hq:
@akshhkaushik Thanks for the review — both findings reproduce, and the temp-file leak is a real defect.

One clarification on scope first, since it affects how I'd like to handle them. Both behaviours predate this PR rather than being introduced by it:

**Read-only input.** The original helper called `tempfile.mkstemp(dir=src.parent)` and then `os.replace(tmp_path, src)`. It needed write access to the source directory both to stage the temp file and to overwrite the source, so a read-only dataset mount was never supported by this function. My change removes the second of those two writes but keeps the first, so it does not regress the case — it inherits it.

**Temp file leak.** The original had no `try`/`finally` either, and `json.loads(line)` sits inside the write loop, so a malformed record partway through the file orphaned the temp file exactly as you showed. Same defect, same cause.

Neither of those makes your points less valid. I just want to be clear about what this PR changes versus what it carries forward.

**Fixed and pushed:** the temp file lifecycle. The write loop and the rename are now wrapped in `try`/`finally` with `Path(tmp_path).unlink(missing_ok=True)`, which cleans up on any failure and is a no-op after a successful rename. Added a regression test using your scenario — a valid first record followed by a malformed one — asserting both that the error propagates and that no staging file is left behind.

Confirmed it catches the defect rather than being written to pass. On `9ac680f`:

```
FAIL: test_malformed_later_record_leaves_no_temporary_file
AssertionError: Lists differ: ['tmpy_cyoc9l.jsonl'] != []
```

With the fix applied:

```
Ran 3 tests in 0.003s
OK
```

**What I'd rather not decide unilaterally:** the caller-owned staging directory. I agree it's the correct destination — writing beside input data is the wrong default, and Ianvs already has a `workspace` concept in `benchmarkingjob.yaml` that looks like the natural home for it. But this PR is already labeled `kind/api-change` for returning a path instead of mutating in place, and threading a staging directory through would be a second signature change plus a decision about where the two in-tree callers source that directory from.

@MooreZheng — would you prefer that scoped into this PR, or kept as a follow-up once the return-value change lands? Happy either way; I'd just rather not expand an API change twice in one PR without a steer.

On the remaining test suggestions: read-only-input I can add as an expected-failure test documenting the current limitation, or leave until the staging-directory question is settled — let me know which is more useful. Repeated and concurrent calls I'd want to treat separately, since `mkstemp` is unique per call but the normalized output filename currently is not, so two concurrent normalizations of the same source would race on the final rename.

Roshan5105labs:
### LFX Pre-Test Review — PR #784

## Review Scope

I reviewed the current PR head a2d5ee4.

This PR changes rename_keys_jsonl() from mutating the caller's JSONL file in place to returning a normalized file path.

The federated training and evaluation callers are updated to use the returned path:

```python
train_dataset_file = rename_keys_jsonl(train_dataset_file)
```
and:
```python
test_dataset_file = rename_keys_jsonl(test_dataset_file)
```
I also verified on the PR branch that the production callers capture the returned path rather than ignoring it.

## Positive Findings

The change improves ownership of caller-provided dataset files by avoiding in-place modification of the original JSONL input.

The implementation also handles three important cases:

already-normalized input returns the original source path;

non-normalized input is written to a separate normalized file;

temporary staging files are cleaned up if normalization fails.

Runtime Verification

I executed the targeted unit tests in the clean Ianvs Python 3.8 environment:
```python
python -m unittest tests/test_dataset_utils.py -v
```
The result was:
```text
test_already_normalized_file_returns_source_path ... ok
test_malformed_later_record_leaves_no_temporary_file ... ok
test_normalization_preserves_source_file ... ok

Ran 3 tests in 0.041s

OK
```
image

These tests verify:

preservation of the source dataset;

correct behavior for already-normalized input;

cleanup of temporary files after malformed input.

Relationship to the FedLLM Restoration Problem

This PR improves one part of the state-ownership problem associated with #492 by preventing preprocessing from mutating caller-owned dataset files.

However, it does not address the separate client-artifact isolation concern in FedLLM-PEFT, where multiple client estimators retain the same configured save_dir.

Therefore, this PR improves the dataset-ownership boundary but should not be considered complete restoration of the broader FedLLM example by itself.

## Recommendation

Classification: Accept for the behavior reviewed here; broader FedLLM restoration remains incomplete.

The targeted normalization behavior is supported by passing regression tests, and the relevant production callers correctly use the returned normalized path.

The remaining shared-save_dir concern should be tracked as a separate restoration dependency rather than expanding this PR beyond its current responsibility.

subham-hq:
Thanks for taking the time to check this out and run it rather than reading the diff.

The Python 3.8 run is the part I most needed and could not do myself — I have been developing on 3.12, and `Path.unlink(missing_ok=True)` in the cleanup path requires 3.8 or later. It is in range, but "in range according to the docs" and "passes on the version the install guide actually targets" are different claims, and only one of them was mine to make. Good to have that closed.

Confirming the two production callers capture the returned path was the other useful check. That was the failure mode that would have made this change silently inert: the mutation stops, the normalized file gets written, and the callers keep reading the untouched original. Verified on the branch is better than verified by me.

Agreed on the `save_dir` scoping, and I think the distinction is worth stating explicitly since both get called "state ownership": this PR is about not mutating an input the caller owns, whereas the FedLLM-PEFT concern is about several client estimators sharing one output location. Different direction, different fix, and folding the second into a PR already carrying `kind/api-change` for the return value would make it harder to review rather than more complete. Tracking it against #492 seems right.

One thing still open on this PR, so an Accept does not quietly close over it: @akshhkaushik raised earlier that the staging file is still written beside the source via `mkstemp(dir=src.parent)`, which fails on a read-only dataset mount. That behaviour predates the PR — the original needed write access to that directory twice over, to stage and then to overwrite — so this change does not regress it, but it does not fix it either. Moving staging to a caller-supplied directory would be a second signature change, and I have left that for @MooreZheng to direct rather than deciding it myself.

PR #785, opened 2026-08-13
docs: fix wrong resources/third_party path in setup instructions

purunjay-bhardwaj:
## What this PR does

The non-Docker setup docs (`docs/guides/quick-start.md`, `docs/guides/how-to-install-ianvs.md`) and ~17 individual example READMEs instruct users to install the bundled Sedna wheel via:

```
pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl
```

That path doesn't exist — the wheel actually lives at the repo-root `resources/third_party/`, not `examples/resources/third_party/`. Running the documented command fails immediately on a clean clone with a file-not-found error, blocking `ianvs -v` from even importing.

I also found the same class of bug for a few other bundled assets referenced with an incorrect `examples/` prefix (`sedna-llm.zip`, `sedna.zip`, `algorithms/common.py`, the FPN TensorFlow wheel) in the `government`, `smart_coding`, and `bdd` example READMEs, and corrected those too. Additionally fixed a stale `sedna-0.4.1` reference in the `llm-edge-benchmark-suite` README to the current `0.6.0.1` wheel that's actually shipped in the repo.

## How I verified it

On a fresh clone:
- `pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl` → fails (path doesn't exist)
- `pip install resources/third_party/sedna-0.6.0.1-py3-none-any.whl` → succeeds, and `ianvs -v` runs afterward

This is a pure documentation/path fix — no code changes.


Nikhil-VS1811:
Hi, I reviewed this documentation path fix for the LFX pre-test to highlight how onboarding blockers effectively break examples for new contributors. Full review here: https://github.com/kubeedge/ianvs/discussions/937

Shikhar-Kesharwani:
**Review depth: Analysis**

The path corrections in this PR are well-supported and correctly address the documented `resources/third_party` path mismatch. However, there is a separate restoration gap that remains unresolved and can still prevent a clean checkout from being fully runnable.

The PR itself identifies two remaining references that are not merely using the wrong path:

* `examples/bdd/.../task_remodeling.py` references `examples/resources/yolo_model/`
* `examples/llm_simple_qa/README.md` references `opencompass-0.2.5` and `sedna-with-jsonl.zip`

The important distinction is that these assets are reported as missing under **both** `examples/resources/` and the repository-level `resources/` tree. Therefore, changing the path prefix would not resolve these references.

This creates a separate reproducibility/restoration problem: a contributor following the affected example instructions from a clean checkout can still reach a missing-resource failure even after the path corrections in this PR are applied.

I recommend tracking these as separate missing-asset restoration blockers rather than treating them as additional path corrections. The affected documentation/code should either reference assets that are actually shipped and versioned, or the required assets should be restored/pinned with clear acquisition instructions. A verification check should also confirm that every referenced local resource exists before declaring the example setup complete.

**Evidence:** the PR's own inspection identifies the two references and explicitly notes that the corresponding resources are absent from both resource locations.

**Verification boundary:** this finding is supported by the repository/path evidence documented in the PR. I am not claiming end-to-end execution of the affected examples because the referenced assets are currently unavailable.

Issue #369, opened 2026-02-23
[PRE-TEST] Critical Tensor Geometry Failures and Metric Inconsistencies in cloud_VLA_finetune Benchmark


satyam-coder07:
**1. Background**

**Introduction**
The examples/cloud_VLA_finetune benchmark is a cutting-edge addition to KubeEdge-Ianvs, designed to evaluate Vision-Language-Action (VLA) models in robotics scenarios. It leverages the singletasklearning paradigm to simulate how foundation models (like OpenVLA or RT-2) can be fine-tuned and deployed across the cloud-edge continuum.

**The Problem**
While the inference phase completes, the benchmark suffers from a fundamental pipeline crash during evaluation. After a deep dive into the testcasecontroller, I identified that the current implementation assumes a static output shape, which is incompatible with the variable-length nature of VLA action sequences.

The example doesn't just fail due to dependencies; it fails due to a logical mismatch in how Ianvs handles multimodal sequence data.

**Debug Process and Logs**
I initiated the benchmark and traced the failure to the metric computation layer. The crash is a sequential result of two main failures:

**Failure 1: Tensor Broadcasting Mismatch**
The model predicts a sequence of actions (e.g., a trajectory of 16 steps), but due to edge-simulated data truncation or variable-length ground truths, the evaluation tensor comparison fails.

```python
RuntimeError: The size of tensor a (12) must match the size of tensor b (16) at non-singleton dimension 1
```
**Failure 2: Lack of Sequence Masking**
Even when shapes are coerced, the Mean Squared Error (MSE) calculation includes "padding zeros" as actual data, which artificially inflates model accuracy and invalidates the benchmark's scientific integrity.

**2. Goals**
**What I Would Contribute**

- Implement Dynamic Padding: Integrate torch.nn.utils.rnn.pad_sequence into the evaluation algorithm to handle variable-length VLA outputs gracefully.

- Standardise Metric Interfaces: Modify the local algorithm script to ensure it passes sequence-length metadata to the ianvs/core metrics module.

- Configurable Sequence Constraints: Update testenv.yaml to allow users to define max_sequence_length and padding_strategy directly from the YAML.

- Hardware-Agnostic Validation: Fix the hardcoded CUDA assumptions in the VLA fine-tuning script to allow developers on Mac/CPU environments to validate their pipelines.

**3. Scope
Expected Users**

Robotics Researchers: Who need to benchmark foundation models (VLAs) for real-world manipulation tasks.

[...]

PR #786, opened 2026-08-13
fix(cloud_VLA_finetune): replace hardcoded personal cluster paths with relative paths

purunjay-bhardwaj:
## What this PR does

Every config file in `examples/cloud_VLA_finetune/singletask_learning_bench` hardcodes the original author's private HPC cluster path (`/inspire/hdd/global_user/chaimingxu-240108540141/jwq-test/ianvs/...`) for `workspace`, `testenv`, algorithm `url`, and dataset index fields in `benchmarkingjob.yaml`, `vla_algorithm.yaml`, and `testenv.yaml`. None of these paths exist on any other machine, so the example cannot run for anyone but the original author. The README's own "Run Ianvs" command also uses a leading-slash absolute path (`ianvs -f /examples/...`) instead of the relative-path convention every other example uses.

This PR rewrites all of these to the standard `./examples/...` relative-path convention, matching every other example in the repo, and fixes the README's leading-slash typo.

Checked this isn't already covered: issue #515 (hardcoded-path tracking issue) is scoped via `grep -rn "/home/" examples/`, which doesn't match this example's `/inspire/hdd/...` paths. Issue #369 (the only other open issue on this example) explicitly disclaims covering "surface-level bugs... broken paths."

## How I verified it

Ran `ianvs -f ./examples/cloud_VLA_finetune/singletask_learning_bench/benchmarkingjob.yaml` before and after:
- **Before:** fails immediately — can't load the hardcoded YAML paths.
- **After:** YAML/URL resolution succeeds; the run now progresses to a distinct, pre-existing, and separate failure — `testenv.yaml` points at a `test_data.txt` index file and `algorithm.yaml`'s `initial_model_url` points at an OpenVLA-7B checkpoint, neither of which is committed to the repo. That's a missing-dataset/model-asset problem, not a path problem, and is out of scope for this PR — noted here for visibility rather than papered over.


diegodevxd:
This patch fixes four YAML/README entry points, but it does not eliminate the private cluster dependency from the runnable code. The PR head still contains active `/inspire/hdd/...` values in `testalgorithms/vla_dataselect/basemodel.py`, `finetuning.py`, `generation.py`, and `vla_algorithm.yaml`; notably `COMPONENT_DIR` points at a checked-in example component, while model/data/output roots are external assets.

Please classify these separately. Convert checked-in component paths to a repository-derived path, and make external model/data/output locations explicit parameters or documented environment overrides. Otherwise the configuration will load further but module initialization remains bound to the original cluster.

PR #787, opened 2026-08-13
fix(smart_coding/issue): replace hardcoded workspace path with relative path

purunjay-bhardwaj:
## What this PR does

`examples/smart_coding/smart_coding_learning_bench/issue/benchmarkingjob.yaml` hardcoded `workspace: "/root/xieb/ianvs/workspace"`, the original author's personal path, which doesn't exist on any other machine.

This example wasn't caught by #515 / #643 (which fixed every other hardcoded path in this example, including the sibling `comment/benchmarkingjob.yaml`) because #515 was scoped via `grep -rn "/home/"`, which misses this `/root/`-prefixed path. Fixed to `"./workspace"`, matching the convention #643 already established for the rest of this example and every other example in the repo.

## Verification

Confirmed via diff against #643's file list that this specific file was the one variant left behind by that fix.


akshhkaushik (commented):
### Correct local fix, but make the open dependency explicit

The three changed values in the `issue` benchmarking job now resolve from the repository root: workspace is relative, and both referenced YAML files exist.

The sibling `comment` job on this PR head still contains three `/home/xiebo/ianvs/...` values. I checked them through Ianvs's local-file predicate; both referenced files return false from a fresh checkout. The current PR is therefore not independently consistent with its description that the rest of the example has already been fixed.

```text
issue:   testenv=True, algorithm=True
comment: testenv=False, algorithm=False
```

Recommendation: either document and enforce the sibling repair as a merge dependency, or include its three path fixes here after coordinating with that work's author. I would accept the changed line itself; the requested revision is about landing order and the claim of standalone restoration, not the relative-path choice.

Verification boundary: parsed both benchmarking configs and checked referenced paths; no LLM/model/API execution.

PR #488, opened 2026-05-28
[LFX Term : 01]  : Restoring City-Scape Sythia Semantic Segmentation 

NishantSinghhhhh:
**What type of PR is this?**

  /kind bug
  /kind failing-test

  **What this PR does / why we need it**:

  The `examples/cityscapes-synthia/lifelong_learning_bench/semantic-segmentation` benchmark was completely broken and could not be run end-to-end. This PR fixes **10 distinct 
  bugs** across configuration files, the RFNet utility modules, and the core lifelong learning paradigm so that the benchmark runs successfully on a local machine.

  The fixes fall into five categories:

  **1. Stale paths and wrong config keys (benchmarkingjob.yaml, rfnet_algorithm.yaml, testenv.yaml)**
  All three YAML files still referenced the old directory layout (`semantic_segmentation/lifelong_learning_bench/`) and a previous developer's home directory 
  (`/home/shijing.hu/`). In addition, `testenv.yaml` used `train_url`/`test_url` as dataset keys, but `sedna`'s `Dataset.process_dataset()` only recognises 
  `train_index`/`test_index` — passing the wrong key silently falls through to a `NotImplementedError`.

  **2. Hardcoded CUDA and worker count (RFNet/utils/args.py)**
  `TrainArgs` and `ValArgs` hardcoded `cuda=True` and `no_cuda=False`, crashing immediately on any machine without a GPU (`RuntimeError: No CUDA GPUs are available`). Both 
  classes also set `workers=4`, which causes `DataLoader` forked workers to deadlock when spawned from inside the ianvs benchmarking process on Linux. Fixed by using 
  `torch.cuda.is_available()` and `workers=0`.

  **3. Deprecated torchvision API (RFNet/utils/summaries.py)**
  `torchvision.utils.make_grid()` renamed the `range=` parameter to `value_range=` in version 0.9.0 and removed the old name in later versions, causing `TypeError: make_grid() 
  got an unexpected keyword argument 'range'` at the first visualisation step. Also switched the TensorBoard import from the third-party `tensorboardX` (not in requirements) to
   the built-in `torch.utils.tensorboard`.

  **4. Disk exhaustion from accumulated checkpoints (RFNet/utils/saver.py)**
  `Saver.__init__` incremented `experiment_N` on every run and never cleaned up old directories. With one ~500 MB checkpoint saved per epoch, repeated runs filled the entire 
[...]

PR #771, opened 2026-08-12
[LFX Term 2 2026] Add example validation framework

ken6078:
**What type of PR is this?**
/kind feature
/kind test

**What this PR does / why we need it**:
This PR restores the llm_simple_qa example and introduces an automated example validation framework.

  It includes:

  - Fixes and supporting scripts required to run llm_simple_qa reliably.
  - Static validation for example configurations and dependencies.
  - Dynamic smoke tests using mock Hugging Face and OpenAI-compatible runtimes.
  - Tiered validation workflows for pull requests, pushes, and scheduled health checks.
  - Regression detection and status reporting for all registered examples.
  - Documentation covering validation rules, classifications, status meanings, and local usage.

  Ianvs examples currently lack consistent automated checks, making configuration errors, broken dependencies, and runtime regressions difficult to detect before they are merged. This framework provides earlier
  feedback, reduces reliance on external model services during CI, and makes the health and support status of each example visible and reproducible.

**Which issue(s) this PR fixes**:

Fixes #233  
Fixes #247
Fixes #331  
Fixes #368 
Fixes #387 
Fixes #383  
Related to #230, #452, #541, #625

PR #788, opened 2026-08-13
fix(cityscapes-synthia): fix stale paths and unsupported keys in fulltest/smalltest/traveltest variant configs

purunjay-bhardwaj:
## What this PR does

The three "variant" configs for the cityscapes-synthia semantic-segmentation example (`benchmarkingjob-fulltest.yaml`, `-smalltest.yaml`, `-traveltest.yaml`, and their paired `testenv-full/small/travel.yaml`) had two bugs:

1. They still referenced `examples/semantic_segmentation/lifelong_learning_bench/...`, a directory that no longer exists — the example was moved/renamed to `examples/cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/...`.
2. The testenv files used `train_url`/`test_url` keys that `core/testenvmanager/dataset/dataset.py`'s `process_dataset()` doesn't recognize, unconditionally raising `NotImplementedError`.

This is a sibling gap to #488, which fixed only the default (non-suffixed) `benchmarkingjob.yaml` / `testenv.yaml` / `rfnet_algorithm.yaml` for this example — confirmed by diffing #488's exact file list, and these `*-fulltest`/`*-smalltest`/`*-traveltest` variants were left untouched by it.

## Verification

- Confirmed the referenced `examples/semantic_segmentation/...` paths don't exist in the current tree.
- Reproduced the `NotImplementedError` from the unsupported `train_url`/`test_url` keys directly against `process_dataset()` before the fix, confirmed it's gone after.


sheikhayaan:
The numbers below are measured before and after applying the patch, so they're observed and not
estimated. This fixes most of what it targets, and the files it misses fall into a pattern that
makes finishing the job straightforward.

Using the already-merged #771 static check on the `benchmarkingjob-smalltest.yaml` unit:

| | broken refs | exit |
|---|---:|---:|
| before | 30 | 1 |
| after `patch -p1` | **9** | **1** |

So a real two-thirds reduction, and the unit still doesn't pass. All 9 survivors are in files
this PR didn't touch:

```
benchmarkingjob.yaml            -> L9, L10, L11, L24
testalgorithms/rfnet/rfnet_algorithm.yaml -> L28, L43, L56
testenv/testenv.yaml            -> L18, L33
```

The pattern there is exact, and useful. This PR fixed every **variant** file (`-fulltest`,
`-smalltest`, `-traveltest`, `-new-split`, `testenv-full`, `testenv-small`, `testenv-travel`)
and missed all three **base** files, the ones whose names carry no suffix. Adding
`benchmarkingjob.yaml`, `rfnet_algorithm.yaml` and `testenv/testenv.yaml` should close the
unit.

That split is a property of the file list rather than of any one revision, so the force-push to
`5bbc391` doesn't change it: while the diff touches only suffixed files, the three unsuffixed
ones keep their stale references whatever else moves. On a clean checkout those three still hold
30 references to `examples/semantic_segmentation/...`, which is where the "before" figure above
comes from.

Something for whoever picks that up: the surviving references point at
`examples/semantic_segmentation/lifelong_learning_bench/...`, a directory name matching neither
the current layout nor the stale prefix in the variant files. This example appears to have been
renamed at least twice, and the base files were left behind at the earlier generation, so the
three remaining edits aren't the same string substitution as the ones already made here.

**Recommendation: minor revision.** Correct and substantial as far as it goes. Extend it to the
three base files above and the unit's static check should reach exit 0. That's the criterion
I'd hold it to, having measured the before and after myself.

suhaan-24:
The key rename and the path corrections here look right. `core/testenvmanager/dataset/dataset.py`
assigns `train_url`/`test_url` inside `process_dataset()` rather than reading them from config, so
the old keys reached:

```python
else:
    raise NotImplementedError('not one of train_index/train_data/train_data_info')
```

and the run aborted in `test_env.prepare()` before any model was built.

## One thing worth checking across the three variants

This PR updates `benchmarkingjob-fulltest.yaml`, `-smalltest.yaml` and `-traveltest.yaml`
together, which is the right instinct — but those three share a `testenv/` directory with several
`testenv*.yaml` variants, and the algorithm YAMLs reference them by path. It would be worth
confirming that every `testenv*` variant in this Example has been renamed, not only the ones
reachable from the three benchmarkingjob files edited here. A partially migrated Example is harder
to diagnose than an unmigrated one, because the failure then depends on which config the user
happens to select.

## Note on the workspace path

The change from `/ianvs/lifelong_learning_bench/workspace` to `./workspace/semantic-segmentation`
is a real improvement — the absolute form assumes a directory that exists only on the original
author's machine, and on macOS it cannot even be created at the filesystem root without disabling
SIP. I hit exactly this on the sibling curb-detection Example and had to repoint the paths to run
it at all.

Worth noting for whoever reviews the dataset side: `testenv.yaml` in this Example still carries an
absolute dataset path of the same shape. The workspace fix here makes the output location portable
but the input location remains machine-specific, so the Example is not yet runnable from a fresh
clone on that basis alone.

## Note on recurrence

Core ignores unrecognised dataset keys without comment, so a config using the old names fails with
`NotImplementedError` naming the three *accepted* keys but not the one that was supplied. That is
why this class of breakage tends to be found by running rather than by reading.

## Verification boundary

**Not executed.** This Example (`cityscapes-synthia/lifelong_learning_bench/semantic-segmentation`)
requires the 47.5 GB `semantic_segmentation_dataset.zip`, which I did not download. The
`NotImplementedError` path and the SIP/absolute-path behaviour are both from executed runs of the
sibling `curb-detection` Example in the same family, not of this one.

rajdhruvsingh:
Reviewing this alongside the existing thread — credit to @sheikhayaan for the before/after measurement via the merged validator (30→9 broken refs) and the variant-vs-base file split, and to @suhaan-24 for confirming the key rename "looks right" and flagging the cross-testenv-variant migration-completeness question. This comment adds three things beyond both:

1. Schema census, file-by-file: checked all seven patched files directly — active train_index:/test_index: keys present, dead train_url/test_url keys removed, across every variant. This gives @suhaan-24's "looks right" assessment file-level confirmation rather than a spot check.
2. Mode preservation: diffed all seven files for mode: changes — zero found. Every active mode value survives untouched, which matters because sibling config-repair PRs elsewhere in this restoration effort have silently dropped mode: keys during path fixes.
3. On @sheikhayaan's suggested remedy (extend this PR to the three base files it doesn't touch): the three base-file references are already carried by another open PR's diff, file-for-file matched (4 in benchmarkingjob.yaml, 3 in rfnet_algorithm.yaml, 2 in testenv.yaml — the exact nine lines @sheikhayaan identified as missing). Extending this PR to those files would duplicate that work. Recommend keeping this PR scoped to the variant files as-is, and resolving the base trio through the other PR.

phantom-712:
The path and key corrections in the suffixed variants are technically aligned
with the current loader contract. The old directory prefix cannot resolve in a
fresh checkout, and `process_dataset()` does not treat `train_url`/`test_url`
as equivalent to the supported input fields.

The acceptance test should enumerate the actual reachable graph rather than
checking only the three top-level benchmarking jobs:

```text
benchmarkingjob-fulltest.yaml  -> its testenv -> its algorithm -> module URLs
benchmarkingjob-smalltest.yaml -> its testenv -> its algorithm -> module URLs
benchmarkingjob-traveltest.yaml-> its testenv -> its algorithm -> module URLs
```

For each edge, assert that the referenced file exists and that the dataset
fields are accepted. Also scan the unsuffixed base files separately and report
them as remaining scope if they are intentionally handled elsewhere. A partial
migration is harder to diagnose than a uniformly broken configuration because
the result depends on which variant a user selects.

The workspace correction improves portability, but an environment-variable or
relative dataset path still needs a supplied external dataset. Static path
success must not be presented as an end-to-end Cityscapes run.

Recommendation: minor revision. Keep the variant-local changes, attach the
enumerated config/schema test, and explicitly link the owner/follow-up for the
untouched unsuffixed files.

PR #789, opened 2026-08-13
fix(core): expand ~ in dataset path fields before local-file/dir checks

purunjay-bhardwaj:
## What this PR does

`core/common/utils.py`'s `is_local_file()` / `is_local_dir()`, and `core/testenvmanager/dataset/dataset.py`'s `_check_dataset_url()` path via `_parse_config()`, never expand `~` in a configured path before checking it. A dataset path like `~/data/index.txt` is therefore neither found as a local file nor recognized as absolute, and `Dataset._check_dataset_url()` raises:

```
ValueError: dataset file(~/data/index.txt) is not a local file and not a absolute path
```

even when the file genuinely exists at the expanded location.

This isn't speculative — two examples already carry a manual comment warning contributors to avoid `~` in their configs for exactly this reason (`Cloud_Robotics/.../benchmarkingjob.yaml`, `cloud-edge-collaborative-inference-for-llm/benchmarkingjob.yaml`), which is evidence of a real, previously-worked-around core bug rather than something users should have to remember.

Fix: `os.path.expanduser()` in `is_local_file`/`is_local_dir`, and in `Dataset._parse_config()` for the path-bearing config fields (`train_url`, `test_url`, `train_index`, `test_index`, `train_data`, `test_data`, `train_data_info`, `test_data_info`). This is backward-compatible — `expanduser()` is a no-op on any path that doesn't start with `~`.

## Verification

- Live before/after repro: a real file at an expanded `~` path raises `ValueError` on `main`, resolves and loads successfully with the fix.
- No new pylint warnings introduced.


diegodevxd:
There is a check/consume inconsistency in this patch. `is_local_file(~/â€¦yaml)` becomes true because it expands the tilde, but `yaml2dict(~/â€¦yaml)` still executes `open(url, rb)` on the raw string. I reproduced this using a real YAML file:

```text
is_local_file=True
FileNotFoundError: [Errno 2] No such file or directory: '~/...yaml'
```

This affects more than datasets: `core/cmd/benchmarking.py`, `BenchmarkingJob._parse_testenv_config()`, and `TestCaseController._parse_algorithms_config()` all check with `is_local_file()` and then call `yaml2dict()` using the original value.

Please centralize supported local-path normalization and apply it before both predicates and consumers (`yaml2dict`, `py2dict`, `load_module`, plus dataset fields). Add tests for a top-level tilde YAML and nested testenv/algorithm URLs; otherwise this patch accepts a path that the next Core operation cannot load.

akshita317:
`os.path.expanduser` on the path fields is a sensible portability fix, and threading it through `is_local_file` / `is_local_dir` in `core/common/utils.py` as well as `_parse_config` is the right pairing — a `~` path would otherwise fail the local-file check before it ever reached the loader.

One finding on the membership of `_PATH_FIELDS`.

The tuple includes `train_url` and `test_url`, but those two are not input paths — Core writes them. `core/testenvmanager/dataset/dataset.py:158`:

```python
self.train_url = self._process_index_file(self.train_index)
```

`process_dataset()` assigns both from `train_index` / `train_data` / `train_data_info`, so any user-supplied value is overwritten before use, and `expanduser` on it has no lasting effect. The other six entries (`train_index`, `test_index`, `train_data`, `test_data`, `train_data_info`, `test_data_info`) are genuine inputs and are correctly included.

Dropping the two output fields would make the tuple express the intended invariant exactly.

**Compatibility note:** PR #758 also modifies `_parse_config`, adding a `_DEPRECATED_FIELDS` check with a `continue` for these same two fields. The branches merge cleanly, but the combined method reads:

```python
if attr in self._DEPRECATED_FIELDS:   # {train_url, test_url}
    LOGGER.warning(...)
    continue                           # <-- exits before the expansion below
if attr in self.__dict__:
    if attr in self._PATH_FIELDS and isinstance(value, str):
        value = os.path.expanduser(value)
```

so those two `_PATH_FIELDS` entries become dead code with no conflict marker to signal it. Since #758 treats them as deprecated inputs and this PR treats them as live input paths, it would be worth agreeing one model before either merges.

diegodevxd:
Executed follow-up on this PR head `1c35f2c6ba1e7806183346411329ebfb9e33a467` with a real YAML placed under HOME:

```text
RAW=~/contract.yaml
EXPANDED=C:\\Users\\...\\contract.yaml
PREDICATE=True
CONSUMER_ERROR=FileNotFoundError
```

`is_local_file()` expands `~`, but `yaml2dict()` opens the raw string. A local red/green regression test passes after the minimal consumer change `open(os.path.expanduser(url), ...)`; relative, `./relative`, and absolute paths remain valid, while nonexistent paths remain nonexistent.

Please apply one shared normalisation contract to filesystem consumers (`yaml2dict`, `py2dict`, and `load_module`) rather than adding caller-local workarounds. This does not require changing the existing repo/CWD-relative semantics.

Prachi194agrawal:
**Setup**

```
$ git fetch origin pull/789/head:pr-789 && git checkout pr-789
$ git log -1 --format="%H %s"
1c35f2c6ba1e7806183346411329ebfb9e33a467 fix(core): expand "~" in dataset/file path checks instead of silently rejecting them
```

`pandas` and `pyyaml` need to be installed first -`dataset.py:20` imports
pandas unconditionally at module load, so without it you get an import-time
`ModuleNotFoundError` before any of this PR's code runs. I also stubbed
`sys.modules["sedna"]` / `sys.modules["sedna.datasources"]` to isolate the
import; the PyPI package named `sedna` is an unrelated project and doesn't
provide `sedna.datasources`. Neither affects the code path under test.

**Result**

```
target after expansion: /home/prachi/nonexistent_test_file_xyz.txt | exists on disk: False
RESULT: ACCEPTED; train_index = /home/prachi/nonexistent_test_file_xyz.txt
```


**What this confirms.** The fix is real and the gap it closes is real before
this PR, `core/` had zero `expanduser` / `normpath` / `realpath` calls
(`grep -rn "expanduser\|normpath\|realpath" core/` returns 0 for each on
`main`), so a `testenv.yaml` pointing at `~/data/index.txt` was rejected even
when the file genuinely existed. `os.path.isfile('~/data/x')` is False on a
literal tilde and `os.path.isabs('~/data/x')` is also False, so the old code
hit the raise. That's a straightforward false-rejection bug and this fixes it.

**One side effect worth a line in the description.** The expanded path is
then checked by `_check_dataset_url` (`dataset.py:78-79`), whose accept
condition is `exists OR is_absolute`:

```python
if not utils.is_local_file(url) and not os.path.isabs(url):
    raise ValueError(...)
```

Once `~/foo` expands to `/home/user/foo`, it satisfies `is_absolute` whether
or not it exists  which is what the run above shows. So a *non-existent*
`~/...` path is now accepted, where previously it was rejected. The old
rejection was accidental rather than principled, so I don't think this is a
regression to fix in this PR. But someone reading the title could reasonably
conclude it closes existence validation for tilde paths, and it doesn't.

**One question on coverage.** From the diff, expansion is applied to a fixed
list of named path fields. Is that list intended to be exhaustive? If a path
field exists outside it, `~` would work in some `testenv.yaml` keys and not
others, which is harder to debug than uniform non-support because the failure
is silent and field-dependent.

**Recommendation: minor revision**  the code is a safe correctness
improvement and I'd be glad to see it land. Just a note in the description
that it doesn't close the non-existence-tolerant check in
`_check_dataset_url`, so it isn't mistaken for a full fix of #515 / #604.

Environment: Ubuntu 24.04.3, Python 3.12.3, base commit `37a9c60`, ⟦25-08-26⟧.

Nikhil-VS1811:
Hi, I reviewed this PR for the LFX pre-test. The path expansion is clean, but be aware it currently conflicts with the continue logic introduced in PR #758 for deprecated fields. Full details in my review:  https://github.com/kubeedge/ianvs/discussions/937

barakahabimbola (commented):
There’s a check/consume mismatch in this patch.

`is_local_file()` now expands `~` before checking the path, but `yaml2dict()` still calls `open(url, "rb")` with the original value. For example, `~/config.yaml` can pass `is_local_file()` after this change, while `yaml2dict()` still tries to open the literal `~/config.yaml` and raises `FileNotFoundError`.

Could the path normalization be applied before the filesystem consumer as well? A regression test with a real YAML file under `~` would verify that the expanded path is actually usable end-to-end.

Liyuchen0118 (commented):
**Recommendation**

Major revision.

**Tested head**

`1c35f2c6ba1e7806183346411329ebfb9e33a467`

**Validation method**

Focused path-consumer assertion matrix run against the tested PR head. Test type: source-level consumer behavior; not an Ianvs end-to-end runtime test. Harness exit: `0`.

**Observed result**

- `~/x.yaml`: predicate `True`; `yaml2dict()` raises `FileNotFoundError`.
- `~/x.py`: predicate `True`; `py2dict()` raises `ModuleNotFoundError`.
- `load_module("~/x.py")` raises `RuntimeError`.
- Ordinary relative and absolute YAML paths pass unchanged.
- A missing home path returns `False`.
- `~otheruser/x.yaml` returns `False`.
- File-as-directory and directory-as-file checks return `False`.
- A symlink YAML path passes and loads successfully.
- A nested home-relative value remains raw before the Dataset boundary.

Harness exit `0` means the expected defect matrix was successfully reproduced.

**Impact**

Validation and consumption interpret the same value differently. No active home-relative configuration was found in the selected Examples or current repository, so this is preventive Core hardening. It should not block Example-local repairs.

**Requested change**

1. Normalize at shared consumer boundaries, including `yaml2dict`, `py2dict`, `load_module`, and Dataset input processing.
2. Add tests for home-relative YAML/Python, ordinary paths, missing paths, `~otheruser`, nested URL values, symlinks, and file/directory mismatches.
3. Confirm that output-only Dataset fields are not rejected as missing input files.

**Prior discussion**

- https://github.com/kubeedge/ianvs/pull/789#issuecomment-5381275534
- https://github.com/kubeedge/ianvs/pull/789#issuecomment-5381813715
- https://github.com/kubeedge/ianvs/pull/789#issuecomment-5382578199

**Additional evidence**

Relevant consumers are in `core/common/utils.py:61-100` and Dataset handling in `core/testenvmanager/dataset/dataset.py:75-140`.

PR #790, opened 2026-08-13
Fix: examples/llm_simple_qa: broken config paths and CUDA-only model loading

### LINKED ISSUE #331: examples/llm_simple_qa` example not runnable (broken config paths and hard-coded local paths)
## Description

The `examples/llm_simple_qa` example currently fails to run because its README and YAML configurations reference paths that do not exist in the repository and include hard-coded absolute paths tied to the original author’s local environment.

---

## Steps to Reproduce

1. Clone the repository and install `ianvs`.
2. Run the command shown in the README:

   ```bash
   ianvs -f examples/llm/singletask_learning_bench/simple_qa/benchmarkingjob.yaml
   ```
3. Alternatively, try running the example YAML directly:

   ```bash
   ianvs -f examples/llm_simple_qa/benchmarkingjob.yaml
   ```

---

## Actual Result

* The README command points to a YAML path that does not exist.
* `examples/llm_simple_qa/benchmarkingjob.yaml` references:

  * `./examples/llm/singletask_learning_bench/simple_qa/testenv/testenv.yaml`
  * `./examples/llm/singletask_learning_bench/simple_qa/testalgorithms/...`
    These paths are not present in the repository, causing config loading to fail.
* Additional portability issues:

  * Hard-coded absolute paths such as `/home/icyfeather/...` for workspace, datasets, and models
  * `basemodel.py` assumes `device="cuda"` and a local model path, preventing execution on CPU-only systems

---

## Expected Result

The example should run using repository-correct relative paths and without requiring author-specific filesystem paths or a CUDA-only setup.

---

## Suggested Fix

* Update all YAML and README references from
  `./examples/llm/singletask_learning_bench/simple_qa/...`
  to
  `./examples/llm_simple_qa/...`
* Replace absolute workspace, dataset, and model paths with:

  * relative paths, or
  * environment variable overrides with safe defaults
* Update `basemodel.py` to support CPU fallback when CUDA is not available

---

## Environment

* OS: macOS / Linux
* Branch: `main`
* Example: `examples/llm_simple_qa`

I need your opinion on that @MooreZheng

a-reda-ok:
**What type of PR is this?**
/kind bug
/kind cleanup

This PR makes the examples/llm_simple_qa example runnable from the repository by removing author-local absolute paths and adding CPU-safe model loading. Without these changes the example references non-existent paths (examples/llm/singletask_learning_bench/...) and forces CUDA and an author-specific model directory, preventing contributors and CI from running the example on typical machines.

Main changes:

- Replace absolute/incorrect paths with repo-relative paths in README.md, benchmarkingjob.yaml, testenv/testenv.yaml, and gen_algorithm.yaml.
- Make basemodel.py detect CUDA and fall back to CPU; allow overriding model location via QWEN_MODEL_PATH environment variable.
- Update op_eval.py to read model path from QWEN_MODEL_PATH and set run_cfg num_gpus based on availability.

Fixes #331


akshhkaushik (commented):
### Request changes — the PR head fails `py_compile`

The portability intent is useful, but the edited BaseModel cannot be imported. The new device assignment contains a literal leading `+`:

```python
+device = "cuda" if torch.cuda.is_available() else "cpu"
```

Compiling the exact PR-head file produces:

```text
File "basemodel.py", line 30
    +device = "cuda" if torch.cuda.is_available() else "cpu"
SyntaxError: cannot assign to expression here
```

This blocks every CPU and CUDA path before model loading, so it is the first merge gate. After removing it, the duplicate `if device == "cuda"`/`else` branches can be reduced to one `self.model.to(device)` call, and the CPU behavior should be tested with a stubbed `from_pretrained` to avoid downloading a model.

Recommendation: fix the syntax and add `py_compile` plus a CPU import/constructor smoke test. Rebase as well: GitHub currently reports the PR as conflicting.

Verification boundary: compiled the exact PR-head source; did not download Qwen or run OpenCompass.

Hiteshsai007:
The hardcoded `/home/icyfeather/...` workspace/testenv/algorithm paths and the stale
`examples/llm/...` references absolutely need fixing, and the CUDA-only model load is a real bug —
those hunks are welcome.

I ran this example end-to-end on current `main` with the real model (CPU), so two of the changes
here look like they'd move it **backwards** versus what already runs on main:

1. **Dataset paths.** The PR points `testenv` `train_data`/`test_data` at
   `./examples/llm_simple_qa/train_data/data.jsonl`, but `scripts/02_prepare_dataset.py` writes to
   `./dataset/llm_simple_qa/train_data/data.jsonl` — and current `main`'s testenv already references
   that `./dataset/...` location. With this PR's path, the data the prep script writes isn't where
   the testenv reads, so a clean run fails at dataset loading. Recommend keeping the `./dataset/
   llm_simple_qa/...` paths that are on main.

2. **Model loading.** The PR loads `./models/Qwen2-0.5B-Instruct` (via `QWEN_MODEL_PATH`, default
   not shipped), whereas main uses the HF id `Qwen/Qwen2.5-0.5B-Instruct`, which downloads
   automatically. The PR's default path doesn't exist in the repo and isn't fetched, so a fresh
   `ianvs -f` fails at `from_pretrained` unless the user manually places an older Qwen2 checkpoint;
   it also drops the `mps` branch (this PR is cuda→cpu only; main is cuda→mps→cpu). If offline/local
   loading is wanted, keep the HF id as the default and only fall back to a local path when
   `QWEN_MODEL_PATH` is explicitly set.

Net: the path/typo/absolute-path fixes are good; the dataset-path and model-id hunks conflict with
the already-working main version (this branch looks to be behind the example that now runs) and
should be reconciled so they don't regress it.

a-reda-ok:
Closing this PR as the underlying issue was addressed and resolved in #771. Thanks for the feedback on this!

ALbahrani5:
### Technical Review for LFX Term 3 Pre-test

**Status:** Major Revision Required (Blocking)  
**Summary:** The intent to fix absolute paths and make model loading CPU-safe is welcome. However, the current PR head contains a critical syntax error that prevents the module from being imported at all.

---

### 🔴 Critical Blocking Issue

The following line in `basemodel.py` has a leading `+` (a unified diff artifact) which is invalid Python syntax:

```python
# Incorrect (SyntaxError: invalid syntax)
+device = "cuda" if torch.cuda.is_available() else "cpu"

# Correct
device = "cuda" if torch.cuda.is_available() else "cpu"

PR #791, opened 2026-08-14
fix: add missing colorlog and PyYAML dependencies to requirements.txt

### LINKED ISSUE #413: [Dependencies] Add missing colorlog and PyYAML to requirements.txt

blackdragoon26:
**What type of PR is this?**

/kind bug

This PR fixes the broken fresh-install experience by adding two missing runtime dependencies to the root `requirements.txt`:

| Package | Requirement | Import location | Import statement |
| --- | --- | --- | --- |
| `colorlog` | `colorlog>=4.7.2 # MIT` | `core/common/log.py` | `import colorlog` |
| `PyYAML` | `PyYAML>=6.0 # MIT` | `core/common/utils.py` | `import yaml` |

Both packages are imported unconditionally at module level. Since they were absent from `requirements.txt`, following the installation guide could leave the `ianvs` CLI failing with `ModuleNotFoundError` unless another package happened to install them.

The dependency list remains alphabetized, the existing `prettytable~=2.5.0 # BSD` license annotation is preserved, and the two new dependencies include their MIT annotations.

Fixes #413

- Verified that both `import colorlog` and `import yaml` succeed after installing the declared dependencies.
- `git diff --check` passes.
- No runtime code is modified; this is a dependency-metadata fix only.

```release-note
NONE
```


akshita317:
The underlying problem is real and reproducible: on a clean virtualenv with only the declared `requirements.txt` installed, importing core fails, because `core/common/log.py` imports `colorlog` and `core/common/utils.py` imports `yaml`, neither of which is declared.

Two findings.

**1. The `prettytable` license annotation is dropped.**

```diff
-prettytable~=2.5.0 # BSD
+prettytable~=2.5.0
```

`requirements.txt` is the only per-dependency record of license provenance in this repository, and license scanning is one of the stated outcomes of #230. The removal looks incidental to the reordering rather than intentional. It would also be worth annotating the two new lines in the same style (`colorlog # MIT`, `PyYAML # MIT`) so the convention stays complete.

**2. Several open PRs add the same dependency with different specifiers**, and they all edit adjacent lines of this 7-line file:

| PR | specifier |
|---|---|
| #674 | `colorlog` |
| #468 | `colorlog>=6.0.0`, `pyyaml>=6.0.0` |
| this PR | `colorlog>=4.7.2`, `PyYAML>=6.0` |
| #793 | `colorlog # MIT` |
| #682 | `colorlog`, `PyYAML` |

PR #468 and this PR specify incompatible floors for `colorlog` (`>=6.0.0` vs `>=4.7.2`), and `pyyaml` vs `PyYAML` differ in casing across PRs. At most one of these can merge without conflict, so it may be worth consolidating into a single PR with an agreed floor.

Related: **#682** adds `install_requires=_infos.basic_dependencies` to `setup.py`. Without it, `requirements.txt` only affects from-source installs — `pip install ianvs` still resolves no dependencies at all, so this fix and #682 are complementary and probably want to land together.

blackdragoon26:
@akshita317 
I am not sure, what you are trying to convey with this message.
Do review the AI usage Policy for the repository.

> The underlying problem is real and reproducible: on a clean virtualenv with only the declared `requirements.txt` installed, importing core fails, because `core/common/log.py` imports `colorlog` and `core/common/utils.py` imports `yaml`, neither of which is declared.
> 
> Two findings.
> 
> **1. The `prettytable` license annotation is dropped.**
> 
> ```diff
> -prettytable~=2.5.0 # BSD
> +prettytable~=2.5.0
> ```
> 
> `requirements.txt` is the only per-dependency record of license provenance in this repository, and license scanning is one of the stated outcomes of #230. The removal looks incidental to the reordering rather than intentional. It would also be worth annotating the two new lines in the same style (`colorlog # MIT`, `PyYAML # MIT`) so the convention stays complete.
> 
> **2. Several open PRs add the same dependency with different specifiers**, and they all edit adjacent lines of this 7-line file:
> 
> PR	specifier
> #674	`colorlog`
> #468	`colorlog>=6.0.0`, `pyyaml>=6.0.0`
> this PR	`colorlog>=4.7.2`, `PyYAML>=6.0`
> #793	`colorlog # MIT`
> #682	`colorlog`, `PyYAML`
> PR #468 and this PR specify incompatible floors for `colorlog` (`>=6.0.0` vs `>=4.7.2`), and `pyyaml` vs `PyYAML` differ in casing across PRs. At most one of these can merge without conflict, so it may be worth consolidating into a single PR with an agreed floor.
> 
> Related: **#682** adds `install_requires=_infos.basic_dependencies` to `setup.py`. Without it, `requirements.txt` only affects from-source installs — `pip install ianvs` still resolves no dependencies at all, so this fix and #682 are complementary and probably want to land together.

akshita317:
Fair enough — that was way too long for a two-line PR, sorry. Let me cut it to the one thing that actually matters.

The reordering dropped the `# BSD` comment off the prettytable line:

```
-prettytable~=2.5.0 # BSD
+prettytable~=2.5.0
```

I don't think that was deliberate. It's the only place this repo records a license per dependency, so it'd be worth putting back, and maybe adding `# MIT` to the two new lines while you're in there. That's the only change I'd suggest.

Everything else looks right. I checked the failure it fixes — clean venv, only the declared requirements installed:

```
File "core/common/log.py", line 18, in <module>
    import colorlog
```

So the fix is definitely needed.

The other stuff I mentioned wasn't a request, just a heads-up: #674, #468, #793 and #682 all add colorlog to the same few lines of this file, so whichever one merges first will conflict with the others. #468 uses `>=6.0.0` where yours uses `>=4.7.2`. Seemed worth someone knowing before four people end up rebasing.

Also small thing — your PR description says `colorlog==4.7.2` and `PyYAML==6.0`, but the diff has `>=4.7.2` and `>=6.0`. I think the text predates your force-push.

On the policy point — I went looking for contribution rules on this in CONTRIBUTING.md and over in kubeedge/community and couldn't find anything covering it. If there's a policy I've missed, link me and I'll follow it.

kavaljeetsingh-dev:
Tested this directly — it's complete, both `colorlog` and `PyYAML` are covered:

```
$ python3 -c "import core.common.log, core.common.utils"
# both import cleanly
```

One small thing worth a cleanup pass before merge: the new `requirements.txt` line has a Windows CRLF line ending in a file that's otherwise LF-only, which is a minor but easy fix.

Nikhil-VS1811:
Hi, I tested this PR for the LFX pre-test and confirmed it successfully resolves the root ModuleNotFoundErrors for the CLI. Full review here:  https://github.com/kubeedge/ianvs/discussions/937

AmanBasu20:
**Recommendation: Minor Revision.**

**Analysis:** This is the most architecturally sound fix. It places `colorlog==4.7.2` and `PyYAML==6.0` in the root `requirements.txt` and sorts them alphabetically. This resolves the root problem for all examples simultaneously.

**Regression Risk / Revision Required:** While this fixes the `pip` installation, it fails to update the `setup.py` `install_requires` mapping. Therefore, if a user runs `pip install .` instead of `pip install -r requirements.txt`, the installation will still silently fail to pull `colorlog`.

AmanBasu20:
This PR places `colorlog==4.7.2` and `PyYAML==6.0` in root `requirements.txt`, sorted alphabetically, resolving the Task 1 root problem for all examples simultaneously rather than one — architecturally the most sound of the three dependency PRs reviewed here.

**Regression risk / revision required:** It does not update `setup.py`'s `install_requires` mapping, so `pip install .` (as opposed to `pip install -r requirements.txt`) will still silently omit `colorlog`. It also drops the `# BSD` license annotation on the `prettytable` line during reordering — `requirements.txt` is this repository's only per-dependency license record, and license scanning is a stated outcome of #230, so this is a real (if minor) regression, not a nitpick.

**Recommendation: Minor Revision** — restore the `# BSD` annotation (and annotate the two new lines, e.g. `colorlog # MIT`, `PyYAML # MIT`), and pair this PR with a follow-up wiring `install_requires=_infos.basic_dependencies` into `setup.py` so `pip install .` and `pip install -r requirements.txt` behave identically.

**Note on version floor:** #468 specifies `colorlog>=6.0.0` where this PR specifies `>=4.7.2`. Both PRs currently edit adjacent lines of the same file — recommend consolidating on a single agreed floor before merge to avoid downstream resolver inconsistency (see comment on #468 for the specific risk this creates).

PR #792, opened 2026-08-14
fix(bdd): correct curb detection model metric

Aryansingh-ai:
## What type of PR is this?

/kind bug

## What this PR does / why we need it

Fixes a broken metric reference in the BDD curb-detection example.

The `testenv.yaml` configuration referenced an `accuracy.py` file that does not
exist in the example. The example provides `map.py`, which registers the
`map` metric through `ClassFactory`.

This PR updates the model evaluation configuration to reference the existing
`map.py` implementation.

### Changes

- Changed model metric from `accuracy` to `map`
- Updated the metric URL to the existing `map.py`
- Verified `map.py` registers the `map` metric
- Verified the referenced file exists
- Verified `git diff --check` passes

## Which issue(s) does this PR fix?
NA


akshhkaushik (commented):
### Blocked — the selected metric sits behind a broken `multi-inference` branch

Changing curb detection's model metric from accuracy to its existing mAP function is locally reasonable. The same config retains `mode: "multi-inference"`, and current `LifelongLearning.run()` has no executable body for that mode.

I invoked the real control path on this PR head with the configured mode:

```text
UnboundLocalError: cannot access local variable 'test_res' where it is not associated with a value
```

Recommendation: first decide whether `multi-inference` is supported. If supported, implement and test the branch with BDD's metric result shape. If unsupported, migrate the BDD config to a supported mode and document the semantic change before evaluating this metric edit.

Verification boundary: executed the real paradigm control flow with external training/inference methods unreachable on this branch; did not load the BDD dataset or detector.

Submission context: [submission discussion](https://github.com/kubeedge/ianvs/discussions/924) ([Bonus cross-target finding](https://github.com/kubeedge/ianvs/discussions/924#discussioncomment-18147196)).

Issue #742, opened 2026-08-10
[PRE-TASK] Restoring examples/yaoba: An Orphaned Paradigm and a Chain of Infrastructure Gaps


hibaa23:
[PRE-TASK] Restoring `examples/yaoba`: An Orphaned Paradigm and a Chain of Infrastructure Gaps

## Background

I'm applying for the LFX Mentorship 2026 Term 3 "Comprehensive Example Restoration for KubeEdge Ianvs: Phase IV." To evaluate where I could contribute, I picked `examples/yaoba` — the oldest untouched example in the repository (last modified 2023-06-16, predating all prior restoration phases) — and attempted a full local setup and run on Windows, since I noticed the project's installation guide currently only documents a Linux workflow.

What started as an environment setup turned into a chain of failures that traces back to a single root cause: **the `singletasklearning_acboost` paradigm used by `yaoba/singletask_learning_boost` is not registered anywhere in Ianvs's core dispatch logic.** The example likely cannot run on the current `main` branch, regardless of environment. Below is the full investigation trail, the evidence for each issue, and a proposed restoration plan.

## Investigation Environment

- OS: Windows 11, PowerShell
- Python 3.9.25 (isolated conda environment, matching the version pinned in the example's README)
- torch 1.13.0+cpu, mmcv-full 1.7.1, mmdet 2.28.2 (versions pinned by the example)

### 1. Root cause: orphaned paradigm (blocking, affects the whole example)

`examples/yaoba/singletask_learning_boost/testalgorithms/algorithm.yaml` declares:
```yaml
paradigm_type: "singletasklearning_acboost"
```

This value does not exist in `core/common/constant.py`'s `ParadigmType` enum, which only defines:
```
singletasklearning, incrementallearning, multiedgeinference,
lifelonglearning, federatedlearning,
federatedclassincrementallearning, jointinference
```

Running the benchmark confirms this fails validation:
```
ValueError: not support paradigm(singletasklearning_acboost).the following paradigms
can be selected: ['singletasklearning', 'incrementallearning', 'multiedgeinference',
'lifelonglearning', 'federatedlearning', 'federatedclassincrementallearning', 'jointinference']
```

[...]

PR #793, opened 2026-08-14
Fix: add missing colorlog dependency to requirements.txtAdd missing colorlog dependency to requirements.txt

hibaa23:
**What type of PR is this?**
/kind bug

`core/common/log.py` imports `colorlog` unconditionally, but it is not listed in the root `requirements.txt`. As a result, a fresh install following the official install guide (`docs/guides/how-to-install-ianvs.md`) fails immediately when running any benchmark:
```
File "benchmarking.py", line 20, in <module>
from core.common.log import LOGGER
File "core/common/log.py", line 18, in <module>
import colorlog
```
This affects every example in the repository, not just one, since `core.common.log` is imported at the very top of `benchmarking.py`'s import chain. This PR adds `colorlog` to `requirements.txt` so a clean install works out of the box.

Found while investigating `examples/yaoba` for LFX Mentorship 2026 Term 3 pre-task work (see #742 for the full investigation).

Related to #742


kavaljeetsingh-dev:
Tested this directly. Same gap as the other colorlog-only PRs on this issue: `core/common/utils.py` also has an undeclared `yaml` import that this doesn't cover:

```
$ python3 -c "import core.common.utils"
ModuleNotFoundError: No module named 'yaml'
```

#791 and #468 (also open, same issue) both include PyYAML as well.

hibaa23:
Good catch — added PyYAML alongside colorlog, same gap (core/common/utils.py imports yaml directly). Pushed.

sheikhayaan:
Worth confirming this is a hard blocker and not a tidy-up, because that changes how urgently it wants merging. One scope gap below that I think deserves a line in the description.

The blocker part first. benchmarking.py:20, the first import of the entrypoint, is from core.common.log import LOGGER, and core/common/log.py:18 is import colorlog. requirements.txt on main contains only prettytable~=2.5.0, scikit-learn, numpy, pandas, tqdm, matplotlib and onnx. On a clean environment:

$ python3 -c "import benchmarking"

Ianvs cannot start at all after a documented install. That's a total cold-start failure, not a degraded path, and it seems worth saying explicitly because a one-line requirements change reads like housekeeping.

The PyYAML line added after @kavaljeetsingh-dev's review closes the second import gap, so as of 1fb2419 both undeclared runtime imports on the pip install -r path are covered. What follows is a third gap on a different install path, which is why I think it belongs in the description instead of a follow-up.

A scope gap that neither this PR nor #674 mentions. setup.py computes a basic_dependencies property from requirements.txt but never passes install_requires to setup(). The keyword doesn't appear in the call at all. So pip install ianvs installs no dependencies whatsoever. This PR fixes the pip install -r path and leaves the packaged path broken. Worth one line in the description so the remaining half doesn't get lost when this merges and the issue closes.

Recommendation: accepted as it is. Minimal, correct, and it matches the file's existing style by carrying a licence comment (colorlog # MIT) the way the prettytable line does. I'd merge this one in preference to #674, and I've put the reasoning on that PR instead of here.

TanishChahal:
Reviewing the dependency routing here for the LFX pre-test. The recent observation in this thread about `setup.py` dropping `install_requires` is spot on, but it highlights a deeper structural flaw: relying on `requirements.txt` to dynamically feed `setup.py` is a brittle, deprecated anti-pattern. This manual syncing is exactly what caused the `colorlog` and `PyYAML` drift in the first place. I’ve detailed a migration path to declarative PEP-517 `pyproject.toml` isolation in my root-cause analysis here: https://github.com/kubeedge/ianvs/discussions/911.

ALbahrani5:
Same technical fix as #674 (colorlog + yaml declaration), functionally redundant. Per the existing review thread on #674, that PR has priority on age and explicit issue linkage to #673. Recommend closing this as a duplicate unless #674 stalls. Full analysis: [https://github.com/kubeedge/ianvs/discussions/971]

ALbahrani5:
This adds fastapi and pydantic to the shared root requirements.txt with no version bounds. Open Issue #666 documents that an unbounded/downgraded pydantic breaks the ianvs command entirely (ImportError: cannot import name 'IncEx'). Recommend pinning lower bounds (e.g. pydantic>=2.9) before merge, since this changes a Core-shared manifest affecting every example's install, not just curb-detection. Full analysis: [https://github.com/kubeedge/ianvs/discussions/971]

PR #794, opened 2026-08-14
fix(Cloud_Robotics): vendor missing RFNet tree and fix config paths (…

### LINKED ISSUE #719: bug: Cloud_Robotics/Semantic_Segmentation is missing its vendored RFNet/ source tree, unlike 4 sibling examples
**What happened**:

`examples/Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation/testalgorithms/basemodel.py` does:

```python
from RFNet.dataloaders import make_data_loader
from RFNet.dataloaders import custom_transforms as tr
from RFNet.utils.lr_scheduler import LR_Scheduler
from RFNet.train import Trainer
from RFNet.eval import Validator
from RFNet.utils.saver import Saver
from RFNet.eval import load_my_state_dict
import RFNet.train_config as train_cfgs
import RFNet.eval_config as valid_cfgs
```

`RFNet` is not a PyPI package - it is Ianvs' own bundled implementation of the RFNet segmentation architecture. **Four other examples in this repository vendor a full local copy of it** at a `testalgorithms/rfnet/RFNet/` subdirectory:

- `examples/robot/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet/`
- `examples/cityscapes/singletask_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet/`
- `examples/cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet/`
- `examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testalgorithms/rfnet/RFNet/`
- (a fifth, `cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/`, has the `rfnet/` folder too)

`Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation/testalgorithms/` contains only `algorithm.yaml` and `basemodel.py` - **no `rfnet/` subdirectory at all**. The import fails immediately:

```
ModuleNotFoundError: No module named 'RFNet'
```

`basemodel.py` successfully imports the RFNet architecture code, the same way it does in the four sibling examples listed above.

```python
import sys
sys.path.insert(0, "examples/Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation/testalgorithms")
import basemodel
```
```
ModuleNotFoundError: No module named 'RFNet'
```

Also confirms there's no `requirements.txt` for this sub-example at all: reproducing the above first requires manually installing `torch` and `torchvision`, neither of which are declared anywhere for this example.

This is the deepest of several compounding bugs in this sub-example - even after its config-path bugs are fixed (filed separately, `benchmarkingjob.yaml`/`testenv.yaml`/`algorithm.yaml` all reference several different misspelled variants of this directory's own path), this import is the true root blocker.

This doesn't look like a documentation gap ("please `pip install RFNet`") - there's no such package to install. It looks like the vendored `rfnet/RFNet/` source tree that every sibling RFNet-based example ships was simply never copied into this one, most likely lost during a rename/refactor (consistent with the config files elsewhere in this same sub-example referencing at least six different misspelled variants of its own directory name, suggesting the example was restructured at some point without every file being updated to match).

The fix is most likely: pick the closest-matching vendored copy from one of the four sibling examples above, diff it against the others to confirm they haven't diverged in ways that matter for this dataset, and copy it into `Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation/testalgorithms/rfnet/RFNet/`, then verify `basemodel.py`'s hyperparameter wiring (`momentum`, `learning_rate`, `epochs`) matches what the restored `RFNet.train_config`/`eval_config` expect.

Environment:
- Ianvs version: `36ec008` (main)
- Python version: 3.10.19
- OS: Ubuntu 24.04.3 LTS, Linux 7.0.0-28-generic x86_64

abhishekkamble12:
### What type of PR is this?
/kind bug

Fixes `ModuleNotFoundError: No module named 'RFNet'` in `Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation`.

1. **Vendors the complete RFNet source tree**: Adds the missing `RFNet` package (`dataloaders`, `models`, `utils`, `train.py`, `eval.py`) under `testalgorithms/rfnet/RFNet/` matching sibling RFNet-based benchmarks.
2. **Modernizes `basemodel.py`**: Updates `testalgorithms/rfnet/basemodel.py` to use `from RFNet.utils.args import TrainArgs, ValArgs` instead of legacy non-existent `train_config` / `eval_config`.
3. **Fixes broken configuration paths**: Resolves misspelled/inconsistent paths across `algorithm.yaml`, `benchmarkingjob.yaml`, `testenv.yaml`, and `Readme.md`.

Fixes #719

Tested imports and verified directory structure matches sibling benchmarks (`robot`, `cityscapes-synthia`).

### Does this PR introduce a user-facing change?
```release-note
Fix missing RFNet source tree and broken configuration paths in Cloud_Robotics Semantic_Segmentation benchmark.

### CODE REVIEW

**Comment by sheikhayaan** [comment]
Flagging a structural point before line-by-line review, because I don't think it can be
resolved inside the diff.

This PR vendors a complete RFNet implementation into a new example directory. Four RFNet trees
already exist in `examples/`:

```
examples/cityscapes-synthia/lifelong_learning_bench/curb-detection/testalgorithms/rfnet/RFNet
examples/cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet
examples/cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/testalgorithms/rfnet/RFNet
examples/robot/lifelong_learning_bench/semantic-segmentation/testalgorithms/rfnet/RFNet
```

157 `.py` files between them. This would make five.

The part worth weighing isn't duplication as such. It's that the existing four **have already
diverged, silently.** Comparing by content across the four:

```
test.py                copies=4  distinct=1     identical
sedna_train.py         copies=4  distinct=1     identical
utils/lr_scheduler.py  copies=4  distinct=1     identical
utils/loss.py          copies=4  distinct=2
predict.py             copies=4  distinct=2
utils/args.py          copies=4  distinct=3
train.py               copies=4  distinct=4     all four differ
utils/metrics.py       copies=4  distinct=4     all four differ
```

`train.py` and `utils/metrics.py` exist in four mutually different versions, with no record
anywhere of which differences are deliberate example-specific tuning and which are drift from
an unsynchronised copy. That's the state a fifth copy would be joining.

Two consequences. Any model-level fix has to be applied five times, and whoever applies it has
to judge, per file, whether a difference between copies is intentional. With `metrics.py`
already at four variants, that judgement isn't recoverable from the repository. The second is
that this is part of why config path drift spread as far as it did: I measured 284 broken
references across 18 example units, and the two largest contributors are `cityscapes-synthia`
at 135 and `robot` at 74, which are precisely the trees carrying these copies. Duplicated
subtrees multiply the surface a rename has to be chased through.

To be fair to the author, vendoring is the established pattern here and there's no shared
location for model code to use instead, so "don't vendor" isn't actionable on its own. The
`+4366/-170` size is a consequence of missing structure, not of anything done wrong.

Suggestions:

1. **Preferred:** land a shared location first, either a top-level `models/rfnet/` or an
   importable package, and have this example consume it. That's larger than this PR and
   probably needs a maintainer decision, so it may not be reasonable to block on.
2. **Minimum, if it merges as-is:** record provenance in the new tree: which existing copy it
   came from, at what commit, and what was intentionally changed. A short `PROVENANCE.md`
   beside the vendored code would make the fifth copy the first one that's diffable against its
   origin, and stop the divergence widening any further.
3. Consider a tracking issue for consolidating the existing four, using the comparison above as
   the starting inventory.

Mechanically: at 41 files this also needs an entry in
`.github/workflows/validator/data/example_inventory.yaml`, or it inherits the default
`unvalidated` status and CI won't check it at all, per #836.

**The objection with real force, and how far it goes.** The strongest reading against me is
that these four trees aren't duplicates at all, but four legitimately different models, forked
on purpose for four different benchmarks (curb detection versus semantic segmentation,
cityscapes-synthia versus robot), and what I'm calling silent divergence is per-example tuning
working as intended. Of my four reviews this is the one where I think the counter-argument
genuinely bites, and the file-level data supports it partially: if the trees were meant to be
identical, `test.py`, `sedna_train.py` and `utils/lr_scheduler.py` wouldn't be byte-identical
across all four while `train.py` differs in all four. Divergence concentrated in the training
entry point is exactly the signature of deliberate per-task tuning.

Where the evidence stops supporting it is `utils/metrics.py`. Metric computation is
task-independent. mIoU is mIoU whether the classes are curbs or road surfaces, so four mutually
different versions of it can't be explained as task tuning, and that's the file I'd point at as
unintended drift. So I've narrowed the claim: I'm not arguing the four trees should be one
tree, and I withdraw any implication that all the divergence is accidental. I'm arguing that
nobody can currently tell *which* differences are deliberate, because there's no record, and
that's a property of the repository, not a criticism of this author. It's also why my minimum
ask is provenance rather than consolidation. Provenance is what makes the deliberate/accidental
distinction recoverable, and it's cheap.

**Two interactions before this lands.** #788 is currently repairing 135 broken references
inside `cityscapes-synthia`, which is where three of the four existing RFNet trees live; adding
a fifth tree while that repair is in flight enlarges the surface #788's successor has to chase,
so I'd sequence this after #788 instead of beside it. And if #683 or #687 lands (both remove a
self-referential `ianvs` dependency from example requirement files), any new example
requirements added here should be written to the post-#683 convention, or this PR reintroduces
the defect those two just removed.

**Recommendation: major revision.** To be explicit about what is and isn't being asked: the
example itself looks like a genuine contribution and I'm not asking for it to be rewritten or
withdrawn, and I'm not blocking on the shared-model-location work in suggestion 1, which is
larger than this PR and needs a maintainer decision. What I'd want before merge is the minimum
from suggestion 2, a `PROVENANCE.md` naming the source copy, its commit and the intentional
changes, plus the `example_inventory.yaml` entry, without which CI won't check 41 new files at
all. Both are small. With those two, I'd move to **accepted as it is**.

asadulla-h:
Reviewing this as part of my LFX pre-test (full analysis in my pre-test Discussion, linked below).

This PR correctly resolves the missing RFNet tree for Cloud_Robotics (#719) using a portable relative-path pattern:

python3 -c "from mypath import Path; print(Path.db_root_dir('cityscapes'))"

Output:
./dataset/

Recommendation: accepted as it is. One note for context — the sibling examples this vendored tree is modeled on (robot, cityscapes-synthia) are currently attempting to merge byte-identical trees with the old broken absolute paths (see #297). This fix could be a good reference for cleaning those up too, since the pattern here is correct and portable.

Full analysis: https://github.com/kubeedge/ianvs/discussions/923

rajdhruvsingh:
Credit to @sheikhayaan for the structural flag here — four RFNet trees already exist in this repository (157 .py files between them), and her content comparison shows they've already diverged silently, with train.py and utils/metrics.py each existing in four mutually different versions. This PR would add a fifth copy to that set.

Two additions: first, the PR description doesn't currently include a license note for the vendored RFNet source — worth adding, since it's a full subtree import and maintainers reviewing dependency provenance will want that stated up front rather than having to check. Second, credit to @asadulla-h's review for noting that this PR's vendored copy already uses a portable relative-path pattern (verified: it resolves dataset paths without hardcoding a machine-specific location) — that pattern is worth preserving specifically if a shared-location consolidation of the existing four trees is ever pursued, since it's already correct where the others aren't.

asadulla-h:
Restoration path follow-up (full details in my pre-test Discussion, linked below).

Used your relative-path pattern directly to fix robot's mypath.py as part of my Task 4 restoration plan — applied it, then verified the fix with a real before/after check (total broken-path count dropped by exactly 7, no side effects). This pattern is doing double duty now: correct for Cloud_Robotics, and the reference implementation for restoring robot and cityscapes-synthia too.

Full path: https://github.com/kubeedge/ianvs/discussions/923

SurbhiD404:
Independently verified on this branch (2e36750):

1. File count correction: the vendored RFNet tree contains 34 total files (33 .py), not 157 as cited in earlier discussion.

2. mypath.py's portable relative-path pattern confirmed in code: Path.db_root_dir('cityscapes') returns './dataset/'.

3. Ran the basemodel.py import chain directly - resolves cleanly through numpy and PIL, only fails at torch (expected, no ML environment installed; not a code defect).

Recommendation: major revision - path fixes and portable pattern are solid, but vendoring needs PROVENANCE.md, an example_inventory.yaml entry, and should sequence after #788.

SurbhiD404:
Repair Boundary Finding: Two orthogonal problems, two boundaries. (1) Path repairs (Class O + A) - Example-local fix is correct; the portable relative-path pattern here is the reference implementation. (2) RFNet vendoring - separate boundary (Shared utility); self-verified the vendored tree is 34 files (not 157 as previously cited), still adds a copy without provenance. Scoped separately from the path-repair argument.

PR #795, opened 2026-08-14
Fix invalid YAML indentation in yaoba README documentation

hibaa23:
**What type of PR is this?**
/kind documentation

The `examples/yaoba/singletask_learning_boost` README contains two YAML configuration snippets with indentation errors that make them invalid YAML, which could mislead contributors copying these examples:

1. A stray ` type;` line inside the `testenv.yaml` example block, unrelated to any key.
2. Inconsistent indentation on the `resource_dir` entry in the `algorithm.yaml` example block (one extra leading space compared to sibling entries).

This PR corrects both snippets so they are valid, copy-pasteable YAML.

Found while investigating `examples/yaoba` for LFX Mentorship 2026 Term 3 pre-task work (see #742).

Related to #742


31groot:
### Request changes :  the repaired README still removes the `resource_dir` value

**Review depth: Analysis**

The PR correctly removes the stray `type;` line and fixes the indentation in the README's YAML examples. However, the resulting `algorithm.yaml` snippet is still not a complete copy-pasteable configuration.

The `resource_dir` entry was changed from:

```yaml
- resource_dir:
    values:
      - "examples/yaoba/singletask_learning_boost/resource"
```

to:

```yaml
- resource_dir:
```

This parses as:

```text
Parsed resource_dir entry: {'resource_dir': None}
resource_dir value: None
README CONFIG GAP REPRODUCED
```

This is more than a formatting issue. The README instructs users to configure `resource_dir`, while the corresponding shipped `algorithm.yaml` still defines it with a concrete `values` list. A contributor copying the README therefore gets a semantically incomplete configuration.

A parser expecting the usual structure:

```yaml
{"resource_dir": {"values": [...]}}
```

will instead receive:

```yaml
{"resource_dir": None}
```

**Recommendation:** restore the `values` block for `resource_dir` in the README snippet so that it matches the shipped `algorithm.yaml`. Please also add a documentation regression check that parses the YAML example and verifies that its structure matches the shipped configuration.

**Why this is distinct:** the original defect was invalid indentation and a stray token. The remaining issue is that the repair removes a required configuration value, leaving a syntactically valid but semantically incomplete example.

**Verification boundary:** I compared the README snippet with the corresponding shipped `algorithm.yaml` and parsed the submitted snippet with PyYAML, reproducing `resource_dir: None`. I did not run the full Yaoba benchmark.

PR #796, opened 2026-08-14
Fix hardcoded absolute path in yaoba testenv.yaml

hibaa23:
**What type of PR is this?**
/kind bug

`examples/yaoba/singletask_learning_boost/testenv/testenv.yaml` hardcodes `image_folder_url` to an absolute path from the original author's local machine:

    image_folder_url: "/home/wjj/wjj/Public/code/ianvs/dataset/yaoba/images"

This path does not exist on any other contributor's machine, so the example cannot run out of the box for anyone else. This PR replaces it with a relative path consistent with the other dataset fields already declared in the same file (`train_url`, `test_url`, etc.).

Found while investigating `examples/yaoba` for LFX Mentorship 2026 Term 3 pre-task work (see #742).

Related to #742


diegodevxd:
Changing `image_folder_url` is a reasonable local portability repair, but this field is not in #789's `_PATH_FIELDS` and is not among #835's checked dataset keys. A future path validator could therefore pass while this example-specific required input regresses.

Please document whether `image_folder_url` is required by the Yaoba runtime and add it to the appropriate example schema/validator rule (with a clear policy for locally absent datasets). This keeps the local fix from becoming an untested special case.

akshita317:
The intent is right — `/home/wjj/wjj/Public/code/ianvs/dataset/yaoba/images` is a developer-local path and should be repo-relative.

@diegodevxd asked above whether `image_folder_url` is required by the Yaoba runtime. I traced it, and the answer is **yes, it is required — and it is currently discarded before it ever reaches the runtime.** Details below, since that also determines whether this PR has any effect today.

Two findings on the effect.

**1. The change currently has no runtime effect.**

`image_folder_url` is not declared in `Dataset.__init__` (`core/testenvmanager/dataset/dataset.py:45-54`), and `Dataset._parse_config` (L71-73) assigns only keys already present on the instance:

```python
    if attr in self.__dict__:
        self.__dict__[attr] = value
```

so the key is discarded whatever its value. Verified on this branch:

```
BRANCH: pr-796
  image_folder_url in YAML = 'examples/yaoba/singletask_learning_boost/dataset/images'
  L49: self.dataset.image_folder_url -> AttributeError: 'Dataset' object has no attribute 'image_folder_url'
  RESULT: 3/3 attribute reads still fail on branch 'pr-796'
```

Same 3/3 as unmodified `main`. `core/.../singletask_learning_active_boost.py:49` reads the attribute, so the example fails there either way.

The path correction becomes correct *and effective* once `image_folder_url`, `known_dataset_url`, `unknown_dataset_url` and `val_url` are added to `Dataset.__init__` — so this change is worth keeping, just not sufficient alone.

**2. PR #824 changes the same line.**

PR #824 makes the identical edit to this file, plus the paradigm dispatch registration. Attempting both:

```
CONFLICT (content): Merge conflict in examples/yaoba/singletask_learning_boost/testenv/testenv.yaml
```

Worth coordinating with @31groot so one of the two carries the config change.

hibaa23:
Thanks both for tracing this — makes sense that the value has nowhere
to land yet.

Agreed this PR isn't sufficient alone: the corrected path only takes
effect once `image_folder_url` (and the other three dropped keys) are
registered on `Dataset.__init__`. I'll leave this open since the value
itself is still correct, and becomes live as soon as that Core-level
fix lands — whether from #824, a dedicated Core PR, or otherwise.

Re: the #824 conflict — happy to defer the config-line edit to
whichever PR ends up carrying the Core registration fix, to avoid
duplicating it.

31groot:
Thank you, you're right, and I should have caught it myself before claiming the path fix in #824 was meaningful. I reproduced the exact failure you described and pushed the actual fix.

**Reproduced your finding independently:**
```
AttributeError: 'Dataset' object has no attribute 'image_folder_url'
```
confirmed against `Dataset.__init__` / `_parse_config()` exactly as you described — `image_folder_url` (and `known_dataset_url`, `unknown_dataset_url`, `val_url`) were never declared, so `_parse_config`'s `if attr in self.__dict__` guard silently drops them regardless of value.

**Fix pushed to #824:** added all four attributes to `Dataset.__init__` with empty-string defaults (no-op for every other example that doesn't set them). Re-ran your reproduction case against the patched branch:
```
image_folder_url = 'examples/yaoba/singletask_learning_boost/dataset/images'
known_dataset_url = 'examples/yaoba/singletask_learning_boost/resource/json/known_train.json'
unknown_dataset_url = 'examples/yaoba/singletask_learning_boost/resource/json/NG_labeled.json'
val_url = 'examples/yaoba/singletask_learning_boost/resource/json/known_train.json'
```
All four now load correctly through the real `testenv.yaml` → `Dataset._parse_config()` path.

**Re: the merge conflict** : agreed we shouldn't duplicate the config-line edit. Since #824 now also carries the Core-level `Dataset.__init__` registration (not just the path-value correction), I'd suggest #824 carries both and this PR can drop its `testenv.yaml` line once that's confirmed, happy to coordinate whichever order maintainers prefer. Appreciate you catching this; it was a real gap in what I'd verified.

hibaa23:
Sounds right to me — happy to drop the testenv.yaml line from this PR once #824's Dataset.__init__ registration lands, to avoid the duplicate. Nice fix.

PR #799, opened 2026-08-14
[PHASE IV] fix(core/algorithm): raise ValueError on duplicate module type instead of silently dropping it

### LINKED ISSUE #797: [PHASE IV] core/algorithm: duplicate module `type` silently overwrites earlier module in get_full_combinations
**What happened**:

`Algorithm._parse_modules_config()` in `core/testcasecontroller/algorithm/algorithm.py` builds a `modules_list` of `(module.type, [module_variants])` tuples and passes it directly to `get_full_combinations()` in `core/testcasecontroller/generation_assistant/generation_assistant.py`. That function builds its output with `dict(zip(name_list, combination_value_list))`, so if two modules in the config end up with the same `.type` string, the resulting dict silently keeps only the last one — an entire module variant vanishes from every generated test case combination with no error, warning, or log message.

If two modules in the `modules` section of an algorithm config share the same `type`, Ianvs should raise a clear `ValueError` identifying the duplicated type(s), rather than silently dropping one of them.

```python
from core.testcasecontroller.generation_assistant.generation_assistant import get_full_combinations

modules_list = [
    ("basemodel", ["module_variant_A"]),
    ("basemodel", ["module_variant_B"]),
]

result = get_full_combinations(modules_list)
print(result)
# Actual: [{'basemodel': 'module_variant_B'}]
# module_variant_A silently disappears — no error, no warning.
```

Affected file: `core/testcasecontroller/algorithm/algorithm.py`, `_parse_modules_config()` (around line 191).

Root cause is generic to `get_full_combinations()`'s dict-based combination building, but the fix is best placed caller-side in `_parse_modules_config()` since that's where the semantic meaning of "module type" is known, allowing a precise error message.

A fix would involve: after building `modules_list`, checking for duplicate `module.type` values and raising `ValueError` listing the duplicated type(s) before calling `get_full_combinations()`.

Found while investigating for LFX Phase IV.

hayyanbaig:
**What type of PR is this?**

/kind bug

`Algorithm._parse_modules_config()` passed `modules_list` straight into `get_full_combinations()` with no check for duplicate module types. Since the combination step builds a dict keyed by type (`dict(zip(name_list, combination_value_list))`), a duplicate type silently overwrote an earlier module with no error — a user could lose an entire module variant from their benchmark run without any indication anything went wrong.

This PR adds a check in `_parse_modules_config()` that raises a clear `ValueError` listing the duplicated type(s) before `get_full_combinations()` is ever called, instead of silently dropping one of the modules.

Fixes #797

**Testing done**:

Reproduced the silent overwrite with an isolated call to `get_full_combinations()` using two `("basemodel", [...])` tuples — confirmed only the second survived in the output dict. After the fix, constructing a `modules_list` with duplicate types now raises `ValueError` listing the duplicated type(s) before `get_full_combinations()` is ever called. Verified against a real, unmodified example config (llm-agent) that `_parse_modules_config()` still parses cleanly with no false positive when module types are unique.


akshhkaushik (commented):
### Accept, with a focused regression test

I executed `_parse_modules_config()` with two `basemodel` entries. The patched code raises the intended explanatory error instead of letting `get_full_combinations()` overwrite one key:

```text
ValueError: Duplicate module type(s) ['basemodel'] ...
```

I also scanned every tracked algorithm YAML on the PR base and found zero existing duplicate-type lists, so the new guard does not reject a committed example.

Recommendation: merge. Please add the synthetic duplicate case as a unit test so the behavior is protected. The current set construction is quadratic because it calls `module_types.count()` for every element; config lists are small today, so this is not a blocker, but a one-pass `Counter`/seen set would be clearer.

Verification boundary: executed the real parser and scanned tracked configs; no model modules were loaded.

karyxx:
I pulled this PR branch and tested it locally against a few algorithm benchmark setups as part of my LFX Mentorship Term 3 pre-test investigation.

### What I tested:
I set up a benchmark configuration to compare two candidate base model architectures (`FPN_Model` vs. `ResNet_Model`) inside a single algorithm config:

```python
from core.testcasecontroller.algorithm.algorithm import Algorithm

multi_model_config = {
    "algorithm": {
        "paradigm_type": "singletasklearning",
        "modules": [
            {"type": "basemodel", "name": "FPN_Model", "url": "./testalgorithms/fpn/basemodel.py"},
            {"type": "basemodel", "name": "ResNet_Model", "url": "./testalgorithms/resnet/basemodel.py"}
        ]
    }
}

Algorithm("multi_model_benchmark", multi_model_config)
```

**Output on this branch:**
```text
ValueError: Duplicate module type(s) ['basemodel'] found in algorithm config; get_full_combinations() would silently keep only the last module of each duplicated type, dropping the earlier one from every generated test case. Each module 'type' in the 'modules' list must be unique.
```

---

### My findings and analysis:

1. **The Multi-Model Benchmarking Dilemma:**  
   There is a plausible alternative view that duplicate `module.type` entries should simply be treated as invalid syntax and rejected with a fatal error. While raising a `ValueError` succeeds at preventing the silent data loss in `dict(zip(...))`, it creates a major functional limitation for Ianvs as a benchmarking framework.  
   In ML benchmarking, evaluating multiple model architectures of the same role (for example, comparing `FPN` vs. `ResNet`, or testing different YOLO backbones) within a single benchmark matrix is a core requirement. Throwing a fatal `ValueError` prevents users from running multi-model comparisons within a single configuration, forcing them to manually duplicate their setup across multiple YAML files.

2. **Caller vs. Callee Placement:**  
   The underlying defect lives in `core/testcasecontroller/generation_assistant/generation_assistant.py` line 48, where `dict(zip(name_list, ...))` collapses duplicate keys. Placing the check only in `Algorithm._parse_modules_config()` leaves the shared `get_full_combinations()` helper unhardened against key collisions from other callers.

3. **Performance:**  
   The duplicate check `{t for t in module_types if module_types.count(t) > 1}` does repeated scans over the list ($O(N^2)$). A single-pass set tracker (`seen = set(); {t for t in module_types if t in seen or seen.add(t)}`) is cleaner and more efficient.

---

### Recommendation:
**major revision (fatal errors need significant coding)**  
*Justification:* While this patch succeeds at stopping silent data loss, throwing a fatal error blocks valid multi-model comparative benchmarking. I recommend revising the fix to group duplicate module keys directly in `generation_assistant.py` (e.g. via `defaultdict(list)`) so that multiple candidate models can be benchmarked automatically without breaking single-file workflows.

ishansurdi:
I tested this against the reported input:
 
```
ValueError: Duplicate module type(s) ['basemodel'] found in algorithm config; ...
```
 
and the single-module control, same branch:
 
```
valid control combos: 2
```
 
Resolves #797 within its stated scope, no change to valid input. The multi-model-comparison concern and the `O(n²)` scan are already covered well in the existing reviews here, so I won't repeat them — I'd only add: I checked whether this closes the underlying helper too, since `get_full_combinations()` has a second caller. It doesn't. `Module._parse_hyperparameters()`
has the identical collision for duplicate hyperparameter names, and it's still open on this
branch:
 
```
generated=4 unique=2
```
 
Worth a shared-helper follow-up regardless of how the module-type question above is resolved, since it's the same defect surfacing through a second caller.

PR #800, opened 2026-08-14
[PHASE IV] fix(core/module): raise clear ValueError when hyperparameter 'values' is missing

### LINKED ISSUE #798: [PHASE IV] core/module: missing `values` key in hyperparameter config crashes with opaque TypeError
**What happened**:

`Module._parse_hyperparameters()` in `core/testcasecontroller/algorithm/module/module.py` reads `hp_values = hp_config[1].get("values")`, which silently returns `None` if a hyperparameter block in the YAML omits the `values` key. That `None` is passed straight through to `get_full_combinations()`, which calls `itertools.product(*values_list)`. Passing `None` to `product()` raises:

```
TypeError: 'NoneType' object is not iterable
```

This error carries no reference to which hyperparameter caused it, no reference to the module being configured, and no indication that the actual problem is a missing `values:` key in the YAML. A user with several hyperparameters across multiple modules has to manually bisect their config to find the mistake.

If a hyperparameter's config is missing a non-empty `values` list, Ianvs should raise a clear `ValueError` naming the specific hyperparameter and explaining what's missing, at config-parsing time rather than deep inside `itertools.product()`.

```python
from core.testcasecontroller.generation_assistant.generation_assistant import get_full_combinations

# Simulates a hyperparameter block with no 'values' key, e.g.:
# hyperparameters:
#   - learning_rate: {}
result = get_full_combinations([("learning_rate", None)])
# Actual: TypeError: 'NoneType' object is not iterable
# No mention of "learning_rate" anywhere in the error.
```

Affected file: `core/testcasecontroller/algorithm/module/module.py`, `_parse_hyperparameters()` (around line 158).

A fix would involve: validating `hp_values` is a non-empty list immediately after reading it from the config, before appending it to `hp_name_values_list`, and raising `ValueError` naming the specific hyperparameter if it isn't.

Found while investigating for LFX Phase IV.

hayyanbaig:
**What type of PR is this?**

/kind bug

`_parse_hyperparameters()` passed a hyperparameter's `values` straight through to `get_full_combinations()` with no validation. If a hyperparameter block in the YAML omitted the `values` key, `.get("values")` silently returned `None`, which then crashed deep inside `itertools.product()` with a bare `TypeError: 'NoneType' object is not iterable` — an error that names neither the hyperparameter nor the module responsible.

This PR adds a check right after `hp_values` is read, raising a `ValueError` that explicitly names the affected hyperparameter and explains what's missing from its config.

Fixes #798

**Testing done**:

Reproduced the original crash with a hyperparameter config missing `values`, confirming the error gave zero indication of the actual cause. After the fix, the same input now raises a clear `ValueError` naming the specific hyperparameter (`learning_rate` in the reproduction) instead of an opaque `TypeError`. Verified against a real, unmodified example config (llm-agent) that hyperparameter parsing still works cleanly with no false positive when `values` is present and valid.


Arijit429:

Recommendation: Accept as is

This PR closes the fourth site of the same defect family documented in Task 1 — unguarded config access producing an opaque error instead of a clear one:

- Outer key, TestEnv — testenv.py:63 — fixed by #854 (open)
- Outer key, Algorithm — algorithm.py:160 — fixed by #854 (open)
- Inner field, paradigm_type — algorithm.py — fixed by #704 (open)
- Hyperparameter values — module.py — fixed by this PR (#800)

All four independently authored, none references any of the others, none has merged. The isinstance + presence-check shape matches #854's fix exactly — correct, minimal, one file, no scope creep.

Verified: py_compile on pr-800 branch passes clean. The author's own reproduction confirms the TypeError → clear ValueError improvement.

MukandKrishna (commented):
**Recommendation: changes requested.** This patch improves the error for a present hyperparameter mapping whose `values` key is missing, but the same malformed-input family still bypasses the new guard when the mapping itself is missing.

For example, this valid YAML shape:

```yaml
hyperparameters:
  - learning_rate:
```

loads as `[{"learning_rate": None}]`. On PR head `c8dd24033ab9c60e26a7ff048d85f7d460fdd2e8`, `_parse_hyperparameters()` calls `.get("values")` before validating the nested value:

```python
hp_values = hp_config[1].get("values")
```

I executed both boundary cases against the exact PR file:

```text
missing values inside mapping: ValueError: hyperparameter 'learning_rate' is missing a non-empty 'values' list in its config (got None). ...
missing nested mapping (YAML: '- learning_rate:'): AttributeError: 'NoneType' object has no attribute 'get'
```

Therefore, the new clear-error contract currently applies to `{"learning_rate": {}}`, but not to the closely equivalent `{"learning_rate": None}` produced by an empty YAML key.

Please validate that `hp_config[1]` is a mapping before calling `.get()`, then retain the existing non-empty-list check for its `values` field. A focused regression test should cover both YAML forms and assert that both raise the documented `ValueError` with the hyperparameter name, rather than leaking `AttributeError`.

Verification boundary: I executed `_parse_hyperparameters()` from the exact PR head with minimal import stubs because this parser path does not require the Ianvs runtime or Sedna behavior. The full benchmark pipeline was not run.

Arijit429:
Update to my earlier review: another reviewer has independently found and executed a second boundary case — when the YAML has `- learning_rate:` (empty key), hp_config[1] is None and .get() raises AttributeError instead of the documented ValueError. This confirms the fix is incomplete for the None-mapping case. My original Accept recommendation should be revised to Minor Revision: add a None/mapping check before the .get("values") call, and a regression test covering both YAML forms.

karyxx:
I pulled this branch and tested the fix locally against several edge-case hyperparameter configs as part of my LFX Mentorship Term 3 pre-test investigation.

### What I tested:
I tested the parsing flow across different YAML hyperparameter structures:

```python

# Case 1: Missing values key inside mapping ({'learning_rate': {'val': [0.001]}})
config_missing_val = {
    "type": "basemodel", "name": "FPN", "url": "test.py",
    "hyperparameters": [{"learning_rate": {"val": [0.001, 0.01]}}]
}
Module(config_missing_val)
```

**Output on this branch:**
```text
ValueError: hyperparameter 'learning_rate' is missing a non-empty 'values' list in its config (got None). Add a 'values:' entry with at least one value under 'learning_rate'.
```

---

### My thoughts and feedback:

1. **Clear improvement on the missing key error:**  
   The check `if not isinstance(hp_values, list) or not hp_values:` is a big improvement. Instead of an unhandled `TypeError: 'NoneType' object is not iterable` deep in `itertools.product`, users get an immediate `ValueError` naming the exact parameter (`learning_rate`).

2. **Pre-get mapping validation (`AttributeError` on empty keys or scalars):**  
   In `_parse_hyperparameters()`, line 165 calls `hp_config[1].get("values")`.  
   If a user writes an empty key in YAML:
   ```yaml
   hyperparameters:
     - learning_rate:
   ```
   PyYAML parses this as `{"learning_rate": None}`. Calling `.get()` on `None` throws `AttributeError: 'NoneType' object has no attribute 'get'` before ever reaching the new `isinstance(hp_values, list)` check.  
   Similarly, writing a direct scalar (`- learning_rate: 0.001`) causes `AttributeError: 'float' object has no attribute 'get'`.  
   We should check that `isinstance(hp_config[1], dict)` before attempting `.get("values")`.

3. **Scalar auto-coercion vs. strict rejection:**  
   For single-value parameters (like `learning_rate: 0.001`), automatically wrapping scalars into single-item lists (`hp_values = [hp_config[1]] if not isinstance(hp_config[1], dict) else ...`) would make the YAML interface much more forgiving and user-friendly.

---

### Recommendation:
**minor revision (fixes without coding effort, e.g., docs, squash commits)**  
*Justification:* targeted fix that solves the primary `TypeError`. It just needs a quick guard ensuring `hp_config[1]` is a `dict` before calling `.get("values")` to prevent unhandled `AttributeError` on empty or scalar YAML keys.

ishansurdi:
I tested the two cases named in #798:
 
```
missing values key:   ValueError: hyperparameter 'learning_rate' is missing a non-empty 'values' list ...
values: "ab" (string): ValueError, no longer silently expanded into ['a', 'b']
```
 
Both correct. The null-mapping `AttributeError` case is already reported twice in this thread and I can confirm it independently reproduces - I won't re-detail it. Two further gaps I tested that aren't yet mentioned here:
 
```
other_hyperparameters: {}
  TypeError: 'NoneType' object is not iterable
  -> this PR's isinstance() check lives only in the else branch; it's never reached when
     hp_name == "other_hyperparameters".
 
two learning_rate entries, different values
  generated=4 unique=2, first two values silently dropped
  -> this PR validates value shape, not key uniqueness; the loss is one level up in
     get_full_combinations().
```
 
I'd fold both into whatever revision addresses the null-mapping case, since all three are the same class of gap - shapes that reach `_parse_hyperparameters()` and aren't yet covered.

PR #802, opened 2026-08-15
fix(core): fix race condition in FCIL helper_function client routing

### LINKED ISSUE #801: [Bug] Race condition in FederatedClassIncrementalLearning causes cross-client state corruption in helper_function
**What happened**:

In the Federated Class-Incremental Learning (FCIL) subsystem (`core/testcasecontroller/algorithm/paradigm/federated_learning/federated_class_incremental_learning.py`), a multithreading race condition causes cross-client state corruption during round-by-round training and helper synchronization.

### 1. Non-deterministic Thread Completion Causes Client State Swapping
During each training round, `FederatedLearning.train()` spawns concurrent threads for each client:
```python
# core/testcasecontroller/algorithm/paradigm/federated_learning/federated_learning.py lines 254-260
for idx in range(self.clients_number):
    client_thread = Thread(
        target=self.client_train,
        args=(idx, train_datasets, None),
        kwargs=kwargs,
    )
    client_thread.start()
```

In `FederatedClassIncrementalLearning.client_train()`, each thread finishes training and appends its `train_info` into `self.train_infos`:
```python
# core/testcasecontroller/algorithm/paradigm/federated_learning/federated_class_incremental_learning.py lines 214-219
train_info = super().client_train(
    client_idx, train_datasets, validation_datasets, **kwargs
)
with self.lock:
    self.train_infos.append(train_info)
```

Because client training threads run concurrently and have variable computation/I-O durations, they complete in non-deterministic order. `self.train_infos` is populated in **completion order**, not **client ID order**.

Subsequently, `helper_function()` assumes `self.train_infos` is strictly ordered by client index `i`:
```python
# core/testcasecontroller/algorithm/paradigm/federated_learning/federated_class_incremental_learning.py lines 228-230
for i in range(self.clients_number):
    helper_info = self.aggregator.helper_function(train_infos[i])
    self.clients[i].helper_function(helper_info)
```

If Client 2 finishes before Client 0, `train_infos[0]` contains Client 2's `train_info` (`{"client_id": 2, ...}`). In the loop:
- Client 0 (`self.clients[0]`) receives Client 2's aggregated helper information (e.g. task prototypes, exemplar data, or class boundary state).
- Client 2 (`self.clients[2]`) receives Client 0's helper information.

This leads to silent cross-contamination of client state in multi-client federated class-incremental training.

---

### 2. Unguarded Dictionary Lookup on `model_eval` Crashes with `KeyError` / `AttributeError`
In `FederatedClassIncrementalLearning.__init__()`:
```python
metrics_dict = kwargs.get("model_eval", {})["model_metric"]
_, accuracy_func = get_metric_func(metrics_dict)
self.accuracy_func = accuracy_func
```
- If `model_eval` is not provided in `kwargs` (or is empty), `kwargs.get("model_eval", {})["model_metric"]` crashes immediately with `KeyError: 'model_metric'`.
- If `model_metric` contains empty/uninitialized fields (e.g., default `{"name": "", "url": ""}`), `get_metric_func()` attempts `getattr(sys.modules[__name__], "_func")` and crashes with `AttributeError`.

---

1. Helper state routing in `helper_function()` must match each `train_info` to its corresponding client by `client_id` (or index through a client-keyed dictionary), ensuring client `i` always receives helper state derived from client `i`'s training outputs regardless of thread completion order.
2. `__init__()` should defensively validate `model_eval` and `model_metric`, providing safe fallbacks or raising informative `ValueError` exceptions if required metrics are missing.

---

Run the following minimal Python script demonstrating how concurrent thread execution corrupts client state assignment in the current `helper_function` logic:

```python
import threading
import time

clients_number = 3
train_infos = []
lock = threading.RLock()

def simulated_client_train(client_idx):
    # Simulate variable execution duration (e.g. client 2 finishes first, client 0 last)
    delay = 0.2 if client_idx == 0 else (0.1 if client_idx == 1 else 0.01)
    time.sleep(delay)
    train_info = {
        "client_id": client_idx,
        "exemplars": f"exemplars_for_client_{client_idx}",
        "num_samples": 100
    }
    with lock:
        train_infos.append(train_info)

threads = [
    threading.Thread(target=simulated_client_train, args=(i,))
    for i in range(clients_number)
]
for t in threads:
    t.start()
for t in threads:
    t.join()

print("Completion order in train_infos:", [info["client_id"] for info in train_infos])

# How FederatedClassIncrementalLearning.helper_function currently routes info:
print("\n--- FCI helper_function routing ---")
for i in range(clients_number):
    assigned_info = train_infos[i]
    print(f"Client {i} incorrectly receives state from client {assigned_info['client_id']}")
```

**Observed Output**:
```
Completion order in train_infos: [2, 1, 0]

--- FCI helper_function routing ---
Client 0 incorrectly receives state from client 2
Client 1 incorrectly receives state from client 1
Client 2 incorrectly receives state from client 0
```

---

- `core/testcasecontroller/algorithm/paradigm/federated_learning/federated_class_incremental_learning.py` (`client_train`, `helper_function`, `__init__`)
- `core/testcasecontroller/algorithm/paradigm/federated_learning/federated_learning.py`

### Root Cause
1. `FederatedClassIncrementalLearning.client_train` appends `train_info` to a shared list `self.train_infos` inside a threaded callback without preserving client positional ordering.
2. `FederatedClassIncrementalLearning.helper_function` assumes `train_infos[i]` corresponds to `self.clients[i]`, ignoring the `train_info["client_id"]` attribute attached during `client_train`.

### Impact
- **State Corruption**: In continual federated learning algorithms (such as GLFC or FedClassIncremental), server helper functions calculate class-specific exemplars, generative replay samples, or distillation targets per client. Mismatching this mapping corrupts client training updates and degrades model convergence.
- **Reliability**: Benchmarks crash with unhandled `KeyError` / `AttributeError` when `model_eval` configuration dictionaries are missing or default-initialized.

### High-level Solution Direction
1. Store client training outputs in a thread-safe mapping keyed by `client_id` (e.g. `self.train_infos[client_idx] = train_info` or a dict `{client_id: train_info}`), rather than an unordered append.
2. In `helper_function()`, retrieve the helper info for each client using its explicit `client_id`.
3. Add defensive `.get()` lookups and validation for `model_eval` and `model_metric` in `FederatedClassIncrementalLearning.__init__()`.
4. Add unit regression tests verifying deterministic client-to-helper state mapping under simulated multi-threaded execution.

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

This PR resolves a multithreading race condition and defensive initialization gaps in the Federated Class-Incremental Learning (FCIL) subsystem:

1. **Deterministic Client State Routing in `helper_function()`**:
   - In `FederatedClassIncrementalLearning.client_train()`, client training info is now tracked directly by client index (`self.train_infos[client_idx] = train_info`) in a thread-safe manner.
   - `helper_function()` routes server helper state to each client using an explicit `client_id` mapping (`train_info_map[i]`), ensuring that client `i` always receives the helper state computed from client `i`'s training updates regardless of thread completion order.

2. **Defensive `model_eval` / `model_metric` Initialization**:
   - In `FederatedClassIncrementalLearning.__init__()`, guarded dictionary access for `model_eval` and `model_metric` to avoid `KeyError` when `model_eval` is omitted and `AttributeError` when `name` is uninitialized.

3. **Safe Aggregation Lookup in `FederatedLearning.__init__()`**:
   - Safely unpacked `self.aggregator` with tuple type-checking to prevent `TypeError` when the aggregation module instance is `None`.

4. **Unit Tests**:
   - Added `tests/test_federated_class_incremental_learning.py` verifying out-of-order list routing, dictionary routing, concurrent multi-threaded `client_train` state recording, and missing/empty `model_eval` initialization (5 tests passed).

Fixes #801

### DIFF
--- a/core/testcasecontroller/algorithm/paradigm/federated_learning/federated_class_incremental_learning.py
+++ b/core/testcasecontroller/algorithm/paradigm/federated_learning/federated_class_incremental_learning.py
@@ -61,13 +61,21 @@ class FederatedClassIncrementalLearning(FederatedLearning):
         }
 
         self.aggregate_clients = []
-        self.train_infos = []
+        self.train_infos = {}
 
         self.forget_rate_metrics = []
         self.accuracy_per_round = []
-        metrics_dict = kwargs.get("model_eval", {})["model_metric"]
-        _, accuracy_func = get_metric_func(metrics_dict)
-        self.accuracy_func = accuracy_func
+        self.accuracy_func = None
+        model_eval = kwargs.get("model_eval")
+        if isinstance(model_eval, dict):
+            metrics_dict = model_eval.get("model_metric")
+            try:
+                _, accuracy_func = get_metric_func(metrics_dict)
+                self.accuracy_func = accuracy_func
+            except Exception as err:  # pylint: disable=broad-except
+                LOGGER.warning(
+                    f"Failed to initialize accuracy_func from {metrics_dict}: {err}"
+                )
 
     def task_definition(self, dataset_files, task_id):
         """Define the task for the class incremental learning paradigm
@@ -215,18 +223,42 @@ class FederatedClassIncrementalLearning(FederatedLearning):
             client_idx, train_datasets, validation_datasets, **kwargs
         )
         with self.lock:
-            self.train_infos.append(train_info)
+            self.train_infos[client_idx] = train_info
 
     def helper_function(self, train_infos):
         """helper function for FCI Method
            Many of the FCI algorithms need server to perform some operations
            after the training of each round e.g data generation, model update etc.
         Args:
-            train_infos (list of dict): the train info that the clients want to send to the server
+            train_infos (dict or list of dict): the train info that the clients
+                want to send to the server
         """
+        if isinstance(train_infos, dict):
+            train_info_map = train_infos
+        else:
+            train_info_map = {
+                info.get("client_id", idx): info
+                for idx, info in enumerate(train_infos)
+                if isinstance(info, dict)
+            }
+
+        expected_clients = set(range(self.clients_number))
+        actual_clients = set(train_info_map.keys())
+        missing_clients = expected_clients - actual_clients
+        unexpected_clients = actual_clients - expected_clients
+        if missing_clients or unexpected_clients:
+            err_msg = []
+            if missing_clients:
+                err_msg.append(f"missing client(s) {sorted(list(missing_clients))}")
+            if unexpected_clients:
+                err_msg.append(f"unexpected client(s) {sorted(list(unexpected_clients))}")
+            raise RuntimeError(
+                f"invalid train_infos in helper_function: {', '.join(err_msg)}, "
+                f"expected {sorted(list(expected_clients))}"
+            )
 
         for i in range(self.clients_number):
-            helper_info = self.aggregator.helper_function(train_infos[i])
+            helper_info = self.aggregator.helper_function(train_info_map[i])
             self.clients[i].helper_function(helper_info)
         LOGGER.info("finish helper function")
 
--- a/core/testcasecontroller/algorithm/paradigm/federated_learning/federated_learning.py
+++ b/core/testcasecontroller/algorithm/paradigm/federated_learning/federated_learning.py
@@ -75,7 +75,11 @@ class FederatedLearning(ParadigmBase):
         self.aggregate_clients = []
         self.clients_number = kwargs.get("client_number", 1)
         self.mode = kwargs.get("if_mode_llm", False)
-        _, self.aggregator = self.module_instances.get(ModuleType.AGGREGATION.value)
+        agg_val = self.module_instances.get(ModuleType.AGGREGATION.value)
+        if isinstance(agg_val, (list, tuple)) and len(agg_val) == 2:
+            _, self.aggregator = agg_val
+        else:
+            self.aggregator = agg_val
         if self.mode:
             LOGGER.info("Using LLM multi GPU mode for Federated Learning")
             self.gpu_num = kwargs.get("gpu_num", 1)
@@ -250,17 +254,27 @@ class FederatedLearning(ParadigmBase):
             train_datasets (list): train data for each client
         """
         client_threads = []
+        errors = []
         LOGGER.info(f"len(self.clients): {len(self.clients)}")
         for idx in range(self.clients_number):
-            client_thread = Thread(
-                target=self.client_train,
-                args=(idx, train_datasets, None),
-                kwargs=kwargs,
-            )
+            def _worker(client_idx=idx):
+                try:
+                    self.client_train(client_idx, train_datasets, None, **kwargs)
+                except Exception as err:  # pylint: disable=broad-except
+                    with self.lock:
+                        errors.append((client_idx, err))
+
+            client_thread = Thread(target=_worker)
             client_thread.start()
             client_threads.append(client_thread)
         for thread in client_threads:
             thread.join()
+
+        if errors:
+            raise RuntimeError(
+                f"client training failed for client(s): "
+                f"{[(idx, str(err)) for idx, err in errors]}"
+            )
         LOGGER.info("finish training")
 
     # pylint: disable=unused-argument
@@ -309,6 +323,7 @@ class FederatedLearning(ParadigmBase):
         Args:
             global_weights (list): aggregated weights
         """
+        # pylint: disable=E1101
         for client in self.clients:
             client.set_weights(global_weights)
         LOGGER.info("finish send weights to clients")
--- a/tests/test_federated_class_incremental_learning.py
+++ b/tests/test_federated_class_incremental_learning.py
@@ -1,0 +1,226 @@
+# Copyright 2022 The KubeEdge Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for FederatedClassIncrementalLearning paradigm."""
+
+# pylint: disable=missing-class-docstring,missing-function-docstring,too-few-public-methods,wrong-import-position,line-too-long,import-outside-toplevel,protected-access,unnecessary-dunder-call
+
+import os
+import sys
+import threading
+import time
+import unittest
+from unittest.mock import patch
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+from core.testcasecontroller.algorithm.paradigm.federated_learning.federated_class_incremental_learning import (
+    FederatedClassIncrementalLearning,
+)
+
+
+class MockClient:
+    def __init__(self, client_id):
+        self.client_id = client_id
+        self.received_helper_info = None
+
+    def helper_function(self, helper_info):
+        self.received_helper_info = helper_info
+
+
+class MockAggregator:
+    def helper_function(self, train_info):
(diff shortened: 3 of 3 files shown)


Hiteshsai007:
## Major revision — missing client results are silently skipped after worker failure

**Decision: request changes.** Keying `train_infos` by `client_idx` fixes completion-order routing, but the new `if i in train_info_map` branch silently accepts an incomplete round.

At [`federated_class_incremental_learning.py` lines 236–248](https://github.com/kubeedge/ianvs/blob/a6c3b41f5d932637e6f57abedc20e4aa9427346a/core/testcasecontroller/algorithm/paradigm/federated_learning/federated_class_incremental_learning.py#L236-L248), a missing client key causes that client's helper update to be skipped without an error:

```python
for i in range(self.clients_number):
    if i in train_info_map:
        helper_info = self.aggregator.helper_function(train_info_map[i])
        self.clients[i].helper_function(helper_info)
```

This matters because `FederatedLearning.train()` runs `client_train` in plain `threading.Thread` workers and only joins them; exceptions raised inside a worker are not propagated to the caller. A failed client can therefore leave its dictionary entry absent. The PR then lets the round proceed through helper processing with only a subset of clients, and `run()` clears `train_infos` afterward. That turns a client-training failure into silent partial state distribution.

I executed the PR's exact `helper_function` body with `clients_number=2` and `train_infos={0: {'client_id': 0}}`:

```text
client 0 calls: [{'client_id': 0}]
client 1 calls: []
method returned without error despite missing client 1
```

**Recommendation:** after all worker joins and before aggregation/helper processing, compare the produced client IDs with `set(range(self.clients_number))` and fail with an actionable error listing missing clients. Better still, wrap worker targets to capture and re-raise exceptions in the coordinator. `helper_function` should reject incomplete or unexpected key sets rather than silently skipping them. Add a test where one worker raises and assert that the round fails before aggregation/helper distribution.

**Verification boundary:** I executed the exact changed helper method with an incomplete map and inspected the real thread-launch/join path statically. I did not run a full FCIL benchmark or prove how every aggregator behaves with the separately incomplete `aggregate_clients` list.

bhuvan-somisetty:
Thanks for the thorough review and edge case catch @Hiteshsai007! Completely agree that worker failures shouldn't lead to silent partial state distribution. Updated the PR to:
1. Wrap `FederatedLearning.train()` worker targets to capture worker exceptions and re-raise a `RuntimeError` in the coordinator listing failed clients.
2. Validate that `train_info_map` in `helper_function()` contains all expected client IDs `0..clients_number-1` and raise an informative `RuntimeError` if any are missing.
3. Added unit tests covering worker failure propagation and incomplete `train_infos` validation under `tests/test_federated_class_incremental_learning.py` (7 tests pass cleanly).

Hiteshsai007:
## Follow-up verification — original blocking finding addressed

Thank you for the update. I re-reviewed current head [`998290e`](https://github.com/kubeedge/ianvs/commit/998290ebd9b23629a3ebb28a5c99b7a034768506) against my earlier incomplete-round finding.

The revision now addresses both parts of the recommendation:

1. Worker targets capture exceptions under the lock, and `FederatedLearning.train()` raises a coordinator-level `RuntimeError` after all workers are joined.
2. `FederatedClassIncrementalLearning.helper_function()` compares actual and expected client IDs and raises before distributing helper state when a result is missing.
3. The added tests cover both an incomplete `train_infos` map and a client worker raising during training.

I executed the exact updated `helper_function()` body with two expected clients and only client 0 present:

```text
Expected clients: [0, 1]
Provided train_infos: {0: {'client_id': 0}}
RuntimeError: incomplete train_infos in helper_function: missing client(s) [1], expected [0, 1]
```

This is the fail-fast behavior requested in my original review, so **my original blocking finding is resolved at `998290e`**.

One non-blocking hardening suggestion: also reject unexpected client IDs (`actual_clients - expected_clients`) rather than checking only missing IDs. That is not required to resolve the reported silent-partial-round defect.

**Verification boundary:** I executed the exact updated helper method with controlled collaborators and inspected the worker-exception path and added tests statically. I did not run the full FCIL benchmark or the complete test module with all Ianvs/Sedna dependencies.

**Follow-up execution at updated commit `998290e`:**

figure4

bhuvan-somisetty:
Thanks for the follow-up verification @Hiteshsai007! Great suggestion on the hardening — I've updated `helper_function()` to validate and reject unexpected client IDs (`actual_clients - expected_clients`) as well and added a unit test covering it. All 8 tests pass cleanly.

Hiteshsai007:
Thank you — I rechecked updated commit [`89c2a01`](https://github.com/kubeedge/ianvs/commit/89c2a013de1185d9ec137e1045c8337334aed861). The additional validation now rejects both missing and unexpected client IDs, and the new `test_helper_function_unexpected_train_infos_raises` covers the hardening suggestion.

I also executed the exact updated `helper_function()` with expected clients `[0, 1]` and supplied keys `[0, 1, 5]`:

```text
RuntimeError: invalid train_infos in helper_function: unexpected client(s) [5], expected [0, 1]
```

That resolves my remaining non-blocking suggestion. **All findings and follow-up suggestions from my review are addressed at `89c2a01`.**

**Verification boundary:** I executed the exact updated helper method for the unexpected-ID case and inspected the added regression test statically. I did not independently run the complete eight-test module or the full FCIL benchmark.

bhuvan-somisetty:
Thank you @Hiteshsai007 for the thorough reviews, test executions, and valuable feedback throughout!

Hiteshsai007:
You’re welcome, and thank you for implementing the requested fail-fast handling and follow-up client-ID validation.

I appreciate the clear updates and regression coverage. The final verification and addressed findings are now recorded in the related submission evidence.

bhuvan-somisetty:
Thanks @Hiteshsai007! Glad all the fail-fast handling, worker exception propagation, and client ID validations are verified and in place. Appreciate the great review and guidance throughout!

Roshan5105labs:
### LFX Pre-Test Review — PR #802

## Review Scope

I reviewed the current PR head 89c2a01.

The PR addresses the FCIL client-routing problem by changing client training information from completion-ordered list storage to client-ID keyed storage.

The relevant change stores the result as:

self.train_infos[client_idx] = train_info

instead of relying on append order.

The PR also validates missing and unexpected client IDs before helper processing.

## Positive Findings

This directly addresses the identity assumption behind #801: client state is now associated with an explicit client ID rather than the nondeterministic order in which worker threads complete.

I also inspected the added regression coverage. The PR includes tests covering:

out-of-order result routing;

dictionary-based result routing;

concurrent client result population;

missing client information;

unexpected client information;

worker-training failure.

These tests cover the important failure modes introduced by concurrent FCIL execution.

## Cross-PR Interaction

While reviewing this PR together with the other eligible restoration PRs, I found substantial overlap with PR #810.

Both #802 and #810 modify the shared FederatedLearning.train() path to:

catch exceptions raised during client training;

associate the failure with the corresponding client ID;

wait for worker threads to complete;

propagate the failure back to the coordinating thread.

I compared the current heads of #802 and #810 using Git merge analysis.

The current revisions modify the same regions of the shared federated implementation and produce merge-conflict markers when combined.

image

This means the interaction is not only conceptual duplication; the two current implementations also require explicit coordination before integration.

## Verification

I attempted to run the FCIL unit tests from this PR in a clean Python 3.8 environment with the bundled Sedna 0.6.0.1 package and Ianvs requirements installed.

Test execution was blocked during Ianvs import by a Windows ONNX DLL initialization error before the FCIL test cases themselves were reached.

Therefore, my runtime verification for this PR is limited to the import boundary. I additionally inspected the implementation and regression tests statically.

## Recommendation

Classification: Minor revision / coordination required before merge.

The FCIL client-identity repair and associated regression coverage are useful and directly address #801.

However, the generic worker-failure handling overlaps with PR #810. I recommend selecting one canonical implementation for worker-failure propagation in the shared federated controller, then rebasing or adjusting the other PR to remove the duplicated Core responsibility.

The FCIL-specific identity mapping and validation in #802 can remain independently valuable after that coordination.

bhuvan-somisetty:
Thanks @Roshan5105labs.

I have addressed the Python 3.8/3.9 pylint static verification failure by resolving broad exception handling pragmas, docstring line length formatting, and mock test imports.

All 8 unit tests in `tests/test_federated_class_incremental_learning.py` pass and the code rates at 10.00/10 with pylint. The updated commit has been force-pushed.

PR #803, opened 2026-08-15
fix(incremental-learning): prevent stale label reuse in hard example retraining

Aryansingh-ai:
**Problem**

In `IncrementalLearning._get_train_dataset`, the `label` variable used to write ground-truth labels for hard-example retraining data is not reset per loop iteration.

**Root cause**

for old, new in hard_examples:
    index = np.where(data_labels.x == old)
    if len(index[0]) == 1:
        label = data_labels.y[index[0][0]]
    file.write(f"{new} {label}\n")

When a hard example's original path does not match exactly one entry in the source index file (0 matches from a path mismatch, or >1 from duplicate entries — both plausible in real datasets), the `if` body is skipped and the previous iteration's `label` is silently reused and written for the current, unrelated image.

**Impact**

Silently corrupts the training data fed into incremental retraining — a mislabeled hard example biases the newly trained model with zero error, warning, or log trace. Affects any `incremental_learning` paradigm run where hard examples include an unmatched or duplicated path.

**Fix**

`label` is now reset to `None` each iteration. Any hard example without a unique matching label is skipped with a logged warning instead of silently reusing a stale value.

**Validation**

Isolated reproduction of the exact loop logic confirmed the defect and the fix (unmatched sample silently inherited a previous sample's label before the fix; after the fix it is skipped with a warning). `pylint` on the changed file: 9.83/10. `py_compile` passes.


whozahm3d:
### Task 2 — Code Review & Merge Conflict Analysis

**Recommendation:** **Minor Revision**

#### Review Findings:
1. **Guard Semantics:** The strict guard `if len(index[0]) == 1:` correctly handles zero matches, multi-matches, and `NoneType` labels by skipping and logging warnings. This is safer than PR #843's `> 0` check.
2. **Scope Limitation & Merge Conflict:** This PR resolves only 1 of the 4 core defects in Issue #842. Merging alongside PR #843 produces a direct merge conflict in `_get_train_dataset()`. Its guard logic should be folded into a revised PR #843.

MukandKrishna (commented):
**Recommendation: major revision.** Resetting `label`, checking strict cardinality, and logging are all improvements over stale reuse. The remaining problem is that the method reports success after changing the requested training population.

Executed results from this PR's exact method:

```text
CASE=missing_after_valid
WARNING=skip hard example(missing.jpg): no unique matching label found in (labels.txt).
OUTPUT='new-valid.jpg cat\n'

CASE=duplicate_identity
WARNING=skip hard example(duplicate.jpg): no unique matching label found in (labels.txt).
OUTPUT=''
```

The first case requests two hard examples and returns a normal path containing one; the second returns a normal path to an empty index. The caller has no machine-readable indication that the retraining input was degraded.

Please prevalidate the whole batch and fail before publishing a partial file. If degraded continuation is an intentional product decision, return structured written/missing/duplicate counts and require an explicit configured threshold; a log warning alone is insufficient for benchmark integrity. A regression test should assert this policy.

Executed against PR head `cd0d04763e41a6ee063443692c1fe2740afe6be2`; full model training was not run. Related reproduction context: [Task 1](https://github.com/kubeedge/ianvs/discussions/921#discussioncomment-18145877) and the [Issue #842 follow-up](https://github.com/kubeedge/ianvs/issues/842#issuecomment-5408109234).

pratikpanda2006:
**Review:**

- Fixes label corruption in `_get_train_dataset()` by resetting `label = None` at the top of each loop iteration — correct fix for the stale-label-reuse symptom described in #842.
- Scope: example/core-local patch to the same function PR #843 also modifies. Does not address the metric-unpacking `TypeError` or the `_inference()` dict/list mismatch — only one of the four defects in #842.
- Overlap risk: if both #803 and #843 merge, `_get_train_dataset()` gets two independent patches to the same code path, which risks a silent merge conflict or duplicated logic.
- Recommendation: reject / close as superseded once #843 merges, since #843's fix to the same function is more defensive (explicit `len(index[0]) > 0` check vs. a bare reset).

PR #806, opened 2026-08-15
fix(ianvs): cleanup sys.path after dynamic module loading

### LINKED ISSUE #805: [Bug] Fix sys.path cleanup during dynamic module loading
**What happened**:

While reviewing the Ianvs codebase, I noticed that dynamic module loading in `core/common/utils.py` temporarily modifies `sys.path`.

If module loading fails before cleanup is executed, the modified path may remain in the Python process state. This can potentially affect subsequent imports and lead to difficult-to-debug behavior.

Temporary changes made to `sys.path` during dynamic module loading should always be reverted, regardless of whether the import succeeds or raises an exception.

1. Inspect the module loading logic in `core/common/utils.py`.
2. Trigger a failed dynamic import where `importlib.import_module()` raises an exception.
3. Observe that cleanup code after the import operation is skipped when an exception occurs.
4. The process retains the modified `sys.path` state.

Proposed fix:
- Use `try...finally` to guarantee `sys.path` cleanup.
- Add regression tests covering both successful and failed module loading scenarios.
- Keep existing functionality and public APIs unchanged.

I would like to implement this fix and contribute the corresponding tests.

Ganesh-403:
**What type of PR is this?**

/kind bug

This PR fixes a cleanup issue in `core/common/utils.py` during dynamic module loading.

Currently, `py2dict()` and `load_module()` temporarily modify Python import state using `sys.path`. If module loading fails due to an exception, cleanup code may not execute, leaving the process state modified.

This PR:
- Ensures `sys.path` cleanup always happens using `try...finally`
- Ensures temporary `sys.modules` changes are cleaned up safely
- Adds regression tests covering successful and failed dynamic module loading scenarios
- Adds minimal pytest configuration so the new unit tests can run without collecting unrelated example scripts

These changes improve state isolation and prevent hard-to-debug import contamination issues.

Fixes #805

### DIFF
--- a/core/common/utils.py
+++ b/core/common/utils.py
@@ -63,17 +63,19 @@ def py2dict(url):
     if url.endswith('.py'):
         module_name = os.path.basename(url)[:-3]
         config_dir = os.path.dirname(url)
+        old_sys_path = sys.path.copy()
         sys.path.insert(0, config_dir)
-        mod = import_module(module_name)
-        sys.path.pop(0)
-        raw_dict = {
-            name: value
-            for name, value in mod.__dict__.items()
-            if not name.startswith('__')
-        }
-        sys.modules.pop(module_name)
-
-        return raw_dict
+        try:
+            mod = import_module(module_name)
+            raw_dict = {
+                name: value
+                for name, value in mod.__dict__.items()
+                if not name.startswith('__')
+            }
+            return raw_dict
+        finally:
+            sys.modules.pop(module_name, None)
+            sys.path[:] = old_sys_path
 
     raise RuntimeError('config file must be the py format')
 
@@ -95,9 +97,11 @@ def load_module(url):
     if os.path.isfile(url):
         module_name = module_name.split(".")[0]
 
+    old_sys_path = sys.path.copy()
     sys.path.insert(0, module_path)
     try:
         importlib.import_module(module_name)
-        sys.path.pop(0)
     except Exception as err:
         raise RuntimeError(f"load module(url={url}) failed, error: {err}") from err
+    finally:
+        sys.path[:] = old_sys_path
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,2 @@
+[pytest]
+testpaths = tests
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,47 @@
+import sys
+import pytest
+
+from core.common import utils
+
+def test_py2dict_success(tmp_path):
+    test_file = tmp_path / "good_module.py"
+    test_file.write_text("TEST_VAR = 'value'")
+
+    initial_sys_path = list(sys.path)
+
+    result = utils.py2dict(str(test_file))
+
+    assert result.get("TEST_VAR") == "value"
+    assert sys.path == initial_sys_path
+
+def test_py2dict_failure(tmp_path):
+    test_file = tmp_path / "bad_module.py"
+    test_file.write_text("1 / 0")
+
+    initial_sys_path = list(sys.path)
+
+    with pytest.raises(ZeroDivisionError):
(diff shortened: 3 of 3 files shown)


Ganesh-403:
Updated the PR based on review feedback:
- restored sys.path using snapshot/restore instead of pop(0)
- removed unnecessary pytest pythonpath configuration
- cleaned up tests and made exception assertions specific

pytest now passes with 4 tests.

Omkeswani27:
Tested this branch directly (`git fetch origin pull/806/head:pr-806`):

load_module leaked? False
py2dict leaked? False

Confirms this correctly fixes the sys.path leak via snapshot/restore, the
most robust of three overlapping fixes (also see #729, #759). Separately:
the PR body says "Fixes #805," but #805's actual title is unrelated
(FederatedClassIncrementalLearning race condition); GitHub's own "closes"
widget instead targets a differently-numbered issue. Worth a maintainer's
look before merge. Full analysis:

Yash4616:
Executed against the same reproduction of the module load path. This is the only PR in the open set whose `sys.path` handling is correct when the loaded module itself inserts at index 0.

| | restores after a failed import | correct when the loaded module inserts at index 0 |
|---|---|---|
| `main` | FAIL | FAIL |
| this PR | PASS | PASS, snapshot and restore |
| #729 | PASS | FAIL, removes the module's entry and leaks the inserted one |
| #759 | PASS | FAIL, guard does not match so nothing is removed |

`sys.path[:] = old_sys_path` handles the case the positional approaches miss, and it covers `py2dict()` as well. The case is reachable: `examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/basemodel.py:29` inserts at index 0 during its own import.

Module identity is untouched, so two files sharing a basename still resolve to the first one loaded. That is outside what this PR claims to fix, and I mention it only because the same function is being changed by several open PRs at once.

One correction for the record: a comment on this PR suggests the `Fixes #805` link is wrong on the grounds that #805 is a FederatedClassIncrementalLearning race condition. #805 is *"[Bug] Fix sys.path cleanup during dynamic module loading"*. The race condition issue is #801. The link here is correct.

The DCO check on the current head shows `action_required` and has not run.

**Recommendation: minor revision.** The remaining items are a rebase and getting the check to run, neither of which needs code.

Aryansingh-ai:

I checked out PR #806 and tested the modified `core/common/utils.py` directly.

### 1. Failure-path `sys.path` cleanup — PASS

I forced `load_module()` to fail and compared `sys.path` before and after the call.

Observed result:

```text
BEFORE_LEN: 10
AFTER_LEN: 10
UNCHANGED: True

The temporary module directory was not left in sys.path after the failed import.

2. Basename collision — FAIL

I then tested the separate module-identity property using two files with the same basename:

one/plugin.py → VALUE = "one"
two/plugin.py → VALUE = "two"

Observed result:

AFTER FIRST: one
AFTER SECOND: one
EXPECTED SECOND: two

Therefore, #806 fixes the tested sys.path cleanup behavior but does not fix the independent basename-collision defect from #557.

3. Overlap with #729 / #759

The cleanup behavior substantially overlaps with #729 and #759.

All three PRs independently modify the same Core dynamic-loading cleanup boundary:

#729 → try/finally cleanup
#759 → try/finally cleanup
#806 → snapshot/restore cleanup

From a maintainer perspective, merging all three would be redundant. One coherent cleanup implementation should be selected rather than accumulating multiple competing fixes to the same utility.

The tested cleanup behavior is correct for the specific sys.path failure-path defect.

I recommend:

Clearly scope the PR to the cleanup defect.
Retain regression coverage for failed dynamic imports.
Coordinate with #729 and #759 so that only one implementation of the overlapping cleanup fix is ultimately merged.
Do not consider this PR a solution to the separate #557 module-identity collision.

Issue #332, opened 2026-02-08
[cloud-edge-collaborative-inference-for-llm] Missing dataset blocks example execution - incomplete setup documentation


krrish175-byte:
### Description

The cloud-edge-collaborative-inference-for-llm example cannot be executed due to missing GPQA dataset files and incomplete documentation for dataset preparation. New users attempting to run this example (including LFX Mentorship pre-test participants) are blocked.

### Environment

- **OS**: Ubuntu-based Linux
- **Python**: 3.10
- **Ianvs Version**: Latest from main branch (as of Feb 2026)
- **Example**: `examples/cloud-edge-collaborative-inference-for-llm`

### Steps to Reproduce

1. Clone Ianvs repository
   ```bash
   git clone https://github.com/kubeedge/ianvs.git
   cd ianvs
   ```

2. Install dependencies
   ```bash
   cd examples
   pip3 install -r requirements.txt
   ```

3. Navigate to cloud-edge LLM example
   ```bash
   cd cloud-edge-collaborative-inference-for-llm
   ```

4. Attempt to check dataset
   ```bash
   ls dataset/
   ```

5. Try to run example
   ```bash
   ianvs -f testenv/testenv.yaml
   ```

### Expected Behavior

- Dataset should either:
  - Be included in the repository (at least a sample/toy dataset), OR
  - Have clear, step-by-step instructions for downloading and preparing the GPQA dataset from Kaggle, OR
  - Include an automated download script

- Example should be runnable following the README instructions

### Actual Behavior

**Dataset not found**:
```bash
$ ls dataset/
ls: cannot access 'dataset/': No such file or directory

$ find . -name "dataset" -type d
[no results]
```

**Configuration references missing files**:
```yaml
# testenv/testenv.yaml
dataset:
  train_data: "./dataset/gpqa/train_data/data.json"
  test_data_info: "./dataset/gpqa/test_data/metadata.json"
```

**README mentions dataset but lacks complete instructions**:
- Mentions "creating a custom dataset" but doesn't explain how
- References Kaggle API token but doesn't show dataset download command
- Link to "[Detailed Setup Guide]" appears incomplete

### Root Cause Analysis

The example assumes users know how to:
1. Create a Kaggle account
2. Obtain Kaggle API credentials
3. Install Kaggle CLI
4. Find the correct GPQA dataset on Kaggle
5. Download and structure the dataset correctly

However, these steps are not documented in the README, making the example inaccessible to new users.

### Proposed Fix

[...]

Issue #330, opened 2026-02-07
[Pre-Test] Cloud-edge LLM example reproducibility gaps and benchmark proposal


iashutoshyadav:
## Context

This issue is submitted as part of the **LFX Mentorship 2026 pre-test** for the
KubeEdge / Ianvs project.

It contains:
1. A mini proposal for a cloud-edge speculative decoding benchmark (Task 1)
2. A reproducible investigation of the cloud-edge LLM example (Task 2A)
3. Identified reproducibility and documentation gaps with suggested improvements (Task 2B)

The intent is to provide constructive feedback aligned with KubeEdge contribution guidelines.

---

### 1. Problem Statement

Speculative decoding can accelerate LLM inference by allowing a draft model to
generate multiple tokens that are verified by a stronger model. However, in
cloud-edge collaborative inference, speedups are not guaranteed.

Network RTT, limited bandwidth, and jitter introduce communication overhead into
the decoding critical path. In addition, heterogeneous compute environments
(edge CPUs vs cloud GPUs) can reduce acceptance rates or shift bottlenecks,
making end-to-end performance unpredictable. This motivates a systematic,
reproducible benchmark tailored to cloud-edge constraints.

---

### 2. Benchmark Scope

Key variables to evaluate include:
- Network RTT and bandwidth
- Speculative draft size (K)
- Draft vs verify compute ratio
- Concurrency level
- Prompt length and generation length

---

### 3. Metrics & Methodology

Metrics:
- Time To First Token (TTFT)
- Throughput (tokens/sec)
- End-to-end latency (p50 / p95)

Timing starts when a request reaches the edge and ends when the final token is
returned. Network wait and verification time are included. One-time model loading
is excluded via warmup runs.

---

### 4. High-level Design in Ianvs

The benchmark maps to:
- Benchmark case: cloud-edge speculative decoding
- Scenario: configurable network and compute settings
- Runner: orchestrates edge draft and cloud verification
- Reporter: collects TTFT, throughput, and latency percentiles

Configuration is YAML-based with fixed seeds, fixed prompts, and pinned model
versions for reproducibility.

---

### 5. Milestones (9–12 Weeks)

**Milestone 1 (Weeks 1–3)**  
Baseline decoding and standard speculative decoding integrated into Ianvs.  
Acceptance: reproducible metrics across runs.

**Milestone 2 (Weeks 4–7)**  
Network emulation and parameter sweeps (RTT, K, concurrency).  
[...]

PR #807, opened 2026-08-15
fix(cloud-edge-llm): add GPQA-diamond dataset setup without Kaggle/HF auth

kavaljeetsingh-dev:
## Problem

The README for `cloud-edge-collaborative-inference-for-llm` states that GPQA-diamond dataset setup "follows the same progress" as MMLU-5-shot, but never actually provides a working GPQA download command. The canonical GPQA dataset on HuggingFace (Idavidrein/gpqa) is gated and requires login + accepting terms of use, which blocks users from reproducing this benchmark.

Related: #332, #330

## Fix

Adds `prepare_gpqa_dataset.py`, a helper script that downloads GPQA-diamond from the same public, unauthenticated CSV mirror used by OpenAI's own `simple-evals` benchmarking tool (`https://openaipublic.blob.core.windows.net/simple-evals/gpqa_diamond.csv`), and converts it into Ianvs' expected `data.jsonl` / `metadata.json` format.

- No Kaggle account or HuggingFace login required
- Deterministic answer-choice shuffling (fixed seed) for reproducibility
- Updates README.md with the new setup instructions

## Testing

Ran locally, successfully parsed and wrote 198 questions (matches the known size of GPQA-diamond):

​```
Downloading GPQA-diamond from https://openaipublic.blob.core.windows.net/simple-evals/gpqa_diamond.csv ...
Formatting examples into Ianvs schema ...
Parsed 198 valid questions.
Wrote ..\..\dataset\gpqa\test_data\data.jsonl
Wrote ..\..\dataset\gpqa\test_data\metadata.json
Done. No Kaggle account or HuggingFace login was required.
​```

Fixes #

### DIFF
--- a/examples/cloud-edge-collaborative-inference-for-llm/README.md
+++ b/examples/cloud-edge-collaborative-inference-for-llm/README.md
@@ -192,6 +192,15 @@ kaggle datasets download -d kubeedgeianvs/ianvs-mmlu-5shot
 unzip -o ianvs-mmlu-5shot.zip
 rm -rf ianvs-mmlu-5shot.zip
 ```
+**GPQA-diamond** (no Kaggle account or auth token required):
+
+The canonical GPQA dataset on HuggingFace is gated and requires accepting terms of use before download. To avoid that friction, this example instead uses the same public, unauthenticated CSV mirror that OpenAI's own `simple-evals` benchmarking tool relies on. Run the provided helper script from the example directory:
+
+```bash
+python prepare_gpqa_dataset.py --output-dir ../../dataset/gpqa
+```
+
+This downloads `gpqa_diamond.csv` from `https://openaipublic.blob.core.windows.net/simple-evals/gpqa_diamond.csv`, converts it into Ianvs' expected `data.jsonl` / `metadata.json` format with a fixed random seed (for reproducible answer-choice shuffling), and writes it to `dataset/gpqa/`.
 
 2. Then, check the path of `train_data` and `test_data` in 
 `examples/cloud-edge-collaborative-inference-for-llm/testenv/testenv.yaml`.
--- a/examples/cloud-edge-collaborative-inference-for-llm/prepare_gpqa_dataset.py
+++ b/examples/cloud-edge-collaborative-inference-for-llm/prepare_gpqa_dataset.py
@@ -1,0 +1,231 @@
+# Copyright 2024 The KubeEdge Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Prepare the GPQA-Diamond dataset for the cloud-edge-collaborative-inference-for-llm
+example, without requiring a Kaggle account or a gated HuggingFace login.
+
+Background
+----------
+The canonical GPQA dataset (Idavidrein/gpqa on HuggingFace) is *gated*: users
+must log in and accept terms of use before downloading it. The README of this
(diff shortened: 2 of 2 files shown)


Shikhar-Kesharwani:
### Review — supplementary reproducibility finding

The new public GPQA download path is a meaningful improvement because it removes the Kaggle/Hugging Face authentication barrier, and the deterministic answer-choice shuffling is good for reproducibility.

One remaining reproducibility concern is that the benchmark dataset itself is fetched from a mutable remote CSV URL:

`https://openaipublic.blob.core.windows.net/simple-evals/gpqa_diamond.csv`

The script does not pin the dataset to an immutable revision, verify a checksum, or record the source artifact's digest after download. Therefore, the generated `data.jsonl` can change across runs if the remote CSV is replaced or updated, even though the local answer-choice shuffling uses a fixed seed.

This matters for a benchmark because two users following the same documented setup at different times could evaluate different underlying question sets while believing they are running the same GPQA-diamond configuration. The current test that parses 198 questions verifies the current artifact's shape, but it does not establish artifact identity.

I recommend pinning the dataset to an immutable/versioned artifact where possible, or at minimum publishing and verifying a SHA-256 checksum for the expected CSV and recording the dataset revision/hash in the generated metadata. A regression test could then verify that the downloaded artifact matches the expected digest before conversion.

This is separate from the authentication/setup problem fixed by this PR: the current change makes the dataset easier to obtain, while the remaining concern is ensuring that the newly documented download path remains reproducible and identifies the exact benchmark artifact being evaluated.

31groot:
### Request changes : GPQA output path doesn't match `testenv.yaml`

**Review depth: Analysis**

The PR adds the new GPQA setup command, but the documented output directory does not match the dataset path used by `testenv.yaml`.

The README says to run:

```bash
python prepare_gpqa_dataset.py --output-dir ../../dataset/gpqa
```

From:

```text
examples/cloud-edge-collaborative-inference-for-llm/
```

this resolves to:

```text
<repo>/dataset/gpqa
```

while `testenv.yaml` expects:

```text
<repo>/examples/cloud-edge-collaborative-inference-for-llm/dataset/gpqa
```

I reproduced the mismatch locally:

```text
README output dir : /home/parv/ianvs/dataset/gpqa
testenv data dir  : /home/parv/ianvs/examples/cloud-edge-collaborative-inference-for-llm/dataset/gpqa
same directory    : False
GPQA PATH MISMATCH REPRODUCED
```

So following the README command can generate the dataset successfully, while the benchmark still looks in a different directory and reports missing files.

**Recommendation:** make the README `--output-dir` match the directory consumed by `testenv.yaml` (or update `testenv.yaml` accordingly). A small regression test could run the documented command and verify that the expected dataset files exist where `testenv.yaml` looks for them.

**Verification:** Reproduced locally by resolving both paths from the Example directory. No external download or full benchmark run was required.

Kingthnight:
### Review: malformed GPQA rows can produce a successful partial dataset

I tested the converter locally on Windows with Python 3.12.1 using a small synthetic CSV. I replaced only the download function so the test could run offline.

The CSV contained two rows:

- one complete row
- one malformed row with an empty `Incorrect Answer 3`

The result was:

```text
Input CSV rows: 2
Converted rows: 1
Rejected rows reported by API: none
Conversion returned normally: True
Parsed 1 valid questions.
Rows written by main(): 1
```

The malformed row was silently skipped, while the conversion completed normally and wrote a partial dataset. This could make the benchmark appear ready even though some source questions were lost.

I also tested a row where the correct answer and one incorrect answer had identical text:

```text
Converted row count: 1
Gold response: B
Occurrences of identical answer 'Alpha': 2

A. Gamma
B. Alpha
C. Beta
D. Alpha
```

The converter accepted this row, leaving two indistinguishable choices in the generated question.

Possible improvements:

- fail by default when any input row is rejected
- report the rejected row count, identifiers and reasons
- provide an explicit option such as `--allow-partial` if partial conversion is intended
- validate that all four choices are non-empty and distinct
- validate the complete output before replacing the destination file
- add tests for malformed and duplicate-choice rows

I did not download the live GPQA dataset for this test. I replaced the network call with a local fixture, while using the converter and output-writing code from this PR. The temporary test files were removed afterward.

I am connecting this finding with the other restoration cases in [Discussion #965](https://github.com/kubeedge/ianvs/discussions/965).

PR #808, opened 2026-08-15
fix(federated-learning): fix broken non-iid data partitioning

Aryansingh-ai:
**Problem**

In `partition_data()` (`core/testenvmanager/dataset/utils.py`), selecting the `non-iid` data partitioning strategy for federated learning crashes immediately on any realistic dataset.

**Root cause**

    current_class = random.sample(range(class_num), sample_number)   # a Python list
    indices = np.where(y_data == current_class)[0]

`y_data == current_class` compares a `(N,)` numpy array against a `(sample_number,)` Python list, triggering numpy broadcasting instead of a membership test. Unless `N == sample_number` by coincidence, this raises `ValueError: operands could not be broadcast together`.

**Impact**

The `non-iid` data partitioning option — a standard, documented feature for federated learning benchmarks (`fl_data_setting.data_partition: "non-iid"`) — is completely non-functional. Any benchmark configured to use it crashes before any client data is produced.

**Fix**

Replace the equality comparison with `np.isin(y_data, current_class)`, which correctly selects all samples whose label belongs to the client's sampled class subset.

**Validation**

Isolated reproduction of the original logic reliably reproduces `ValueError: operands could not be broadcast together with shapes (100,) (6,)` on a 100-sample, 10-class synthetic dataset. The fixed logic (verified separately) produces correct per-client partitions with no crash. `pylint` on the changed file: 9.37/10 (identical to the unmodified baseline — no new warnings introduced). `py_compile` passes.


Hiteshsai007:
## Major revision — sample real label values, not positions in the unique-label array

**Decision: request changes.** The `np.isin` change correctly removes the broadcasting error, but the non-IID path can still silently return empty or incomplete client partitions for valid datasets whose labels are not zero-based contiguous integers.

At [`utils.py` lines 83–91](https://github.com/kubeedge/ianvs/blob/28cff7eb87b8e9a4b1d3f6544716f395e5b465bd/core/testenvmanager/dataset/utils.py#L83-L91), `class_num` is computed from `np.unique(y_data)`, but `current_class` is sampled from `range(class_num)`. That range contains **positions** (`0, 1, ...`), not necessarily the actual values present in `y_data`. `np.isin` therefore fixes the array-shape exception while exposing this separate label-domain assumption.

I reproduced the changed predicate against a six-row dataset with labels `[10, 10, 20, 20, 30, 30]`, three classes, `non_iid_ratio=2/3`, and `random.seed(7)`:

```text
unique labels       = [10, 20, 30]
PR selected values  = [1, 0]
PR indices/labels   = [] / []
fixed selected      = [20, 10]
fixed indices/labels= [0, 1, 2, 3] / [10, 10, 20, 20]
```

This matters for 1-based class IDs, sparse numeric IDs, and string labels. The run no longer necessarily crashes, so an empty partition can travel farther into training and make the failure less obvious.

**Recommendation:** retain the unique values and sample from them directly, for example:

```python
classes = np.unique(y_data).tolist()
sample_number = int(len(classes) * non_iid_ratio)
current_class = random.sample(classes, sample_number)
indices = np.where(np.isin(y_data, current_class))[0]
```

Please also validate `sample_number >= 1` (or define the intended zero-class behavior), since truncation currently produces an empty selection whenever `class_num * non_iid_ratio < 1`. A focused test should cover at least labels `[10, 20, 30]` or strings, in addition to the existing zero-based case.

**Verification boundary:** I executed the selection predicate with NumPy using the inputs and seed shown above and confirmed both the empty PR result and the recommended result. I did not run the full federated-learning pipeline or evaluate how downstream trainers report empty-client datasets.

PR #810, opened 2026-08-15
fix(federated-learning): surface client training failures instead of silently swallowing them

### LINKED ISSUE #809: Federated learning silently aggregates with missing/failed clients on training failure
**What happened**:

`FederatedLearning.train()` (`core/testcasecontroller/algorithm/paradigm/federated_learning/federated_learning.py`) runs each client's local training in a separate `Thread`, then joins them with no exception handling:

    for idx in range(self.clients_number):
        client_thread = Thread(target=self.client_train, args=(idx, train_datasets, None), kwargs=kwargs)
        client_thread.start()
        client_threads.append(client_thread)
    for thread in client_threads:
        thread.join()

Python's `threading.Thread` swallows uncaught exceptions raised in its target — they are printed via `threading.excepthook` to stderr but never propagate through `.join()` or raise in the calling thread. `client_train()` only appends the client's result to `self.aggregate_clients` after its training call succeeds:

    def client_train(self, client_idx, train_datasets, validation_datasets, **kwargs):
        train_info = self.clients[client_idx].train(train_datasets[client_idx], validation_datasets, **kwargs)
        ...
        with self.lock:
            self.aggregate_clients.append(agg_client)

If a client's training call raises for any reason (bad data, OOM, a model bug), that client is simply absent from `aggregate_clients`, with zero indication anything went wrong.

`run()` then proceeds unconditionally:

    for r in range(self.rounds):
        train_fn(train_datasets, round=r)
        global_weights = self.aggregator.aggregate(self.aggregate_clients)
        ...

A client training failure should either abort the round or, at minimum, be surfaced as a visible error — not silently vanish from aggregation.

**How to reproduce it**:

Isolated reproduction of the exact threading structure (3 simulated clients, client 1 raises):

    from threading import Thread, RLock

    lock = RLock()
    aggregate_clients = []

    def client_train(client_idx):
        if client_idx == 1:
            raise RuntimeError(f"simulated training failure for client {client_idx}")
        with lock:
            aggregate_clients.append(f"client_{client_idx}_weights")

    def train(clients_number):
        client_threads = []
        for idx in range(clients_number):
            t = Thread(target=client_train, args=(idx,))
            t.start()
            client_threads.append(t)
        for t in client_threads:
            t.join()
        print("train() returned normally -- no exception raised to caller")

    train(3)
    print(f"aggregate_clients: {aggregate_clients}")

Output:

    train() returned normally -- no exception raised to caller
    aggregate_clients: ['client_0_weights', 'client_2_weights']

`train()` returns normally despite client 1 failing; the exception is only visible as an unstructured traceback in stderr via `threading.excepthook`. The benchmark proceeds to aggregate with 2 of 3 configured clients, silently.

**Impact**:

Any real federated learning benchmark run where a client fails mid-training (a realistic occurrence — OOM, corrupted batch, transient model error) will silently produce a leaderboard result for a smaller/different federation than the one configured, with no error, warning, or log distinguishing it from a fully successful run.

**Environment**:

- ianvs main branch, commit 36ec008
- Affects: `core/testcasecontroller/algorithm/paradigm/federated_learning/federated_learning.py`, `FederatedLearning.train()`

I have a minimal fix and will open a PR referencing this issue.

Aryansingh-ai:
**What type of PR is this?**

/kind bug

`FederatedLearning.train()` runs each client's local training in a separate `Thread`, then joins them with no exception handling:

    for idx in range(self.clients_number):
        client_thread = Thread(target=self.client_train, args=(idx, train_datasets, None), kwargs=kwargs)
        client_thread.start()
        client_threads.append(client_thread)
    for thread in client_threads:
        thread.join()

Python's `threading.Thread` swallows uncaught exceptions raised in its target — they are printed via `threading.excepthook` to stderr but never propagate through `.join()` or raise in the calling thread. `client_train()` only appends to `self.aggregate_clients` after its training call succeeds, so if any client's training raises for any reason (bad data, OOM, a model bug, anything), that client is simply absent from `aggregate_clients` with zero indication anything went wrong.

`run()` then unconditionally proceeds to `self.aggregator.aggregate(self.aggregate_clients)` regardless of whether every client actually contributed. The practical effect: a federated learning benchmark can silently run — and report results for — fewer clients than the configuration specifies, every round, with no error surfaced anywhere in the run. This is a silent correctness bug, not a crash: the benchmark completes "successfully" and produces a leaderboard entry that doesn't reflect the configured federation.

**Fix**

Client threads now target a new `_client_train_safe` wrapper that calls the existing `client_train()` inside a `try/except`, and records `(client_idx, exception)` into `self.thread_errors` under the existing `self.lock` if it fails. `self.thread_errors` is initialized in `__init__` (and reset at the start of each `train()` call, so failures don't leak across rounds). After joining all threads, `train()` checks `self.thread_errors`: if any client failed, it raises a `RuntimeError` naming the failed client id(s) and the first underlying error, instead of silently continuing to aggregation.

This is a minimal, contained change — no modification to `client_train()`, `llm_client_train()`, aggregation logic, or any public interface. Existing behavior for the all-clients-succeed case is unchanged.

**Validation**

- Isolated reproduction of the original logic (same threading structure, a simulated failing client 1 of 3) confirms the defect: `train()` returns normally and `aggregate_clients` ends up with only 2 of 3 expected entries, with the exception visible only as an unstructured traceback in stderr.
- The fixed logic, verified with the same reproduction harness, correctly raises `RuntimeError: federated learning client training failed for client(s) [1]; first error: ...` instead of proceeding, and does not affect the success path (all clients succeeding still produces a full `aggregate_clients` list and returns normally).
- `pylint` on the changed file: 9.69/10, vs 9.67/10 on unmodified `main` for the same file — no regressions introduced; the small delta comes from fixing an existing `attribute-defined-outside-init` pattern that this change touched (`self.thread_errors` is now declared in `__init__` rather than only in `train()`).
- `py_compile` passes.

Fixes #809


Omkeswani27:
Supplementary review (Bonus task, LFX 2026 Term 3 pre-test). Note: the PR
body states "Fixes #809," but #809's title ("Fix sys.path cleanup during
dynamic module loading") is unrelated to this PR's actual content
(FederatedLearning thread-failure handling). This is the second issue-link
mismatch found in this general PR range (see also #806 on issue #805) —
possibly a recurring pattern worth flagging. Separately, this PR's use of
per-client Threads confirms Core code in this repo does run concurrent
work against shared state, relevant to a concurrency edge-case flagged
(but unconfirmed) elsewhere. Full write-up:
https://github.com/kubeedge/ianvs/discussions/878#discussioncomment-18122630

suhaan-24:
This is a good catch — an exception raised inside a `Thread` target is written to stderr by the
default excepthook and otherwise vanishes, so a client that fails to train currently leaves the
round looking successful. Capturing into `self.thread_errors` and raising after `join()` is the
right shape.

Checked the implementation and it holds up: `self.lock = RLock()` exists (line 74), `thread_errors`
is reset at the top of `train()` so it does not accumulate across rounds, and the raise sits after
every `thread.join()` — so aggregation never runs on a partially-trained round.

## `llm_train()` has the same pattern and is not covered

Twenty lines below the patched method, `llm_train()` runs its clients through a queue of GPU
workers:

```python
        def _gpu_worker(gpu_idx: int):
            while not task_q.empty():
                try:
                    client_idx = task_q.get_nowait()
                except Empty:
                    break
                LOGGER.info(f"GPU-{gpu_idx} starts training client {client_idx}")
                self.llm_client_train(
                    client_idx=client_idx,
                    train_datasets=train_datasets,
                    validation_datasets=None,
                    device_id=gpu_idx,
                )
```

The `try/except` there guards only `task_q.get_nowait()` against `Empty`. An exception inside
`llm_client_train` propagates out of `_gpu_worker`, is swallowed by its `Thread` exactly as
`client_train`'s was before this fix, and the worker dies silently — the queue then drains through
the remaining workers and the round reports success with some clients never trained.

Since `_client_train_safe` already exists, the same wrapper shape would apply directly:

```python
                try:
                    self.llm_client_train(client_idx=client_idx, ...)
                except Exception as err:  # pylint: disable=broad-except
                    with self.lock:
                        self.thread_errors.append((client_idx, err))
```

plus the same post-join check after the workers are joined.

Worth covering in this PR rather than a follow-up, since the two methods will otherwise diverge in
failure behaviour — `train()` raising and `llm_train()` silently continuing is a confusing pair to
debug.

## Smaller suggestions

- The error message surfaces `self.thread_errors[0][1]` only. With several clients failing for
  different reasons, the first one may not be representative — logging all of them at ERROR before
  raising would keep the detail without lengthening the exception text.
- The `pylint: disable=broad-except` is justified here — the wrapper's job is to catch anything a
  client raises. Worth noting `KeyboardInterrupt` derives from `BaseException` and so is correctly
  not captured, which is the behaviour you want on Ctrl-C.

## Verification boundary

`self.lock`, the reset placement, the post-join raise, and the `llm_train`/`_gpu_worker` body were
all read from this PR's head. I have **not** executed a federated example — neither `cifar100` nor
`federated-llm/fedllm-peft` was run — so the `llm_train` gap is established from the code path, not
from an observed silent failure.

Roshan5105labs:
### LFX Pre-Test Review — PR #810

## Review Scope

I reviewed the current PR head a985e4a.

The PR introduces worker-failure propagation in the shared federated-learning controller by wrapping client training, recording worker exceptions, and raising a RuntimeError from the coordinating thread after all workers have joined.

This directly addresses the failure-propagation problem described in #809.

## Positive Findings

The repair boundary is appropriate.

Worker-thread exception propagation is a responsibility of the shared FederatedLearning controller rather than an Example-local concern, because the coordinator must decide whether a federated round is complete before aggregation or later processing continues.

The PR therefore places the fix at the correct abstraction level.

## Cross-PR Interaction

While reviewing the current eligible PRs together, I found substantial overlap with PR #802.

Both #802 and #810 independently modify the shared FederatedLearning.train() path to:

catch exceptions raised during client training;

associate the exception with the corresponding client ID;

wait for worker completion;

raise from the coordinating thread when one or more clients fail.

I compared the current heads of #802 and #810 using Git merge analysis.

The current revisions modify overlapping regions of the shared federated implementation and produce merge-conflict markers when combined.

image
This means the overlap is both conceptual and mechanical: both PRs currently attempt to own the same Core responsibility.

## Test Coverage

The current #810 diff modifies the shared federated implementation but does not add a dedicated regression test file for this behavior.

PR #802, by comparison, currently includes a worker-failure regression test alongside its FCIL tests.

Whichever implementation becomes canonical should retain explicit regression coverage for worker-failure propagation.

## Verification

I attempted direct execution of the modified federated path in a clean Python 3.8 environment with the bundled Sedna 0.6.0.1 package and Ianvs requirements installed.

Execution was blocked during Ianvs import by a Windows ONNX DLL initialization error before the federated code under review was reached.

Therefore, I do not claim runtime verification of the #810 implementation itself. My conclusions are based on source inspection, test inspection, and the cross-PR Git merge analysis.

## Recommendation

Classification: Major revision / coordination required.

The Core-level design direction is valid, but the current PR now overlaps substantially with #802 and conflicts with it mechanically.

I recommend that maintainers first decide which PR should own generic worker-failure propagation in the shared federated controller, retain or add regression coverage in that canonical implementation, and rebase or adjust the other PR to remove duplicated Core behavior.

PR #811, opened 2026-08-15
perf(cli): defer BenchmarkingJob import so -v and --help are fast

Laithamr05:
**What type of PR is this?**

/kind cleanup

`ianvs -v` prints the version and exits, but takes roughly 5 to 7 seconds to do so on a warm install. `ianvs --help` is the same.

The cause is structural rather than a bug. `core/cmd/benchmarking.py` imports `BenchmarkingJob` at module level, and that single import transitively loads the whole benchmarking stack before `main()` is ever entered: `testenvmanager`, `testcasecontroller` and the full paradigm tree, `storymanager` with `matplotlib.pyplot`, plus `sedna` and `onnx`.

Both `-v` and `--help` are argparse actions that print and call `sys.exit()` **during** `parse_args()`. So for those two flags the entire import cost is paid and then discarded.

Measured with `python -X importtime`, cumulative:

```
    7.49 s  core.cmd.benchmarking
    7.18 s  core.cmd.obj.benchmarkingjob
    2.59 s  core.testenvmanager.testenv
    2.45 s  core.testcasecontroller.algorithm.paradigm
    2.13 s  core.storymanager.rank
```

This PR moves that one import inside `main()`, after `parse_args()` has returned. Six lines in one file.

**Result**

| Command | Before | After |
|---|---|---|
| `ianvs -v` | 5.06 s, 7.02 s | 0.44 s, 0.32 s, 0.29 s |
| `ianvs --help` | comparable | 0.29 s |

**Notes for reviewers**

- The normal `-f` path is unchanged. `ianvs -f /nonexistent.yaml` still prints `not found benchmarking config(/nonexistent.yaml) file in local`, exactly as before. The import simply happens a few microseconds later in the same call.
- A function level import raises pylint `C0415 import-outside-toplevel`, which would fail the lint job. The diff therefore carries a targeted `# pylint: disable=import-outside-toplevel` on that single line, with a comment explaining why the import is deferred. `pylint --max-positional-arguments=10 core` scores 10.00/10, exit code 0, identical to the unpatched baseline.
- Related in spirit but distinct: #450 covers an unconditional `import onnx` in `multiedge_inference.py` that fails for users who do not need it. That is about an import breaking; this is about an import being unnecessary at CLI startup. Different module, different symptom.

Tested on Python 3.14 with the repository `requirements.txt`.

None. Filed directly as a small self contained improvement.


akshhkaushik (commented):
### Accept after adding two small CLI tests

The lazy import works for the two paths this PR targets. I executed the patched CLI for `-v` and `--help`; both exited zero, printed the expected output, and did not import `core.cmd.obj`:

```text
ARGV -v EXIT 0 OBJ_IMPORTED False
ARGV --help EXIT 0 OBJ_IMPORTED False
```

That isolates informational CLI commands from Sedna/ONNX/matplotlib import failures without changing the `-f` job path.

Recommendation: merge after adding subprocess tests for both flags. Assert exit code/output and run them in an environment where importing `core.cmd.obj` would fail or assert the module is absent, as above. That prevents a future top-level import from silently restoring the startup dependency.

Verification boundary: executed both real argparse exit paths; did not run a benchmarking config.

Yahdel:
Focused review from the command-startup side:

I reproduced the intended behavior for this PR's Core command with benchmark-only imports blocked. Both `python -m core.cmd.benchmarking -v` and `--help` completed without reaching `core.cmd.obj`, Sedna, ONNX, matplotlib, or prettytable. A normal `-f` command still reached `core.cmd.obj`, which is the expected behavior: the benchmark runtime is deferred until it is actually needed.

One integration point still matters. Ianvs also has a root-level `benchmarking.py` entry point. With #811 alone, the Core module command passed, but the root command still reached the blocked import. PR #672 makes the wrapper use this fixed Core command. With #672 and #811 together, version and help succeeded from both entry points.

I recommend merging this focused change with a small regression test that covers `-v` and `--help` while a benchmark-only import is unavailable. If the root wrapper remains supported, test that path too or land #672 alongside this PR.

Verification boundary: I tested the two informational commands and a deliberate runtime-loading probe. I did not run a full benchmark workload. Full context: Discussion #961.

Laithamr05:
Thanks for the careful review, and for actually executing both exit paths rather than reading the diff.

Added in ff1a715: `tests/test_cli_lazy_import.py`, one test per flag. Each runs the entry point in a subprocess and asserts three things: exit code 0, the expected stdout (the version string, and `usage: ianvs`), and that `core.cmd.obj` is absent from `sys.modules` afterwards. The subprocess reports the exit code and the import status as JSON on stderr, so stdout stays exactly what a user would see.

I checked that the guard can actually fail, since a regression test that cannot go red is not worth much. Temporarily restoring the top-level import turns both tests red with a message pointing at the cause:

```
AssertionError: True is not false : `ianvs -v` imported core.cmd.obj; the import in
core/cmd/benchmarking.py must stay inside main() so informational flags do not load
sedna, onnx and matplotlib.
```

The suite also goes from 0.13s to 3.23s under that revert, which is the startup cost this PR removes showing up directly in the tests.

Two notes on how it is wired in:

- It uses only `unittest` and `subprocess`, so it adds no test dependency to a repo that does not currently have one, and it parses under the 3.7 grammar for the 3.7 to 3.9 CI matrix.
- I added a `Run unit tests` step to `.github/workflows/main.yaml`, because the guard only prevents a future regression if CI runs it. That is the one part of this change that goes slightly beyond the review comment, so if you would rather keep the workflow untouched and land the tests alone, say so and I will drop that hunk.

`pylint --max-positional-arguments=10 core` is still 10.00/10, and the new files are 10.00/10 as well.

Laithamr05:
Follow-up, since I force-pushed and two things changed underneath this PR.

**Rebased onto latest main.** `main` moved 11 commits and the CI layout was reworked while this was open: `.github/workflows/main.yaml` is gone, split into `static_code_pylint_cicd.yaml.yaml`, `dynamic_code_cicd.yaml` and friends. That deletion collided with my earlier workflow edit and marked the PR conflicting. It is rebased and mergeable again, and the pylint job is untouched.

**I dropped the workflow change.** Git followed the rename and reapplied my `Run unit tests` step into `static_code_pylint_cicd.yaml.yaml`, but that workflow is explicitly the static analysis one, and running a subprocess test suite there cuts against the static and dynamic split you have just introduced. I would rather not guess at that, so this PR now touches only:

```
core/cmd/benchmarking.py
tests/__init__.py
tests/test_cli_lazy_import.py
```

If you want the guard to actually run in CI, and it is only a regression guard if it does, this is the whole step, to drop wherever it belongs in the new layout:

```yaml
      - name: Run unit tests
        run: |
          python -m unittest discover -s '${{github.workspace}}/tests' -t '${{github.workspace}}' -v
```

It needs no new dependency, since the tests use only `unittest` and `subprocess`.

**Also fixed:** the commit message tripped `do-not-merge/invalid-commit-message`, because I had referenced the reviewer and the PR number in the commit body. Both are removed and the label has cleared.

Re-verified after the rebase: both tests pass in 0.13s, `pylint --max-positional-arguments=10 core` is 10.00/10 exit 0, the new files are 10.00/10, and `ianvs -f /nonexistent.yaml` still fails with `not found benchmarking config(/nonexistent.yaml) file in local`, so the job path is unchanged.

Ready for another look.

Laithamr05:
@Yahdel thanks, the root entry point was a real gap and you were right about it. I reproduced it before changing anything:

```
python benchmarking.py -v   ->  exit 0, core.cmd.obj imported: True   (0.94s)
python -m core.cmd.benchmarking -v -> exit 0, core.cmd.obj imported: False  (0.05s)
```

So the deferral in this PR genuinely does not reach the root file, because that file is not a wrapper. It is a near verbatim copy of the pre-patch CLI carrying its own top-level `from core.cmd.obj import BenchmarkingJob`.

**I have not edited `benchmarking.py` here.** PR #672 is open, mergeable, and reduces that file to exactly the four lines that make it delegate. Patching the same file in this PR would collide with someone else's work for no benefit, and the fix belongs there rather than here.

What I did instead: the root path is now covered by two more tests that **skip while the standalone copy remains and activate on their own once it delegates**. The skip condition parses `benchmarking.py` and looks for a module level `core.cmd.obj` import, so nothing has to be remembered or rewired when #672 lands.

I verified that this actually flips rather than staying dormant, by locally replacing the root file with the four line version from #672:

- with the copy in place: `Ran 4 tests, OK (skipped=2)`
- simulating #672 merged: `Ran 4 tests, OK`, all four active
- simulating #672 merged **and** reverting the lazy import: `Ran 4 tests, FAILED (failures=4)`, 0.25s to 4.39s

So after #672 both entry points are guarded, and if either regresses the suite goes red.

Current state: 2 pass, 2 skip with the reason printed. `pylint --max-positional-arguments=10 core` 10.00/10 exit 0, new files 10.00/10, parses under the 3.7 grammar for the 3.7 to 3.9 matrix, and `-f` behaviour is unchanged.

One note for @akshhkaushik as well: the earlier `do-not-merge/invalid-commit-message` was mine, from naming a reviewer and the PR number inside a commit body. Both are gone and the label has cleared.

PR #813, opened 2026-08-16
test(ianvs): add core utility unit tests

Ganesh-403:
**What type of PR is this?**

**What this PR does / why we need it**:

This PR adds foundational unit tests for Ianvs core utilities to improve test coverage and establish a testing baseline.

The added tests cover deterministic functions in `core/common/utils.py`:

- `is_local_file`
- `is_local_dir`
- `get_file_format`
- `parse_kwargs`

The tests use pytest fixtures such as `tmp_path` to keep filesystem-related tests isolated and reproducible.

These tests help:
- improve confidence when modifying core utilities
- catch regressions earlier
- provide a foundation for expanding test coverage across Ianvs

**Which issue(s) this PR fixes**:

Fixes #812

PR #815, opened 2026-08-16
ci(ianvs): run pytest in github actions

### LINKED ISSUE #814: Add pytest workflow to run unit tests in CI

Add pytest execution to the existing GitHub Actions CI workflow.

The workflow should:
- install the required test dependencies
- run the existing pytest tests
- report failures when unit tests do not pass

**Why is this needed**:

Ianvs now contains pytest-based unit tests, but they are currently only executed manually by contributors.

Running pytest automatically in CI will:
- catch regressions before merging changes
- provide immediate feedback to contributors
- improve confidence in future code modifications

Ganesh-403:
**What type of PR is this?**

/kind test

This PR integrates pytest execution into the existing GitHub Actions CI workflow.

Previously, CI only performed pylint checks and did not execute the pytest-based unit tests introduced in the repository.

Changes:
- Added `pytest` to CI-only tool installation.
- Added a pytest execution step in `.github/workflows/main.yaml`.
- The workflow safely skips pytest when the `tests/` directory is not available, allowing compatibility with the current branch state while tests are introduced.

Running tests automatically helps detect regressions earlier and improves confidence in future contributions.

Fixes #814

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -27,7 +27,7 @@ jobs:
           sudo apt-get update
           sudo apt-get install libgl1-mesa-glx -y
           python -m pip install pip==24.0
-          python -m pip install  pylint
+          python -m pip install pylint "pytest<8"
           python -m pip install ${{github.workspace}}/resources/third_party/*
           python -m pip install -r ${{github.workspace}}/requirements.txt
       - name: Analysing code of core with pylint
@@ -38,4 +38,11 @@ jobs:
             pylint --max-positional-arguments=10 '${{github.workspace}}/core'
           else
             pylint '${{github.workspace}}/core'
+          fi
+      - name: Run pytest
+        run: |
+          if [ -d "tests" ]; then
+            PYTHONPATH=. python -m pytest tests
+          else
+            echo "tests/ directory not found. Skipping pytest."
           fi


Ganesh-403:
Thanks for the review. Updated the workflow to pin pytest for Python 3.7 compatibility and restricted pytest execution to the dedicated tests directory.

greenfield01:
Related to #813 this PR's pytest step is conditional on a tests/ directory existing, which #813 is what actually creates. Tested both merged together on current main; confirmed the handoff works correctly (full run in my comment on #813).

Two smaller things I noticed reading the workflow file directly:

The Python version matrix (.github/workflows/static_code_pylint_cicd.yaml.yaml) only covers 3.7/3.8/3.9. I ran the test suite locally on 3.10.21 without issue, but CI itself never verifies anything past 3.9, Python 3.7 reached end-of-life mid-2023, so the matrix may be worth revisiting.
CI installs pytest<8 specifically. I ran the same tests locally with pytest 9.1.1 unpinned and they passed fine, not sure if the pin is intentional (compatibility reason not documented in the PR) or just stale.

Neither blocks this PR, just flagging in case they're useful before merge.

Arijit429:

Recommendation: Minor Revision (revised from Major Revision)

Acknowledged: the conditional guard is intentional — this PR is designed to pair with #813 which adds the tests/ directory. The handoff design is correct.

Remaining concern: without an explicit note in the PR body linking to #813 as a dependency, a maintainer merging this standalone would see a CI step that always skips silently. The dependency relationship should be documented in the PR description so merge order is clear.

Also flagging: the Python matrix only covers 3.7-3.9. Python 3.7 is EOL since mid-2023. Worth updating to 3.9/3.10/3.11 before merge.

Yash4616:
Reviewed against `main` at `37a9c60`. Merges clean, and the diff lands on `static_code_pylint_cicd.yaml.yaml`, which is where `.github/workflows/main.yaml` was renamed in `7f7776a`.

The dependency on a `tests/` directory has already been established in this thread and I am not restating it.

One addition on sequencing. There is more than one open PR that would create `tests/`. PR #835, which adds example config path validation, also ships `tests/test_check_example_paths.py`, so it would satisfy the same guard this PR waits on. That matters because #815 and #835 do not merge together cleanly: both remap onto `static_code_pylint_cicd.yaml.yaml` and conflict there. Whichever lands first, the second needs manual resolution, so the pairing that would actually make pytest execute is not a straightforward merge in either order.

Worth flagging for whoever sequences this area rather than treating it as a defect in this PR.

rajdhruvsingh:
Credit to @greenfield01 (tested this PR merged together with its companion tests/-creating PR, confirmed the handoff works; flagged the CI Python matrix stopping at 3.9 and an unexplained pytest<8 pin) and @Arijit429 (confirms the dependency on that companion PR is intentional design, recommends documenting it explicitly in the PR description).

One addition: right now this PR builds test-running infrastructure for a suite that doesn't exist yet on its own — it's correctly designed to go green-but-empty until the companion PR lands. Worth noting for whoever reviews it that once populated, this is also a natural landing spot for promoting any future paradigm-level execution harness into CI as an actual regression gate, not just a pytest runner — currently nothing in this repository exercises the lifelong-learning paradigm's own code path in CI at all.

PR #818, opened 2026-08-17
Fix #445: guard regex match and update deprecated addmm_ signature in M3L basemodel

### LINKED ISSUE #445: [MOT17] M3L basemodel.py: unguarded regex crashes on non-standard checkpoints; deprecated `addmm_` API removed in PyTorch ≥ 2.0

nitbig:
**What type of PR is this?**
/kind bug

This PR fixes two bugs in the M3L ReID algorithm (`testalgorithms/reid/m3l/basemodel.py`) that block the ReID benchmark job from completing:

1. **Unguarded regex crash on non-standard checkpoint filenames**
   `load()` used `re.compile(...).search(model_url).group(1)` without checking
   whether the search actually matched. If a checkpoint filename didn't follow
   the expected `_<arch>.pth` pattern, `.search()` returned `None`, and calling
   `.group(1)` on `None` crashed with an unhandled `AttributeError: 'NoneType'
   object has no attribute 'group'`: a confusing error with no indication of
   what actually went wrong.

   Fix: added a `None` check after the regex match, raising a clear `ValueError`
   that tells the user the filename didn't match the expected pattern.

2. **Deprecated `addmm_` positional signature**
   `_pairwise_distance()` called `addmm_` using a positional argument form
   (`addmm_(1, -2, x, y.t())`) that was removed in PyTorch 2.0+, causing
   `TypeError: addmm_() received an invalid combination of arguments` on any
   modern PyTorch install.

   Fix: updated the call to the current keyword-argument signature
   (`addmm_(x, y.t(), beta=1, alpha=-2)`).

Both fixes are confined to `m3l/basemodel.py` and require no changes to Ianvs core.

**Testing**:
- Verified the file compiles cleanly: `python3 -m py_compile basemodel.py`
- Isolated and ran both fixed methods directly against PyTorch 2.13.0:
  - Regex guard: confirmed a standard filename (`market_resnet.pth`) still
    extracts the architecture correctly, and a non-standard filename
    (`market_best_model.bin`, the exact repro case from the issue) now raises
    a clean `ValueError` instead of crashing with `AttributeError`.
  - `addmm_` fix: confirmed it runs without `TypeError` on PyTorch 2.13.0, and
    cross-checked the output numerically against `torch.cdist` (PyTorch's own
    squared-Euclidean-distance function): max difference was `3e-7`, i.e.
    floating-point noise, confirming the fix is numerically correct and not
    just crash-free.
- Note: this testing isolated the two affected methods, since the full file
  depends on `sedna` and Ianvs's internal `reid` package, which aren't
  installable outside the full Ianvs environment. Full end-to-end run of
  `predict()` also requires a GPU due to a separate, pre-existing
  `self.model.cuda()` call unrelated to this issue: out of scope for this fix.

Fixes #445


akshhkaushik (commented):
### Minor revision — the guarded regex still rejects supported digit-bearing architectures

The `addmm_` keyword-signature update is correct, and raising a named `ValueError` is better than calling `.group()` on `None`. The retained regex is narrower than the bundled model registry:

```python
re.compile(r"_([a-zA-Z]+)\.pth")
```

The same example's registry lists `resnet18`, `resnet34`, `resnet50`, `resnet101`, and `resnet152`. Executing the exact regex shows the mismatch:

```text
model_IBNMeta.pth => IBNMeta
model_resnet50.pth => None
market_IBNMeta.pth.tar => IBNMeta
```

So a checkpoint for a supported digit-bearing architecture receives the new “cannot infer” error. Recommendation: parse the filename suffix against the actual registry identifier alphabet (at least letters, digits, underscore, hyphen), decide explicitly whether `.pth.tar` is supported, and add both cases as tests. The current configured `market_IBNMeta.pth.tar` still matches because the regex is unanchored, which should be intentional rather than accidental.

Verification boundary: executed the exact regex and inspected the in-tree model registry/config; did not load PyTorch weights.

Issue #819, opened 2026-08-17
Simulation Controller (core/testcasecontroller/simulation): 13 verified breakages in the 2022 implementation


Prachi194agrawal:
**What happened**:

I audited the simulation controller added in 2022 through PR #39 while
preparing a proposal for the LFX Mentorship project tracked by #348. I checked
the code under `core/testcasecontroller/simulation/` and
`core/testcasecontroller/simulation_system_admin/` and wrote a read-only
verification script for the findings.

On `main` at commit `36ec008`, the script reports 13 findings. They include
runtime defects, validation gaps, portability limitations, and missing test
coverage:

| ID | Finding | Current status |
|---|---|---|
| B1 | Docker and kind fallback branches are unreachable because `subprocess.run(..., check=True)` raises before the return-code checks | Addressed by open PR #656 |
| B2 | YAML booleans pass the node-count type checks because `bool` subclasses `int` | Addressed by open PR #698 |
| B3 | Unknown simulation keys are silently ignored | Addressed by open PR #698 |
| B4 | Empty required values, including `cluster_name`, pass type validation | Not covered by the open parser PR |
| B5 | CPU detection depends on `lscpu \| grep CPU:` and fails on current `util-linux` output | Addressed by open PR #656 |
| B6 | Build and cleanup use different Sedna branch names (`master` and `main`) | Addressed by open PR #656; both URLs currently resolve, so this is an inconsistency rather than a current outage |
| B7 | The cleanup function is exported but never called, so a completed or failed run can leave its kind cluster behind | Addressed by open PR #745 |
| B8 | The kind installer is pinned to an old amd64-only binary | Addressed in part by open PR #777 |
| B9 | No shipped example contains a `simulation:` block | Open documentation and coverage gap |
| B11 | The Sedna installer does not translate `aarch64` to the KubeEdge release asset name | Upstream portability limitation |
| B12 | Ianvs accepts node counts above the Sedna backend limits | Not found in an existing issue or PR before this report |
| B13 | Ianvs has no peak-memory, CPU-utilization, or wall-time leaderboard metrics | Open feature gap |
| B14 | One failing test case aborts the loop before completed results are returned to the ranking layer | Open fault-containment gap |

The most immediate configuration problem is B12. The Sedna all-in-one script
sets `MAX_CLOUD_WORKER_NODES=2` and `MAX_EDGE_WORKER_NODES=3`, but Ianvs accepts
any integer for `cloud_number` and `edge_number`. For example,
[...]

PR #35, opened 2022-08-31
[OSPP] Proposal for Edge Intelligence Benchmark for Edge-Cloud Collaborative Lifelong Detection

iszhyang:
Proposal for Edge Intelligence Benchmark for Edge-Cloud Collaborative Lifelong Detection

As a very promising community in edge AI Benchmarking, ianvs still has the following problems to be solved

The automatic dataset download and management features are missing, and users still need tedious operations to use the dataset.
Repeated deployment of the edge-cloud collaboration environment is too heavy. The distributed collaborative system simulation feature of ianvs needs to be developed urgently.
This proposal provides features in datasets, edge-cloud collaborative AI system simulation, etc. for the current version of Ianvs.

This PR is a modification of the last [PR](https://github.com/kubeedge/ianvs/pull/24)

PR #820, opened 2026-08-17
Feature/simulation sandbox

Prachi194agrawal:
/kind feature
/kind design

This PR restores the Ianvs simulation-controller work introduced by Zhang Yang during OSPP 2022 in #35 and #39, and turns it into an opt-in sandbox execution path for test cases. It also adds system-resource metrics that can be displayed alongside algorithm metrics in the existing leaderboard pipeline.

The implementation is split into two tiers:

- **Process tier:** the primary implementation in this branch. Each test case runs in a separate worker process with an allowlisted environment, a private temporary workspace, optional per-test-case virtual environment, wall-clock timeout, `RLIMIT_AS` memory limit, and CPU affinity. A test-case failure no longer terminates the Ianvs parent process. With the default `fail_fast: false`, successful results from other test cases are retained.
- **Cluster tier:** an initial orchestration foundation for `kind` + KubeEdge + Sedna. It adds environment lifecycle ownership, image-build and Job-manifest helpers, `kubectl` deployment/deletion, and a ConfigMap result protocol. This path has not been validated on a live cluster and is not presented as production-ready in this PR.

The `sandbox` block is optional. When it is absent or disabled, `TestCaseController.run_testcases()` continues through the legacy inline execution path. The changed failure-containment behavior is therefore limited to users who opt into the sandbox.

**Main changes:**

- Harden simulation configuration parsing:
  - reject booleans as node counts;
  - reject unknown keys instead of silently dropping them;
  - reject empty or invalid RFC 1123 cluster names;
  - normalize the special KubeEdge version value `latest`;
  - validate the Sedna all-in-one backend limit of 1–2 cloud workers and 1–3 edge nodes before provisioning starts.
- Replace fragile legacy host checks with explicit capability reporting based on `shutil.which()`, `os.cpu_count()`/CPU affinity, available memory, and platform checks.
- Move simulation lifecycle ownership into `SimulationEnvironmentAdministrator` and call teardown from a `finally` block after a cluster backend has been provisioned.
- Add the process sandbox, worker protocol, timeout escalation, process-tree profiling, and cleanup of temporary runtimes.
- Add PSS where available, with summed RSS as a recorded fallback, plus CPU time, CPU utilization, wall time, memory headroom, exit status, and OOM classification.
- Expose selected system metrics through the existing ranking and leaderboard path.
- Add the cluster-tier environment and Job administrator foundation.
- Remove `core/testcasecontroller/simulation_system_admin/`, whose responsibilities are superseded by the new environment administrator and sandbox modules.
- Add an annotated example, a detailed design document, 60 simulation-focused tests, and `verify_legacy_simulation.py` for reproducing the legacy findings against a pristine checkout.

**Failure behavior:**

- Sandbox disabled: existing inline behavior is preserved.
- Sandbox enabled with `fail_fast: false`: failed test cases are logged and omitted from the successful-result set; later test cases continue and earlier successful results are preserved.
- Sandbox enabled with `fail_fast: true`: the first failed test case raises after sandbox cleanup.

**Verification performed:**

```text
$ .venv/bin/python -m pytest tests/simulation -q
............................................................             [100%]
60 passed in 0.19s

$ .venv/bin/python -m pylint core --max-positional-arguments=10

$ git diff --check main...HEAD
# no output
```

These checks were run locally with Python 3.10.19. The synthetic memory-limit run documented in the proposal demonstrates a 2 GiB allocation being rejected under a 256 MiB `RLIMIT_AS` limit while a second test case succeeds, but that scenario is not yet part of the automated test suite.

**Known limitations and follow-up work:**

- cgroup v2 enforcement is not implemented. Memory enforcement currently uses `RLIMIT_AS`, which limits virtual address space rather than resident memory and can be too restrictive for ML runtimes.
- CPU affinity restricts the worker to a set of CPUs; it is not equivalent to a fractional CPU-time quota. Fractional `cpus` values are therefore not precisely enforced.
- The cluster tier has not been tested end to end. The current branch still needs a complete worker image build context, a way to load or publish the image for `kind`, in-container worker dependencies, Kubernetes RBAC for result ConfigMaps, and validation on supported KubeEdge/Sedna versions.
- Cleanup is guaranteed after a successfully provisioned cluster, but cleanup after a partially failed Sedna installer run still needs to be hardened.
- Failed-test profiles are logged but are not yet persisted as leaderboard rows.
- A real Ianvs paradigm has not yet been run end to end through the process sandbox in this branch.
- `verify_legacy_simulation.py` is an evidence script for a pristine legacy checkout; it is not yet a post-fix CI regression test.

Because #526 is an active proposal in the same design area, this PR is intended to make the implementation and trade-offs concrete for review. I am happy to split the process tier, legacy fixes, metrics integration, and cluster work into smaller PRs after maintainer direction.

The design rationale, legacy-defect table, configuration contract, and roadmap are documented in:

`docs/proposals/simulation/sandbox-engine/ianvs-simulation-sandbox.md`

**Which issue(s) this PR is related to:**

Related to #819
Related to #348


IIITManjeet (commented):
Nice PR. The process-level approach fits much better here than the old 2022 `kind`-only path, and I like that `RLIMIT_AS` is documented as a fallback rather than being treated as exact accounting.

I went through the profiler's exit paths against `279a5c0` and found two things worth looking at.

## 1. A timeout kill can be reported as OOM

`detect_oom_kill()` treats `-9`/`137` as OOM:

```python
_OOM_EXIT_CODES = {-9, 137}
```

But `SIGKILL` can also come from the sandbox itself. On timeout, `_kill_process_group()` sends `SIGTERM`, waits 5 seconds, then escalates to `SIGKILL`.

Normally the worker exits on `SIGTERM` (`-15`), so this is not every timeout. The issue is when the worker survives the full grace period—for example, while stuck in I/O or a driver call. That `-9` is then reported as an OOM kill.

The bigger issue is that the timeout message:

```python
profile.error = f"test case exceeded its {timeout}s timeout and was killed."
```

gets replaced by the OOM message, and `oom_killed: true` is also sent to the leaderboard.

Since the parent already knows it initiated the timeout kill, I think this can be handled cleanly with a `killed_by_timeout` flag and checked before classifying the exit as OOM. The stderr OOM markers can remain unchanged for genuine kernel OOMs.

One related question: should `RLIMIT_AS` / `MemoryError` be distinguished from a kernel OOM? They are different failure modes, although classifying both as memory-related may be intentional.

## 2. `memory_source` can report `pss` for a mixed total

In `_tree_snapshot()`, `source` becomes `"pss"` as soon as one process provides PSS, but processes that fall back to RSS are still added to the same total:

```python
if pss is not None:
    total_memory += pss
    source = "pss"
else:
    total_memory += proc.memory_info().rss
```

So if PSS lookup fails for part of the tree, the total is a PSS/RSS mix but is reported as `"pss"`.

This only affects the partial-fallback case, but reporting `"mixed"` when any process falls back would make `memory_source` accurate.

Happy to send patches for either of these if useful.

PR #821, opened 2026-08-17
fix(simulation): fix 5 host-environment check bugs in simulation_syst…

iron-prog:
**What type of PR is this?**
/kind bug
/kind test

Fixes 5 bugs in `core/testcasecontroller/simulation_system_admin/simulation_system_admin.py`, the host-environment check module restored from the 2022 simulation proposal (PR #35 / #39). This is groundwork for the LFX Fall 2026 Term 3 project "KubeEdge-Ianvs Simulation Sandbox: Environment-Isolated Execution" (#348).

`@RudraDudhat2509` and `@DsThakurRawat` identified these 5 breakages first, in comments on #348 — this PR implements tested fixes for all five and credits their analysis directly:

1. `check_host_docker()` silently reported success even when Docker was unreachable — piping through `head` meant `check=True` validated `head`'s exit code, not Docker's. Fixed by calling `docker version` directly.
2. `check_host_kind()` had the same root cause with the opposite symptom: it crashed instead of reaching its own auto-install fallback. Same fix applied.
3. `get_host_number_of_cpus()` crashed on modern util-linux, which emits `CPU(s):` rather than the `CPU:` label the grep pattern expected. Fixed the pattern and anchored it so it can't match the `NUMA node0 CPU(s):` line instead.
4. Both host-check parsers were reading the raw `repr()` of a `bytes` object instead of decoded text. Now both decode explicitly via `.decode("utf-8")`.
5. The Sedna installer URL disagreed between build (`master`) and teardown (`main`). Introduced a single `SEDNA_INSTALL_SCRIPT_URL` constant used by both, so they can't drift apart again.

All 5 bugs were reproduced live before writing the fix, not inferred from reading the code alone.

**Verification**:

Added `tests/test_simulation_system_admin.py` — 12 unit tests via `unittest.mock`, one or more per bug. Verified against the pre-fix code that 8 of 12 fail there, confirming these are real regression tests, not tautologies.

```
$ python -m unittest tests.test_simulation_system_admin -v
...
Ran 12 tests in 0.006s

OK
```

`pylint core/testcasecontroller/simulation_system_admin/simulation_system_admin.py` → 10.00/10, matching CI's `pylint core/` check.

**Scope notes**:

- Left the `enviroment`/`destory` typos in existing function names alone — fixing them would change the public interface consumed by `core/cmd/obj/benchmarkingjob.py` and inflate this diff beyond the 5 reported bugs. Happy to do a separate cosmetic PR if maintainers want it.
- The install fallbacks (`curl | bash`) still check the pipeline's exit code rather than curl's specifically — a latent instance of the same bug class as #1/#2, but outside the 5 identified breakages and not exercised by this PR's tests. Flagging it here rather than leaving it for a future contributor to rediscover.

Related to #348


f2025376141-arch:
### Supplementary Review — PR #821

**Review Depth: Architectural Insight**

PR #821 improves host-environment validation for the simulation infrastructure. My additional concern is the reliability of the installation fallback path.

If the download step fails but a later command determines the final status, an incomplete installation could potentially be treated as successful. This creates a restoration risk because Example execution depends on the environment being completely and correctly prepared.

I recommend explicitly checking the download command before executing the downloaded content and adding a regression test for a failed download. This would make the setup fail clearly instead of allowing a partial installation to continue.

This finding is distinct from the host-environment defects already addressed by this PR because it focuses on failure propagation in the installation fallback.

**Verification Boundary:** This review is based on the PR implementation and documented behavior. I do not claim a live installation or simulation-cluster reproduction.

Issue #812, opened 2026-08-16
Add foundational unit tests for Ianvs core utilities


Ganesh-403:
**What would you like to be added/modified**:

Add foundational pytest-based unit tests for Ianvs core utilities

The initial tests will focus on deterministic and isolated components such as:
- utility functions
- input validation logic
- error handling paths
- core module behavior that does not require external services

The test suite should follow the existing repository structure and avoid integration tests or network-dependent scenarios.

**Why is this needed**:

Currently, Ianvs core modules have limited unit test coverage, making it harder to verify behavior and prevent regressions during future development.

Adding focused unit tests will:
- improve confidence when modifying core components
- catch regressions earlier
- provide a foundation for expanding test coverage across the project

PR #638, opened 2026-08-02
fix(dataset): resolve tempfile leaks, ZeroDivisionError in splitting, and file IO guardrails

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Fixes several systemic defects in the dataset management subsystem (`core/testenvmanager/dataset/dataset.py` and `core/testenvmanager/testenv/testenv.py`):

1. **Resource Leak & Temp Directory Lifecycle**:
   Updated dataset index processing (`_process_txt_index_file`, `_process_index_file`, `process_dataset`) and dataset splitting methods to accept workspace output directory contexts. When `output_dir` is provided, intermediate index files and split datasets are written directly to managed workspace directories (`os.makedirs(output_dir, exist_ok=True)`) rather than leaking untracked directories in system `/tmp`.

2. **Defensive Validation for Splitting Algorithms (`ZeroDivisionError`)**:
   Added input parameter validation in `Dataset.split_dataset()`:
   - Validates `times >= 1` and `0.0 < ratio < 1.0`.
   - Enforces `times >= 2` for `city_splitting` (preventing `times - 1` from reaching `0` and causing a `ZeroDivisionError`).
   - Validates supported splitting methods explicitly.

3. **File I/O Guardrails**:
   - `_write_data_file()` automatically creates parent directories (`os.makedirs(os.path.dirname(data_file), exist_ok=True)`) before writing files to prevent `FileNotFoundError`.
   - `_read_data_file()` and `_write_data_file()` throw explicit `ValueError` exceptions for unsupported data formats instead of returning `None` and triggering `TypeError`.

4. **Unit Tests**:
   Added `tests/test_dataset.py` to test dataset parameter validation, splitting algorithms (`default`, `city_splitting`), parent directory creation, and error handling.

**Which issue(s) this PR fixes**:

Fixes #637

**Special notes for your reviewer**:

NONE

**Does this PR introduce a user-facing change?**:

```release-note
Fix tempfile resource leaks in dataset splitting, prevent ZeroDivisionError in city_splitting, and add missing parent directory creation during dataset file writing.
```

PR #825, opened 2026-08-18
fix(core/lifelong): fail fast when splitting_method under-produces dataset splits

### LINKED ISSUE #822: [Bug] Lifelong learning: `splitting_method: "default"` produces `rounds` dataset splits but `run()` indexes `rounds + 1`, raising IndexError on the final round
**What happened**:

`LifelongLearning.run()` requests `rounds` dataset splits and then iterates over `rounds + 1` of them. The four splitting methods in `core/testenvmanager/dataset/dataset.py` disagree about how many splits they return, so whether a run survives depends on which `splitting_method` the algorithm config happens to use.

Both the `no-inference` branch (line 93) and the `hard-example-mining` branch (line 167) do this:

```python
rounds = self.incremental_rounds
...
dataset_files = self._split_dataset(splitting_dataset_times=rounds)
...
for r in range(rounds + 1):
    train_dataset_file, eval_dataset_file = dataset_files[r]
```

`_split_dataset` forwards `times=rounds` to `Dataset.split_dataset`, which dispatches on `splitting_method`:

| `splitting_method` | dispatches to | appends before the loop? | splits returned |
|---|---|---|---|
| `default` | `_splitting_more_times` | no | **`times`** |
| `city_splitting` | `_city_splitting` | yes | `times + 1` |
| `fwt_splitting` | `_fwt_splitting` | yes | `times + 1` |
| `hard-example_splitting` | `_hard_example_splitting` | yes | `times + 1` |

All four share the same `while index <= times:` loop body, but three of them append one extra entry *before* that loop and `_splitting_more_times` does not. So with `splitting_method: "default"` the loop indexes one past the end:

```
splitting_method=default                produced=2  loop needs=3  -> IndexError at r=2
splitting_method=fwt/city/hard-example  produced=3  loop needs=3  -> ok
```

The failure lands on the last iteration, so every preceding round has already run its training before the job dies. It also surfaces as a generic `RuntimeError("(paradigm=lifelonglearning) pipeline runs failed, error: list index out of range")` from `TestCase.run`, which does not point at the config field that caused it.

`"default"` is also the fallback in `Algorithm.__init__` (`core/testcasecontroller/algorithm/algorithm.py:66`), so an algorithm.yaml that omits `splitting_method` entirely gets the failing path.

The number of splits returned should be consistent across splitting methods for a given `times`, or `run()` should validate `len(dataset_files)` against the number of rounds it intends to iterate and fail early with a message naming `splitting_method` — rather than raising `IndexError` after the final round of training has completed.

Run any lifelong-learning benchmark that combines `mode: "no-inference"` (or `"hard-example-mining"`) with `splitting_method: "default"`. Shipped examples in this state on `main` (36ec008):

- `examples/robot/lifelong_learning_bench/semantic-segmentation` — `testalgorithms/rfnet/rfnet_algorithm.yaml`, `rfnet_algorithm-simple.yaml`, `sam_algorithm.yaml`, `sam_vit_algorithm.yaml`, `vit_algorithm.yaml`
- `examples/cityscapes-synthia/lifelong_learning_bench/semantic-segmentation` — `testalgorithms/rfnet/rfnet_algorithm.yaml`, `rfnet_algorithm-new-split.yaml`
- `examples/robot-cityscapes-synthia/lifelong_learning_bench/semantic-segmentation` — `testalgorithms/erfnet/test_algorithm.yaml`

The mismatch can also be seen without a dataset, by comparing the split counts each method produces against the `range(rounds + 1)` the caller iterates.

Relevant code:
- `core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py:94,100` and `:168,174`
- `core/testenvmanager/dataset/dataset.py:223` (`default` -> `_splitting_more_times`), `:299`, `:346`, `:404`, `:477`

I checked open and closed issues and PRs for prior reports and did not find this one. The nearby reports cover different defects in the same area, so listing them to be explicit:

- #591 / #592 — `_city_splitting` `ZeroDivisionError` when `times=1`; different method, different failure.
- #321 — reports two `IndexError`s, both elsewhere: `task_remodeling.py:78` (`self.models[0]` on an empty list) and `accuracy.py:50` (tuple indexing).
- #106 — a traceback through `_split_dataset`, but the failure is `os.makedirs` inside `dataset_output_dir()`.
- #367 / #476 — BDD curb-detection reproducibility; covers the separate `mode: "multi-inference"` handler gap, not the split count.

I would like to work on this.

/assign

IIITManjeet:
**What type of PR is this?**

/kind bug

`LifelongLearning.run()` requests `rounds` dataset splits and then iterates `range(rounds + 1)` in the `no-inference` and `hard-example-mining` branches, because those flows index a round-0 baseline entry.

The four splitting methods do not agree on how many splits they return for a given `times`. `_fwt_splitting`, `_city_splitting` and `_hard_example_splitting` each append a round-0 baseline *before* their shared `while index <= times:` loop and return `times + 1`. `_splitting_more_times` — which `method == "default"` dispatches to — has only the in-loop append and returns `times`:

| `splitting_method` | dispatches to | round-0 baseline | splits returned |
|---|---|---|---|
| `default` | `_splitting_more_times` | no | `times` |
| `city_splitting` | `_city_splitting` | yes | `times + 1` |
| `fwt_splitting` | `_fwt_splitting` | yes | `times + 1` |
| `hard-example_splitting` | `_hard_example_splitting` | yes | `times + 1` |

So under `default` the loop indexed one past the end and raised `IndexError: list index out of range` on the **final** round — after every earlier round had already completed its training — surfacing only as a generic `RuntimeError("(paradigm=lifelonglearning) pipeline runs failed, ...")` from `TestCase.run`, which does not name the configuration field responsible.

`"default"` is also the fallback in `Algorithm.__init__`, so an algorithm.yaml that omits `splitting_method` gets the failing path. Eight shipped configs across three examples are currently in this state (listed in #822).

**Approach**

The splitters cannot simply be aligned to all return `times + 1`, because two other paradigms depend on the current `default` count:

- `IncrementalLearning.run()` iterates `range(1, rounds + 1)` indexing `dataset_files[r - 1]`
- `FederatedClassIncrementalLearning.run()` iterates `range(self.incremental_rounds)`

Both require exactly `times` entries, so changing `_splitting_more_times` would break them.

This PR therefore keeps the change inside `LifelongLearning`: `_split_dataset` gains an optional `required_splits` argument, and the two branches that index a round-0 baseline pass `rounds + 1`. When the configured splitting method cannot supply that many, the run now fails **before any training starts**, with a message naming `splitting_method` and both the required and produced counts.

The `multi-inference` branch indexes `dataset_files[r - 1]` and needs only `rounds` entries, so it is deliberately left unchanged.

Behaviour, using the same shapes the real splitters produce:

```
  splitting_method=default
    before fix: IndexError at r=2  <-- after all prior rounds trained
    after  fix: RuntimeError before training -> lifelong learning needs 3 dataset
                splits for 2 incremental rounds, but splitting_method produced 2.

  splitting_method=fwt/city/hard-example
    before fix: ran to completion
    after  fix: proceeds
```

Configurations that work today are unaffected.

**A question for maintainers**

This PR turns a late, opaque crash into an early, actionable one, but it does not make `default` usable with these two modes. Making those eight shipped configs run needs a decision that felt like yours rather than mine, since it changes what data each round sees:

1. **Give `default` a round-0 baseline too**, in a way scoped to lifelong learning so `IncrementalLearning` and `FederatedClassIncrementalLearning` keep receiving `times` entries.
2. **Change the shipped configs** to a splitting method that already supplies the baseline (`fwt_splitting` / `city_splitting` / `hard-example_splitting`).

Option 1 alters the per-round data distribution and therefore published benchmark numbers; option 2 leaves `default` permanently incompatible with these modes. Happy to implement either in a follow-up — please let me know which you prefer.

Fixes #822

**Does this PR introduce a user-facing change?**

A lifelong-learning job that combines `splitting_method: "default"` with `mode: "no-inference"` or `"hard-example-mining"` now fails immediately with an explanatory error, instead of raising `IndexError` after the final round of training. No behaviour change for runs that succeed today.

**Testing**

`pylint` on the changed file with the CI settings (`--max-positional-arguments=10`) reports 9.80/10, unchanged from `main` — the only remaining message is the pre-existing `sedna` import error, which CI resolves by installing `resources/third_party/*`.

I have not added a unit test here because the repository has no `tests/` tree on `main` yet and #812 / #813 / #814 / #815 are actively introducing one. Happy to add coverage for this once that lands, or in this PR if you would rather not wait.


AdityaKumarSethia:
Reviewed alongside #638 and the unguarded-division cluster covered in [Discussion #873](https://github.com/kubeedge/ianvs/discussions/873), noting this is adjacent but a genuinely different bug, not folded into that cluster's Root Problem.

Confirmed the described mismatch is real: `"default"` (`_splitting_more_times`) returns `times` splits, while `city_splitting`/`fwt_splitting`/`hard-example_splitting` each prepend a baseline chunk and return `times + 1`. `run()`'s `no-inference` and `hard-example-mining` branches both call `_split_dataset(splitting_dataset_times=rounds)` and then index up to `dataset_files[rounds]`, which under-produces exactly when `"default"` is the configured method, this PR's `required_splits=rounds + 1` check catches that mismatch before the indexing happens.

Worth flagging one adjacent case while reviewing: `_splitting_more_times` (the method whose output this PR is validating) has its own separate, unguarded `times=0` crash (`ZeroDivisionError`, not the `IndexError`/short-list case this PR addresses), currently latent, not triggered by any committed config. That one's addressed by #638's input validation, not by this PR, since this PR checks the result of splitting, not the `times` parameter going in. The two are complementary, not overlapping, worth landing together rather than either alone, since #638 stops a bad `times` value early and this PR catches the separate return-count mismatch that can still occur even with a valid `times`.

rajdhruvsingh:
Read the full diff against upstream/main@37a9c60a9747af0cfe3170f84249bc349c56e8d5. This is a well-built fail-fast guard with one framing problem and one untested boundary. Recommendation: minor revision, then accept.

What it does (verified): adds optional required_splits to _split_dataset(), passed as rounds + 1 from the no-inference and hard-example-mining branches; raises a named RuntimeError naming the offending splitting_method before training starts if the splitter under-produces. Caller-safe: the parameter defaults to None and the implicit-default mode branch is untouched.

What it does not do (verified by grep over the diff): there is no change to range(rounds + 1) or any dataset_files[...] access — the off-by-one indexing behind #822 remains in the code. This PR converts a late IndexError (after earlier rounds have trained) into an early actionable error; it does not make default usable. Credit where due: the commit message says exactly this — the PR doesn't oversell itself in its own text, and the guard's error message helpfully names which splitters supply the round-0 baseline.

Framing correction — the affected-config census. The description repeats #822's figure ("Eight shipped configs across three examples"). A repo-wide sweep of active (uncommented) splitting_method: "default" in lifelong-learning examples finds four:

cityscapes-synthia/scene-based-unknown-task-recognition/curb-detection/.../rfnet_algorithm.yaml   ← not in #822's list
cityscapes-synthia/lifelong_learning_bench/semantic-segmentation/.../rfnet_algorithm.yaml          ← in #822's list
cityscapes-synthia/lifelong_learning_bench/curb-detection/.../rfnet_algorithm.yaml                  ← not in #822's list
bdd/lifelong_learning_bench/curb-detection/.../yolo_algorithm.yaml                                  ← not in #822's list

Six of the eight configs #822 lists ship with the key commented out or re-pointed on current main (all five robot/* algorithm yamls use hard-example_splitting/fwt_splitting; rfnet_algorithm-new-split.yaml uses city_splitting; robot-cityscapes-synthia's erfnet yaml uses fwt_splitting). The census cuts both ways: smaller than claimed and missing three live hits. Your own body makes the stronger argument anyway — "default" is the Algorithm.__init__ fallback for omitted keys, so no static census can bound the affected set; that observation argues for keeping exactly this kind of runtime guard rather than config-list-based reasoning. Suggest updating the description's count to the verified four (or citing the fallback argument alone).

Untested edge case: rounds == 0 (required_splits=1) isn't addressed by the PR and wasn't executed in my testing either — worth one sentence or one assertion in a test before merge.

AdityaKumarSethia:
Good catch on the census discrepancy, that number wasn't something I'd cited, but it's useful to have corrected regardless since it'll come up again. Agree the runtime guard is the more defensible approach exactly because a static config count can't bound this, `"default"` being the fallback for any omitted key means the affected set isn't enumerable from configs alone. The `rounds == 0` gap is worth a one-line test before merge, as you say.

f2025376141-arch:
### Supplementary Review — PR #825

**Review Depth: Architectural Insight**

PR #825 improves Lifelong Learning validation by detecting insufficient dataset splits before the previous late `IndexError` occurs.

My additional concern is regression coverage across the different split-count contracts used by the supported configurations. The PR distinguishes between paths requiring `times` splits and branches requiring `rounds + 1` entries because they use a round-0 baseline.

I recommend adding parameterized tests covering the different splitter and consumer combinations, including the default path, city splitting, FWT splitting, hard-example splitting, and the multi-inference path. The tests should verify the expected split-count contract rather than only checking that an error is raised.

This is distinct from the existing fix because it focuses on preventing future changes from breaking another paradigm's expected splitter contract.

**Alternative explanation:** globally changing every splitter to return `times + 1` could appear simpler, but this could change the existing contract for paradigms that intentionally consume `times` entries. Consumer-side validation is therefore the safer boundary.

**Verification Boundary:** This review is based on the PR implementation and documented split-count behavior. I do not claim full benchmark or dataset execution.

AdityaKumarSethia:
Traced the actual call sites to map out the contract you're describing, three consumers, confirmed on `main`:
```
core/testcasecontroller/algorithm/paradigm/federated_learning/federated_learning.py:130 -> _split_dataset(splitting_dataset_times=split_time)
core/testcasecontroller/algorithm/paradigm/incremental_learning/incremental_learning.py:82 -> _preprocess_dataset(splitting_dataset_times=rounds)
core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py:94,168,264 -> _split_dataset(splitting_dataset_times=rounds), three separate call sites across the no-inference, hard-example-mining, and third mode branches
```
All three currently call with `times`-shaped expectations, none currently pass anything indicating a `rounds + 1` baseline expectation, that distinction lives entirely in `Dataset.split_dataset()`'s internal dispatch (`city_splitting`/`fwt_splitting`/`hard-example_splitting` prepend a baseline chunk, `"default"` doesn't), not in how any consumer calls it.

That actually strengthens your point about consumer-side validation being the safer boundary: none of the three call sites currently disambiguate which contract they expect, they all just pass a round count and trust whatever the configured `splitting_method` happens to return. A parameterized test matrix covering these three real call sites against all four splitting methods would catch exactly the class of mismatch this PR is guarding against, and would do it with real coverage data instead of us reasoning about it from the dispatch code alone.

ishansurdi:
I tested the guard by driving `_split_dataset()` directly (dataset stubbed to control the produced count only - the method under test is untouched):
 
```
produced=2 required=None -> OK, returned 2
produced=2 required=3    -> RuntimeError: lifelong learning needs 3 dataset splits for 2 incremental rounds, but splitting_method(default) produced 2. ...
produced=3 required=3    -> OK, returned 3
produced=4 required=3    -> OK, returned 4
```
 
Works as intended and backward compatible with no requirement passed. Note this test is scoped to `_split_dataset()` alone, so it's unaffected by the earlier-firing `#284` crash raised elsewhere in this thread — that's a separate blocker in `run()`'s broader pipeline, not in the method this PR changes.
 
The `produced=4` case is a real gap: the check is `<`, not `!=`, so an over-producing splitter would pass silently and the caller  -  which always consumes exactly `range(rounds + 1)` entries - would just ignore the surplus. None of the four built-in methods over-produces today, so this isn't a reproduced failure, just an untested edge in the contract as written.
 
On the affected-config count being discussed in this thread - I independently paired each active `splitting_method: "default"` config against its testenv's `model_metric.mode`:
 
```
bdd/curb-detection                        mode=multi-inference    (unguarded by design)
cityscapes-synthia/curb-detection         mode=<missing>          (falls to unguarded elif)
cityscapes-synthia/semantic-segmentation  mode=no-inference        (guarded — this PR applies)
cityscapes-synthia/scene-based-curb       mode=<missing>          (falls to unguarded elif)
```
 
This matches the corrected count of four already reached in this thread, and narrows it further: of those four, only the semantic-segmentation config's mode currently reaches a guarded branch. Worth reflecting the `splitting_method` + `model_metric.mode` pairing in the PR description, not just the splitting method alone, since that pairing is what actually
determines reach.

f2025376141-arch:
> I tested the guard by driving `_split_dataset()` directly (dataset stubbed to control the produced count only - the method under test is untouched):
> 
> ```
> produced=2 required=None -> OK, returned 2
> produced=2 required=3    -> RuntimeError: lifelong learning needs 3 dataset splits for 2 incremental rounds, but splitting_method(default) produced 2. ...
> produced=3 required=3    -> OK, returned 3
> produced=4 required=3    -> OK, returned 4
> ```
> 
> Works as intended and backward compatible with no requirement passed. Note this test is scoped to `_split_dataset()` alone, so it's unaffected by the earlier-firing `#284` crash raised elsewhere in this thread — that's a separate blocker in `run()`'s broader pipeline, not in the method this PR changes.
> 
> The `produced=4` case is a real gap: the check is `<`, not `!=`, so an over-producing splitter would pass silently and the caller - which always consumes exactly `range(rounds + 1)` entries - would just ignore the surplus. None of the four built-in methods over-produces today, so this isn't a reproduced failure, just an untested edge in the contract as written.
> 
> On the affected-config count being discussed in this thread - I independently paired each active `splitting_method: "default"` config against its testenv's `model_metric.mode`:
> 
> ```
> bdd/curb-detection                        mode=multi-inference    (unguarded by design)
> cityscapes-synthia/curb-detection         mode=<missing>          (falls to unguarded elif)
> cityscapes-synthia/semantic-segmentation  mode=no-inference        (guarded — this PR applies)
> cityscapes-synthia/scene-based-curb       mode=<missing>          (falls to unguarded elif)
> ```
> 
> This matches the corrected count of four already reached in this thread, and narrows it further: of those four, only the semantic-segmentation config's mode currently reaches a guarded branch. Worth reflecting the `splitting_method` + `model_metric.mode` pairing in the PR description, not just the splitting method alone, since that pairing is what actually determines reach.

Thanks for the detailed review! I agree that the different split-count contracts need regression coverage. I’ll add parameterized tests for the relevant splitter/consumer combinations, including both the "times" and "rounds + 1" cases. I also agree that consumer-side validation is the safer approach.

Thanks for the helpful feedback!

pratikpanda2006:
**Review:**

- Correctly resolves #822 by validating `round_idx < len(dataset_splits)` before indexing and raising a clear error instead of letting the loop hit `IndexError` on the final round.
- Scope: touches `lifelong_learning.py` only — doesn't change the underlying split-count inconsistency between the four `splitting_method` implementations in `dataset.py`, but does prevent the crash symptom for all of them.
- Regression risk: low — this only adds a boundary check, doesn't alter the training/eval logic itself.
- Recommendation: accept as-is.

phantom-712:
The consumer-scoped guard is the correct architectural direction. The
LifelongLearning `no-inference` and `hard-example-mining` branches index an
initial state plus incremental rounds, while other paradigms consume the
splitter output with different indexing assumptions. A global change to the
default splitter would risk changing those consumers' data contracts.

The test matrix should exercise the real call boundary, not only a stubbed
list-length helper:

```text
consumer mode                    splitter                  times=1,2,3
LifelongLearning baseline path   default/city/fwt/hard     assert required count
IncrementalLearning              default and configured    assert existing count
FCIL                             configured default        assert existing count
```

The guard should document whether it rejects only under-production or requires
exact equality. If the caller always consumes exactly `rounds + 1` entries,
an over-producing splitter currently passes a `< required` check and its extra
entry is silently ignored. That is not a reproduced built-in failure, but it is
an untested contract boundary and should be an explicit choice.

The Cityscapes splitter also needs its own parameterized case: it emits an
initial pair and then decrements `times`, so its count should be measured from
the implementation rather than inferred from the generic splitter table.

Recommendation: minor revision. Keep the fail-fast diagnostic, add the
parameterized consumer/splitter tests, correct the Cityscapes wording, and
state clearly that a separate configuration or LL-scoped behavior decision is
still required before the default Example is runnable. The bounded
working-tree run completed the LL path with 12 official-format training
records, 4 test records, three states, and two updates, so the guard's
consumer placement composes with the later stages. This is fixture-level
evidence and not execution of PR #825 as-is or of the full benchmark.

A five-round official-resource Robot RFNet simple run was also completed on the
repaired composed tree. It returned Ianvs code 0, produced a five-column
all-task matrix, and retained 1,641 output files including train and evaluation
indices, predictions, and model artifacts. The observed accuracy was
`0.002232849365268865`. This confirms the consumer-specific count decision in
one real-resource composition, but it is not execution of PR #825 unchanged
and it does not establish comparative benchmark quality.

Issue #595, opened 2026-07-30
[Bug] Configuration field validation in core components bypasses type checks due to logical 'and' in _check_fields()


bhuvan-somisetty:
### Bug Description

In several `core` configuration management modules (such as `Rank`, `BenchmarkingJob`, `Algorithm`, and `Module`), the `_check_fields()` method is used to validate required config parameters and their data types during initialization.

However, the validation expressions use boolean `and` instead of `or` when verifying field presence and data types.

For example, in `core/storymanager/rank/rank.py`:
```python
if not self.sort_by and not isinstance(self.sort_by, list):
    raise ValueError(
        f"rank's sort_by({self.sort_by}) must be provided and be list type."
    )
```

### Problem Analysis

When `sort_by` is provided with an invalid, non-list type (e.g. a string `"accuracy"` or a dictionary `{"acc": "ascend"}`):
1. `not self.sort_by` evaluates to `False` (since non-empty strings/dicts are truthy in Python).
2. `not isinstance(self.sort_by, list)` evaluates to `True`.
3. `False and True` evaluates to `False`.

Because `and` was used instead of `or`, no `ValueError` is raised when an invalid type is passed in the configuration YAML. The invalid type silently bypasses initialization checks and causes obscure runtime exceptions (like `AttributeError` or `TypeError`) much later during pipeline execution.

Additionally, in `Rank._check_fields()`:
```python
if not self.save_mode and not isinstance(self.save_mode, list):
    raise ValueError(
        f"rank's save_mode({self.save_mode}) "
        f"must be provided and be list type."
    )
```
Here, `self.save_mode` is treated as a string throughout the codebase (e.g. `"selected_and_all"`, `"selected_only"`), but `_check_fields` checks it against `list` and prints an inaccurate error message stating it must be of list type.

Similar `and`-based validation logic exists in:
- `core/cmd/obj/benchmarkingjob.py`
- `core/testcasecontroller/algorithm/algorithm.py`
- `core/testcasecontroller/algorithm/module/module.py`

### How to Reproduce

1. Instantiate `Rank` with a configuration dictionary containing an invalid type for `sort_by` (e.g., `Rank({"sort_by": "accuracy"})`).
2. Observe that `_check_fields()` does not raise a `ValueError`.
3. Running downstream ranking operations triggers obscure errors when iterating over string elements instead of dictionary items in a list.

### Expected Behavior

[...]

PR #829, opened 2026-08-18
fix(core/storymanager/rank): validate save_mode value and fail loudly on unsupported modes

### LINKED ISSUE #828: [Bug] Unsupported save_mode silently skips all leaderboard persistence, and the save_mode type check tests for list instead of str
**What happened**:

`Rank.save()` in `core/storymanager/rank/rank.py` dispatches on `save_mode` using three independent `if` statements with no fallback branch, and `_check_fields()` does not validate that `save_mode` holds one of the supported values.

As a result, an unsupported `save_mode` — for example the single-character typo `"selected_and_al"` — passes configuration validation, matches none of the three branches in `save()`, and the benchmarking job **completes with a normal exit status while writing no leaderboard files at all**.

There are two distinct defects.

**1. The `save_mode` type check is incorrect and effectively unreachable**

```python
# core/storymanager/rank/rank.py:88
    raise ValueError(
        f"must be provided and be list type."
    )
```

`save_mode` is declared as `self.save_mode: str = "selected_and_all"` (line 46) and consumed via string comparison (lines 275-285), but the check tests for `list`. Because the `and` short-circuits on the truthiness test, any non-empty value passes regardless of type:

```python

def cfg(sm):
    return {"sort_by": [{"acc": "descend"}], "save_mode": sm,
            "selected_dataitem": {"paradigms": ["all"], "modules": ["all"],
                                  "hyperparameters": ["all"], "metrics": ["all"]}}

Rank(cfg("selected_and_al"))     # accepted
Rank(cfg("SELECTED_AND_ALL"))    # accepted
Rank(cfg(12345))                 # accepted
Rank(cfg(["selected_only"]))     # accepted
```

**2. `save()` silently persists nothing for an unsupported mode**

```python
# core/storymanager/rank/rank.py:275-285
if self.save_mode == "selected_and_all": ...
if self.save_mode == "selected_only": ...
if self.save_mode == "selected_and_all_and_picture": ...
# no else
```

Reproduction of the full user-facing chain:

```python
import os, tempfile
from unittest.mock import patch

out = tempfile.mkdtemp()
r = Rank(cfg("selected_and_al"))          # typo accepted at config time

with patch.object(Rank, "_prepare", lambda self, tc, tr, od: (
        setattr(self, "all_rank_file", os.path.join(od, "all_rank.csv")),
        setattr(self, "selected_rank_file", os.path.join(od, "selected_rank.csv")))):
    r.save([], {}, out)

print(os.listdir(out))                    # [] -> nothing written, no error raised

r.visualization = {"mode": "selected_only", "method": "print_table"}
r.plot()
```

The run reports no error during `save()`. The failure surfaces only afterwards in `plot()`:

```
RuntimeError: process visualization(method=print_table) of rank file(.../selected_rank.csv)
failed, error: [Errno 2] No such file or directory: '.../selected_rank.csv'
```

That message names the visualization method and points away from the actual cause, so a user debugging it is led toward `print_table` rather than the `save_mode` typo. Losing an entire benchmarking run's leaderboard output without an error is worse than a crash, because a long GPU-backed run appears to have succeeded.

- `save_mode` should be validated as a non-empty `str`, matching how it is declared and consumed, and the error message should say "string" rather than "list".
- An unsupported `save_mode` value should be rejected at config-parse time with a message listing the supported values.
- `save()` should never silently skip persistence; an unrecognised mode should raise rather than fall through all branches.

Run either snippet above against `main`, or set `save_mode: "selected_and_al"` under `benchmarkingjob.rank` in any example's `benchmarkingjob.yaml` and run `ianvs -f <that file>`.

- Affected component: `core/storymanager/rank/rank.py`
- The accepted value set is well defined, because it is what `save()` dispatches on: `selected_and_all`, `selected_only`, `selected_and_all_and_picture`. (An earlier version of this line said all three are "used by examples in-tree". That is not accurate — measured across the 47 configs carrying a `benchmarkingjob.rank` block, 45 use `selected_and_all`, 1 uses `selected_and_all_and_picture`, 1 falls back to the class default, and **0 use `selected_only`**. Corrected after review — thanks @azaynul10. The set is defined by the code, not by usage.)
- Enforcing that set cannot break any shipped example: 0 of the 47 configs declare a value outside it, and the one config that passes no `save_mode` to `Rank` falls back to a valid default.

**Relationship to existing issues** — I checked the tracker before filing:

- #659 and #595 cover the `_check_fields()` **type**-checking logic (`and` vs `or`, and `save_mode` being a `str` not a `list`). This report is about the **value** never being validated against the supported set, and about `save()` silently writing nothing — neither is mentioned in those issues, and none of the open PRs touching `_check_fields()` adds value validation or an `else` branch to the `save()` dispatch.
- #559 covers `selected_and_all_and_picture` crashing when the Matrix metric is missing, which is a different code path (`_draw_pictures`).

I have a fix ready and would like to work on this — please assign it to me.

/kind bug

abhishekKokadwar:
**What type of PR is this?**

/kind bug

`Rank.save()` dispatches on `save_mode` using three independent `if` statements with no fallback branch, and `_check_fields()` never validated that `save_mode` holds one of the supported values. An unsupported value — for example the single-character typo `"selected_and_al"` — passed validation, matched none of the branches, and the benchmarking job **completed with a normal exit status while writing no leaderboard files at all**.

The failure then surfaced only later in `plot()` as:

```
RuntimeError: process visualization(method=print_table) of rank file(.../selected_rank.csv)
failed, error: [Errno 2] No such file or directory: '.../selected_rank.csv'
```

which names the visualization method and points away from the actual `save_mode` misconfiguration. Silently discarding an entire run's leaderboard output is worse than crashing, since a long GPU-backed benchmarking run appears to have succeeded.

Separately, the existing type check was both wrong and effectively unreachable:

```python
```

`save_mode` is declared as `str` (line 46) and consumed via string comparison (lines 275-285), but the check tested for `list`. Because the `and` short-circuits on the truthiness test, any non-empty value passed regardless of type — `12345` and `["selected_only"]` were both accepted.

This PR:

- adds `Rank.VALID_SAVE_MODES` as a single source of truth for the three supported modes;
- corrects the type check to `str` (and fixes the error message, which said "list") and validates the value against `VALID_SAVE_MODES`, so a bad mode now fails fast at config-parse time with a message listing the supported values;
- converts the `save()` dispatch to `if`/`elif`/`else`, so an unsupported mode can never silently skip persistence.

**Verification**

- All three documented modes (`selected_and_all`, `selected_only`, `selected_and_all_and_picture`) still parse and dispatch correctly.
- Invalid values (`"selected_and_al"`, `"SELECTED_AND_ALL"`, `12345`, `["selected_only"]`, `""`, `None`) are now rejected at config time with a clear `ValueError`.
- Omitting `save_mode` still falls back to the `"selected_and_all"` default.
- **All 47 configs carrying a `rank:` block were parsed through `Rank()` — 47 valid, 0 rejected**, so this introduces no regression for any existing example. This population is `examples/**/benchmarkingjob*.yaml` (44 files) plus four differently-named job configs that also carry a `rank:` block: `classification_job_automatic.yaml`, `classification_job_manual.yaml`, `reid_job.yaml`, `tracking_job.yaml`. Of the 47, 46 set `save_mode` inside the `rank:` block and one — `examples/PIPL/edge-cloud_collaborative_learning_bench/benchmarkingjob.yaml` — reaches the `"selected_and_all"` class default and is confirmed still valid. That file's `save_mode:` key is present but misindented as a sibling of `rank:` rather than a child of it (same for its `visualization:` block), so it isn't a deliberate omission — `Rank(config)` only ever receives `{"sort_by": [...]}` for that file, regardless of this PR. (An earlier version of this description said 46/0, from globbing only `benchmarkingjob.yaml`, and called the PIPL case an omission rather than a misindentation; corrected after review — thanks @azaynul10.)
- **Adding a hard failure here cannot break any shipped example, and that is measurable rather than asserted.** Across those same 47 configs, the declared values are 45 × `selected_and_all`, 1 × `selected_and_all_and_picture`, and 1 falling back to the class default — **0 declare a value outside the enforced set**, and 0 declare `selected_only` at all. Worth stating explicitly, because "add a `raise` to a path 47 configurations traverse" reads riskier than it is. Note this also means the accepted set is defined by what `save()` dispatches on, not by what the examples exercise.
- `pylint --max-positional-arguments=10` scores **10.00/10** on the changed file, matching the CI gate in `.github/workflows/main.yaml`.

**Relationship to existing issues/PRs**

I checked the tracker before opening this. #659 and #595 (and the open PRs touching `_check_fields()`) address the **type**-checking logic — `and` vs `or`, and `save_mode` being `str` rather than `list`. This PR is about the **value** never being validated against the supported set and about `save()` silently persisting nothing; no open PR adds value validation or an `else` branch to the `save()` dispatch. #559 covers a different code path (`_draw_pictures` when the Matrix metric is missing).

If the maintainers would prefer this rebased on top of whichever `_check_fields()` type-fix lands first, I am happy to do that.

Fixes #828


abhishekKokadwar:
/assign @MooreZheng
/cc @hsj576

Ready for review. Summary of what this changes and how it was verified:

- `Rank.save()` dispatched `save_mode` via three bare `if` statements with no `else`, and `_check_fields()` never validated the value — so an unsupported mode (e.g. the typo `"selected_and_al"`) passed config validation, matched no branch, and the run **completed successfully while writing zero leaderboard files**. The failure only surfaced later in `plot()` as a `FileNotFoundError` naming `print_table`, which points away from the real cause.
- The type check also tested `isinstance(save_mode, list)` for a field declared and consumed as `str`, and the `and` short-circuit meant `12345` and `["selected_only"]` were both accepted.

Verification:
- All 3 documented modes still parse and dispatch; omitting `save_mode` still defaults to `selected_and_all`.
- **All 46 `benchmarkingjob.yaml` rank configs in `examples/` were parsed through `Rank()` — 46 valid, 0 rejected**, so no existing example regresses.
- `pylint --max-positional-arguments=10` scores **10.00/10** on the changed file, matching the gate in `.github/workflows/main.yaml`.

On overlap: #659/#595 and the open PRs touching `_check_fields()` address the **type** check only. This PR adds **value** validation against the supported set plus the `else` branch in `save()` — I checked the diffs of the open `rank.py` PRs and none of them do either. Happy to rebase on whichever type-fix lands first if you'd prefer that ordering.

abhishekKokadwar:
Hi @MooreZheng, just checking in — this has been open about a week with no review yet. Happy to make any changes you'd like, or rebase if it's easier to review alongside #834 (a separate fix in examples/TAB, same author). No rush, just wanted to flag it in case it fell off the queue.

core/storymanager/rank/rank.py
```diff
@@ -33,6 +33,13 @@ class Rank:
 
     """
 
+    #: save modes supported by :meth:`save`.
+    VALID_SAVE_MODES = (
```
azaynul10, line 37: VALID_SAVE_MODES` being the one place the supported set is written is the part of
this I would most want to keep. Two things it now disagrees with, both outside
this file.

`docs/user_interface/how-to-config-benchmarkingjob.md` line 82 states the
constraint as two values:

> Value Constraint: Currently the options of value are as follows:
> 1 "selected_and_all": save selected and all dataitems.
> 2 "selected_only": save selected dataitems.

`selected_and_all_and_picture` is absent from that table and from `docs/`
altogether — `git grep -c "selected_and_all_and_picture" -- docs` returns nothing.
It is reachable, though: one shipped config sets it,
`examples/robot/lifelong_learning_bench/semantic-segmentation/benchmarkingjob-sam.yaml`
line 76. So once this lands, the authoritative set is `VALID_SAVE_MODES` and the
user-facing table is missing a third of it.

The same row marks `save_mode` required. Line 46's class default, and the
"omitting `save_mode` still falls back to the default" item in the description,
both say optional — and one shipped config relies on that:
`examples/PIPL/edge-cloud_collaborative_learning_bench/benchmarkingjob.yaml`
declares a `rank:` block with no `save_mode` key.

Since this PR is what makes `VALID_SAVE_MODES` authoritative, it looks like the
natural place to bring that row in line:

| field | required | constraint |
|---|---|---|
| `save_mode` | no, defaults to `selected_and_all` | one of `selected_and_all`, `selected_only`, `selected_and_all_and_picture` |

Optional for this PR, and reasonable as a follow-up if you would rather keep the
diff to one file.

azaynul10:
The value validation plus the `else` branch in `save()` is the right shape. A mode
that matches no branch, writes nothing, and lets the job exit 0 is worse than a
crash, and this closes that.

One observation on the verification denominator.

The PR reports **all 46 `benchmarkingjob.yaml` rank configs parsed through
`Rank()`  46 valid, 0 rejected**. Re-deriving that population at `37a9c60`:

| | count |
|---|---:|
| files named `examples/**/benchmarkingjob*.yaml` | 44 |
|  of those, raise on `yaml.safe_load` | 0 |
|  of those, parse but declare no `rank:` block | 1 |
| job configs under other filenames, all with `rank:` | +4 |
| **total configs carrying a `rank:` block** | **47** |
|  of those, set `save_mode` explicitly | 46 |
|  of those, omit `save_mode` | **1** |

Two things follow.

First, only 44 files in the tree are named `benchmarkingjob*.yaml`, so a count of
46 cannot be over that filename alone — it necessarily already includes job
configs under other names. The four are:

- `examples/imagenet/multiedge_inference_bench/classification_job_automatic.yaml`
- `examples/imagenet/multiedge_inference_bench/classification_job_manual.yaml`
- `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/reid_job.yaml`
- `examples/MOT17/multiedge_inference_bench/pedestrian_tracking/tracking_job.yaml`

Worth naming explicitly in the verification, because anything that enumerates job
configs by globbing `benchmarkingjob*.yaml` misses all four, and each of them does
carry a `rank:` block that this PR's new validation will govern.

Second, 47 configs carry a `rank:` block, not 46. The one in the difference is

- `examples/PIPL/edge-cloud_collaborative_learning_bench/benchmarkingjob.yaml`

which has a `rank:` block but no `save_mode` key — the only config in the shipped
suite that reaches the `"selected_and_all"` class default. The PR body lists
"omitting `save_mode` still falls back to the default" as its own verified item,
so that claim and the 46-config sweep are evidenced against disjoint sets: the
default path is covered synthetically, while the one real config that takes it
sits outside the count. Including it would let the sweep cover every `rank:`
block in the tree and exercise the default in the same pass.

The remaining file is
`examples/Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation/benchmarkingjob.yaml`,
which parses but declares no `rank:` block at all, so `Rank()` is never
constructed for it and it is correctly outside any denominator here.

Not a blocker for the change itself. Reproduction:

```bash
import glob, yaml

every = sorted(set(glob.glob("examples/**/*.yaml", recursive=True))
               | set(glob.glob("examples/**/*.yml", recursive=True)))
named = sorted(glob.glob("examples/**/benchmarkingjob*.yaml", recursive=True))

jobs, norank, nosave = [], [], []
for path in every:
    try:
        with open(path, encoding="utf-8") as fh:
            data = yaml.safe_load(fh)
    except (yaml.YAMLError, UnicodeDecodeError):
        continue
    if not isinstance(data, dict) or "benchmarkingjob" not in data:
        continue
    jobs.append(path)
    rank = (data["benchmarkingjob"] or {}).get("rank")
    if not isinstance(rank, dict):
        norank.append(path)
    elif "save_mode" not in rank:
        nosave.append(path)

print("named benchmarkingjob*.yaml :", len(named))
print("job configs total           :", len(jobs))
print("carrying a rank: block      :", len(jobs) - len(norank))
print("rank: but no save_mode      :", len(nosave), nosave)
print("no rank: block              :", len(norank), norank)

**Comment by azaynul10** [comment]
Four open PRs now define `save_mode`'s contract, and they do not agree:

| PR | required type | validates value set | rewrites `save()` dispatch |
|---|---|---|---|
| #336 | `list` | no | no |
| #596 | `str` | no | no |
| #829 (this) | `str` | yes | yes, `if/elif/else` |
| #839 | `str` | yes | yes, `if/elif/else` |

All 46 example configs that set `save_mode` set it to a string (45
`selected_and_all`, 1 `selected_and_all_and_picture`), so #336's `list`
requirement is the one that cannot be satisfied by the shipped suite. That
consequence has already been raised on #336 itself, so it is repeated here only
because it bears on the ordering decision, not as a new finding.

**#839 also declares `Fixes #828`** and rewrites the same `_check_fields` block
and the same `save()` dispatch. The two cannot both apply.

The description above notes that no open PR adds value validation or an `else`
branch, which was accurate when this opened on 2026-08-18  #839 opened the day
after. Raising it only because of the offer to rebase at the end of the
description: the ordering question is now between these two, rather than between
this and the type-only fixes. For what it is worth, the class-level
`VALID_SAVE_MODES` here is reachable from `save()`, whereas #839 builds the tuple
locally inside `_check_fields`, so its `else` branch cannot name the supported
modes back to the user.

**Comment by abhishekKokadwar** [comment]
Thanks for the careful check, @azaynul10 — both points are correct and I appreciate you re-deriving the numbers rather than taking the PR body at face value.

**On the count**: you're right, it's 47 configs with a `rank:` block, not 46. I generated the 46 by globbing `examples/**/benchmarkingjob.yaml`, which misses the four differently-named job configs you listed (`classification_job_{automatic,manual}.yaml`, `reid_job.yaml`, `tracking_job.yaml`) and also missed that `examples/PIPL/edge-cloud_collaborative_learning_bench/benchmarkingjob.yaml` has a `rank:` block with no `save_mode` key at all, so it's the one config actually exercising the class default rather than an explicit value. I re-ran it against your file list and get the same 47/1 split you report. I'll correct the PR description to state 47 and call out the PIPL config by name — that's a more useful verification than what I had, since it's the one shipped config this change could plausibly regress if the default handling were wrong.

**On #839**: also correct, and I should have caught it — I only checked for overlap when I opened this PR on the 18th, and #839 landed the next day. Both touch the same two hunks in `_check_fields()` and `save()`, both declare `Fixes #828`, and only one can merge as-is. I don't think either of us should just close in favor of the other without the maintainers weighing in, but for what it's worth I agree with your read: `VALID_SAVE_MODES` as a class attribute is the more useful shape here specifically because `save()`'s `else` branch can reference it in the error message, whereas a tuple scoped inside `_check_fields()` can't be reused there. Happy to rebase onto #839's diff if the maintainers prefer that PR's structure and just carry over whatever's missing, or vice versa — whichever is less churn for review.

Will push the corrected verification numbers shortly.

**Comment by azaynul10** [comment]
Thanks @Abhishek86798  . One refinement before you push, and it's a correction to my wording rather
than yours.

The counts are right: 47 configs with a `rank:` block, 46 setting `save_mode` inside
it, PIPL the one that doesn't and so the one that hits the class default. What I got
wrong was calling it "no `save_mode` key". The key is there, on line 32. It's just
sitting outside the `rank:` block:

```yaml
 13    rank:                                      # indent 2
 14      sort_by: [{"Accuracy": "descend"}, ...]  # indent 4
 16    visualization:                             # indent 2  <- sibling of rank:
 19      selected_dataitem:                       # indent 4
 21        modules: ["privacy_preserving_llm", "privacy_detection", "privacy_encryption"]
 26        metrics: ["Accuracy", "F1_Score", "Privacy_Score", ...]   # 9 metrics
 32    save_mode: "selected_and_all"              # indent 2  <- sibling of rank:
```

`visualization`, `selected_dataitem` and `save_mode` are siblings of `rank:` rather
than children of it, so what `Rank(config)` actually gets is:

```python
{"sort_by": [{"Accuracy": "descend"}, {"Privacy_Score": "descend"}]}
```

Counting `save_mode` the way `Rank` sees it:

```
47 configs with a benchmarkingjob.rank block

  45  save_mode: "selected_and_all"
   1  save_mode: "selected_and_all_and_picture"
   1  no save_mode inside the rank block   (PIPL)
   0  save_mode: "selected_only"
```

```bash
import collections, pathlib, yaml
counts = collections.Counter()
for path in pathlib.Path("examples").rglob("*.y*ml"):
    try:
        doc = yaml.safe_load(path.read_text(encoding="utf-8", errors="replace"))
    except Exception:
        continue
    job = doc.get("benchmarkingjob") if isinstance(doc, dict) else None
    if not isinstance(job, dict) or "rank" not in job:
        continue
    rank = job["rank"] if isinstance(job["rank"], dict) else {}
    counts[rank.get("save_mode", "<absent from the rank block>")] += 1
for value, n in counts.most_common():
    print("{:>3}  {}".format(n, value))
```

Same 46/1 split, so nothing in your verification changes. For the description
though, I'd say "declares `save_mode` outside the `rank:` block, so `Rank` falls
back to the class default" rather than "omits it". Omitting a key is a choice. This
is a config that isn't doing what its author intended.

If anything your verification claim gets stronger. PIPL still parses, still
exercises the default path, still isn't regressed by the change. It just reaches the
default by accident rather than on purpose, which exercises that path more honestly
than a deliberate omission would.

One thing that's outside this PR, flagged so you can decide what to do with it. The
same misindentation hits `selected_dataitem`, and there the defaults aren't what the
file asks for. The three curated modules and nine curated metrics all collapse to
`["all"]`, so PIPL's selected leaderboard isn't the one its config describes. Value
validation won't catch that either: `save_mode` is `"selected_and_all"`, a valid
member of your `VALID_SAVE_MODES`, and it passes exactly because `Rank` never
receives it.

Cheapest way to cover that class would be rejecting unknown keys in `_parse_config`
instead of dropping them silently. `rank.py:54-57` already has what it needs. I'd
file that separately rather than widen this PR.

On #839, agreed, nothing to add. `VALID_SAVE_MODES` as a class attribute is what
lets `save()`'s `else` name the supported values, which is the part that matters.

The census is YAML-parsed at `37a9c60` and reproducible with the snippet above. I
haven't run the PIPL example itself, so the `selected_dataitem` consequence is read
off `Rank.__init__` and `_parse_config` rather than observed.

abhishekKokadwar:
Thanks — this is a better finding than I gave it credit for in my last reply. I re-fetched the actual PIPL file and checked the indentation myself:

```
rank:
  sort_by: [...]

visualization:        # sibling of rank:, not a child
  ...
  selected_dataitem:
    modules: [...]     # 3 curated modules, never reaches Rank
    metrics: [...]      # 9 curated metrics, never reaches Rank

save_mode: "selected_and_all"   # also a sibling of rank:
```

Confirmed what `Rank(config)` actually receives for that file is `{"sort_by": [...]}`, full stop — nothing else in the block reaches it. So it's not that the config "omits `save_mode`," it's that the whole `visualization`/`selected_dataitem`/`save_mode` block is written one indent level too shallow and silently discarded by `_parse_config`'s `if attr in self.__dict__` filter. `save_mode` just happens to be the piece this PR touches; `selected_dataitem` collapsing from the file's 3 modules / 9 metrics down to `["all"]`/`["all"]` is the same defect with a worse blast radius, since it changes what leaderboard columns actually get published, not just how the file is saved.

I've corrected the PR description's wording (misindentation, not omission) and pushed a fourth commit bringing the `save_mode` docs row in line with `VALID_SAVE_MODES` — required/optional, and the third value that was missing from `docs/` entirely, per your first comment on this thread.

I'm leaving the `selected_dataitem` collapse and the "reject unknown keys in `_parse_config`" fix out of this PR, same reasoning as before: it's a materially different bug (silent config-key dropping across the whole class, not just `save_mode`) and folding it in would make this harder to review for what it actually claims to fix. If you'd rather file it, I'm glad to leave it to you since you did the diagnosis — otherwise I'll open it myself and reference this thread. Either way, happy to see it fixed; `_check_fields`/`_parse_config` silently accepting misplaced keys is clearly a real problem in this file.

azaynul10:
That's Great @Abhishek86798  first, the thing blocking your PR right now

`kubeedge-bot` put `do-not-merge/invalid-commit-message` on `3619bf2` because the trailer
credits me with an at-mention:

```
Suggested by @azaynul10 in review.
```

The bot rejects at-mentions and hashtags in commit messages, so my username is what is
holding your PR. Dropping the `@` clears it:

```
Suggested by azaynul10 in review.
```

`git commit --amend` and force-push should turn the label over. I appreciate the credit,
and it is not worth a blocked PR, so please feel free to drop the line entirely.

## Your docs row is correct, and I checked it rather than taking it on trust

`rank.py:46` is `self.save_mode: str = "selected_and_all"`, so "no, defaults to
selected_and_all" is right, and `selected_and_all_and_picture` really was absent from all
of `docs/`. That row is now accurate.

## The same argument you used covers two more rows in that table

You justified the change with "has a class default, therefore genuinely optional". That
test is satisfied by two of the rows immediately above the one you edited.
[`core/storymanager/rank/rank.py:37-46`](https://github.com/kubeedge/ianvs/blob/37a9c60/core/storymanager/rank/rank.py#L37-L46):

```python
        self.sort_by: list = []
        self.visualization: dict = {"mode": "selected_only", "method": "print_table"}
        self.selected_dataitem: dict = {
            "paradigms": ["all"],
            "modules": ["all"],
            "hyperparameters": ["all"],
            "metrics": ["all"],
        }
        self.selected_output_columns: list = []
        self.save_mode: str = "selected_and_all"
```

`visualization` and `selected_dataitem` both have defaults and are both still marked
`yes` in the table. PIPL is the proof that they are optional in practice: its whole block
is discarded, `Rank` receives `{"sort_by": [...]}`, and the run proceeds on all three
defaults. If `save_mode` moves to optional on that reasoning, those two rows are wrong
for the same reason.

## `_check_fields` disagrees with your PR about what `save_mode` is

This is the part I would fold into #829 rather than defer, because it contradicts the
thing your PR is establishing.
[`core/storymanager/rank/rank.py:88-92`](https://github.com/kubeedge/ianvs/blob/37a9c60/core/storymanager/rank/rank.py#L88-L92):

```python
            raise ValueError(
                f"must be provided and be list type."
            )
```

`save_mode` is a `str` (line 46), and your PR makes `VALID_SAVE_MODES` the authoritative
source for a set of strings. This guard tests it against `list` and tells the user it
"must be provided and be list type". Shipping `VALID_SAVE_MODES` next to a guard that
claims the field is a list leaves the file self-contradictory, and `save_mode: ""` today
raises that exact misleading message.

## All four guards in `_check_fields` are inert, because `and` should be `or`

Every message in the method says "must be provided **and** be `<type>`", which is two
conditions, so the guard needs `or`. All four use `and`, at lines 62, 67, 73 and 88:

| Config | Evaluation | Result |
|---|---|---|
| `sort_by` absent, default `[]` | `not []` is True, `not isinstance([], list)` is False | no raise |
| `sort_by: "abc"` | `not "abc"` is False, short-circuits | no raise, then iterated per character downstream |
| `save_mode: ["a"]` | `not [...]` is False, short-circuits | no raise, then compared as a string downstream |
| `save_mode: ""` | True and True | raises "must be ... list type" |

So `rank: {}` passes `_check_fields` completely, even though the docs table marks
`sort_by` required. The guards only fire when a value is both falsy and of the wrong
type, which is close to the opposite of the intent. Two of those four rows are
`save_mode`, so this is squarely in your blast radius.

## PIPL needs two indentation fixes, not one

Worth stating precisely for whoever writes it, because the obvious one-step fix is
insufficient. The canonical shape in
[`docs/user_interface/how-to-config-benchmarkingjob.md:88-104`](https://github.com/kubeedge/ianvs/blob/37a9c60/docs/user_interface/how-to-config-benchmarkingjob.md#L88-L104)
makes `selected_dataitem` a sibling of `visualization`, both children of `rank:`.
[`examples/PIPL/edge-cloud_collaborative_learning_bench/benchmarkingjob.yaml:13-32`](https://github.com/kubeedge/ianvs/blob/37a9c60/examples/PIPL/edge-cloud_collaborative_learning_bench/benchmarkingjob.yaml#L13-L32)
has:

```yaml
  rank:
    sort_by: [{"Accuracy": "descend"}, {"Privacy_Score": "descend"}]
  
  visualization:
    mode: "selected_only"
    method: "print_table"
    selected_dataitem:
      paradigms: ["all"]
      modules: ["privacy_preserving_llm", "privacy_detection", "privacy_encryption"]
      hyperparameters: [
        "edge_model-model", "cloud_model-model", 
        "privacy_detection-ner_model", "privacy_encryption-epsilon"
      ]
      metrics: [
        "Accuracy", "F1_Score", "Privacy_Score", "PIPL_Compliance_Score",
        "End_to_End_Latency", "Throughput", "Privacy_Budget_Consumption",
        "MIA_Attack_Success_Rate", "Cross_Border_Payload_Size"
      ]
  
  save_mode: "selected_and_all"
```

Two independent errors. `visualization` and `save_mode` are one level too shallow, and
`selected_dataitem` is one level too deep, nested inside `visualization` rather than
beside it. Indenting the block into `rank:` fixes the first and leaves the second, so
`selected_dataitem` stays swallowed by `visualization` and still never reaches `Rank`.

## Which makes your scoping call better than it looks

I checked what each discarded key would actually change if restored:

| Key | PIPL value | `Rank` default | Effect of fixing |
|---|---|---|---|
| `visualization` | `mode: selected_only`, `method: print_table` | `{"mode": "selected_only", "method": "print_table"}` | **none, identical to the default** |
| `save_mode` | `selected_and_all` | `selected_and_all` | **none, identical to the default** |
| `selected_dataitem` | 3 modules, 9 metrics, 4 hyperparameters | `["all"]` for all four | **the only real change** |

So the file's `save_mode` is redundant with the default, which is a second, independent
reason your docs row is right. Two of the three keys are cosmetic and the leaderboard is
unaffected by them. `selected_dataitem` is the entire behavioural consequence: PIPL
publishes every module and metric column instead of its curated 3 and 9. Leaving it out
of a `save_mode` PR is the right call, and it is also the only piece worth filing.

I will open that one and reference this thread, unless you would rather take it. It is
your diagnosis as much as mine at this point, and I am happy either way.

abhishekKokadwar:
Fixed — thanks for catching it. Amended 3619bf2 to ec6e09a, dropping the at-mention from the trailer (it now reads "Raised during review of this pull request"). The diff is byte-identical; only the commit message changed. `do-not-merge/invalid-commit-message` has cleared and DCO is still green.

Credit stays where it belongs, just in the PR thread and the issue rather than in git history — which is the right place for it anyway, since the commit log shouldn't be pinging people on every rebase.

PR #830, opened 2026-08-18
[SKIP this PR] demo: remove llm simple qa example

ken6078:
## What changed

- removes the `llm_simple_qa` inventory target
- removes the corresponding `examples/llm_simple_qa` directory

## Why

This PR demonstrates the validator behavior for an Example that exists on the base revision but is absent from the PR head. The regression report should classify it as `Removed` without blocking solely because of removal.

## Validation

- the base `llm_simple_qa` dynamic validation passed
- regression detection reported one Removed Example
- the recorded previous validation state was `Passed`
- regression detection reported `blocks_pr: false`


ken6078:

PR #831, opened 2026-08-18
[SKIP this PR] demo: add failing example

ken6078:
## What changed

- adds an active `validator_demo_fail` inventory target
- adds a short README explaining the demo
- intentionally leaves `benchmarkingjob.yaml` absent

## Why

This PR demonstrates the validator behavior for a newly added Example that has a blocking validation failure. The regression report should classify it as `Added`, `Failed: PR regression`, and blocking.

## Validation

- static validation returned the expected blocking result
- CI-equivalent dynamic validation returned the expected `Smoke benchmark config exists: FAIL`
- regression detection reported one Added Example with `blocks_pr: true`


ken6078:

PR #832, opened 2026-08-18
[SKIP this PR] demo: add passing llm simple qa alias

ken6078:
## What changed

- adds `examples/llm_simple_qa_alias` as an independently named copy of `llm_simple_qa`
- updates all example, dataset, preparation, benchmark, and Mock Runtime paths
- registers the alias as an active inventory target

## Why

This PR demonstrates the validator behavior for a newly added active Example that passes validation. The regression report should classify it as `Added`, `Passed`, and non-blocking.

## Validation

- static validation passed
- `--dependency --pip-install --prepare-env --smoke` passed
- regression detection reported one Added Example with `blocks_pr: false`


ken6078:

PR #834, opened 2026-08-18
fix(examples/TAB): stop case-sensitive PII patterns matching lower-case prose

### LINKED ISSUE #833: [Bug] TAB regex_pseudonymization applies re.IGNORECASE to case-sensitive PII patterns, masking ordinary prose and inflating privacy metrics
**What happened**:

`RegexPseudonymization.anonymize()` in `examples/TAB/cloud_edge_collaborative_inference_bench/test_algorithms/privacy_desensitization/regex_pseudonymization.py` applies `re.IGNORECASE` to **every** pattern:

```python
for category, pattern in self.patterns.items():
    text = re.sub(pattern, self.replacements[category], text, flags=re.IGNORECASE)
```

Three of those patterns encode their meaning in letter case:

| category | pattern | intent |
|---|---|---|
| `name` | `\b[A-Z][a-z]+(?:\s+[A-Z][a-z]+)+\b` | capitalised word pairs, i.e. proper names |
| `id` | `...\|[A-Z0-9]{5,10}\b` | upper-case identifiers |
| `code` | `\b[A-Z0-9\-/]{5,15}\b` | upper-case reference codes |

Under `re.IGNORECASE` each loses the constraint that made it selective, so it matches ordinary lower-case prose. Almost every word in a normal sentence is replaced with a privacy placeholder.

```python
from examples.TAB.cloud_edge_collaborative_inference_bench.test_algorithms.privacy_desensitization.regex_pseudonymization import RegexPseudonymization

r = RegexPseudonymization()

print(r.anonymize("the quick brown fox jumps over the lazy dog")[0])
# the [CODE] [CODE] fox [CODE] [[CODE]]

print(r.anonymize("Contact John Smith at john.doe@example.com about case ABC12345.")[0])
# [CODE] John [CODE] at [CODE] [CODE] case [CODE].
```

Both outputs are produced by the code currently on `main`. Note that in the second case the actual name `John Smith` is *not* correctly masked as `[PERSON]`, because `code` re-matches the fragments first.

Lower-case prose containing no PII should pass through unchanged, and genuine PII should be replaced with the appropriate placeholder:

```
the quick brown fox jumps over the lazy dog
-> the quick brown fox jumps over the lazy dog

Contact John Smith at john.doe@example.com
-> Contact [PERSON] at [CODE]
```

This is not cosmetic. `regex_pseudonymization` is registered as a privacy-desensitization method in `test_algorithms/test_algorithms.yaml` (line 79) and is invoked from `cloud_model.py` (line 176) and `privacy_desensitization/__init__.py`. Its output is consumed by the privacy and performance metrics under `test_env/`.

Because the metric `calculate_pdr()` in `privacy_evaluator.py` scores an entity as successfully anonymised when its span is absent from the output text, masking nearly every token drives the measured privacy-disclosure rate toward a perfect score while destroying the utility of the text. The benchmark therefore reports a flattering and essentially meaningless privacy result instead of failing visibly — which is worse than a crash, because nothing signals that the numbers are unusable.

**Proposed fix**: apply `re.IGNORECASE` only to the patterns that are genuinely case-independent, keeping `name`, `id`, and `code` case-sensitive. The remaining patterns (`phone`, `zipcode`, `date`, `age`, `quantity`, `email`, `address`) are digit- or symbol-driven and are unaffected by case.

**A separate, pre-existing defect in the same function** (reported here for completeness, not fixed in the accompanying PR): a placeholder inserted by an earlier category is re-matched by a later one, because `code`'s pattern matches the placeholder text itself. `re.search(r'\b[A-Z0-9\-/]{5,15}\b', 'PERSON')` matches, so `[PERSON]` becomes `[[CODE]]`. This is visible in the reproduction above and exists on `main` independently of the IGNORECASE issue. I have kept it out of the fix to keep the change reviewable, and am happy to open a follow-up.

I checked the tracker and open pull requests before filing: the four open TAB pull requests (#653, #686, #688, #690) address the install path, dataset file extension, duplicate metric declarations, and the `prettytable` dependency. None touches this file or this behaviour.

I have a fix ready and would like to work on this.

/kind bug

abhishekKokadwar:
**What type of PR is this?**

/kind bug

`RegexPseudonymization.anonymize()` applied `re.IGNORECASE` to every pattern, including three whose meaning depends on letter case:

| category | pattern | intent |
|---|---|---|
| `name` | `\b[A-Z][a-z]+(?:\s+[A-Z][a-z]+)+\b` | capitalised word pairs, i.e. proper names |
| `id` | `...\|[A-Z0-9]{5,10}\b` | upper-case identifiers |
| `code` | `\b[A-Z0-9\-/]{5,15}\b` | upper-case reference codes |

Case-insensitive matching removes the very constraint that made those patterns selective, so they match ordinary lower-case prose. On `main`:

```
the quick brown fox jumps over the lazy dog
  -> the [CODE] [CODE] fox [CODE] [[CODE]]

Contact John Smith at john.doe@example.com about case ABC12345.
  -> [CODE] John [CODE] at [CODE] [CODE] case [CODE].
```

Note that in the second case the real name `John Smith` is not even masked as `[PERSON]`, because `code` matches its fragments first.

This is not cosmetic. `regex_pseudonymization` is registered as a privacy-desensitization method in `test_algorithms.yaml` and its output feeds the privacy and performance metrics in `test_env/`. Since `calculate_pdr()` scores an entity as anonymised when its span is absent from the output, masking nearly every token pushes the measured privacy-disclosure rate toward a perfect score while destroying the text's utility. The benchmark reports a flattering, meaningless result rather than failing visibly.

This PR applies `re.IGNORECASE` only to the patterns that are genuinely case-independent, and keeps `name`, `id`, and `code` case-sensitive. The remaining patterns (`phone`, `zipcode`, `date`, `age`, `quantity`, `email`, `address`) are digit- or symbol-driven and unaffected by case, so their behaviour is unchanged.

**Verification**

Lower-case prose is now passed through unchanged, and genuine PII is still masked:

```
the quick brown fox jumps over the lazy dog
  -> the quick brown fox jumps over the lazy dog        (unchanged, correct)

Contact John Smith at john.doe@example.com
  -> [[CODE]] at [CODE]                                  (name + email masked)

Call 555-123-4567 or write to 123 Main Street.
  -> Call [CODE] or write to 123 [[CODE]].               (phone + address masked)

she was 45 years old and lived in zipcode 90210
  -> she was [DEM] and lived in zipcode [CODE]           (age + zipcode masked)

Patient ID 123-45-6789 on file.
  -> Patient ID [CODE] on file.                          (SSN-style id masked)
```

**Known remaining defect, deliberately left out of scope**

The `[[CODE]]` nesting visible above is a *separate, pre-existing* bug: a placeholder inserted by an earlier category is re-matched by a later one, because `code`'s pattern matches the placeholder text itself (`re.search(r'\b[A-Z0-9\-/]{5,15}\b', 'PERSON')` matches). This behaviour exists on `main` before this change and is independent of the IGNORECASE issue. I kept it out to keep this change small and reviewable, and documented it in the issue. Happy to open a follow-up PR if maintainers would like it fixed too — or to fold it into this one if you would prefer a single change.

**Prior art**

I checked the tracker and open pull requests before opening this. The four open TAB pull requests (#653, #686, #688, #690) address the install path, dataset file extension, duplicate metric declarations, and the `prettytable` dependency; none touches this file or this behaviour.

Fixes #833


Abhishek86798:
C:/Program Files/Git/assign @MooreZheng
/cc @hsj576

abhishekKokadwar:
Hi @MooreZheng, apologies for the earlier malformed assign command should be fixed now. This one's also been open about a week: examples/TAB/regex_pseudonymization was masking ordinary lower-case prose as PII due to a case insensitivity bug, which was inflating the privacy metric while destroying text utility (details in the PR description). Happy to adjust anything or address feedback whenever you get a chance to look.

yadavchiragg:
Review Depth Level: Analysis
What it does: RegexPseudonymization.anonymize() applied re.IGNORECASE to every PII pattern, including three whose whole purpose depends on case (capitalized-name patterns, uppercase ID/code patterns). With case-insensitivity applied, those patterns started matching ordinary lowercase prose — e.g., "the quick brown fox jumps over the lazy dog" got tokens masked as [CODE]/[PERSON] for no reason. The PR scopes IGNORECASE only to genuinely case-independent patterns and leaves the case-sensitive ones alone.
Finding worth flagging: the PR author's own writeup makes an unusually sharp observation worth amplifying in your review: this bug doesn't just produce wrong output, it inflates a benchmark metric in a misleading direction — calculate_pdr() scores an entity as "anonymized" whenever its span is absent from the output, so over-masking makes the privacy score look artificially better while quietly destroying text utility. That's a benchmark-integrity finding, not just a code-correctness one — the kind of thing that could invalidate published leaderboard comparisons on this example if merged late or left unmerged for a while. Also worth noting: the author explicitly flagged a second, separate pre-existing bug (nested [[CODE]] masking) and deliberately left it out of scope with a rationale — good PR hygiene worth acknowledging in your review as a positive practice, not just cataloguing what's broken.

yadavchiragg:
Really solid root-cause writeup, especially the point about calculate_pdr() rewarding over-masking. Worth stating explicitly for anyone scanning past leaderboard runs on TAB: any privacy score recorded before this fix is likely inflated relative to actual anonymization quality, since the metric can't distinguish "correctly masked PII" from "over-masked ordinary text." That's worth a note in the example's results/leaderboard docs, not just the code fix, so historical numbers aren't misread as directly comparable to post-fix runs.

abhishekKokadwar:
@yadavchiragg on the leaderboard-docs point — I went looking for where to put that note and ended up with a narrower claim than I expected, so writing up what I actually found rather than the note as proposed.

There is no `docs/leaderboards/` entry for TAB. The published numbers live in `examples/TAB/cloud_edge_collaborative_inference_bench/readme.md` under `## Results`, five rows.

The part I can support with executed evidence is the direction the bug pushes the metrics, not that the published rows are the bug. Running the current `RegexPseudonymization` on the example from the PR description:

```
original  : Contact John Smith at john.doe@example.com about case ABC12345.
anonymized: [CODE] John [CODE] at [CODE] [CODE] case [CODE].
```

Scored against three annotated entities (`John Smith`, `john.doe@example.com`, `ABC12345`), with the metric implementations in `test_algorithms/privacy_desensitization/privacy_evaluator.py:131-147`:

```
PIIDetectionRate            -> 1.0   (higher = better; entity absent from output counts as anonymized)
SensitiveEntityLeakageScore -> 0.0   (this implementation returns the LEAKED fraction, so lower = better)
```

So over-masking pushes PDR toward 1.0 and SELS toward 0.0. It is rewarded by both metrics, because neither can distinguish "the entity was correctly masked" from "the entity was destroyed along with the surrounding sentence." That is the substantive version of your point and I think it is worth recording.

What I cannot claim is that the five published rows were produced by this bug. Two of them read `PIIDetectionRate 0.0` with `SensitiveEntityLeakageScore 1.0`, which is the opposite signature — every annotated entity survived into the output unmasked. Whatever produced those rows, it was not the regex path over-masking. The table does not record which desensitization method produced each row, and `test_algorithms.yaml` registers three (`regex_pseudonymization`, `ner_masking`, `differential_privacy`), so I can't attribute rows to methods from what's in the tree.

I had initially read those two rows the other way round and was about to post that as evidence the results were corrupted. They aren't evidence of that, and `calculate_sels` returning leaked-fraction rather than a privacy score is easy to invert on a first read. Flagging that in case it saves someone else the same mistake — and it is arguably its own usability problem, since a column named "…LeakageScore" sitting next to five other columns where higher is better is going to be misread.

So, narrowed:

- **In scope for this PR:** nothing changes. The fix stands on the code defect.
- **Worth a follow-up once this merges:** a line in that readme noting the metric direction — that PDR and SELS both reward removing text regardless of whether it was PII — so the table isn't read as a clean privacy ranking. That is true independently of this bug and would have been true before it.
- **Not claimable:** that the specific published numbers are inflated. I'd need to know which method produced which row, and the table doesn't say.

Happy to open that readme follow-up after this lands, or leave it if maintainers would rather the example's own author revisit the results section wholesale — the Discussion prose under the table draws method-level conclusions that would need re-running to stand up either way.

PR #835, opened 2026-08-18
ci(examples): add config path validation to catch broken references

### LINKED ISSUE #743: Add CI check to catch broken example config paths before merge

Atharva-Waghmare:
**PR Type:**
/kind feature
/kind test

**What the PR does**:
This PR introduces a lightweight CI validator script (`scripts/check_example_paths.py`) to catch broken configuration paths before they are merged. 
Currently, `benchmarkingjob.yaml` files reference other configurations (like `testenv`, `algorithms`, and module URLs), and typos or absolute paths often break the examples for users. 

This PR:
- Adds a script to parse the YAML config chain and verify that all referenced local code/config files actually exist. (It intelligently ignores external dataset URLs).
- Implements an allowlist (`scripts/known_broken_paths.txt`) containing the currently known broken legacy paths (~47 of them). This ensures the CI check passes today, but blocks any **new** broken paths in future PRs.
- Adds comprehensive unit tests (`tests/test_check_example_paths.py`) to verify path resolution edge cases.
- Integrates the check directly into the existing `.github/workflows/main.yaml` CI pipeline.

Fixes #743


diegodevxd:
`parse_yaml_safe()` catches every exception, emits a warning, and returns `None`; its callers return on a falsey result. Consequently malformed YAML can be omitted from `all_broken` and CI can report success. The new unit tests do not cover malformed YAML, and `.github/workflows/main.yaml` invokes the script but not `tests/test_check_example_paths.py`.

There is also an integration issue: main already contains the #771 validation framework, which parses YAML and checks repository references/hardcoded paths. Adding a second checker with a different resolver, severity model, and allowlist produces divergent CI truth.

Please move the useful chain traversal into the existing #771 validator, make malformed YAML a blocking structural error, run its tests in CI, and give each temporary allowlist entry an owner/removal condition. External datasets can remain non-blocking classifications; invalid configuration should not be silently skipped.

akshita317:
The explicit `resolve_path()` handling of the `./`-prefix, absolute and relative cases is clear, and shipping `tests/test_check_example_paths.py` alongside the validator is good practice — a CI check that nothing tests is its own risk.

Two findings.

**1. This substantially overlaps PR #744**, which is also open and adds the same capability:

| | this PR | #744 |
|---|---|---|
| workflow | `.github/workflows/main.yaml` | `.github/workflows/main.yaml` |
| validator | `scripts/check_example_paths.py` | `scripts/validate_example_configs.py` |
| suppression file | `scripts/known_broken_paths.txt` | `scripts/example_config_baseline.txt` |
| method | walk `examples/`, parse `benchmarkingjob*`, resolve `testenv` + algorithm `url`, check existence | identical |

Both modify the same workflow file, so they cannot both merge cleanly, and running both would scan the repository twice per PR.

Each has something the other lacks: this PR has unit tests; #744 ships a generated baseline documenting ~30 currently-broken references with their resolved paths, which is independently useful. Combining this PR's script and tests with #744's generated baseline would give a single stronger check — probably better than closing either.

**2. Coverage note.** Path-existence validation does not catch the case where the path is valid and the key is discarded. `Dataset._parse_config` (`core/testenvmanager/dataset/dataset.py:71-73`) assigns only keys already declared on the instance, so `examples/yaoba/**/testenv.yaml` — whose `image_folder_url`, `known_dataset_url` and `unknown_dataset_url` all resolve fine — passes this check while those three values are silently dropped. `core/.../singletask_learning_active_boost.py:49,61,62` reads them, so the run fails with `AttributeError` despite green CI.

A schema assertion (every `self.dataset.<attr>` read in `core/` must be declared in `Dataset.__init__`) is about ten lines and would close that gap. On current `main` it reports `['image_folder_url', 'known_dataset_url', 'unknown_dataset_url', 'val_url']`. Given this PR already has a test harness, it might be a natural place to add it.

diegodevxd:
Executed follow-up on this PR head `5a2419391286a51342563cedee829acf5c0d7cee` through the real checker, using a malformed `benchmarkingjob` fixture:

```text
[WARN] Could not parse ...benchmarkingjob_malformed.yaml
NEW broken paths:     0
All broken paths are in the allowlist. CI passes.
EXIT=0
```

The parser warning does not become a broken record, so invalid YAML can produce a false-success result. In a local temporary patch, recording `YAML syntax: invalid YAML` whenever `parse_yaml_safe()` returns `None` made the same fixture exit 1. A separate valid chain with an absent external dataset still exited 0, so this change can block syntax errors without reclassifying external assets as structural failures.

Please add this regression coverage and make YAML parse failure an explicit blocking checker result.

sheikhayaan:
@diegodevxd's point about integrating into the #771 framework and @akshita317's #744 comparison both cover the scope question better than I could, so I won't restate them. I want to raise something narrower about the checker's traversal, because I think it affects whether the committed baseline can be trusted.

The allowlist can only contain breakage the walker was able to reach.

check_benchmarkingjob records a broken testenv reference and then gates the descent on that same path resolving:

python
testenv_resolved = resolve_path(testenv_raw, repo_root)
check_path_exists(testenv_raw, repo_root, filepath, "testenv", broken_paths, verbose)
if testenv_resolved.exists():                      # False for exactly the broken units
    check_testenv(testenv_resolved, repo_root, broken_paths, verbose)

and identically for algorithms:

python
if algo_resolved.exists():
    check_algorithm(algo_resolved, repo_root, broken_paths, verbose)

So when the top-level link is the broken thing, the checker records that one reference and never opens the file — every metrics[].url, model_eval.model_metric.url and modules[].url inside the unreachable subtree goes uninspected. This is a sibling of the silent-skip path @diegodevxd found in parse_yaml_safe(), but with a different trigger: his is malformed YAML, this is a well-formed file that was never reached.

The field distribution of the 47 committed entries shows it directly:

 19  testenv                        <- top level, from benchmarkingjob.yaml
 19  algorithms[N].url              <- top level, from benchmarkingjob.yaml
  5  model_eval.model_metric.url    <- requires descending into testenv.yaml
  3  metrics[N].url                 <- requires descending into testenv.yaml
  1  modules[N].url                 <- requires descending into an algorithm file

Nineteen algorithm files recorded as unreachable, and exactly one modules[].url entry in the whole allowlist. The 9 deeper entries all come from units whose chain did resolve: the cifar100/* testenvs and phys_scene_gen/.../algorithmenv.yaml. Wherever the chain broke, nothing below it was ever looked at.

Concrete instance. examples/cityscapes/singletask_learning_bench/semantic-segmentation has four broken references, all four in fields this script already knows how to check. The allowlist contains two:

benchmarkingjob.yaml:testenv:./examples/.../singletask_learning_bench/testenv/testenv.yaml
benchmarkingjob.yaml:algorithms[0].url:./examples/.../testalgorithms/rfnet/rfnet_algorithm.yaml

Missing, and both verified absent on disk:

rfnet_algorithm.yaml : modules[0].url  -> ./examples/.../testalgorithms/rfnet/basemodel.py   MISSING
testenv/testenv.yaml : metrics[0].url  -> ./examples/.../testenv/map.py                      MISSING

Why this matters for the gate and not just for the count. Because the baseline is a snapshot of reachable breakage, repairing a top-level link unmasks entries that were never baselined. A contributor who fixes only the two benchmarkingjob.yaml references, which is precisely the scope of #581 as filed, removes two allowlisted entries and simultaneously surfaces two that aren't in known_broken_paths.txt. CI then fails a PR that reduced the amount of broken configuration, and blames the contributor who improved things. That'll recur for each of the 19 masked units, and it's a poor first experience for exactly the kind of contribution this gate exists to encourage.

For contrast, the merged #771 check doesn't hit this, because it scans the files it finds in the example directory instead of following the chain, so unreachability can't hide anything from it. On this example it reports 4 of 4. I'm not arguing chain-following is wrong. It's strictly more precise about which field is at fault, which a directory scan can't tell you. The problem is chain-following combined with a baseline allowlist.

Suggestions:

If the chain walk stays, make masking explicit. When testenv_resolved.exists() is False, emit a distinct record, something like <subtree-unverified>:<path>, and baseline that. Unmasking then becomes a visible, reviewable baseline diff instead of a surprise CI failure, and the 19 masked units become countable.
Alternatively, walk the example directory for config YAMLs regardless of reachability, so deeper files get checked even when their parent link is stale.
Either way this interacts with the consolidation @diegodevxd and @akshita317 are proposing, and whichever implementation wins shouldn't inherit the gating.
The branch now has a real merge conflict against main, not just a stale mergeable_state, so a rebase is needed before any of this can be assessed on CI.

The counter-argument, and what decides it. The strongest reading against my analysis is that the gating is deliberate scoping and not a defect: once a top-level link is known broken, descending would only produce noisy, redundant errors about a subtree the contributor already has to fix, so reporting the root cause once and stopping is arguably better UX. I took that seriously, and as checker behaviour it's a reasonable design. What the evidence says is that it doesn't hold in combination with a committed allowlist, which is what this PR adds.

If the intent were "report the root cause once", the masked subtree would be reported as unverified. It isn't: the 19 units this allowlist records with a broken testenv link produce no record at all of what was skipped, so the information is lost and not merely deferred. Deliberate scoping would also be stable under repair, and this isn't: fixing the parent increases the number of findings, which is the opposite of the monotonic behaviour a baseline gate needs. The 19-to-1 ratio of algorithms[].url entries to modules[].url entries is the fingerprint of that instability, and it's measured, not inferred. Finally, a scoping choice would be documented; known_broken_paths.txt is presented as an inventory of broken paths with nothing distinguishing "checked and broken" from "never looked at", so a future reader can't tell the two apart.

So I'll accept the gating may be intentional in the checker. What I'm arguing is that it's unsound in a baseline. That's why suggestion 1 keeps the gating and only asks that masking be recorded, instead of asking for the traversal to be rewritten.

The merge interaction I'd most want a maintainer to see. #835, #744 and the merged #771 are three sources of CI truth for one property, and they already disagree on the answer. @akshita317 notes #744 ships a generated baseline of about 30 broken references; this PR commits 47; and #771 reports 4 of 4 on the cityscapes unit where this PR's baseline records 2. Three walkers, three counts, one property. The 30-versus-47 gap is worth someone checking independently of the overlap question, because two baselines for the same repository state should not differ by half unless the walkers disagree about what counts as reachable, which is the defect above.

If #835 merges while #771 is already running, a contributor gets a green from one and a red from the other with no way to tell which is authoritative. The order matters too: if #582 merges first, #835's baseline needs regenerating before it lands or it ships two stale entries; if #835 merges first, #582 turns CI red while making the repo better. Neither is a defect in either PR alone. It's a defect in merging both without deciding which check owns the property.

Recommendation: major revision. The underlying contribution is valuable and the defect class it targets is real. What I'd want before merge: record masked subtrees as explicit entries (suggestion 1) so the baseline is monotonic under repair; regenerate the baseline against whatever the merge order turns out to be; and state in the PR description how this coexists with the already-merged #771, since right now the two can return different verdicts for the same example. The rebase off dirty is needed regardless.

Yash4616:

The `parse_yaml_safe()` exception handling has already been reported in this thread, with an executed reproduction against this head, and I am not restating it.

Two additions.

The job pins the interpreter directly:

```yaml
      - name: Set up Python 3.9
        uses: actions/setup-python@v3
        with:
          python-version: "3.9"
```

Two other open PRs in this area are trying to move the CI matrix off 3.9 as an EOL-adjacent version. Hard-coding 3.9 in a new job works against that, and means this check will keep running on 3.9 after the rest of the matrix moves. Taking the version from the matrix, or from a single shared value, would avoid a second migration later. Note also that `actions/setup-python@v3` is well behind the `@v7` used elsewhere in the current workflows.

Second, this PR ships `tests/test_check_example_paths.py`, which makes it the PR that creates the top-level `tests/` directory. That has a consequence outside its own scope: PR #815 adds a pytest step guarded on `tests/` existing, so this PR would satisfy that guard. The two conflict with each other in `static_code_pylint_cicd.yaml.yaml`, so they cannot simply both be merged. Flagging it as sequencing information.

Recommendation: minor revision, on top of the findings already in this thread.

Prachi194agrawal:
Read through this one carefully - the YAML-chain resolver and the unit tests
are solid work, and `known_broken_paths.txt` is the right shape for landing a
gate on a repo that already carries debt. Blocking new breakage while
tolerating known debt is exactly the correct approach.

One thing worth checking before this goes further: much of this overlaps
infrastructure that landed on `main` on 2026-08-21 as #771
(`ken6078/feat/example-validator-pr`, merge commit `37a9c60`).
`.github/workflows/validator/static_validator.py` already implements
`_check_repo_path_references` and `_check_hardcoded_paths`, and it's wired
into `static_code_requirement_cicd.yaml` on `pull_request` for `examples/**`.

Running it against unmodified `main`:

```
ERROR     18  Repository path references exist
ERROR      1  benchmarkingjob.yaml exists
WARNING   32  Hardcoded local path check
total inventory entries: 48
entries with >=1 ERROR: 13
```

So the detection this PR adds is largely already running. This PR was opened
2026-08-18, three days before that merge, so the overlap looks like timing
rather than anything anyone missed.

**The concrete risk of shipping both.** Two independent path resolvers means
two definitions of "valid," maintained separately: this PR's
`known_broken_paths.txt` and the merged validator's `example_inventory.yaml`.
Those can disagree, and a config could pass one gate while failing the other.
Worth noting the existing inventory has already drifted on its own - it's
missing `examples/cifar100/fci_ssl/fedavg`, and its `examples/aoa` entry
points one directory level above where `benchmarkingjob.yaml` actually lives
(`examples/aoa/single_task_bench/`), which is the source of that single
`ERROR 1 benchmarkingjob.yaml exists` above. A second, separately-maintained
list doubles that surface.

**Suggestion, rather than a request.** Would it be feasible to rebase this so
the allowlist logic extends `services/regression_detector.py`'s existing
baseline-debt mechanism, instead of adding a parallel checker? The
`known_broken_paths.txt` idea seems to me the genuinely valuable part -
`regression_detector.py` models baseline debt implicitly by recomputing base
results per PR, and an explicit version-controlled list is more reviewable
than that. It would be a shame to lose it just because it arrived alongside a
duplicate detector.

Happy to be corrected if there's a reason to keep the two separate that I've
missed.

**Recommendation: major revision** — redirect to extend the merged validator
rather than parallel it. The design thinking here is right; the placement is
what I'd change.

Also relevant for whoever triages: #744, #652 and #851 are all open against
#743 with the same overall approach, and all four currently edit `main.yaml`,
so they conflict with each other as well as with this.

Environment: Ubuntu 24.04.3, Python 3.12.3, commit `37a9c60`.

asadulla-h:
Reviewing this as part of my LFX pre-test (full analysis in my pre-test Discussion, linked below).

This is a good architectural move — elevating path validation from the Example layer to CI. I ran the script both ways to understand the allowlist's behavior:

Without --allowlist:
Total broken paths: 47
Known (allowlisted): 0
NEW broken paths: 47

With --allowlist scripts/known_broken_paths.txt:
Total broken paths: 47
Known (allowlisted): 47
NEW broken paths: 0
All broken paths are in the allowlist. CI passes.

Worth noting: every broken path still prints as [BROKEN] in the per-file output either way — nothing is hidden from someone reading full logs. What changes is only the final CI verdict, which goes green while all 47 known-broken references (including the Examples I cover in my Task 1 finding) remain permanently unresolved as long as nobody reads past the summary line.

Two suggestions: (1) an expiry/shrink-tracking mechanism so the allowlist can't just grow forever, and (2) extend the script to also check vendored Python source for hardcoded absolute paths (/home/, /tmp/, C:\Users\) — currently it only walks .yaml/.yml files, so it wouldn't catch the mypath.py-style bug I found in Task 1 at all.

Full analysis: https://github.com/kubeedge/ianvs/discussions/923

Karan825:
Independent review of #835

Recommendation: major revision

This PR is valuable because it moves configuration-path checking into the shared CI/validation layer. It adds `scripts/check_example_paths.py`, tests, an allowlist for known broken paths, and CI integration.

The important architectural question is what path semantics the validator is validating.

Ianvs Core currently resolves relative local paths through `os.path.isfile(url)`, so relative paths depend on the process working directory unless the caller establishes a documented base directory.

I independently reproduced that behavior with a minimal local test: the same relative path resolved successfully from one working directory and failed from another.

Therefore, a validator that runs under one assumed CI working directory can potentially validate a path that behaves differently when the Example is invoked elsewhere.

A second concern is the `known_broken_paths.txt` allowlist. It is useful for preventing newly introduced broken paths while legacy failures remain, but it means the CI gate does not yet imply that all accepted Examples are actually path-clean.

I recommend:
1. defining an explicit path-resolution/base-directory contract;
2. testing representative configurations from more than one working directory where practical; and
3. treating the broken-path allowlist as temporary technical debt with explicit removal criteria.

This would make the CI validator a preventive restoration mechanism rather than only a guard against newly introduced broken references.

rajdhruvsingh:
This thread has seven substantive reviews already covering most of the surface area here — credit to @diegodevxd (malformed-YAML handling producing false-green CI, reproduced directly), @akshita317 (overlap with a sibling PR, contrasted feature-by-feature, plus a key-discard finding below), @sheikhayaan (the masking defect where a broken top-level link prevents the walker from checking its own subtree — the deepest finding here, with 19 of 47 allowlist entries shown to be affected), @Yash4616 (Python version pin and tests/-directory collision concerns), @Prachi194agrawal (redirecting toward extending the already-merged validator's baseline mechanism, with its live output quoted against the current base branch), @asadulla-h (confirmed allowlist on/off behavior via direct execution), and @Karan825 (working-directory-dependent path resolution risk).

One connection across two of these findings that I don't think has been stated yet: @akshita317's key-discard finding (a YAML key not declared on the target class's __init__ gets silently dropped, so a valid-looking config still crashes downstream) and the value-overwrite class of defect this repository's broader schema work has been tracking are the same underlying failure mode — silent loss of a config-supplied value — reached by two different mechanisms. Both are invisible to path-existence checking, which is what this PR (like the already-merged validator) performs. A single generalized rule — "for every class-declared config attribute, trace whether the YAML-supplied value survives unmodified to first read" — would close both defect classes with one check instead of two separate proposals, which might be worth considering however the consolidation question here gets resolved.

ROGUESA1LOR:
### Review finding: dataset configuration is outside the validator's coverage

The new validator checks the `benchmarkingjob → testenv → algorithm → module` chain, but `check_testenv()` does not inspect the dataset fields:

- `train_index`
- `test_index`
- `train_data`
- `test_data`
- `train_data_info`
- `test_data_info`

I independently reproduced this with a minimal configuration containing nonexistent `train_index` and `test_index` values. The validator reported:

`Total broken paths: 0`
`NEW broken paths: 0`

and exited successfully.

This means a configuration can pass this CI check while Ianvs later fails during dataset processing.

I recommend extending the validation coverage to dataset references, while keeping genuinely external/downloaded datasets distinguishable from repository-local configuration errors.

Verification boundary: this finding is based on static inspection plus an independent synthetic reproduction; no external dataset or GPU was required.

31groot:
### Request changes :  dataset paths are not checked

**Review depth: Analysis**

The new CI validator catches several local Example paths, but it doesn't check dataset references such as:

```text
train_data
test_data
train_index
test_index
train_data_info
test_data_info
```

I verified this against PR head `5a24193`:

```text
checks_testenv: True
checks_metrics: True
checks_model_metric: True
checks_dataset_train_data: False
checks_dataset_test_data: False
checks_dataset_train_index: False
checks_dataset_test_index: False

Validator explicitly handles dataset paths: False
```

I also reproduced the blind spot with a minimal Example containing valid config and algorithm paths but intentionally missing:

```yaml
train_data: "./examples/demo/dataset/DOES_NOT_EXIST.txt"
test_data: "./examples/demo/dataset/ALSO_MISSING.txt"
```

The validator still reported:

```text
[OK] benchmarkingjob.yaml: testenv
[OK] benchmarkingjob.yaml: algorithms[0].url
broken paths reported: []
DATASET PATH BLIND SPOT REPRODUCED
```

So broken local dataset paths can currently pass the new CI check and only fail when the Example actually runs.

**Recommendation:** extend the validator to check the local dataset paths used by the Example, while ignoring external URLs. Please add a regression test with a missing local dataset file and verify that the validator reports it.

**Verification:** Reproduced locally against PR head `5a24193` using an isolated temporary Example configuration.

rahulnr107:
### Review — PR #835

---

### Summary

PR #835 introduces repository-wide validation for broken Example configuration paths.

It adds:

* a `check-example-paths` CI job;
* `scripts/check_example_paths.py`;
* an allowlist for existing known broken references;
* unit tests for path resolution and configuration-chain validation.

---

### Technical Assessment

The general approach is useful because it moves configuration-path validation earlier in the lifecycle:

```text
Example configuration
        ↓
CI path validation
        ↓
Broken reference detected
        ↓
Prevent new regression
```

The validator follows references from benchmarking jobs into TestEnv files, algorithm configurations, metric URLs and module URLs.

This can detect invalid configuration references before an Example reaches benchmark runtime.

---

### Positive Design Choice — Existing Broken Path Allowlist

The allowlist provides a practical migration strategy.

The intended behavior is:

```text
Existing broken reference
        ↓
Present in allowlist
        ↓
Reported, but CI may continue
```

For a new broken reference:

```text
New broken reference
        ↓
Not present in allowlist
        ↓
Validator fails
```

This allows path validation to be introduced without requiring every existing broken Example to be repaired in the same PR.

However, allowlist entries should be removed when the corresponding configurations are fixed. Otherwise, existing technical debt could remain permanently accepted.

---

### Independent Verification — Unit Tests

I executed the PR’s targeted test suite locally on Windows:

```powershell
python -m pytest tests/test_check_example_paths.py -v
```

Environment:

```text
Windows
Python 3.13.14
pytest 9.1.1
```

Observed result:

```text
collected 7 items

test_resolve_path PASSED
test_valid_config_chain PASSED
test_broken_testenv_path PASSED
test_broken_module_url PASSED
test_absolute_path_detected PASSED
test_missing_optional_model_eval PASSED
test_load_allowlist PASSED

7 passed in 0.18s
```

The tests cover:

* relative path resolution;
* operating-system-specific absolute paths;
* valid configuration chains;
* broken TestEnv references;
* broken module references;
* absolute missing paths;
* missing optional `model_eval`;
* allowlist loading.

Screenshot (80)


---

### Independent Verification — Repository Validator

I then executed the repository-wide validator introduced by the PR:

```powershell
python scripts/check_example_paths.py --allowlist scripts/known_broken_paths.txt
```

The validator loaded:

```text
47 known broken path(s)
```

and found:

```text
44 benchmarkingjob file(s) to check
```

The final summary was:

```text
============================================================
Config Path Validation Summary
============================================================
Files checked:        44
Total broken paths:   47
Known (allowlisted):  47
NEW broken paths:     0
============================================================

All broken paths are in the allowlist. CI passes.
```

However, during the same execution, the validator also reported:

```text
[WARN] Could not parse
examples\cifar100\fci_ssl\fedavg\benchmarkingjob.yaml
```

with the YAML parsing error:

```text
while scanning a simple key
...
could not find expected ':'
```
Screenshot (81)


---

### Main Finding — YAML Parse Failures Can Produce a False-Green Result

The validator’s `parse_yaml_safe()` function catches YAML parsing exceptions, prints a warning and returns `None`.

The caller then stops processing that configuration without recording the parse failure as a validation failure.

The resulting behavior is:

```text
Malformed YAML
      ↓
Parsing fails
      ↓
Warning only
      ↓
Configuration skipped
      ↓
No new broken path recorded
      ↓
Validator can still report success
```

I reproduced this behavior against the current repository state.

The validator encountered a benchmarking job that it could not parse, but its final result still reported:

```text
NEW broken paths: 0
All broken paths are in the allowlist. CI passes.
```

This creates a false-green condition.

A configuration that cannot be parsed cannot have its path references validated. Unlike an allowlisted broken path, the malformed YAML file is not deliberately recorded and accepted. It is skipped because parsing failed.

For a CI validation tool, this can hide configurations that are invalid or cannot be inspected.

---

### Cross-Platform Observation

The validator contains operating-system-dependent path handling.

Its unit tests explicitly distinguish between Windows and non-Windows behavior using logic based on:

```python
os.name == "nt"
```

However, the proposed GitHub Actions job runs only on Ubuntu.

My local Windows execution confirms that the current Windows-specific tests pass.

Adding Windows to the CI matrix would protect this behavior automatically against future platform-specific path regressions.

---

### Recommendation

**Major revision**

The overall validation approach is useful, but I consider the YAML parse-error handling blocking for a CI validation tool.

A malformed configuration currently produces a warning and is skipped, allowing the validator to report success even though the file was not validated.

I recommend:

1. treating YAML parsing failures as explicit validation failures;
2. returning a non-zero status when an unapproved parse failure occurs;
3. adding a regression test that creates malformed YAML and verifies that validation fails;
4. optionally providing a separate, auditable mechanism if specific existing parse failures must temporarily be accepted;
5. adding Windows coverage to the CI job because the validator contains OS-dependent path behavior.

---

### Verification Boundary

I executed:

* the seven targeted validator unit tests;
* the repository-wide configuration validator on Windows.

I did not execute all 44 discovered benchmark Examples themselves.

Therefore, this verification demonstrates the behavior of the validator, not successful runtime execution of every Example it scans.

---

asadulla-h:
Following up as part of my LFX pre-test (full Repair Boundary analysis in my pre-test Discussion, linked below) — this one is directly actionable for this PR, so I built and tested the extension rather than just describing it.

This script currently only validates YAML config references (testenv/algorithms[0].url). I extended it to also scan vendored Python source for hardcoded absolute paths, using the same structure and allowlist flow already in this PR:

def find_vendored_python_files(examples_dir):
    results = []
    for root, _dirs, files in os.walk(examples_dir):
        for filename in files:
            if filename.endswith(".py"):
                results.append(Path(root) / filename)
    return sorted(results)

def check_python_hardcoded_paths(filepath, repo_root, broken_paths, verbose=False):
    HARDCODED_PATH_PATTERN = re.compile(
        r"""['"](/home/[^'"]+|/tmp/[^'"]+|/root/[^'"]+|[A-Za-z]:\\[^'"]+)['"]"""
    )
    with open(filepath, "r", encoding="utf-8") as fh:
        for lineno, line in enumerate(fh, start=1):
            match = HARDCODED_PATH_PATTERN.search(line)
            if match:
                rel_source = filepath.relative_to(repo_root)
                broken_paths.append({"source": str(rel_source), "field": f"line {lineno}", "path": match.group(1)})
                print(f"  [BROKEN] {rel_source}: line {lineno} -> {match.group(1)}")

Running this against the full repo: 156 total broken paths found (vs. 47 with YAML-only checking) — a 109-instance gap, across 9+ Examples beyond the ones I originally scoped in Task 1 (bdd, cifar100, cityscapes, government, government_rag, llm_simple_qa, smart_coding, yaoba, robot-cityscapes-synthia).

I also confirmed this doesn't just flag everything indiscriminately — running it against #794's already-fixed mypath.py (./dataset/) produces zero matches.

Recommendation: this regex is a starting point, not final — happy to open a draft PR extending this one if useful, but wanted to surface the scale of the gap first.

Full analysis: https://github.com/kubeedge/ianvs/discussions/923

asadulla-h:
Restoration path follow-up (full details in my pre-test Discussion, linked below).

Used the Python-source extension I built on top of this PR in Task 3 as the actual Verification Gate for restoring robot — ran it before and after applying the path fix (via git stash/stash pop for a controlled same-session comparison) and got a clean -7 delta, exactly matching the number of hardcoded lines fixed. This is functioning as the CI-level regression gate my restoration path relies on.

Full path: https://github.com/kubeedge/ianvs/discussions/923

shrav784:
This thread is very thorough already — not restating the duplication/masking/dataset-coverage findings above.

One connection I don't see made yet: several here (@diegodevxd, @Prachi194agrawal) suggest consolidating this into the already-merged #771 validator framework. Worth flagging before that happens — #771's own CI-facing aggregation (report_generator.py's `CombinedReport.passed`, which gates the actual CI exit code at line 323) still has zero concept of "executed" vs "skipped," independent of the `dynamic_example_result()` display function discussed on #836. So consolidating into #771 as-is would inherit that same blind spot rather than fix it. Full trace: [Discussion #973]

Atharva-Waghmare:
Thank you all — this is an unusually thorough review and I want to act on it properly. Planned revision order: (1) fix the YAML-parse-failure false-green and the baseline-masking non-monotonicity — these are correctness bugs in the checker itself; (2) extend coverage to dataset fields and add the Dataset schema assertion; (3) update the PR description to state explicitly how this coexists with the merged https://github.com/kubeedge/ianvs/pull/771, and I'm open to consolidating with https://github.com/kubeedge/ianvs/pull/744 or extending regression_detector.py instead of a parallel checker, whichever maintainers prefer; (4) the smaller CI/process items alongside. @asadulla-h — welcome you opening the hardcoded-path scanner as its own follow-on PR against this; no need to duplicate that here.

PR #839, opened 2026-08-19
fix(storymanager): correct save_mode validation in rank.py

### LINKED ISSUE #828: [Bug] Unsupported save_mode silently skips all leaderboard persistence, and the save_mode type check tests for list instead of str

Arijit429:
Fixes #828

`save_mode` validation in `core/storymanager/rank/rank.py` has two bugs that cause silent failures:

1. Wrong type check — `isinstance(self.save_mode, list)` but `save_mode` is always a string
2. Wrong operator — `and` instead of `or` means a typo'd string value like `"selcted_only"` makes `not self.save_mode` false, short-circuiting the entire check with no exception raised

Net effect: a typo in `save_mode` in `benchmarkingjob.yaml` causes Ianvs to run the full benchmark and then silently produce no leaderboard output — no CSV, no error message, just an empty output directory.

**Fix**:

```python
valid_save_modes = ("selected_and_all", "selected_only", "selected_and_all_and_picture")
if not isinstance(self.save_mode, str) or self.save_mode not in valid_save_modes:
    raise ValueError(
        f"rank's save_mode({self.save_mode}) must be one of {valid_save_modes}."
    )
```

This validates against the correct type and the exact set of accepted values, so unsupported values fail loudly at config-parse time instead of silently at save time.

Fixes #828

### DIFF
--- a/core/storymanager/rank/rank.py
+++ b/core/storymanager/rank/rank.py
@@ -276,13 +276,18 @@ class Rank:
             self._save_all()
             self._save_selected(test_cases, test_results)
 
-        if self.save_mode == "selected_only":
+        elif self.save_mode == "selected_only":
             self._save_selected(test_cases, test_results)
 
-        if self.save_mode == "selected_and_all_and_picture":
+        elif self.save_mode == "selected_and_all_and_picture":
             self._save_all()
             self._save_selected(test_cases, test_results)
             self._draw_pictures(test_cases, test_results)
+
+        else:
+            raise ValueError(
+                f"rank's save_mode({self.save_mode}) is not supported by save()."
+            )
 
     def plot(self):
         """
--- a/core/storymanager/rank/rank.py
+++ b/core/storymanager/rank/rank.py
@@ -85,10 +85,10 @@ class Rank:
         if not self.selected_dataitem.get("metrics"):
             raise ValueError("not found metrics of selected_dataitem in rank.")
 
-        if not self.save_mode and not isinstance(self.save_mode, list):
+        valid_save_modes = ("selected_and_all", "selected_only", "selected_and_all_and_picture")
+        if not isinstance(self.save_mode, str) or self.save_mode not in valid_save_modes:
             raise ValueError(
-                f"rank's save_mode({self.save_mode}) "
-                f"must be provided and be list type."
+                f"rank's save_mode({self.save_mode}) must be one of {valid_save_modes}."
             )
 
     @classmethod
--- a/core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning_active_boost.py
+++ b/core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning_active_boost.py
@@ -48,8 +48,8 @@ class SingleTaskLearningACBoost(SingleTaskLearning):
         # Load test set data
         img_prefix = self.dataset.image_folder_url
         ann_file_path = self.dataset.test_url
-        with open(ann_file_path, mode="r", encoding="utf-8") as f:
-            ann_file = json.load(f)
+        with open(ann_file_path, mode="r", encoding="utf-8") as file_handle:
+            ann_file = json.load(file_handle)
         test_set = []
         for i in ann_file['images']:
             test_set.append(os.path.join(img_prefix, i['file_name']))
--- a/core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning_active_boost.py
+++ b/core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning_active_boost.py
@@ -48,7 +48,8 @@ class SingleTaskLearningACBoost(SingleTaskLearning):
         # Load test set data
         img_prefix = self.dataset.image_folder_url
         ann_file_path = self.dataset.test_url
-        ann_file = json.load(open(ann_file_path, mode="r", encoding="utf-8"))
+        with open(ann_file_path, mode="r", encoding="utf-8") as f:
+            ann_file = json.load(f)
         test_set = []
         for i in ann_file['images']:
             test_set.append(os.path.join(img_prefix, i['file_name']))
--- a/core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning_tta.py
+++ b/core/testcasecontroller/algorithm/paradigm/singletask_learning/singletask_learning_tta.py
@@ -45,8 +45,8 @@ class SingleTaskLearningTTA(SingleTaskLearning):
         # Load test set data
         img_prefix = self.dataset.image_folder_url
         ann_file_path = self.dataset.test_url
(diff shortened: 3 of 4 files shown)


yadavchiragg:
Review Depth Level: Analysis
What it does: Fixes two compounding bugs in save_mode validation: (1) isinstance(self.save_mode, list) checked the wrong type — save_mode is always a string, never a list; (2) and was used instead of or, so a typo'd value like "selcted_only" made the whole guard evaluate False and silently skip validation with no exception raised. Net effect: a single typo in a YAML config causes Ianvs to run the entire benchmark and then silently produce no leaderboard output — no CSV, no error, just an empty directory.
Finding worth flagging: this is the third instance of an and/or inversion bug found across this pretest cycle (alongside #336 and its duplicate #849) — but in a different core file (storymanager/rank.py vs. _check_fields()). Three independent instances of the same category of logic error, in different modules, authored by different people, strongly suggests this isn't isolated code-review-slip risk but a recurring pattern worth a project-wide static-analysis pass (e.g., a linter rule flagging and/or inside boolean validation guards) — a genuinely new, broader finding beyond anything in your Task 1–4 scope.

yadavchiragg:
Good catch on the compounding bug (wrong type check + and/or inversion stacking to fully silence the error). Worth noting for maintainers: this is now the third known instance of the and/or inversion pattern in Ianvs Core, alongside the _check_fields() family (#336/#590/#596/#849). Three independent occurrences across different files suggests a static-analysis/lint rule catching "and inside a not/isinstance validation guard" would be more durable than continuing to catch these one at a time.

Laiiiiba:
**Review Minor revision**

I reviewed the industrialEI `llio_fusion/utils.py` changes, focusing specifically on the visualization path rather than the validation logic already discussed in the existing review.

There is a potential backend-ordering issue in `visualize()`: `matplotlib.pyplot` is imported at module load time, and `plt.figure()` is created before the function later calls:

    matplotlib.use("TkAgg")

Matplotlib backend selection is normally expected to happen before `pyplot` is imported and before figures are created. Calling `matplotlib.use()` at this point may therefore be too late to reliably switch the backend, and can cause problems in headless/CI environments rather than providing the intended Tk backend.

I recommend moving the backend selection to an appropriate initialization point before `pyplot` is imported, or otherwise making the visualization backend selection explicit and safe for headless execution.

The existing optional-import handling is useful, but it does not by itself address this later backend-selection ordering.

 **Minor revision.**

My review here is limited to this visualization/backend issue; I am not duplicating the existing discussion about the validation logic.

Issue #840, opened 2026-08-19
Fresh-install failure summary for cloud-edge-collaborative-inference-for-llm: both documented install routes fail


zjr060424-lab:
## 1. Environment

- OS: Windows 11 Pro (build 26100); path-related findings are OS-independent (verified against repo state on `main`)
- Python: 3.11 (clean venv created for this walkthrough; cross-checked on 3.13)
- pip 26.2.1 / setuptools 84.0.0 / wheel 0.48.0
- Ianvs: `main` @ 36ec008 (fresh clone)

## 2. Steps to Reproduce

Followed the [Quick Start guide](https://ianvs.readthedocs.io/en/latest/guides/quick-start.html) (same content as `docs/guides/quick-start.md`), "Detailed Setup Guide" section, command by command, on a completely clean virtual environment.

## 3. Findings

<details><summary><b>Finding 1 (hard failure)</b>: the documented sedna wheel path does not exist — manual install route breaks at step 1</summary>

Quick-start step:

```bash
pip install examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl
```

Output:

```
WARNING: Requirement 'examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl' looks like a filename, but the file does not exist
Processing ./examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl
ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '<repo>/examples/resources/third_party/sedna-0.6.0.1-py3-none-any.whl'
```

The wheel actually lives at `resources/third_party/sedna-0.6.0.1-py3-none-any.whl` (repo root) — the directory was moved, and CI already uses the new location (`.github/workflows/main.yaml` line 31: `pip install ${{github.workspace}}/resources/third_party/*`), but the docs were never updated. Installing from the real path succeeds.

</details>

<details><summary><b>Finding 2 (hard failure)</b>: the Docker route is broken by the same stale path — plus 17 more files referencing it</summary>

The quick start's recommended "Docker-Based Setup" cannot work either: `examples/cloud-edge-collaborative-inference-for-llm/Dockerfile` line 27 runs the same install against the nonexistent path (the Dockerfile clones `kubeedge/ianvs` inside the image, so this is guaranteed by current repo state, independent of the build context):

```dockerfile
RUN /bin/bash -c "source activate $CONDA_ENV && \
[...]

PR #841, opened 2026-08-19
docs: fix quick-start install paths after third_party directory move

zjr060424-lab:
## What this PR does

Fixes the quick-start critical install path in 4 files:

| File | Change |
|---|---|
| `docs/guides/quick-start.md` | `examples/resources/third_party/` → `resources/third_party/`; `python setup.py install` → `pip install .` |
| `docs/guides/how-to-install-ianvs.md` | same two fixes |
| `examples/cloud-edge-collaborative-inference-for-llm/README.md` | same two fixes |
| `examples/cloud-edge-collaborative-inference-for-llm/Dockerfile` | same two fixes — `docker build` currently fails at the sedna install layer |

## Why

The third-party wheel directory moved from `examples/resources/third_party/` to `resources/third_party/` (CI already uses the new path — see `.github/workflows/main.yaml` line 31), but the install guides and the quick-start example's README/Dockerfile were never updated. As a result **both documented install routes (manual and Docker) fail on a fresh clone**. Full logs and root-cause analysis in #840.

`python setup.py install` is deprecated by setuptools; `pip install .` is the supported equivalent and verified working on the same environments (Python 3.11/3.13, setuptools 84 — logs in the issue).

## Scope note

`grep -r "examples/resources/third_party"` shows ~17 more files (mostly example READMEs and `build.envd`) referencing the stale path. This PR intentionally fixes only the quick-start critical path; I'm happy to sweep the rest in a follow-up PR if maintainers agree.

Addresses the quick-start portion of #840


mwemmz:
Diffed this directly. Confirms the install path fix is correct: `./examples/resources/third_party/*` does not exist in the current tree — verified by attempting the install myself while setting up this pretest's environment; the wheel now lives at `./resources/third_party/`, exactly where this PR points. Also confirmed `python setup.py install` → `python -m pip install .` is the right call, since `setup.py install` is deprecated in current `setuptools`.

One thing worth checking before merge: this PR updates the Dockerfile for `examples/cloud-edge-collaborative-inference-for-llm` specifically. Given the same `third_party` move likely affected other examples' setup instructions too, it may be worth a repo-wide `grep -r "examples/resources/third_party"` to confirm no other Dockerfile or install script has the same stale reference before considering this fully resolved repo-wide (scope of this PR itself looks correct and mergeable regardless).

hibaa23:

Confirmed the diff is exactly correct for the 4 files it touches — verified against `main` @ 36ec008, the path substitution and `setup.py install` → `pip install .` modernization are both accurate and low-risk.

One gap worth tracking explicitly: an extension-unfiltered repo-wide `git grep -rln "resources/third_party" 36ec008` returns 23 matches (not just `.md`/`.py` — it also catches `build.envd` and this example's own Dockerfile); after excluding the 2 already using the corrected path, 21 files across 15 other example families still reference the stale path, of which this PR fixes 4. Recommend either linking a tracking issue for the remaining 17 files now, or converting the "happy to do this in a second PR" note into a concrete follow-up issue.

mwemmz:
Following up on @hibaa23's point above (repo-wide scope of the stale `resources/third_party` path). Re-ran the check independently against current `main` (`37a9c60`, 2026-08-21):

```
git grep -l "examples/resources/third_party" -- '*.md' '*.py'
```

Confirms **19 files still reference the stale path**: the two core docs (`how-to-install-ianvs.md`, `quick-start.md`) plus one `docs/proposals/chore/` design doc, and README/readme files across 13 separate example families (Cloud_Robotics ×2, TAB, bdd, cityscapes-synthia — 3 variants, cloud-edge-collaborative-inference-for-llm, cloud_VLA_finetune, imagenet, industrialEI, llm-edge-benchmark-suite, phys_scene_gen, robot-cityscapes-synthia, robot, RoboDK Palletizing). Scope note: this `.md`/`.py`-filtered count includes 3 files this PR itself fixes (both guides plus the cloud-edge LLM example's README), so 16 would remain after merge under this same filter — consistent with @hibaa23's higher unfiltered count, which additionally catches `build.envd`/Dockerfile references.

This is the same defect pattern this PR fixes (a `resources/` directory move with no reference update), just recurring at a scale beyond any single PR's reasonable diff size. +1 on converting this into a tracked follow-up issue rather than leaving it as a comment thread — happy to help enumerate the full file list if useful.

AmanBasu20:
Confirmed the diff is correct for the 4 files it touches, consistent with the earlier reviews from `mwemmz` and `hibaa23` — the wheel path substitution and `setup.py install` → `pip install .` modernization are both accurate.

One connection worth surfacing: the documentation drift this PR fixes is causally upstream of the `colorlog` masking mechanism described in #840 Finding 3. A contributor who successfully installs the *old, documented* wheel path (before this PR merges) gets `colorlog` for free as Sedna's transitive dependency, which is precisely why #840 describes that bug as intermittent rather than universal. Once this PR (or the full ~17–21-file sweep) lands and contributors move to the corrected path, that transitive masking disappears — meaning this documentation fix, independent of any `requirements.txt` change, will make the Core dependency gap tracked in #840/#742 *newly visible* to contributors who were previously shielded from it by the stale path.

Recommend sequencing this PR and a `colorlog`-in-`requirements.txt` fix (see #791) together, rather than merging them independently, to avoid a window where contributors hit a *new* crash right after this otherwise-correct documentation fix lands.

PR #843, opened 2026-08-20
fix(core): fix evaluation metric callable, dataset extraction, and inference aggregation in incremental learning

### LINKED ISSUE #842: [Bug] Evaluation TypeError, UnboundLocalError on hard example generation, and inference update failures in IncrementalLearning
**What happened**:

During an evaluation of the Incremental Learning paradigm (`core/testcasecontroller/algorithm/paradigm/incremental_learning/incremental_learning.py`), several interrelated correctness and reliability defects were identified across model evaluation, hard example dataset creation, inference aggregation, and update triggering:

### 1. `TypeError: 'tuple' object is not callable` in `_eval()`
In `IncrementalLearning._eval()` (line 186):
```python
job = self.build_paradigm_job(ParadigmType.INCREMENTAL_LEARNING.value)
eval_dataset = self.dataset.load_data(data_index_file, "eval")
eval_results = job.evaluate(eval_dataset, metric=get_metric_func(model_metric))
```
`get_metric_func(model_metric)` returns a 2-tuple `(name, metric_func)` (defined in `core/testcasecontroller/metrics/metrics.py`). Sedna's `IncrementalLearning.evaluate(metric=...)` expects a callable metric function. Passing the unpacked tuple causes a runtime exception:
`TypeError: 'tuple' object is not callable`
(In contrast, `lifelong_learning.py` and `testcase.py` unpack `_, metric_func = get_metric_func(...)` before passing to evaluate).

### 2. `UnboundLocalError` and Stale Label Assignment in `_get_train_dataset()`
In `IncrementalLearning._get_train_dataset()` (lines 155-160):
```python
with open(train_dataset_file, "w", encoding="utf-8") as file:
    for old, new in hard_examples:
        index = np.where(data_labels.x == old)
        if len(index[0]) == 1:
            label = data_labels.y[index[0][0]]
        file.write(f"{new} {label}\n")
```
- If the first hard example has 0 or >1 matches in `data_labels.x`, `label` is never assigned, raising `UnboundLocalError: local variable 'label' referenced before assignment`.
- If a subsequent hard example has 0 or >1 matches, `label` is not updated and retains the previous sample's label, silently writing incorrect ground-truth labels into the retraining dataset index.

### 3. `TypeError` / `ValueError` on Dictionary Update in `_inference()`
In `IncrementalLearning._inference()` (lines 136-140):
```python
inference_results = {}
for _, data in enumerate(inference_dataset_x):
    res, _, is_hard_example = job.inference([data])
    inference_results.update(res)
```
`inference_results` is initialized as a dictionary `{}`. When Sedna estimators return prediction lists or numpy arrays (e.g. `[class_id]`), calling `dict.update(res)` fails with `TypeError: 'list' object is not iterable` or `ValueError: dictionary update sequence element #0 has length 1; 2 is required`.

### 4. `TypeError` on `NoneType` Subtraction in `_trigger_model_update()`
In `IncrementalLearning._trigger_model_update()` (lines 215-220):
```python
metric_values = [0, 0]
for i, result in enumerate(eval_results):
    metrics = result.get("metrics")
    metric_values[i] = metrics.get(metric_name)

metric_delta = metric_values[0] - metric_values[1]
```
If `metrics` is `None` or `metric_name` is missing from the evaluation dictionary, `metric_values[i]` is `None`, causing `TypeError: unsupported operand type(s) for -: 'NoneType' and 'NoneType'`.

---

1. `_eval()` should unpack `get_metric_func(model_metric)` and pass `metric_func` (callable) to `job.evaluate(metric=...)`.
2. `_get_train_dataset()` should handle label lookup safely without `UnboundLocalError` or reusing stale labels from previous loop iterations.
3. `_inference()` should collect prediction results into a list structure (or properly format dictionary updates if key-value mappings are used), consistent with other learning paradigms.
4. `_trigger_model_update()` should validate that metric values are numeric before performing subtraction.

---

Run the following standalone script demonstrating the failures:

```python
import numpy as np

# 1. Reproduce metric tuple passing
def get_metric_func(metric_dict):
    return metric_dict.get("name"), lambda y_true, y_pred: 0.95

metric_arg = get_metric_func({"name": "accuracy"})
try:
    metric_arg([1, 0], [1, 0])
except TypeError as e:
    print("Defect 1 caught:", e)

# 2. Reproduce UnboundLocalError / stale label
hard_examples = [("img1.jpg", "new1.jpg")]
data_labels_x = np.array(["img2.jpg"])
data_labels_y = np.array(["cat"])

try:
    for old, new in hard_examples:
        index = np.where(data_labels_x == old)
        if len(index[0]) == 1:
            label = data_labels_y[index[0][0]]
        line = f"{new} {label}\n"
except UnboundLocalError as e:
    print("Defect 2 caught:", e)

# 3. Reproduce dict.update with list prediction
inference_results = {}
res = [0]
try:
    inference_results.update(res)
except (TypeError, ValueError) as e:
    print("Defect 3 caught:", e)
```

**Observed Output**:
```
Defect 1 caught: 'tuple' object is not callable
Defect 2 caught: cannot access local variable 'label' where it is not associated with a value
Defect 3 caught: cannot convert dictionary update sequence element #0 to a sequence
```

---

- `core/testcasecontroller/algorithm/paradigm/incremental_learning/incremental_learning.py` (`_eval`, `_get_train_dataset`, `_inference`, `_trigger_model_update`)

### Root Cause
1. `_eval` omits unpacking `get_metric_func()`, passing `(name, func)` directly into `job.evaluate(metric=...)`.
2. `_get_train_dataset` conditionally binds `label` only on exact single matches without handling missing matches or resetting `label` per loop iteration.
3. `_inference` assumes `res` is a dictionary or dictionary-convertible sequence for `update()`.
4. `_trigger_model_update` assumes `metrics.get(metric_name)` always returns a non-None float.

### Acceptance Criteria
1. `_eval()` correctly unpacks `metric_func` and passes it to Sedna's evaluation interface.
2. `_get_train_dataset()` resolves sample labels safely and raises clear errors or applies safe defaults when a sample is missing from `data_labels`.
3. `_inference()` supports list and array predictions from estimators.
4. `_trigger_model_update()` validates metric values prior to computing delta.
5. Unit tests added under `tests/` covering incremental learning evaluation, dataset extraction, and model update triggering.

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

This PR resolves four interrelated correctness and reliability defects in the Incremental Learning paradigm (`core/testcasecontroller/algorithm/paradigm/incremental_learning/incremental_learning.py`):

1. **Fix `TypeError: 'tuple' object is not callable` in `_eval()`**:
   - `get_metric_func(model_metric)` returns a 2-tuple `(name, metric_func)`. `_eval()` now unpacks `_, metric_func = get_metric_func(...)` and passes the callable `metric_func` to Sedna's `job.evaluate(metric=...)`.

2. **Fix `UnboundLocalError` & Stale Label Corruption in `_get_train_dataset()`**:
   - Safely checks `len(index[0]) > 0` before assigning `label = data_labels.y[index[0][0]]`, avoiding unassigned variable crashes or reusing stale labels from previous iterations when a sample is missing from `data_labels`. Also writes temporary files to `self.dataset_output_dir()` to avoid workspace leaks.

3. **Fix `TypeError` / `ValueError` on Inference Aggregation in `_inference()`**:
   - Changed `inference_results` to a list structure and extended/appended predictions appropriately, preventing `dict.update()` crashes when Sedna estimators return list or array predictions.

4. **Defensive Metric Validation in `_trigger_model_update()`**:
   - Validates that `eval_results` is a 2-element sequence and extracts numeric metric values safely (supporting nested `metrics` dicts, direct dict values, or scalar numbers) before computing the comparison delta.

5. **Unit Tests**:
   - Added unit test suite `tests/test_incremental_learning.py` covering callable metric passing, missing label handling in hard examples, list inference aggregation, and model update triggering logic (5 tests passed).

Fixes #842

### DIFF
--- a/core/testcasecontroller/algorithm/paradigm/incremental_learning/incremental_learning.py
+++ b/core/testcasecontroller/algorithm/paradigm/incremental_learning/incremental_learning.py
@@ -137,7 +137,10 @@ class IncrementalLearning(ParadigmBase):
         hard_examples = []
         for _, data in enumerate(inference_dataset_x):
             res, _, is_hard_example = job.inference([data])
-            inference_results.update(res)
+            if isinstance(res, dict):
+                inference_results.update(res)
+            else:
+                inference_results[data] = res
             if is_hard_example:
                 shutil.copy(data, hard_example_saved_dir)
                 new_hard_example = os.path.join(hard_example_saved_dir, os.path.basename(data))
@@ -150,18 +153,34 @@ class IncrementalLearning(ParadigmBase):
         # pylint: disable=W0012
         # pylint: disable=E0606
         data_labels = self.dataset.load_data(data_label_file, "train label")
-        temp_dir = tempfile.mkdtemp()
+
+        missing_or_ambiguous = []
+        rows = []
+        for old, new in hard_examples:
+            index = np.where(data_labels.x == old)
+            if len(index[0]) == 1:
+                label = data_labels.y[index[0][0]]
+                rows.append((new, label))
+            else:
+                missing_or_ambiguous.append(old)
+
+        if missing_or_ambiguous:
+            raise ValueError(
+                f"Hard example(s) not found or ambiguous in train label dataset: "
+                f"{missing_or_ambiguous}"
+            )
+
+        temp_dir = tempfile.mkdtemp(dir=self.dataset_output_dir())
         train_dataset_file = os.path.join(temp_dir, os.path.basename(data_label_file))
+
         with open(train_dataset_file, "w", encoding="utf-8") as file:
-            for old, new in hard_examples:
-                index = np.where(data_labels.x == old)
-                if len(index[0]) == 1:
-                    label = data_labels.y[index[0][0]]
+            for new, label in rows:
                 file.write(f"{new} {label}\n")
 
         return train_dataset_file
 
     def _train(self, model, data_index_file, rounds):
+        # pylint: disable=E1101
         train_output_dir = os.path.join(self.workspace, f"output/train/{rounds}")
         if not is_local_dir(train_output_dir):
             os.makedirs(train_output_dir)
@@ -183,7 +202,8 @@ class IncrementalLearning(ParadigmBase):
 
         job = self.build_paradigm_job(ParadigmType.INCREMENTAL_LEARNING.value)
         eval_dataset = self.dataset.load_data(data_index_file, "eval")
-        eval_results = job.evaluate(eval_dataset, metric=get_metric_func(model_metric))
+        _, metric_func = get_metric_func(model_metric)
+        eval_results = job.evaluate(eval_dataset, metric=metric_func)
         del job
 
         return eval_results
@@ -208,14 +228,30 @@ class IncrementalLearning(ParadigmBase):
 
         operator_func = operator_map[operator]
 
-        if len(eval_results) != 2:
+        if not isinstance(eval_results, (list, tuple)) or len(eval_results) != 2:
             raise RuntimeError(f"two models of evaluation should have two results."
                             f" the eval results: {eval_results}")
 
-        metric_values = [0, 0]
+        metric_values = [None, None]
         for i, result in enumerate(eval_results):
-            metrics = result.get("metrics")
-            metric_values[i] = metrics.get(metric_name)
+            if isinstance(result, dict):
+                metrics = result.get("metrics")
+                if isinstance(metrics, dict) and metric_name in metrics:
+                    val = metrics.get(metric_name)
+                    if isinstance(val, (int, float)):
+                        metric_values[i] = float(val)
+                elif metric_name in result:
+                    val = result.get(metric_name)
+                    if isinstance(val, (int, float)):
+                        metric_values[i] = float(val)
+            elif isinstance(result, (int, float)):
+                metric_values[i] = float(result)
+
+        if metric_values[0] is None or metric_values[1] is None:
+            raise RuntimeError(
+                f"could not extract valid numeric metric '{metric_name}' "
+                f"from eval_results: {eval_results}"
+            )
 
         metric_delta = metric_values[0] - metric_values[1]
         return operator_func(metric_delta, threshold)
--- a/tests/test_incremental_learning.py
+++ b/tests/test_incremental_learning.py
@@ -1,0 +1,249 @@
+# Copyright 2022 The KubeEdge Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for IncrementalLearning paradigm."""
+
+# pylint: disable=protected-access,unused-argument,missing-class-docstring,missing-function-docstring,wrong-import-position,too-few-public-methods,unnecessary-lambda-assignment,unused-variable,line-too-long
+
+import os
+import sys
+import unittest
+from unittest.mock import patch
+import numpy as np
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+from core.testcasecontroller.algorithm.paradigm.incremental_learning.incremental_learning import (
+    IncrementalLearning,
+)
+
+
+class MockDataset:
+    def __init__(self, x=None, y=None):
+        self.x = x if x is not None else []
+        self.y = y if y is not None else []
+        self.train_url = "/tmp/train.txt"
+        self.test_url = "/tmp/test.txt"
+
+    def load_data(self, data_url, data_type=None):
+        return self
+
+
+class MockJob:
+    def __init__(self, inference_return=None, evaluate_return=None):
+        self.inference_return = inference_return
+        self.evaluate_return = evaluate_return
+        self.evaluate_metric_passed = None
+
+    def inference(self, data):
+        if self.inference_return is not None:
+            return self.inference_return
+        return [0], None, False
+
+    def evaluate(self, eval_dataset, metric=None):
+        self.evaluate_metric_passed = metric
+        return self.evaluate_return or [{"metrics": {"accuracy": 0.9}}, {"metrics": {"accuracy": 0.8}}]
+
+
+class TestIncrementalLearning(unittest.TestCase):
+    """Test suite for IncrementalLearning paradigm bug fixes."""
+
+    def setUp(self):
+        self.workspace = "/tmp/test_workspace"
+
+    def test_eval_passes_callable_metric(self):
+        """_eval must unpack get_metric_func and pass a callable metric to job.evaluate."""
+        paradigm = object.__new__(IncrementalLearning)
+        paradigm.workspace = self.workspace
+        paradigm.dataset = MockDataset()
+        mock_job = MockJob()
+        paradigm.build_paradigm_job = lambda _paradigm_type: mock_job
+        paradigm.model_eval_config = {
+            "model_metric": {
+                "name": "accuracy",
+                "url": None
+            }
+        }
+
+        mock_metric_fn = lambda y_true, y_pred: 0.95
+        with patch(
+            "core.testcasecontroller.algorithm.paradigm.incremental_learning.incremental_learning.get_metric_func",
+            return_value=("accuracy", mock_metric_fn)
+        ):
+            eval_results = paradigm._eval("new_model_url", "old_model_url", "/tmp/eval.txt")
+
+        # Verify that the metric passed to job.evaluate is callable, not a tuple
+        self.assertTrue(callable(mock_job.evaluate_metric_passed))
+        self.assertNotIsInstance(mock_job.evaluate_metric_passed, tuple)
+        self.assertEqual(mock_job.evaluate_metric_passed, mock_metric_fn)
+
+    def test_get_train_dataset_success(self):
+        """_get_train_dataset must write all hard example rows when all samples match uniquely."""
+        paradigm = object.__new__(IncrementalLearning)
+        paradigm.workspace = self.workspace
+        paradigm.dataset_output_dir = lambda: os.path.dirname(os.path.abspath(__file__))
+
+        hard_examples = [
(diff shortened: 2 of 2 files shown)


whozahm3d:
### Task 2 — Code Review & Novel Finding

**Recommendation:** **Major Revision**

#### Review Findings:
1. **Novel Finding — Type Contract Violation in `inference_results`:** This PR changes `inference_results` from `{}` (`dict`) to `[]` (`list`). Downstream callers in the orchestration loop expect string path key lookups (`inference_results[sample_path]`). Changing this return type to a list raises `TypeError: list indices must be integers or slices, not str` at runtime. The dictionary structure must be preserved using `isinstance(res, dict)` guards.
2. **Ambiguous Multi-Match Guard:** The check `if len(index[0]) > 0` handles zero matches but silently accepts duplicate matches (`> 1`), picking `index[0][0]` without warning and risking writing corrupted label data.
3. **Metric Unpacking:** Metric tuple unpacking `_, metric_func = get_metric_func(model_metric)` is correctly implemented.

bhuvan-somisetty:
Thanks for the detailed review @whozahm3d! Reconciled the patch in commit `b05b1e9`: `inference_results` preserves the `dict` interface with fallback mapping for non-dict returns (`inference_results[data] = res`), and `_get_train_dataset` applies strict `len(index[0]) == 1` label lookup with warnings on missing/ambiguous labels. All 5 unit tests pass cleanly.

whozahm3d:
Thanks @bhuvan-somisetty for the swift update and force-push (`a05b1e9`)! The fallback mapping for non-dict returns alongside strict `len(index[0]) == 1` label guards provides a resilient fix that resolves all core paradigm issues without breaking downstream contract callers. Looks fully ready for merge.

bhuvan-somisetty:
Thank you @whozahm3d! Appreciate the helpful review and collaboration throughout.

MukandKrishna (commented):
**Recommendation: major revision, using this as the consolidation base.** The updated PR correctly preserves the dictionary inference contract, uses strict unique-match detection, unpacks the metric callable, and adds useful tests. The remaining label policy still treats incomplete retraining data as success.

Executed against the current head:

```text
CASE=missing_after_valid
WARNING=Hard example missing.jpg not found or ambiguous in train label dataset. Ignoring.
OUTPUT='new-valid.jpg cat\n'

CASE=duplicate_identity
WARNING=Hard example duplicate.jpg not found or ambiguous in train label dataset. Ignoring.
OUTPUT=''
```

The test `test_get_train_dataset_unbound_local_and_missing_label` supplies two hard examples and explicitly asserts that the one-row result is correct. That prevents stale reuse, but it also codifies silent sample loss. An all-ambiguous batch produces an empty file and moves the failure downstream.

I recommend changing this test and implementation to validate all identities before opening or publishing the output and raise one structured missing/duplicate error. Then consolidate the useful stale-label work here and close the competing PRs rather than resolving their conflicting hunks during merge.

Executed against PR head `b05b1e9e81ae7ec949b43e23a1fafa6549422502`; metric and inference changes were inspected, but no full GPU/model run is claimed. Related reproduction context: [Task 1](https://github.com/kubeedge/ianvs/discussions/921#discussioncomment-18145877), the [Issue #842 follow-up](https://github.com/kubeedge/ianvs/issues/842#issuecomment-5408109234), and the corresponding [PR #803 review](https://github.com/kubeedge/ianvs/pull/803#pullrequestreview-5017121592).

bhuvan-somisetty:
Thanks for the thorough review and suggestion on label verification @MukandKrishna! Updated `_get_train_dataset()` to validate all hard example identities against `data_labels` upfront before writing the dataset file. If any sample is missing or ambiguous, it raises a structured `ValueError` with the list of invalid sample identities rather than silently skipping rows. Added tests covering both valid and missing/ambiguous cases — all 6 tests pass cleanly.

MukandKrishna:
Thanks for the update. I independently rechecked the current head, [`aa2a5b2`](https://github.com/kubeedge/ianvs/commit/aa2a5b22d31c22eef74629d04edc86d21b020dd2).

Confirmed:

- The focused `tests/test_incremental_learning.py` suite passes under `pytest`: **6 passed**.
- Valid hard-example identities produce all expected dataset rows with the correct labels.
- Missing identities raise a `ValueError` that identifies the offending sample.
- Duplicate/ambiguous identities are also rejected with a `ValueError`.
- Validation completes before the dataset file is opened, preventing partial or silently truncated training-index files.

This addresses my original correctness concern about silently skipping unresolved hard examples and breaking input/output cardinality.

One non-blocking lifecycle note: `tempfile.mkdtemp(...)` is currently called before identity validation, so a missing or ambiguous batch leaves an empty temporary directory behind. Creating the temporary directory after successful validation, or cleaning it on failure, would avoid that filesystem artifact. This does not block the correctness fix above.

Verification boundary: I ran the six focused tests using `pytest` with lightweight import-only stubs for the unavailable Sedna runtime classes; the tests themselves mock the runtime collaborators. I did not run the full repository suite or an end-to-end Sedna training workflow.

bhuvan-somisetty:
Thanks for the suggestion on temporary directory lifecycle @MukandKrishna! Updated `_get_train_dataset()` to defer calling `tempfile.mkdtemp()` until after identity validation succeeds, ensuring no empty temporary directories are left behind when validation fails. All 6 tests continue to pass cleanly.

MukandKrishna:
Thanks for addressing the lifecycle follow-up. I independently rechecked the current head, [`23e480f`](https://github.com/kubeedge/ianvs/commit/23e480fc93c9e5de1b162b57313cc1006e9438c7).

Confirmed:

- `tempfile.mkdtemp()` is now called only after the complete identity-validation pass succeeds.
- The focused `tests/test_incremental_learning.py` module still passes under `pytest`: **6 passed**.
- I separately ran the missing/ambiguous-identity test: **1 passed**.
- No temporary directory was created by the validation-failure cases.

This resolves the non-blocking temporary-directory lifecycle note from my previous verification.

Verification boundary: the focused controller tests were executed with lightweight import-only stubs for unavailable Sedna runtime classes that the tests mock and do not exercise. I did not run the full repository suite or an end-to-end Sedna training workflow.

bhuvan-somisetty:
Thank you @MukandKrishna for the prompt re-check and verification of the directory lifecycle fix! Appreciate your helpful reviews and guidance throughout.

greenfield01:
The four correctness fixes and the iterative label-validation improvements across the review cycle look well handled. The metric callable unpacking, dict interface preservation, and deferred tempdir creation all address real issues cleanly.

One thing worth flagging that isn't covered in the existing thread: this PR adds `tests/test_incremental_learning.py` to the repo root `tests/` directory, which puts it in the same location as `tests/test_core_utils.py` from PR #813 and `tests/test_multiedge_inference.py` from PR #872. PR #815 adds a CI step that automatically runs pytest against everything in that directory. None of these four PRs reference each other.

The incremental learning tests stub Sedna at import time, which works cleanly in an isolated run. The question is whether that stub setup is robust enough when pytest discovers and runs all tests in `tests/` together in one session. If #815 merges before this PR, the CI step will pick up these tests automatically, and a stub isolation issue could cause test interference that wouldn't surface in a standalone run. Worth confirming the Sedna stubs are session-scoped rather than module-scoped before this lands, so the combined test suite stays predictable.

bhuvan-somisetty:
Thanks @greenfield01 for the insightful review and for checking test suite isolation!

Confirmed that `tests/test_incremental_learning.py` does not mutate global module state or stub Sedna at the `sys.modules` level. It uses isolated instance-level collaborators (`MockDataset`, `MockJob`) and context-managed `unittest.mock.patch` calls that clean up per-test. It runs cleanly alongside other tests in multi-test discovery without cross-test side effects. Also cleaned up linting and line lengths to a clean 10.00/10 pylint score — all 6 tests continue to pass cleanly.

pratikpanda2006:
**Review:**

- Resolves the root cause in #842: `_eval()` now unpacks `_, metric_func = get_metric_func(...)` before passing to `job.evaluate()`. Confirmed correct against Sedna's documented interface contract.
- Scope beyond example-local: this touches `_eval`, `_get_train_dataset`, `_inference`, and `_trigger_model_update` in the shared paradigm controller — affects every incremental-learning example, not just one.
- Edge case not covered: in `_trigger_model_update()`, if `eval_results` metrics come back with a `None` value (e.g. a metric key missing from the eval dict), `metric_delta = metric_values[0] - metric_values[1]` will still raise `TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'`. Suggest adding an `isinstance(v, (int, float))` guard before the delta computation.
- Recommendation: minor revision (add the numeric guard above); otherwise sound and correctly scoped at the core layer.

MukandKrishna:
Follow-up on the newly raised metric-value boundary: I rechecked current head [`e363cbe`](https://github.com/kubeedge/ianvs/commit/e363cbe31fc4d21ab905e1678b48afd5e493071a).

The specific `NoneType - float` path described above is already guarded. When a nested `metrics` dictionary does not contain the configured metric, `metrics.get(metric_name)` returns `None`, and the existing `None` check raises `RuntimeError` before subtraction. The existing `test_trigger_model_update_missing_metrics_raises` case passes.

However, the check exposed an adjacent unresolved shape: `metric_values` is initialized as `[0, 0]`. If a result is an empty dictionary, or a direct dictionary that lacks both `metrics` and the configured metric key, no assignment occurs and the zero default survives. I executed the exact current method and observed:

```text
nested missing metric -> RuntimeError
empty dictionaries    -> False
direct missing metric -> False
```

The latter two cases silently treat missing metrics as numeric zero rather than reporting extraction failure.

A stronger repair would initialize `metric_values` as `[None, None]`, assign only successfully extracted values, and then validate that both values are real numeric metrics before conversion and subtraction. Focused regression coverage should include:

- missing key inside a nested `metrics` dictionary;
- empty result dictionaries;
- direct dictionaries missing the configured metric;
- nonnumeric metric values.

Verification boundary: I executed these three extraction shapes against current head `e363cbe`. I also ran all six existing `TestIncrementalLearning` unit tests at this head; all six passed. Sedna runtime imports were supplied only through lightweight import stubs because these controller tests mock those collaborators. I did not run the full repository suite or an end-to-end training workflow.

bhuvan-somisetty:
Thanks @MukandKrishna for catching that edge case.

I have updated `_trigger_model_update` to initialize `metric_values` as `[None, None]` and explicitly require successful numeric extraction (`int` or `float`) for both evaluation results. If a result is an empty dictionary, lacks the configured metric key, or contains a non-numeric metric value, it raises `RuntimeError` rather than defaulting to zero.

Added unit test coverage for all these boundary shapes (missing nested keys, empty dictionaries, direct dictionaries without metric, and non-numeric values) in `tests/test_incremental_learning.py`. All tests pass and the code rates 10.00/10 with pylint. The commit has been updated and pushed.

Issue #846, opened 2026-08-20
[Example Audit] examples/aoa/single_task_bench/TForest is permanently non-reproducible — recommend Quarantined classification


Arijit429:
/kind bug
/kind documentation

## TL;DR

examples/aoa/single_task_bench/TForest has no benchmarkingjob.yaml, testenv.yaml, or algorithm.yaml anywhere in its tree, so it was never integrated with the Ianvs benchmarking framework and cannot be run via `ianvs -f`. Its standalone entry point also fails immediately, because the one code path left enabled loads a .joblib file that is not present anywhere in the repository. Tracing why leads to a data-generation function that reads a dataset the example's own README says "is not authorized to be disclosed" — meaning the example is not merely broken, it is permanently non-reproducible by anyone, including its original author, as committed. Alongside this, the directory carries a second, fully separate, PyTorch-based abandoned implementation (other_attempts/) and ~7MB of committed binary/log artifacts, none of which are reachable from any documented or runnable path.

This is a different failure class than the "missing conditional import" bugs currently being fixed across other examples — it's a repo-hygiene and classification problem the CI validation framework proposed in phase-3-2026-term-2/proposal.md needs to handle: an inventory entry that should be classified **Quarantined**, not Broken, because no amount of code fixing restores it.

**Note on method:** I could not execute this end-to-end — the dataset is proprietary and not present in the repo. Every claim below is verified statically: find/git ls-files confirm which files exist and are tracked, and the failure chain is traced by reading the actual call graph in task_main.py/tools.py.

---

## Layer 1 — Never integrated into the Ianvs framework

```bash
$ find examples/aoa -iname "*.yaml"
(no output)
```

There is no benchmarkingjob.yaml, testenv.yaml, or algorithm.yaml anywhere under examples/aoa/. Every other Ianvs example is invoked via `ianvs -f examples/<name>/benchmarkingjob.yaml`. TForest has none — task_main.py is a standalone script, not an Ianvs BaseModel implementation. It has never been runnable as an Ianvs benchmark.

---

## Layer 2 — The standalone script fails on its own first line of execution

The only enabled call in task_main.py:

```python
# task_main.py:395-401
if __name__ == '__main__':
    test_main()
```

test_main() immediately does:

```python
# task_main.py:362-364
def test_main():
    single_tree = joblib.load('single_tree.joblib')
```

single_tree.joblib does not exist anywhere in the working tree:

```bash
[...]

PR #847, opened 2026-08-20
docs: add debugging playbook for common Ianvs example failures

Arijit429:
/kind documentation

Adds a developer-focused debugging playbook to `docs/guides/` covering the most common Ianvs example failure patterns. Each entry is grounded in real bugs found in this codebase — not hypothetical failures — with exact file:line citations, symptom descriptions, root cause analysis, and fix guidance.

Failure patterns covered:

1. Silent save_mode misconfiguration — `core/storymanager/rank/rank.py` (fixed in #[your save_mode PR])
2. Hardcoded False feature flags with missing imports — `pose-estimation-llio/utils.py` pattern (fixed in #635)
3. Self-referential pip dependencies — `ianvs` listed as its own PyPI dependency in two examples
4. Permanently non-reproducible examples — `aoa/TForest` pattern (proprietary dataset, no Ianvs YAML config, filed as #846)
5. `_check_fields` using `and` instead of `or` — same bug shape as the save_mode issue, present in `core/testcasecontroller/algorithm/algorithm.py`

This directly supports Phase IV's documentation modernization deliverable and the CI validation framework proposed in PR #541 — new contributors hitting these failures now have a single reference doc rather than having to trace through the codebase from scratch.

Related to #846

### DIFF
--- a/docs/guides/debugging-common-example-failures.md
+++ b/docs/guides/debugging-common-example-failures.md
@@ -1,0 +1,258 @@
+# Debugging Playbook: Common Ianvs Example Failure Modes
+
+This guide catalogs failure patterns that have actually been found and fixed in the Ianvs codebase and its examples. Each entry cites the real file where the pattern occurred, so you can see the exact bug rather than a hypothetical one.
+
+If your benchmark run is failing, silently producing no output, or behaving unexpectedly, check whether it matches one of the patterns below before assuming the problem is in your own configuration.
+
+---
+
+## 1. Silent `save_mode` misconfiguration
+
+**Where this happened:** `core/storymanager/rank/rank.py`
+
+### Symptom
+
+Your benchmark job runs to completion with no errors, but the rank/leaderboard output files you expect (`all_rank.csv`, `selected_rank.csv`, or the rank pictures) are missing or incomplete — and nothing in the log tells you why.
+
+### Root cause
+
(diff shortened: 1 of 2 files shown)


Shikhar-Kesharwani:
### Review — supplementary documentation finding

The debugging playbook is useful and the failure patterns are grounded in real repository defects. I found one concrete documentation-integrity issue that should be corrected before merge.

The first entry currently says:

`fixed in #[your save_mode PR]`

This is still an unresolved placeholder rather than an actual navigable PR reference. Because this document is intended to be a contributor-facing debugging playbook with exact remediation references, leaving the placeholder makes the first troubleshooting entry incomplete and prevents a contributor from directly following the documented fix.

Please replace the placeholder with the actual fixing PR reference (PR #839), or with the corresponding issue/commit reference if that is the preferred permanent reference.

I would also recommend checking the remaining playbook entries for unresolved placeholders or broken cross-references before merging, since the usefulness of this document depends on contributors being able to follow its remediation references directly.

ankita7Patil:

I reviewed PR #847 as the documentation-focused PR in the related-work set.

The debugging playbook improves contributor guidance for diagnosing common Ianvs example failures, so it supports the root problem by making failures easier to understand and reproduce. However, documentation alone does not enforce the inventory schema, lifecycle/status values, or benchmark coverage identified in Task 1.

I therefore recommend keeping this documentation improvement, while pairing it with the shared inventory/validator changes proposed in the restoration plan.

PR #848, opened 2026-08-20
fix(paradigm): replace UnboundLocalError with clear ValueError for unsupported multi-inference mode

Arijit429:
/kind bug

`LifelongLearning.run()` raises `UnboundLocalError: local variable 'test_res' referenced before assignment` when `mode == 'multi-inference'`.

The branch structure at lines 93-295 is:
- `if mode == 'no-inference'` — assigns test_res
- `elif mode == 'hard-example-mining'` — assigns test_res
- `elif mode != 'multi-inference'` — assigns test_res

When `mode == 'multi-inference'` exactly, none of the three branches match — the third condition `elif mode != 'multi-inference'` is false — so `test_res` is never assigned and line 299 raises `UnboundLocalError`.

This was silenced with `# pylint: disable=E0606` and `# pylint: disable=W0012` rather than fixed — identical suppression pair to `incremental_learning.py:150-151` which had the same class of bug.

Fix:
- Add `else: raise ValueError(f"unsupported mode: {mode}")` before the return statement
- Remove the now-unneeded E0606 and W0012 pylint suppressions

Related to the same bug class fixed in incremental_learning.py

### DIFF
--- a/core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py
+++ b/core/testcasecontroller/algorithm/paradigm/lifelong_learning/lifelong_learning.py
@@ -71,8 +71,6 @@ class LifelongLearning(ParadigmBase):
         # pylint:disable=duplicate-code
         # pylint: disable=R0912
         # pylint: disable=R0915
-        # pylint: disable=W0012
-        # pylint: disable=E0606
         """
         run the test flow of incremental learning paradigm.
 
@@ -295,6 +293,8 @@ class LifelongLearning(ParadigmBase):
             test_res, unseen_task_train_samples = self._inference(self.edge_task_index,
                                                               self.dataset.test_url,
                                                               "test")
+        else:
+            raise ValueError(f"unsupported mode: {mode}")
 
         return test_res, self.system_metric_info
 
--- a/docs/guides/debugging-common-example-failures.md
+++ b/docs/guides/debugging-common-example-failures.md
@@ -1,0 +1,258 @@
+# Debugging Playbook: Common Ianvs Example Failure Modes
+
+This guide catalogs failure patterns that have actually been found and fixed in the Ianvs codebase and its examples. Each entry cites the real file where the pattern occurred, so you can see the exact bug rather than a hypothetical one.
+
+If your benchmark run is failing, silently producing no output, or behaving unexpectedly, check whether it matches one of the patterns below before assuming the problem is in your own configuration.
+
+---
+
+## 1. Silent `save_mode` misconfiguration
+
+**Where this happened:** `core/storymanager/rank/rank.py`
+
+### Symptom
+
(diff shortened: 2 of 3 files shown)


akshhkaushik (commented):
### Request changes — clearer failure is not restoration of a live configured mode

The PR changes `multi-inference` from an accidental unbound-local crash to:

```text
ValueError: unsupported mode: multi-inference
```

I got that result by executing the real patched `LifelongLearning.run()` control path. BDD curb detection currently sets `mode: "multi-inference"`. This PR therefore makes a tracked example fail earlier and more clearly, but does not restore it.

The branch also contains two unrelated commits: LLIO optional-import changes and a new 258-line debugging guide. Those are not needed for the mode fix and widen review/rollback scope.

Recommendation: split the unrelated commits. Then either implement `multi-inference` with a BDD integration test, or remove/migrate every live configuration and document the compatibility break. If the intended PR is only diagnostic improvement, retitle it accordingly rather than claiming the mode is resolved.

Verification boundary: executed the mode dispatch and scanned tracked YAML for mode users; did not run BDD training/inference.

Submission context: [submission discussion](https://github.com/kubeedge/ianvs/discussions/924) ([Bonus cross-target finding](https://github.com/kubeedge/ianvs/discussions/924#discussioncomment-18147196)).

Arijit429:
> ### Request changes — clearer failure is not restoration of a live configured mode
> The PR changes `multi-inference` from an accidental unbound-local crash to:
> 
> ```
> ValueError: unsupported mode: multi-inference
> ```
> 
> I got that result by executing the real patched `LifelongLearning.run()` control path. BDD curb detection currently sets `mode: "multi-inference"`. This PR therefore makes a tracked example fail earlier and more clearly, but does not restore it.
> 
> The branch also contains two unrelated commits: LLIO optional-import changes and a new 258-line debugging guide. Those are not needed for the mode fix and widen review/rollback scope.
> 
> Recommendation: split the unrelated commits. Then either implement `multi-inference` with a BDD integration test, or remove/migrate every live configuration and document the compatibility break. If the intended PR is only diagnostic improvement, retitle it accordingly rather than claiming the mode is resolved.
> 
> Verification boundary: executed the mode dispatch and scanned tracked YAML for mode users; did not run BDD training/inference.
> 
> Submission context: [submission discussion](https://github.com/kubeedge/ianvs/discussions/924) ([Bonus cross-target finding](https://github.com/kubeedge/ianvs/discussions/924#discussioncomment-18147196)).

Fair points — acknowledged on all three:

1. The PR title overclaims. The intent was diagnostic improvement (clear ValueError instead of opaque UnboundLocalError), not full mode restoration. Will retitle to reflect that.

2. The unrelated commits (LLIO optional imports, debugging guide) were on a shared branch — should have been isolated. Will split into separate PRs.

3. Implementing `multi-inference` is out of scope for this fix. Will add a note to the ValueError message pointing to BDD curb detection as a known affected configuration, so contributors understand the compatibility break explicitly.

PR #596, opened 2026-07-30
fix: correct field type validation logic in core modules

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Fixes a bug in configuration field validation (`_check_fields()`) across core components (`Rank`, `BenchmarkingJob`, `Algorithm`, and `Module`). 

When instantiating configuration parser objects, invalid non-empty data types (such as passing a `str` for `sort_by` or an `int` for `name`) were bypassing validation checks without raising a `ValueError`.

In `_check_fields()`, validation conditions were written using boolean `and` instead of `or` between presence and type checks:

```python
if not self.sort_by and not isinstance(self.sort_by, list):
    raise ValueError(...)
```

When an invalid non-empty data type was provided (e.g., `sort_by = "accuracy"`):
1. `not self.sort_by` evaluated to `False` because non-empty strings are truthy in Python.
2. `not isinstance(self.sort_by, list)` evaluated to `True`.
3. `False and True` resolved to `False`, bypassing the `ValueError` check.

As a result, invalid types silently passed initialization and caused obscure downstream runtime errors (such as `AttributeError` or `TypeError`) during pipeline execution.

Additionally, in `Rank._check_fields()`, `save_mode` was checked against `list` instead of `str` (`self.save_mode` is a string such as `"selected_and_all"` or `"selected_only"`).

Changes Introduced:
1. **Logical Operator Fix in Core Components**:
   Replaced boolean `and` with `or` in `_check_fields()` across:
   - `core/storymanager/rank/rank.py`
   - `core/cmd/obj/benchmarkingjob.py`
   - `core/testcasecontroller/algorithm/algorithm.py`
   - `core/testcasecontroller/algorithm/module/module.py`
2. **Save Mode Type Validation**:
   Corrected `Rank._check_fields()` to check `isinstance(self.save_mode, str)` with an updated error message.
3. **Unit Tests**:
   Added `tests/test_check_fields.py` to test field validation against invalid data types for `Rank`, `BenchmarkingJob`, `Algorithm`, and `Module`.

**Which issue(s) this PR fixes**:

Fixes #595

**Special notes for your reviewer**:

NONE

**Does this PR introduce a user-facing change?**:

```release-note
Fix configuration field validation in core components to properly raise ValueError on invalid input types.
```

PR #849, opened 2026-08-21
fix(core): correct and/isinstance inversion in Algorithm._check_fields

Arijit429:
/kind bug

`Algorithm._check_fields()` in `core/testcasecontroller/algorithm/algorithm.py` has the same `and`/`isinstance` inversion bug as `rank.py`. Two fields are affected:

`name` (line 132) — before:
`if not self.name and not isinstance(self.name, str):`

`paradigm_type` (line 136) — identical pattern.

The `and` short-circuits to False for any non-empty value regardless of type — so passing an integer or list as `name` passes validation silently with no error. Changed to `or` in both cases so invalid types are properly rejected at config-parse time.

**Scope:** This PR is intentionally limited to `algorithm.py` only (2 instances: `name` and `paradigm_type`). PR #336 covers all 4 files but has a blocker on `rank.py`'s `save_mode` check — this narrower scope avoids that regression entirely.

**Related PRs:** #336 (broader scope, has save_mode blocker), #590, #596

Related to #846

### DIFF
--- a/core/testcasecontroller/algorithm/algorithm.py
+++ b/core/testcasecontroller/algorithm/algorithm.py
@@ -130,10 +130,10 @@ class Algorithm:
         return None
 
     def _check_fields(self):
-        if not self.name and not isinstance(self.name, str):
+        if not isinstance(self.name, str) or not self.name:
             raise ValueError(f"algorithm name({self.name}) must be provided and be string type.")
 
-        if not self.paradigm_type and not isinstance(self.paradigm_type, str):
+        if not isinstance(self.paradigm_type, str) or not self.paradigm_type:
             raise ValueError(
                 f"algorithm paradigm({self.paradigm_type}) must be provided and be string type.")
 
--- a/examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py
+++ b/examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py
@@ -7,8 +7,20 @@ from functools import wraps
 import time
 from core.common.log import LOGGER
 
-MATPLOTLIB_AVAILABLE = False
-OPEN3D_AVAILABLE = False
+try:
+    import matplotlib
+    import matplotlib.pyplot as plt
+    from matplotlib.patches import Ellipse
+    from matplotlib.collections import PatchCollection
+    MATPLOTLIB_AVAILABLE = True
+except ImportError:
+    MATPLOTLIB_AVAILABLE = False
+
+try:
+    import open3d as o3d
+    OPEN3D_AVAILABLE = True
+except ImportError:
+    OPEN3D_AVAILABLE = False
 
 
 def timeit(func):


akshita317:
The `Algorithm._check_fields` change is correct, and the ordering is better than the plain `or` form used in the other PRs touching this idiom: `not isinstance(self.name, str) or not self.name` establishes the type before applying truthiness, which is the safer evaluation order. Worth keeping that ordering if this gets extended.

One suggestion and two heads-ups.

**Suggestion — worth noting the scope in the description.** The idiom this fixes occurs 10 times across four files:

```
core/cmd/obj/benchmarkingjob.py:54,61
core/storymanager/rank/rank.py:62,67,73,88
core/testcasecontroller/algorithm/algorithm.py:133,136      <-- fixed here
core/testcasecontroller/algorithm/module/module.py:61,69
```

This PR covers the two in `algorithm.py`. That is a perfectly reasonable scope, but since #336, #590 and #596 each fix overlapping subsets with a *different* idiom, saying explicitly in the description that this one is scoped to `algorithm.py` would help whoever sequences them avoid conflicts.

**Heads-up 1 — one caution if the scope is ever widened.** `rank.py:88` checks `isinstance(self.save_mode, list)` while `save_mode` is declared `str` at `rank.py:46`. A mechanical `and` → `or` at that site raises for every example config, so that one needs the declared type corrected, not just the operator. (Already raised on #336 by @coldstarted and @suhaan-24 — flagging it only so it does not get repeated here.)

**Heads-up 2 — the `utils.py` change may collide.** The diff also adds the conditional `matplotlib` / `open3d` imports to `examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py`. That same block is in #661 and #636, so whichever merges first will conflict with the other two. Not a problem with the change itself — the fix is correct in all three — just something a maintainer will want to resolve once rather than three times. Mentioning it in the description would make the overlap visible to reviewers who are only reading the title.

yadavchiragg:
Noting for the record: this PR fixes the same bug as PR #336 (the and/isinstance inversion in _check_fields — same four core modules, same root cause). Worth flagging to maintainers so only one of the two gets merged, since merging both would create duplicate/conflicting commits to the same functions.

yadavchiragg:
Review Depth Level: Observation
What it does: Fixes the identical and→isinstance inversion bug in _check_fields() that PR #336 already fixes — same four core modules, same root cause.
Finding: Two independently-authored PRs (#336, #849) fix the same Core bug. This is a live instance of exactly the "duplicated changes" risk your Task 3 rubric asks about — not hypothetical, actually happening. Maintainers should merge only one; merging both risks conflicting commits to the same functions.

yadavchiragg:
Thanks @akshita317 for the fuller picture — worth updating my note above: this isn't just #849 vs #336, there are four PRs touching overlapping subsets of the same and/isinstance pattern across the four affected files (#336, #590, #596, #849). @coldstarted and @suhaan-24 had already flagged part of this on #336 before my comment; the additional value here is simply having all four enumerated in one place so a maintainer can sequence them rather than merge in an arbitrary order.

Roshan5105labs:
## LFX Bonus Review — PR #849

### Review Depth
**Analysis**

### Scope

I reviewed the target-specific change in PR #849 after separating it from the shared ancestor commit `bf79838`.

The PR changes the validation in `Algorithm._check_fields()` from:

```python
```

to:

```python
if not isinstance(self.name, str) or not self.name:
```

and applies the same correction to `paradigm_type`.

### Technical Finding

The previous condition only rejected a value when it was **both falsy and non-string**.

That leaves two validation gaps:

1. an empty string (`""`) is falsy but still a string, so the previous condition does not reject it;
2. a truthy non-string value such as `123` is not a string but is truthy, so the previous condition also does not reject it.

The new `or` condition correctly rejects either an invalid type or a missing/empty value.

### Independent Verification

I reproduced the exact guard behavior with representative values:

```text
valid string       value='valid-name'   old_rejects=False new_rejects=False
empty string       value=''             old_rejects=False new_rejects=True
truthy non-string  value=123            old_rejects=False new_rejects=True
None               value=None           old_rejects=True  new_rejects=True
```

This confirms that the patch closes both the empty-string case and the truthy non-string case while preserving valid strings.

image


### Regression-Coverage Observation

After isolating the target-specific diff (`bf79838..bonus-849`), I found that PR #849 does not add a dedicated regression test file for `_check_fields()`.

Because the bug is a small boolean validation error with clearly enumerable edge cases, I recommend adding a focused parameterized test covering at least:

- valid non-empty string;
- empty string;
- truthy non-string value;
- `None`.

The same cases should be exercised for both `name` and `paradigm_type`.

### Recommendation

**Classification: Minor revision — logic fix is correct, but focused regression coverage should be added.**

The implementation corrects the validation semantics with a minimal change. A small regression test would protect against reintroducing the same boolean-logic mistake and would document the intended contract of `_check_fields()`.

PR #850, opened 2026-08-21
fix(core): remove shell=True subprocess injection in simulation_system_admin.py

Arijit429:
/kind bug

`simulation_system_admin.py` built shell commands via string interpolation from config-controlled values (`cluster_name`, `kubeedge_version`, `sedna_version`, `cloud_number`, `edge_number`) and executed them with `shell=True`. A malicious or malformed config value could inject arbitrary shell syntax.

Additionally, `check_docker`/`check_kind` used `subprocess.run(..., check=True)` and then branched on `.returncode` — but `check=True` raises `CalledProcessError` on nonzero exit before the returncode branch is ever reached, making the "not installed, try to install" fallback permanently unreachable.

**Fixes**:

1. All subprocess calls converted to list-form args with `shell=False` — config values passed via `env` dict so they can never become shell syntax
2. `curl | bash` pipe converted to `subprocess.Popen`/`subprocess.run` list-form chained calls
3. `check_docker`/`check_kind` changed to `check=False` so the fallback install path is actually reachable
4. `get_host_free_memory_size`/`get_host_number_of_cpus` left untouched — no attacker-controlled input, out of scope

Related to #846

### DIFF
--- a/core/testcasecontroller/simulation_system_admin/simulation_system_admin.py
+++ b/core/testcasecontroller/simulation_system_admin/simulation_system_admin.py
@@ -33,17 +33,15 @@ def check_host_docker():
     if check_docker.returncode != 0:
         # trying to install docker
         LOGGER.info("trying to install docker")
-        curl_proc = subprocess.Popen(
-            ["curl", "-fsSL", "https://get.docker.com"], stdout=subprocess.PIPE)
-        try:
-            subprocess.run(
-                ["bash", "-s", "docker", "--mirror", "Aliyun"],
-                stdin=curl_proc.stdout, check=True)
-        except subprocess.CalledProcessError as err:
-            raise RuntimeError(f"install docker failed, error: {err}.") from err
-        finally:
-            curl_proc.stdout.close()
-            curl_proc.wait()
+        with subprocess.Popen(
+                ["curl", "-fsSL", "https://get.docker.com"],
+                stdout=subprocess.PIPE) as curl_proc:
+            try:
+                subprocess.run(
+                    ["bash", "-s", "docker", "--mirror", "Aliyun"],
+                    stdin=curl_proc.stdout, check=True)
+            except subprocess.CalledProcessError as err:
+                raise RuntimeError(f"install docker failed, error: {err}.") from err
 
         LOGGER.info("successfully installed docker")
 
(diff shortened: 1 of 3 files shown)


Shikhar-Kesharwani:
### Review — supplementary security finding

The removal of `shell=True` from the config-driven subprocesses is correct, and I agree that the two remaining host-resource helpers are outside the direct injection path because they do not interpolate configuration values.

My additional concern is the security boundary around the new installation path. The Docker/Kind fallback still executes a remotely fetched installation script (`curl` output is passed directly to `bash`). Removing shell interpolation prevents config-controlled command injection, but it does not provide integrity or provenance protection for the script being executed.

For a repository-level framework that can automatically install host tooling, this creates a separate supply-chain trust boundary: if the remote installer is modified or compromised, the fallback can still execute arbitrary code with the privileges of the Ianvs process.

I would recommend pinning the installer to a reviewed immutable version/digest, verifying a published checksum or signature before execution, or making the remote installation step explicitly opt-in rather than automatically executing an unpinned remote script.

This is separate from the `shell=True` injection fixed by this PR: the current patch closes the command-construction vulnerability but leaves the remote-script trust boundary unchanged.

Laiiiiba:
### Review  Minor revision

I reviewed the security-related changes with particular attention to the affected command-execution paths.

The direction of the change is sound: removing `shell=True` from the affected execution path reduces the risk associated with shell interpretation and is the appropriate hardening approach.

My remaining concern is verification of failure behavior. I would recommend adding a test covering a non-zero exit from the external `curl` / `bash` operation and asserting that the failure is handled as intended rather than being silently treated as success.

This would make the security change more robust by verifying not only that shell invocation has been removed, but also that command failure is propagated or handled correctly.

 Minor revision.

The underlying security direction looks correct; the main request is targeted failure-path test coverage.

31groot:
### Request changes :  failed `curl` downloads are not propagated by the new installer pipeline

**Review depth: Analysis**

The PR removes the previous `shell=True` pipeline and correctly separates the `curl` and `bash` processes. However, the new implementation does not propagate failures from the `curl` process.

In `build_simulation_enviroment()`, the installer starts `curl` with `subprocess.Popen()` and then waits for the downstream `bash` process:

```python
curl_proc = subprocess.Popen(
    ["curl", "..."],
    stdout=subprocess.PIPE,
)

try:
    subprocess.run(
        ["bash", "-"],
        stdin=curl_proc.stdout,
        env=install_env,
        check=True,
    )
finally:
    curl_proc.stdout.close()
    curl_proc.wait()
```

The return code of `curl_proc` is never checked.

This allows the following state:

```text
curl returncode = 22
bash returncode  = 0
```

while the function can still report a successful installation.

I reproduced this control-flow state against the submitted PR implementation:

```text
BUILD COMPLETED WITHOUT RAISING
curl returncode: 22
bash returncode : 0
CURL FAILURE WAS NOT PROPAGATED
```

The logging/formatter error observed during the harness run came from the minimal `colorlog` test stub used to import Core in the verification environment; it did not affect the subprocess result above.

This is a correctness issue separate from the PR's `shell=True` removal. The refactor changes how the pipeline is executed, but it should still preserve the requirement that a failed download causes the installation to fail.

**Recommendation:** check `curl_proc.returncode` after the downstream process exits and raise if it is non-zero. The installation should only be reported as successful when both `curl` and `bash` succeed.

Please add regression coverage for:

```text
curl succeeds + bash succeeds -> success
curl fails    + bash succeeds -> failure
curl succeeds + bash fails    -> failure
```

**Verification boundary:** I reproduced the `curl != 0` / `bash == 0` condition against the PR's subprocess structure using an isolated mocked harness. I did not perform a real Docker installation or create a real simulation cluster.

PR #851, opened 2026-08-21
ci: add GitHub Actions workflow for example lint and requirements validation

Arijit429:
/kind feature

Adds `.github/workflows/example-lint.yml` — a CI workflow that runs on every PR touching `examples/` and validates:

1. **Pylint across example Python files** — matrix tested on Python 3.9, 3.10, 3.11
2. **Requirements check** — verifies every example directory with a `benchmarkingjob.yaml` also has a `requirements.txt`

Follows the same conventions as the existing `.github/workflows/main.yaml` — pinned pip, `resources/third_party` wheel install, root `requirements.txt`, and the `--max-positional-arguments=10` Python 3.9 workaround from #157.

**Current status — advisory only (continue-on-error: true)**:

Both the pylint and requirements-check steps are currently non-blocking. From the examples audit: only 13/29 example directories have a `requirements.txt`, and the example code hasn't been pylint-cleaned yet. A hard-blocking gate on the full tree would block PRs for pre-existing violations unrelated to the change being reviewed.

The workflow is designed to become blocking incrementally as:
- Example requirements.txt coverage improves (tracked via the dependency manifest fixes)
- Example pylint violations are cleared (tracked via the example restoration PRs)

This approach aligns with the tiered validation strategy in the Phase IV CI proposal (PR #541).

Related to #846

### DIFF
--- a/.github/workflows/example-lint.yml
+++ b/.github/workflows/example-lint.yml
@@ -1,0 +1,59 @@
+name: Example Lint
+
+on:
+  pull_request:
+    paths:
+      - 'examples/**'
+
+jobs:
+  example-pylint:
+    runs-on: ubuntu-22.04
+    name: pylint
+    strategy:
+      matrix:
+        python-version: [ "3.9", "3.10", "3.11" ]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install libgl1-mesa-glx -y
+          python -m pip install pip==24.0
+          python -m pip install pylint
+          python -m pip install ${{github.workspace}}/resources/third_party/*
+          python -m pip install -r ${{github.workspace}}/requirements.txt
+      - name: Analysing example code with pylint
+        # `--max-positional-arguments=10` is set for Python 3.9 to avoid `R0917: too-many-positional-arguments`.
+        # See details at https://github.com/kubeedge/ianvs/issues/157
+        run: |
+          if [ "${{ matrix.python-version }}" = "3.9" ]; then
+            pylint --max-positional-arguments=10 '${{github.workspace}}/examples'
+          else
+            pylint '${{github.workspace}}/examples'
+          fi
+
+  requirements-check:
+    runs-on: ubuntu-22.04
+    name: requirements.txt presence check
+    steps:
+      - uses: actions/checkout@v3
+      - name: Check each example with benchmarkingjob.yaml has requirements.txt
+        run: |
+          missing=0
+          while IFS= read -r -d '' job_file; do
+            example_dir=$(dirname "$job_file")
+            if [ ! -f "$example_dir/requirements.txt" ]; then
+              echo "::error file=$job_file::Missing requirements.txt in $example_dir"
+              missing=1
+            fi
+          done < <(find "${{ github.workspace }}/examples" -name "benchmarkingjob.yaml" -print0)
+
+          if [ "$missing" -eq 1 ]; then
+            echo "One or more example directories with a benchmarkingjob.yaml are missing a requirements.txt."
+            exit 1
+          fi
+          echo "All example directories with benchmarkingjob.yaml have requirements.txt."
--- a/.github/workflows/example-lint.yml
+++ b/.github/workflows/example-lint.yml
@@ -4,6 +4,8 @@ on:
   pull_request:
     paths:
       - 'examples/**'
+      - 'core/**'
+      - 'requirements.txt'
 
 jobs:
   example-pylint:
@@ -13,9 +15,9 @@ jobs:
       matrix:
         python-version: [ "3.9", "3.10", "3.11" ]
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v7
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v7
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
@@ -32,6 +34,7 @@ jobs:
         # continue-on-error: the examples/ backlog is not pylint-clean yet, so this check
         # is currently advisory only. Remove continue-on-error once examples reach a
         # baseline passing state, so this becomes a real PR-blocking gate.
+        # Tracked in kubeedge/ianvs#995.
         continue-on-error: true
         run: |
           if [ "${{ matrix.python-version }}" = "3.9" ]; then
@@ -44,12 +47,13 @@ jobs:
     runs-on: ubuntu-22.04
     name: requirements.txt presence check
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v7
       - name: Check each example with benchmarkingjob.yaml has requirements.txt
         # continue-on-error: most of the examples/ backlog is still missing a
         # requirements.txt, so this check is currently advisory only. Remove
         # continue-on-error once examples reach a baseline passing state, so
         # this becomes a real PR-blocking gate.
+        # Tracked in kubeedge/ianvs#995.
         continue-on-error: true
         run: |
           missing=0
--- a/.github/workflows/example-lint.yml
+++ b/.github/workflows/example-lint.yml
@@ -29,6 +29,10 @@ jobs:
       - name: Analysing example code with pylint
         # `--max-positional-arguments=10` is set for Python 3.9 to avoid `R0917: too-many-positional-arguments`.
         # See details at https://github.com/kubeedge/ianvs/issues/157
+        # continue-on-error: the examples/ backlog is not pylint-clean yet, so this check
+        # is currently advisory only. Remove continue-on-error once examples reach a
+        # baseline passing state, so this becomes a real PR-blocking gate.
+        continue-on-error: true
         run: |
           if [ "${{ matrix.python-version }}" = "3.9" ]; then
             pylint --max-positional-arguments=10 '${{github.workspace}}/examples'
@@ -42,6 +46,11 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Check each example with benchmarkingjob.yaml has requirements.txt
+        # continue-on-error: most of the examples/ backlog is still missing a
+        # requirements.txt, so this check is currently advisory only. Remove
+        # continue-on-error once examples reach a baseline passing state, so
+        # this becomes a real PR-blocking gate.
+        continue-on-error: true
         run: |
           missing=0
           while IFS= read -r -d '' job_file; do
--- a/examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py
+++ b/examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py
@@ -7,8 +7,20 @@ from functools import wraps
 import time
 from core.common.log import LOGGER
 
-MATPLOTLIB_AVAILABLE = False
-OPEN3D_AVAILABLE = False
+try:
+    import matplotlib
+    import matplotlib.pyplot as plt
+    from matplotlib.patches import Ellipse
+    from matplotlib.collections import PatchCollection
+    MATPLOTLIB_AVAILABLE = True
+except ImportError:
+    MATPLOTLIB_AVAILABLE = False
+
+try:
+    import open3d as o3d
+    OPEN3D_AVAILABLE = True
+except ImportError:
+    OPEN3D_AVAILABLE = False
 
 
 def timeit(func):


Arijit429:
I see ken6078's comprehensive validation framework (the merged PR) covers much of this ground. Happy to close #851 if maintainers prefer to build on that framework instead, or to adapt this workflow to complement it.

kavaljeetsingh-dev:
Reviewed the `requirements-check` job in this workflow. It discovers validation targets via:
```
find . -path "*/examples/*" -name "benchmarkingjob.yaml"
```
This can never match anything under `core/`, since `core/` contains no `benchmarkingjob.yaml` by design. This means the same architectural gap already present in the merged `dependency_validator.py` (which scopes checks to `examples/` inventory entries only) is independently repeated here — two separately-authored validation mechanisms both exclude `core/` from dependency checking.

This PR is only 10 commits behind `main` (verified via `git rev-list --count`), so extending its target discovery to also cover `core/` looks like a low-risk, low-effort addition on top of otherwise solid work.

Arijit429:
> Reviewed the `requirements-check` job in this workflow. It discovers validation targets via:
> 
> ```
> find . -path "*/examples/*" -name "benchmarkingjob.yaml"
> ```
> 
> This can never match anything under `core/`, since `core/` contains no `benchmarkingjob.yaml` by design. This means the same architectural gap already present in the merged `dependency_validator.py` (which scopes checks to `examples/` inventory entries only) is independently repeated here — two separately-authored validation mechanisms both exclude `core/` from dependency checking.
> 
> This PR is only 10 commits behind `main` (verified via `git rev-list --count`), so extending its target discovery to also cover `core/` looks like a low-risk, low-effort addition on top of otherwise solid work.
> 
> (Full review for LFX 2026 Term 3: #868)

Good catch — you're right that `BenchmarkingJob._parse_config()` and `Module._parse_config()` carry the same pattern. This PR is intentionally scoped to `testenv.py` and `algorithm.py` as the two entry points that handle every example's config parsing (via `benchmarkingjob.py:115-124` and `testcasecontroller.py:63-78`). The other instances are worth a follow-up PR or can be addressed in the same batch if maintainers prefer. Happy to expand scope if that's the preferred approach.

greenfield01:
Following up on the duplication concern raised in the thread, took a closer look at the overlap with #771's already-merged framework.

dependency_validator.py (part of #771) already runs per-example dependency checks, including whether a requirements.txt exists, as part of its static validation tier. This PR's "requirements check" step (verifying benchmarkingjob.yaml directories have a matching requirements.txt) looks like it's covering overlapping ground, just via a separate lint-focused workflow rather than the validator framework.

Given continue-on-error: true is set on both steps here, running both isn't harmful today, but it does mean the same check could report two independent (and potentially inconsistent) results for the same condition.

Yash4616:
Reviewed against `main` at `37a9c60`. Merges clean, and unlike the other open CI PRs in this area it adds a new workflow file rather than editing one that was renamed in `7f7776a`, so it does not share their conflict surface.

The overlap with the merged validation framework has already been raised in this thread and I am not restating it.

Two points on the gate design.

Both jobs carry `continue-on-error: true`, and the file explains why:

```yaml
        # continue-on-error: the examples/ backlog is not pylint-clean yet, so this check
        # is currently advisory only. Remove continue-on-error once examples reach a
        # baseline passing state, so this becomes a real PR-blocking gate.
```

Documenting the intent is good practice, and the reasoning is sound. The consequence is that the workflow cannot fail a pull request under any circumstance, including a newly introduced problem. Another open PR in this area reaches the same land-now-tighten-later goal with a baseline file: existing breakage is grandfathered, but anything new still fails. That gives both an advisory period and a working gate from day one, where `continue-on-error` gives only the advisory period, and removing it later is a separate decision that nothing currently schedules.

Second, this is a new file and it introduces `actions/checkout@v3` and `actions/setup-python@v3`. The current workflows on `main` use `@v7`. Since this is greenfield rather than an edit to existing steps, matching `@v7` here costs nothing and avoids adding new drift.

Recommendation: minor revision. Consider a baseline in place of `continue-on-error`, and use the action versions already standard on `main`.

suhaan-24:
Useful direction — there is currently no automated check on `examples/**` at all, so anything that
runs on example changes is an improvement over the status quo. A few observations on coverage and
scope.

## What the gate catches, and what it does not

The workflow runs `pylint` across `examples/**` on Python 3.9/3.10/3.11 when example files change.
That catches syntax errors, undefined names and import problems in Python sources.

The most common breakage in these examples is not Python-shaped. Two patterns dominate:

1. **Config values pointing at paths that no longer exist.** For instance a `benchmarkingjob.yaml`
   whose `testenv:` still reads
   `./examples/curb-detection/lifelong_learning_bench/testenv/testenv.yaml` after the directory was
   restructured. That fails at config load with
   `RuntimeError: not found testenv config file(...) in local`, before any Python in the example is
   imported — so `pylint` cannot see it.

2. **Dataset keys the loader does not read.** `core/testenvmanager/dataset/dataset.py` assigns
   `train_url`/`test_url` inside `process_dataset()` rather than reading them from config, so a
   `testenv.yaml` supplying those keys raises
   `NotImplementedError: not one of train_index/train_data/train_data_info`. At `upstream/main`, 26
   of the 48 example `testenv*.yaml` files still declare them.

Both are statically checkable without a dataset, a GPU, or executing anything — which matters here,
because most of these examples cannot run in CI at all. A step that walks `examples/**/*.yaml`,
resolves every `url:`/`testenv:` value against the filesystem, and flags deprecated dataset keys
would cover the two highest-volume failure modes and would run in seconds.

Not asking you to build that here — mainly flagging that the workflow name ("example lint and
requirements") reads broader than what it currently gates, and it would be worth either narrowing
the name or filing a follow-up so the config-validation half is not assumed to be covered.

## Scope note

The PR also rewrites
`examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py` to make the
`matplotlib`/`open3d` imports conditional. That is a good fix in its own right (open issue #635),
but it is an example source change inside a CI PR, and the title does not mention it.

The identical hunk also appears in #683, byte for byte — diffing both branches against their merge
bases shows no difference for that file — so whichever merges second will conflict. Worth deciding
which PR owns it.

## Smaller points

- Pinning `runs-on: ubuntu-22.04` rather than `ubuntu-latest` is a good call for reproducibility.
- Worth confirming whether the job is advisory or blocking on first merge. Running pylint over
  `examples/**` for the first time will almost certainly surface a large existing backlog, and a
  blocking gate would then stop unrelated PRs until it is cleared.

## Verification boundary

**Not executed** — I have not run this workflow or `pylint` across `examples/**`, so I cannot say
how many findings it would produce on first run. The two error messages quoted above are from
executed runs of `cityscapes-synthia/lifelong_learning_bench/curb-detection`; the 26-of-48 count is
a `git grep` over `upstream/main`; the #683 overlap is a diff comparison.

Shikhar-Kesharwani:
## Task 2 Review — PR #851

### Recommendation: major revision

I consider this a critical PR because it introduces repository-level CI validation rather than only fixing one example. However, I also reviewed the complete three-commit PR and noted that it contains an additional example-source change in `examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py`. That change overlaps with the corresponding change discussed in #683, so the PR has both shared-CI scope and example-local scope.

### Root problem versus symptom

The CI portion is directionally relevant to the root problem identified in Task 1 because it attempts to introduce shared validation instead of relying only on individual example fixes.

However, the current workflow does not directly validate the dependency contract between example-local requirements and Ianvs core requirements.

The workflow installs the Ianvs root requirements, while the requirements check only verifies that an example containing `benchmarkingjob.yaml` has a `requirements.txt`. It does not install and validate each example's requirements file or establish that the example dependency set is complete and compatible with the core environment.

Therefore, this PR improves repository-level validation infrastructure but does not, by itself, fully prevent the dependency-boundary failures represented by #644 and #685.

### Enforcement and regression risk

Both validation checks use:

`continue-on-error: true`

This makes them advisory rather than merge-blocking.

I understand the stated reason: the examples tree has an existing lint and requirements-file backlog, so making the checks immediately blocking could reject unrelated changes.

The regression risk is that a newly introduced failure can therefore pass the workflow during the advisory period. A stronger transition mechanism would preserve existing-baseline tolerance while still detecting newly introduced violations.

I therefore consider the current enforcement model incomplete for a long-term regression gate.

### Cross-example impact

Because this is shared CI, its design affects validation of many examples rather than one example.

That makes the dependency-validation scope particularly important: fixing #683 and #688 repairs concrete examples, but future examples can still reproduce similar dependency-boundary problems unless the shared validation layer checks the actual installation relationship between core and example requirements.

### Trigger-boundary finding

My additional review concern is the workflow trigger:

`pull_request: paths: - 'examples/**'`

This means the workflow is selected based on changed paths under `examples/**`.

A change to the repository-level `requirements.txt` can alter the core dependency environment used by examples without itself matching that path filter.

Therefore, a root-level dependency change can bypass this workflow even though it can affect the dependency environment of multiple examples.

This is separate from `continue-on-error`: even if the checks become blocking, the workflow still needs an appropriate trigger strategy for dependency changes outside `examples/**`.

### Multiple-PR interaction and duplicated changes

PRs #683 and #688 address concrete example-level dependency failures, while #851 attempts to establish shared validation.

These changes are complementary in principle: the example fixes repair existing failures, while CI should prevent similar regressions.

However, the additional `utils.py` change in #851 overlaps with the corresponding change in #683. Merging both independently could create duplicated ownership and unnecessary merge/conflict risk. That example-local change should therefore have one clear owning PR rather than being carried in both branches.

I am not claiming the duplicated `utils.py` finding as a new finding; it is already visible in the existing discussion. My focus here is how that duplication affects the scope and merge strategy of the critical CI PR.

### Verification

I inspected the complete three-commit PR and its workflow locally.

I verified that the workflow:

- uses Python 3.9, 3.10, and 3.11;
- installs the Ianvs root requirements;
- runs pylint over the examples;
- checks for the presence of `requirements.txt`;
- uses `continue-on-error: true` for both validation jobs; and
- is triggered by pull requests affecting `examples/**`.

I also compared the PR structure and identified the additional `utils.py` change that overlaps with the example-local change in #683.

I did not execute GitHub-hosted Actions locally, so I do not claim a successful hosted-CI execution.

### Conclusion

The PR provides useful shared CI infrastructure and is relevant to the Task 1 root problem, but I recommend **major revision** before treating it as a complete dependency-regression solution.

The revised validation should define an enforceable baseline, validate the actual relationship between Ianvs core and example dependencies, and ensure that relevant repository-level dependency changes can trigger the appropriate validation.

The example-local `utils.py` change should also have a single clear owning PR to avoid duplicated implementation and merge conflicts.

Shikhar-Kesharwani:
## Task 2 Review — PR #851

### Recommendation: major revision

I consider this a critical PR because it introduces repository-level CI validation rather than only fixing one example. However, I also reviewed the complete three-commit PR and noted that it contains an additional example-source change in `examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py`. That change overlaps with the corresponding change discussed in #683, so the PR has both shared-CI scope and example-local scope.

### Root problem versus symptom

The CI portion is directionally relevant to the root problem identified in Task 1 because it attempts to introduce shared validation instead of relying only on individual example fixes.

However, the current workflow does not directly validate the dependency contract between example-local requirements and Ianvs core requirements.

The workflow installs the Ianvs root requirements, while the requirements check only verifies that an example containing `benchmarkingjob.yaml` has a `requirements.txt`. It does not install and validate each example's requirements file or establish that the example dependency set is complete and compatible with the core environment.

Therefore, this PR improves repository-level validation infrastructure but does not, by itself, fully prevent the dependency-boundary failures represented by #644 and #685.

### Enforcement and regression risk

Both validation checks use:

`continue-on-error: true`

This makes them advisory rather than merge-blocking.

I understand the stated reason: the examples tree has an existing lint and requirements-file backlog, so making the checks immediately blocking could reject unrelated changes.

The regression risk is that a newly introduced failure can therefore pass the workflow during the advisory period. A stronger transition mechanism would preserve existing-baseline tolerance while still detecting newly introduced violations.

I therefore consider the current enforcement model incomplete for a long-term regression gate.

### Cross-example impact

Because this is shared CI, its design affects validation of many examples rather than one example.

That makes the dependency-validation scope particularly important: fixing #683 and #688 repairs concrete examples, but future examples can still reproduce similar dependency-boundary problems unless the shared validation layer checks the actual installation relationship between core and example requirements.

### Trigger-boundary finding

My additional review concern is the workflow trigger:

`pull_request: paths: - 'examples/**'`

This means the workflow is selected based on changed paths under `examples/**`.

A change to the repository-level `requirements.txt` can alter the core dependency environment used by examples without itself matching that path filter.

Therefore, a root-level dependency change can bypass this workflow even though it can affect the dependency environment of multiple examples.

This is separate from `continue-on-error`: even if the checks become blocking, the workflow still needs an appropriate trigger strategy for dependency changes outside `examples/**`.

### Multiple-PR interaction and duplicated changes

PRs #683 and #688 address concrete example-level dependency failures, while #851 attempts to establish shared validation.

These changes are complementary in principle: the example fixes repair existing failures, while CI should prevent similar regressions.

However, the additional `utils.py` change in #851 overlaps with the corresponding change in #683. Merging both independently could create duplicated ownership and unnecessary merge/conflict risk. That example-local change should therefore have one clear owning PR rather than being carried in both branches.

I am not claiming the duplicated `utils.py` finding as a new finding; it is already visible in the existing discussion. My focus here is how that duplication affects the scope and merge strategy of the critical CI PR.

### Verification

I inspected the complete three-commit PR and its workflow locally.

I verified that the workflow:

- uses Python 3.9, 3.10, and 3.11;
- installs the Ianvs root requirements;
- runs pylint over the examples;
- checks for the presence of `requirements.txt`;
- uses `continue-on-error: true` for both validation jobs; and
- is triggered by pull requests affecting `examples/**`.

I also compared the PR structure and identified the additional `utils.py` change that overlaps with the example-local change in #683.

I did not execute GitHub-hosted Actions locally, so I do not claim a successful hosted-CI execution.

### Conclusion

The PR provides useful shared CI infrastructure and is relevant to the Task 1 root problem, but I recommend **major revision** before treating it as a complete dependency-regression solution.

The revised validation should define an enforceable baseline, validate the actual relationship between Ianvs core and example dependencies, and ensure that relevant repository-level dependency changes can trigger the appropriate validation.

The example-local `utils.py` change should also have a single clear owning PR to avoid duplicated implementation and merge conflicts.

Shikhar-Kesharwani:

For PR #851 specifically, the workflow provides shared validation infrastructure, but it should not be treated as the primary repair for the dependency failures identified in #644 and #685.

The current workflow checks for the presence of `requirements.txt` and runs pylint, but the requirements check does not validate the actual dependency relationship between Ianvs Core and each Example. It therefore cannot establish whether an Example's dependency environment is complete or whether an Example incorrectly declares Ianvs itself as an external dependency.

The appropriate boundary for the concrete dependency corrections remains **Dependency / Packaging**, while **CI / Validation** should serve as the enforcement boundary. The affected Example requirements should contain the dependencies appropriate to their defined ownership model, while shared CI should detect violations of that model and prevent recurrence.

A boundary-aligned strategy for this PR is:

1. Define the ownership relationship between Ianvs Core dependencies and Example-specific dependencies.
2. Keep concrete dependency repairs in the affected Example requirements.
3. Extend shared CI validation to check the defined dependency contract rather than only checking for the existence of `requirements.txt`.
4. Ensure relevant repository-level dependency changes can trigger the appropriate validation.
5. Move the relevant validation from advisory to enforceable after an acceptable existing-error baseline has been established.

This boundary limits regression risk. Modifying Ianvs Core runtime code to compensate for Example dependency declarations could affect unrelated Examples that use the same Core functionality. A Dependency / Packaging repair changes installation behavior more narrowly, while CI provides repository-wide regression protection.

I therefore recommend retaining the CI approach as a **secondary enforcement layer**, while keeping the actual dependency repair at the Dependency / Packaging boundary rather than moving the repair into Ianvs Core.

I did not execute GitHub-hosted Actions locally, so this finding is based on inspection of the workflow and its validation scope rather than a claim of successful hosted-CI execution.

christinaliyevav-coder:
I reviewed PR #851 first against its own stated scope.

This PR adds a new repository-level workflow, `.github/workflows/example-lint.yml`, for pull requests that modify `examples/**`. It introduces two shared checks:

- `pylint` over the `examples/` tree on Python 3.9, 3.10 and 3.11;
- verification that directories containing `benchmarkingjob.yaml` also contain a `requirements.txt`.

I think this is useful infrastructure because it moves part of example validation out of individual restoration PRs and into a shared CI layer.

The workflow is intentionally limited. The `requirements-check` job verifies that a manifest exists, while the `pylint` job checks Python source quality. Neither of these mechanisms is designed to validate whether an external API or other runtime prerequisite is actually configured.

I tested that boundary with one concrete example from another restoration PR. A credential expression such as:

```python
api_key = os.getenv("DEEPSEEK_API_KEY") or "<DeepSeek API Key>"
```

is valid Python even though it can hide a missing external-service prerequisite.

Running the extracted credential expressions through `pylint` produced:

```text
------------------------------------
```

I treat this as a **scope boundary**, not as a bug in PR #851. It shows that the new workflow can improve source-level and manifest-level validation without covering runtime/preflight configuration problems.

The existing discussion already raises `continue-on-error: true`, dependency/config coverage, and other workflow concerns, so I am not claiming those observations as new.

For the validation scope stated by this PR, I would keep the workflow. If Ianvs later wants repository-wide checks for external-service prerequisites, I would treat that as a separate preflight/configuration-validation layer rather than adding that responsibility to `pylint`.

My cross-PR finding here is only that the new shared lint gate should not be interpreted as validating external-service readiness: a real missing-credential fallback from another reviewed restoration can remain completely lint-clean.

ankita7Patil:

I reviewed PR #851 as the strongest of the three related PRs because it modifies shared CI validation infrastructure rather than a single example.

The workflow improves example-level validation through Pylint and requirements.txt presence checks, but it does not validate the shared inventory schema, lifecycle/status values, or benchmark coverage identified in Task 1.

Therefore, I recommend extending this workflow with inventory/validator checks before making the validation blocking.

ankita7Patil:
## Task 3/4 — Restoration Cross-Reference

PR #851 is the most important integration point for the restoration plan because it modifies shared CI validation.

The recommended path is to extend this workflow with inventory/schema validation first, keep the new validation warning-only during rollout, and make it blocking only after the existing example baseline is stable.

Full Repair Boundary and Restoration Path:

shubhika123:
## Pre-test Review (LFX Term 3 2026 Phase IV)

**Recommendation: Major Revision**

### Technical Evaluation

This PR introduces `.github/workflows/example-lint.yml` to provide automated validation for the examples across Python 3.9–3.11. The addition is useful, but there are gaps that prevent it from functioning as a reliable regression guard yet.

### Critical Findings

**1. Non-Blocking Execution (`continue-on-error: true`)**

Both the `pylint` and `requirements-check` steps use `continue-on-error: true`. As a result, failures in these checks do not fail the workflow and cannot reliably prevent a problematic change from merging.

The workflow itself explains this is intentional because the existing examples are not yet fully pylint-clean. That's a reasonable interim state, but there should be a defined path toward making the checks blocking.

**Recommendation:** Either (a) add a tracked follow-up issue with a defined condition for removing `continue-on-error: true`, or (b) make the check blocking only for newly-changed files via diff-based linting, so new code is held to the standard immediately without requiring the existing backlog to be fixed first.

### **2. Trigger Boundary Mismatch**

The workflow triggers only on:

```text
examples/**
```

Changes to shared code under `core/**` or the root `requirements.txt` will not trigger this validation workflow at all, meaning changes outside `examples/**` can silently affect the examples this workflow exists to protect, without ever being checked.

**Recommendation:**

```yaml
on:
  pull_request:
    paths:
      - 'examples/**'
      - 'core/**'
      - 'requirements.txt'
      - '.github/workflows/example-lint.yml'
```

### **3. Superficial Requirements Check**

The workflow checks `requirements.txt` existence via `test -f requirements.txt`, confirming only that the file exists, not that the listed packages are actually installable in a clean environment.

**Recommendation:** The workflow should perform an actual dependency installation or resolution check in a clean environment so that invalid or incompatible requirements cause the validation to fail.

### **Additional Finding — Unrelated Changes Bundled in the PR**

The PR also modifies `examples/industrialEI/pose-estimation-llio/testalgorithms/llio_fusion/utils.py`, replacing the hardcoded `MATPLOTLIB_AVAILABLE = False` / `OPEN3D_AVAILABLE = False` with real `try/except ImportError` checks.

This is a legitimate runtime behavior fix for the LLIO example, but it is functionally unrelated to the CI workflow addition. Keeping an infrastructure change and an example runtime fix in the same PR makes both harder to review, revert, or bisect independently.

**Recommendation:** Move the LLIO runtime change into a separate PR, leaving this PR focused on CI infrastructure.

### **Why Major Revision, Not Minor**

Each individual fix above (adding trigger paths, changing `continue-on-error`, improving the requirements check, and splitting the PR) is small in line count. I'm classifying this as **Major Revision** rather than Minor anyway, because the rubric's own framing of "minor revision" is fixes *without coding effort* — docs, squash commits — not small functional or behavioral changes.

A CI workflow whose entire stated purpose is "add automated validation" but which neither blocks on failure nor covers the `core/**` changes most likely to break examples does not yet achieve its stated purpose, regardless of how few lines the fixes would take.

### Final Recommendation

**Major Revision.** The workflow is a useful addition and `continue-on-error` is understandable given the backlog, but the PR needs a defined path to blocking validation, broader trigger coverage, a meaningful requirements/installability check, and separation of the unrelated LLIO change before it functions as a real regression guard.

aryananand-04:
I ran this workflow's logic against some example failures I already had reproduced, to see what it would actually catch. Mixed results, and one part of it is better than I expected.

The pylint job is real new coverage. The lint workflow already on `main` only scopes to `${{github.workspace}}/core`, so nothing currently lints `examples/` at all. That's worth keeping regardless of what happens to the rest of this PR.

The problem is that neither job can fail a PR. Both have `continue-on-error: true`, so whatever they find, the PR still merges. Right now this reports rather than gates.

There's a deeper issue with the lint job specifically. The workflow installs the Ianvs root requirements but not the ML libraries the examples import. With no `torch` or `tensorflow` present, pylint has no signatures to check against, so an API drift break isn't just unreported, it's invisible. I tested against lines I know are broken:

```
basemodel.py:56   (unguarded regex)        -> no output
basemodel.py:153  (addmm_ positional call) -> no output
bert.py:45                                 -> no output
bert.py:146                                -> no output
fpn/basemodel.py:26 (tensorflow.contrib)   -> E0401
```

Only the missing module shows up, and it's one line among 1,993 `E0401`s from a full sweep, so you'd have a hard time picking it out as signal. Non-blocking either way.

The requirements job has a similar shape. It checks that a `requirements.txt` exists, which isn't the same as checking it's complete or installable. An example can ship a manifest, pass this, and still fail on import because something was never listed.

If this is meant to be the long-term regression guard, I'd suggest going in this order:

1. Install the ML libraries the examples import, pinned to whatever the example matrix needs, before linting. Nothing downstream can detect signature drift without this.
2. Add a bare `import` smoke check for any `testalgorithms/` directory importing `torch` or `tensorflow` directly. No inference, no dataset, so no GPU or data needed.
3. Triage the existing 1,993 `E0401`s to sort real signal from environment noise.
4. Drop `continue-on-error: true` only after those three. Flipping it earlier would start rejecting unrelated PRs on noise that was already there, which is worse than what we have now.

One way to check the gate actually gates: take a bug this repo has already fixed, reintroduce it on a throwaway branch, open a dry run PR and see whether CI rejects it. Running the checks against current code proves they execute. It doesn't prove they'd stop a bad merge.

Major revision if this is the long-term guard, though I'd keep the pylint scope change as is.

Ran pylint locally on macOS arm64, Python 3.11, against the example tree. I didn't run the hosted Actions, so nothing here is a claim about a real CI run.

Arijit429:
Thanks everyone for the thorough reviews. Updated the workflow with the following changes:

1. Added `core/**` and `requirements.txt` to trigger paths — changes to core/ or root requirements.txt now trigger the workflow
2. Pinned actions/checkout and actions/setup-python to @v7
3. Added tracking issue #995 to both continue-on-error blocks — this is the explicit reference for when the advisory checks become blocking
4. Confirmed utils.py is not in this branch — the matplotlib/open3d fix (bf79838) was already merged into main before this branch was created, so there is no overlap to resolve here

The pylint coverage this PR adds is additive to the merged dependency_validator.py — that validator checks dependency presence, this workflow lints Python source. Different concerns, no duplication.

The continue-on-error removal path is now tracked in #995.

PR #853, opened 2026-08-21
docs: add blog post on Ianvs example restoration audit findings

Arijit429:
/kind documentation

Adds a blog post documenting the findings from the Ianvs example restoration audit conducted during the LFX Mentorship application period. The post covers:

- Five recurring failure patterns found across examples (with file:line citations)
- Case study: examples/aoa/single_task_bench/TForest — permanently non-reproducible, proposed as Quarantined classification (issue #846)
- What the new CI workflow (PR #851) does and the rationale for starting non-blocking
- Call to action for contributors to help restore remaining examples

The post is intended for submission to the KubeEdge website blog. It links to the debugging playbook in docs/guides/debugging-common-example-failures.md and PR #851 for full technical context.

Related to #846


Shikhar-Kesharwani:
### Review — supplementary headless-environment finding

The conditional imports are a good improvement because they make Matplotlib and Open3D genuinely optional at import time.

There is one remaining portability issue in the same visualization path: `visualize()` unconditionally calls `matplotlib.use("TkAgg")` near the end of the function and then calls `plt.show()`.

This means the new `ImportError`-based optional-dependency handling does not fully make visualization safe in headless environments. A server, Docker container, or CI runner can have Matplotlib installed successfully but still lack Tk/Tkinter or a usable display backend. In that case the import succeeds, but the visualization path can fail later when the `TkAgg` backend is selected or when `plt.show()` attempts to open a GUI window.

This is particularly relevant for this example because Ianvs examples are commonly executed in containerized/headless environments. The dependency check should therefore be separated from GUI-backend availability.

I recommend using a non-interactive backend such as `Agg` for file generation, or selecting `TkAgg` only when an interactive display is explicitly requested and the required GUI environment is available. The code should also avoid making `plt.show()` mandatory when the actual output is already being written with `plt.savefig()`.

This is separate from the optional-import fix in this PR: the current change handles missing Python packages, but the visualization path still assumes a GUI-capable runtime after those imports succeed.

PR #854, opened 2026-08-21
fix(core): replace unguarded config[key] lookups with clear ValueError in _parse_config

Arijit429:
/kind bug

`TestEnv._parse_config()` and `Algorithm._parse_config()` both do unguarded `config[key]` lookups. A typo in the top-level YAML key (e.g. `test_env:` instead of `testenv:`) produces:

`RuntimeError: testenv config file(examples/foo/testenv/testenv.yaml) is not supported, error: 'testenv'`

The `error: 'testenv'` part is just the KeyError repr — it tells you which file is wrong but not what's wrong with it. A new contributor has no way to connect that message to "rename your top-level YAML key."

Before (testenv.py:63):
`config_dict = config[str.lower(TestEnv.__name__)]`

After:
`if not isinstance(config, dict) or expected_key not in config:`
`    raise ValueError(`
`        f"testenv config must have a top-level '{expected_key}' key; "`
`        f"found: {list(config.keys()) if isinstance(config, dict) else config!r}"`
`    )`

Same fix applied to `Algorithm._parse_config()` in algorithm.py. Also covers `yaml2dict` returning None on an empty YAML file — the isinstance guard catches that too.

Related to #846

### DIFF
--- a/core/testcasecontroller/algorithm/algorithm.py
+++ b/core/testcasecontroller/algorithm/algorithm.py
@@ -157,7 +157,13 @@ class Algorithm:
                 f"algorithm initial_model_url({self.initial_model_url}) must be string type.")
 
     def _parse_config(self, config):
-        config_dict = config[str.lower(Algorithm.__name__)]
+        expected_key = str.lower(Algorithm.__name__)
+        if not isinstance(config, dict) or expected_key not in config:
+            raise ValueError(
+                f"algorithm config must have a top-level '{expected_key}' key; "
+                f"found: {list(config.keys()) if isinstance(config, dict) else config!r}"
+            )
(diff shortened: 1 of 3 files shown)


ankita7Patil:

I reviewed PR #854 as the core configuration-parsing fix in the related-work set.

The PR improves error handling by replacing unguarded config-key lookups with a clear ValueError and explicitly reporting the expected and found configuration keys. This is a useful usability improvement, but it does not address the shared inventory schema, lifecycle/status validation, or benchmark coverage gap identified in Task 1.

I therefore recommend keeping this fix, while treating it as complementary to the shared inventory/validator restoration work rather than as a root-problem fix by itself.

PR #860, opened 2026-08-22
fix(examples): remove phantom PDF dependencies, add real PDF extraction, and fix stale paths in GovDoc2Poster

### LINKED ISSUE #857: [License Audit] AGPL-3.0 dependencies in examples conflict with Apache-2.0 project license
**What happened**:
A license scan across example requirements.txt files found two AGPL-3.0 dependencies that directly conflict with Ianvs' Apache-2.0 project license:

1. `PyMuPDF` in `examples/GovDoc2Poster/singletask_learning_bench/requirements.txt` — dual-licensed "GNU AFFERO GPL 3.0 or Artifex Commercial License". Without a paid Artifex commercial license, usage is AGPL-only. AGPL's network copyleft clause requires the entire distributed application to also be AGPL.

2. `ultralytics` in `examples/industrialEI/single_task_learning_bench/deformable_component_manipulation/requirements.txt` — licensed AGPL-3.0. The example also loads `yolov8n.pt` pretrained weights directly, which carry the same AGPL-3.0 license.

All example dependencies should be compatible with Ianvs' Apache-2.0 license. AGPL-3.0 dependencies should either be replaced with permissively-licensed alternatives or explicitly approved by maintainers with documented reasoning.

```bash
pip index versions pymupdf 2>/dev/null | head -1
curl -s https://pypi.org/pypi/PyMuPDF/json | python3 -c "import sys,json; d=json.load(sys.stdin); print(d['info']['license'])"
# Returns: GNU AFFERO GPL 3.0

curl -s https://pypi.org/pypi/ultralytics/json | python3 -c "import sys,json; d=json.load(sys.stdin); print(d['info']['license'])"
# Returns: AGPL-3.0
```

Proposed resolution:
- Replace PyMuPDF with an Apache/MIT-compatible alternative (e.g. pypdf, pdfplumber)
- Replace ultralytics with an alternative or document the AGPL dependency explicitly with maintainer approval
- Add a license compatibility check to the CI validation framework

This finding was part of a broader license scan across all example requirements.txt files conducted as part of the Phase IV example restoration audit.

Arijit429:
/kind bug
/kind cleanup

Three related fixes in `examples/GovDoc2Poster/singletask_learning_bench/`:

**1. Remove phantom PDF dependencies (license fix)**
`requirements.txt` listed `PyMuPDF>=1.23.0` and `PyPDF2>=3.0.0` — neither is imported anywhere in the codebase. PyMuPDF is AGPL-3.0, directly incompatible with Ianvs' Apache-2.0 license (issue #857). Both removed. `pdfplumber` (Apache-2.0 compatible) promoted from optional to required.

**2. Add real PDF text extraction**
`basemodel.py:457` passed `.pdf` file paths directly to `parse_document()` as if they were text — no extraction step existed. Added `_extract_pdf_text()` using pdfplumber and wired it in before `parse_document()`.

**3. Fix stale hardcoded directory name**
`'new_government_agent'` appeared in 5 places (lines 137, 146, 148, 420, 669) — the actual directory is `GovDoc2Poster`. All five corrected.

Fixes #857

### DIFF
--- a/examples/GovDoc2Poster/singletask_learning_bench/requirements.txt
+++ b/examples/GovDoc2Poster/singletask_learning_bench/requirements.txt
@@ -6,11 +6,9 @@ ianvs>=0.1.0
 sedna>=0.1.0
 
 # Document processing (轻量级方案)
-PyPDF2>=3.0.0
-PyMuPDF>=1.23.0  # fitz
 python-docx>=0.8.11
 python-pptx>=0.6.21
-pdfplumber>=0.7.0  # 可选，用于更好的表格提取
+pdfplumber>=0.7.0
 
 # Image processing
 Pillow>=9.0.0
--- a/examples/GovDoc2Poster/singletask_learning_bench/testalgorithms/gen/basemodel.py
+++ b/examples/GovDoc2Poster/singletask_learning_bench/testalgorithms/gen/basemodel.py
@@ -14,6 +14,7 @@ from concurrent.futures import ThreadPoolExecutor, as_completed
 
 # Import necessary libraries
 from dotenv import load_dotenv
+import pdfplumber
 from sedna.common.class_factory import ClassType, ClassFactory
 
 # Import custom tools
@@ -134,7 +135,7 @@ class NewGovernmentPosterAgent:
         else:
             # load poster index if exists so evaluate_only can lookup poster_path for string data items
             try:
-                index_path = os.path.join('examples', 'new_government_agent', 'resources', 'datasets', 'test', 'poster_index.json')
+                index_path = os.path.join('examples', 'GovDoc2Poster', 'resources', 'datasets', 'test', 'poster_index.json')
                 if os.path.exists(index_path):
                     with open(index_path, 'r', encoding='utf-8') as f:
                         self._poster_index = json.load(f)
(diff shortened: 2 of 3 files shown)


yadavchiragg:
Review Depth Level: Analysis
What it does: Three fixes in one PR — (1) removes phantom PyMuPDF/PyPDF2 dependencies that were listed in requirements.txt but never imported anywhere, and drops PyMuPDF specifically because its AGPL-3.0 license is incompatible with Ianvs' Apache-2.0 license; (2) adds real PDF text extraction via pdfplumber — previously, .pdf file paths were passed straight into parse_document() as if they were already text, meaning no extraction ever happened; (3) fixes a stale hardcoded directory name ('new_government_agent' instead of the actual GovDoc2Poster) repeated in 5 separate lines.
Finding worth flagging: finding #2 is the most interesting — a requirements.txt can list a dependency and still be functionally broken if the code never actually calls it correctly. This is a variant of your Task 1 root problem worth naming explicitly: a manifest existing is necessary but not sufficient; #535/#460/#430 were about missing manifests, this is about a present-but-non-functional pipeline despite dependencies technically being listed. Also notable: this PR references a license-compliance issue (#857) as its trigger — a category (Apache-2.0 vs. AGPL-3.0 conflicts) not covered anywhere in your Task 1–4 root-problem cluster, which is good genuine coverage breadth for Bonus.

yadavchiragg:
Nice fix — worth flagging one thing for the description/reviewers: the phantom-dependency removal and the real-extraction fix are actually two different bug classes worth distinguishing explicitly. A `requirements.txt` listing PyMuPDF/PyPDF2 gave the *appearance* of PDF support while `basemodel.py` never called any extraction step — meaning "the manifest is complete" and "the pipeline actually works" are separate claims that can silently diverge. Worth keeping in mind for other examples: a present-and-correct requirements.txt doesn't guarantee the code paths that use those packages are wired up correctly.

Issue #845, opened 2026-08-20
[Proposal] Phase IV: Add pytest suite and schema validation to core/


Arijit429:
## Background

Phase III (PR #541) designed a comprehensive CI validation framework for `examples/`. 
That proposal explicitly scopes itself to examples/ and excludes core/ changes.

## Gap

`core/` currently has:
- Zero test coverage — no pytest.ini, no test_*.py files, no pytest dependency
- `.github/workflows/main.yaml` runs only pylint, no functional tests
- Known bugs that would be caught by tests:
  - `algorithm.py:_check_fields` uses `and` instead of `or`, silently accepting invalid types
  - `utils.py:yaml2dict` uses bare yaml.load with no schema validation
  - `simulation_system_admin.py` builds shell=True subprocess commands via string interpolation from config values — security relevant

## Proposal

Build a pytest suite for core/ as Phase IV work:
- pytest infrastructure setup (pytest.ini, conftest.py, CI integration)
- Unit tests for core paradigms (singletask, incremental, federated, lifelong)
- Schema validation layer for testenv.yaml and algorithm.yaml config parsing
- Fix the concrete bugs above as part of establishing baseline test coverage

## Relationship to existing work

This explicitly builds on Phase III's proposal rather than duplicating it — 
Phase III owns examples/, Phase IV owns core/.

Lu2077:
/assign

Hello @Arijit429! 

I am a prospective mentee for the upcoming LFX Mentorship program. Since I have already contributed to Kubeflow and I am very familiar with the CNCF ecosystem and cloud-native environments. I would like to contribute to this issue. 

I’ve reviewed the description and the gaps mentioned. I would like to start by setting up the baseline fixes for the concrete bugs you identified (like the `and`/`or` logical error in `algorithm.py` or updating the unsafe subprocess calls in `simulation_system_admin.py`). 

I'm ready to put my hands on the code.  
I'll appreciate any comment or feedback, thanks in advance!

Arijit429:
@Lu2077 Thanks for the interest! To clarify — I'm already working on the bugs identified in this issue. The _check_fields and/or logic fix is in my pipeline, and the subprocess security issue in simulation_system_admin.py is also on my list. I'll be opening PRs for these shortly.

Issue #888, opened 2026-08-23
[PHASE IV] Benchmarks report emulated-bfloat16 numbers as real on pre-Ampere GPUs


vjuhhii:
## Background

Ianvs is a benchmarking toolkit: its output is measurements. Two of its LLM examples hardcode
`torch.bfloat16` as the compute dtype, at three call sites. On hardware with compute capability
< 8.0, `bfloat16` has no native execution path — PyTorch emulates it.

Emulation does not raise. The benchmark runs to completion and reports results. Measured on a
GTX 1080 (CC 6.1), a 4096×4096 matmul in `bfloat16` is **1.85× slower than the same matmul in
float32** — the dtype chosen as the fast path is, on this hardware, the slowest of the three
available. (This is a microbenchmark of the arithmetic, not an end-to-end example run; the
end-to-end factor depends on how much of a given benchmark is matmul-bound.)

The measurement Ianvs then publishes does not describe the hardware. It describes an undeclared
dtype assumption. For a benchmarking toolkit this is a more serious failure mode than a crash,
because a crash is visible.

Nothing in the repository can detect this condition:

```console
$ grep -rn "get_device_capability\|is_bf16_supported" --include=*.py .
$   # no matches
```

Zero occurrences, in `core/` and across all 28 examples. Ianvs has no notion of GPU generation.
The only distinction it can express is "cuda" versus "not cuda".

## Goals

1. Stop Ianvs from reporting benchmark results obtained through silently emulated arithmetic.
2. Make each example's real hardware floor explicit — declared in configuration and documentation
   rather than implied by a hardcoded dtype or a transitive dependency.
3. Extend the device contract in `core/` from binary (GPU / no GPU) to capability-aware, so
   paradigms and examples share one source of truth.
4. Add a CI check so a new example cannot reintroduce an undeclared floor.

### Impact

**Primary — invalid measurements, no warning.** Any `bfloat16` benchmark run on pre-Ampere hardware
produces numbers that are partly an artifact of emulation. Nothing in the output distinguishes them
from a valid result. Measured below.

**Secondary — undocumented hard floors.** Two hardware floors are imposed by current example code,
neither documented:

| Floor | Imposed by | Excludes | Failure mode |
|---|---|---|---|
| CC ≥ 8.0 (Ampere, 2020+) | hardcoded `torch.bfloat16` | Pascal, Turing, and older | **silent slowdown** |
[...]

PR #890, opened 2026-08-23
[PHASE IV] test(core): add regression tests for use_gpu device visibility

### LINKED ISSUE #765: [PHASE IV] use_gpu: false in testenv.yaml is a no-op, and use_gpu: true is applied after the base model is already constructed

vjuhhii:
**What type of PR is this?**

/kind test

Adds regression tests for the `use_gpu` device-visibility behaviour introduced in #767. Reviewers
on that PR asked for a test covering the `use_gpu: false` path on a multi-GPU host; these are those
tests, kept in a separate PR so #767 stays at +21/−3.

> **Depends on #767.** These tests assert the behaviour that #767 introduces, so they should land
> after it. On current `main` five of them fail by design — see the verification table below.

These are the first tests in the repository. `tests/` is the location `setup.py` already
anticipates:

```python
packages=find_packages(exclude=["tests", "*.tests", "*.tests.*", "tests.*"])
```

The exclusion has been there all along with nothing to exclude, so this takes the place the project
had already reserved rather than inventing a new convention.

### Why these particular tests

#765 reported two defects, and both are ordering-sensitive:

1. `use_gpu: false` was a silent no-op, so a run configured for CPU inherited whatever devices the
   host exposed.
2. `use_gpu: true` was applied *after* the algorithm module had been instantiated, so AI frameworks
   had already read `CUDA_VISIBLE_DEVICES` and fixed their device list before the restriction took
   effect.

A test that checks the environment variable *after* construction finishes would pass against the
original buggy code and prove nothing. So each case here asserts the value of `CUDA_VISIBLE_DEVICES`
**at the moment the algorithm module is instantiated**, using a stand-in module that records it
inside `get_module_instance()` — the point where, in a real run, `basemodel.py` is imported and a
framework picks its device.

Every case starts from `CUDA_VISIBLE_DEVICES=0,1,2,3` to represent the multi-GPU host that made the
original no-op visible.

| Test | Asserts |
|---|---|
| `test_use_gpu_false_hides_every_device_on_a_multi_gpu_host` | `-1` — the CPU-only run does not inherit host devices |
| `test_use_gpu_true_restricts_to_one_device_before_modules_load` | `0` — applied before the module loads, not after |
| `test_unset_use_gpu_leaves_an_existing_value_alone` | `0,1,2,3` — unchanged |
| `test_unset_use_gpu_does_not_introduce_the_variable` | the variable is not created where none existed |
| `test_set_gpu_visibility_maps_each_state` | the three states, independent of paradigm construction |

The two `unset` cases are invariants rather than regressions: 39 of the 41 `testenv.yaml` files
under `examples/` omit `use_gpu`, and their behaviour must not change.

### Verification

| Branch | Result |
|---|---|
| `main` (without #767) | **5 failed, 2 passed** |
| with #767 applied | **7 passed** |

The two that pass on `main` are the invariants above — correct, since #767 deliberately preserves
that path. The five failures are the three defects plus the two cases that reach
`_set_gpu_visibility`, which does not exist before #767.

### Running them

No new runtime dependencies; the suite needs only `pytest` on top of the environment CI already
builds:

```bash
pip install resources/third_party/*
pip install pytest
pytest tests/
```

### Relationship to #845

#845 proposes building a pytest suite for `core/` as Phase IV work — `pytest.ini`, a shared
`conftest.py`, CI integration, and unit tests across all paradigms.

This PR is deliberately not that. It adds the specific regression tests reviewers asked for on
#767, in the location `setup.py` already reserves, and adds no test infrastructure, no manifest
entry for pytest, and no workflow changes. If #845 lands a broader structure, these tests should
move under it — I would rather have the first cases in place and adapt them than block on the
framework.

### Notes

- `pylint` in CI analyses `core/` only, so these files are outside its scope.
- Wiring `pytest tests/` into a workflow is the natural follow-up. I left it out to keep this PR to
  one thing, and because #845 proposes owning that piece.

None — this PR closes nothing on its own. It adds test coverage for the behaviour fixed by #767,
which is the PR that resolves #765. Related: #765, #767, #845.

### DIFF
--- a/tests/core/test_paradigm_base.py
+++ b/tests/core/test_paradigm_base.py
@@ -1,0 +1,113 @@
+# Copyright 2026 The KubeEdge Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for gpu visibility handling in ParadigmBase.
+
+These cover the two defects reported in #765:
+
+1. `use_gpu: false` was a silent no-op, so a run configured for cpu inherited
+   whatever devices the host exposed.
+2. `use_gpu: true` was applied after the algorithm module had already been
+   instantiated, so AI frameworks had picked their device list before the
+   restriction took effect.
+
+Both are ordering-sensitive, so every case below asserts the value of
+CUDA_VISIBLE_DEVICES *at the moment the algorithm module is instantiated*
+rather than after construction has finished. Asserting afterwards would pass
+even against the original buggy code.
+"""
+
+import os
+
+import pytest
+
+from core.testcasecontroller.algorithm.paradigm.base import ParadigmBase
+
+
+MULTI_GPU_HOST = "0,1,2,3"
+
+
+class RecordingModule:
+    """Stands in for an algorithm module.
+
+    `_get_module_instances()` calls `get_module_instance()` on every configured
+    module; for a real run that is the point where basemodel.py is imported and
+    instantiated, and therefore the point where a framework reads
+    CUDA_VISIBLE_DEVICES. Recording the variable here is what makes these tests
+    sensitive to ordering.
+    """
+
+    def __init__(self):
+        self.visibility_at_instantiation = "not called"
+
+    def get_module_instance(self, module_type):
+        self.visibility_at_instantiation = os.environ.get("CUDA_VISIBLE_DEVICES")
+        return module_type
+
+
+def build_paradigm(tmp_path, **kwargs):
+    """Construct a ParadigmBase and report what the module saw."""
+    module = RecordingModule()
+    ParadigmBase(str(tmp_path), modules={"basemodel": module}, **kwargs)
+    return module.visibility_at_instantiation
+
+
+@pytest.fixture(autouse=True)
+def isolate_cuda_visible_devices(monkeypatch):
+    """Keep each case from leaking CUDA_VISIBLE_DEVICES into the next."""
+    monkeypatch.delenv("CUDA_VISIBLE_DEVICES", raising=False)
+
+
+def test_use_gpu_false_hides_every_device_on_a_multi_gpu_host(tmp_path, monkeypatch):
+    """The original no-op: cpu-only runs must not inherit the host's devices."""
+    monkeypatch.setenv("CUDA_VISIBLE_DEVICES", MULTI_GPU_HOST)
+
+    assert build_paradigm(tmp_path, use_gpu=False) == "-1"
+
+
+def test_use_gpu_true_restricts_to_one_device_before_modules_load(tmp_path, monkeypatch):
+    """The original ordering defect: the module saw all four devices."""
+    monkeypatch.setenv("CUDA_VISIBLE_DEVICES", MULTI_GPU_HOST)
+
+    assert build_paradigm(tmp_path, use_gpu=True) == "0"
+
+
+def test_unset_use_gpu_leaves_an_existing_value_alone(tmp_path, monkeypatch):
+    """Most testenv.yaml files omit the key; their behaviour must not change."""
+    monkeypatch.setenv("CUDA_VISIBLE_DEVICES", MULTI_GPU_HOST)
+
+    assert build_paradigm(tmp_path) == MULTI_GPU_HOST
+
+
+def test_unset_use_gpu_does_not_introduce_the_variable(tmp_path):
+    """Omitting the key must not create CUDA_VISIBLE_DEVICES where none existed."""
+    assert build_paradigm(tmp_path) is None
+    assert "CUDA_VISIBLE_DEVICES" not in os.environ
+
+
+@pytest.mark.parametrize(
+    "use_gpu, expected",
+    [
+        (False, "-1"),
+        (True, "0"),
+        (None, MULTI_GPU_HOST),
+    ],
+)
+def test_set_gpu_visibility_maps_each_state(monkeypatch, use_gpu, expected):
+    """The three states of the setting, independent of paradigm construction."""
+    monkeypatch.setenv("CUDA_VISIBLE_DEVICES", MULTI_GPU_HOST)
+
+    ParadigmBase._set_gpu_visibility(use_gpu)  # pylint: disable=protected-access
+
+    assert os.environ["CUDA_VISIBLE_DEVICES"] == expected
--- a/tests/core/test_paradigm_base.py
+++ b/tests/core/test_paradigm_base.py
@@ -111,3 +111,28 @@ def test_set_gpu_visibility_maps_each_state(monkeypatch, use_gpu, expected):
     ParadigmBase._set_gpu_visibility(use_gpu)  # pylint: disable=protected-access
 
     assert os.environ["CUDA_VISIBLE_DEVICES"] == expected
+
+
+def test_each_construction_applies_the_setting_rather_than_inheriting_it(tmp_path, monkeypatch):
+    """A job builds one paradigm per algorithm, all from a single TestEnv.
+
+    `TestCaseController.build_testcases()` passes the same `test_env` object to
+    every `TestCase`, so cases in one run share a `use_gpu` value. What matters
+    across constructions is therefore that each one applies that value itself
+    instead of trusting whatever the previous construction left in the
+    environment. Raised in review of #890.
+    """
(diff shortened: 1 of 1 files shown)


rahulnr107:
## Review — PR #890

### Summary

PR #890 adds regression tests for the `use_gpu` behavior addressed by PR #767 and Issue #765.

The tests are implemented in:

```text
tests/core/test_paradigm_base.py
```

### Test Coverage

The PR covers the three states of the configuration:

```text
use_gpu = False
use_gpu = True
use_gpu = None
```

It verifies that:

```text
False → CUDA_VISIBLE_DEVICES = "-1"
True  → CUDA_VISIBLE_DEVICES = "0"
None  → existing environment remains unchanged
```

The tests also cover the case where the variable was previously absent.

### Strong Point

The most important aspect of the tests is that they record `CUDA_VISIBLE_DEVICES` at the moment the algorithm module is instantiated.

This is important because the original defect was ordering-sensitive. A test that checked the environment only after construction could pass even if the implementation still configured GPU visibility too late.

The `RecordingModule` approach therefore tests the relevant behavior at the correct boundary without requiring a real GPU.

### Assessment

The tests provide meaningful regression coverage for the Core behavior introduced by PR #767.

### Recommendation

**Accepted as it is.**

### Verification Boundary

The tests use a controlled `RecordingModule` rather than a real AI framework or physical GPU. Therefore they verify the Core ordering and environment-variable contract, but they do not by themselves establish end-to-end GPU behavior with every supported framework.

A separate integration test could provide that additional confidence if required.

### Relationship to PR #767

PR #767 provides the implementation change.

PR #890 provides regression protection for that behavior.

The two PRs are therefore complementary rather than duplicated.

vjuhhii:
Thanks for reviewing, and for stating the verification boundary explicitly — that is exactly the right way to read these. They establish the Core ordering and environment contract; they do not substitute for running a real framework on real hardware.

An integration test would need a GPU runner, which the project does not have today. #888 is my attempt at the next best thing: measure the hardware-dependent behaviour directly and report it, rather than assume it.

Hiteshsai007:
These tests are structured exactly right: asserting `CUDA_VISIBLE_DEVICES` **at the moment the
algorithm module is instantiated** (rather than after construction finishes) is what makes the
ordering-sensitive cases meaningful, and the per-test `monkeypatch` isolation is clean.

Two gaps are worth covering — both involve **multiple constructions in one process**, which is how
ianvs actually runs (one paradigm per TestCase, sequentially):

1. **Cross-TestCase carry-over.** Every case here builds a single `ParadigmBase`. On `pull/767`,
   constructing with `use_gpu=False` sets `CUDA_VISIBLE_DEVICES="-1"` and never restores it, so the
   *next* construction that omits `use_gpu` instantiates its module with `"-1"` still in the
   environment rather than the host value. I traced this with the real `SingleTaskLearning`
   (recording the env at module load across two constructions): after a `use_gpu=False` case, an
   omitted-`use_gpu` case sees `"-1"`. A test shaped like this would lock it down:

   ```python
   monkeypatch.setenv("CUDA_VISIBLE_DEVICES", MULTI_GPU_HOST)
   build_paradigm(tmp_path, use_gpu=False)          # sets "-1", never restored on #767
   assert build_paradigm(tmp_path) == MULTI_GPU_HOST  # currently FAILS (sees "-1")
   ```

2. **Exception path.** If `_get_module_instances()` raises after the visibility value is set,
   there is no `finally`/restore, so the mask persists into the next case. A test that makes the
   module raise mid-construction and then asserts the prior environment is restored would cover it.

These failures are #767's behaviour rather than this PR's, but this test file is the natural place
to pin the **whole** contract (single + sequential + exception) once #767 restores state — otherwise
the regression suite goes green while the two real multi-TestCase paths stay open.

vjuhhii:
Thanks — I looked into both, and they land differently.

Sequential construction: added, though the reachable version is narrower than described. The scenario of one case setting -1 and a later case omitting use_gpu cannot arise through ianvs configuration: TestCaseController.build_testcases() passes the same test_env object to every TestCase it appends, so all cases in a run share one use_gpu value. They cannot disagree about it.

What does matter across constructions is that each one applies the setting itself rather than trusting what the previous one left behind, and that was genuinely untested. Added in 1295573: one case asserts the value is re-applied after something else has restored the host's devices in between, and one covers the gpu path over repeated construction.

Environment restoration on failure: real, but currently unreachable. There is no try/finally around the visibility change, so a failure mid-construction does leave the mask in place. But run_testcases() raises RuntimeError on the first failing case, which aborts the job — there is no later case to be affected. It would become a genuine defect the moment ianvs grows a continue-on-failure mode, so it is worth having on the record; I would rather not add cleanup code for a path that cannot be exercised today, and would defer to maintainers if they see it differently.

vjuhhii:
Correction to something I claimed in the description: I wrote that these are the
first tests in the repository. That is true of `main`, but #648 has been open
since 4 August and also creates `tests/`, with `tests/test_metrics.py`. It got
there first and I should have found it before writing that line.

No file conflict — different filenames — but the two take different approaches to
imports. #648 does `sys.path.insert(0, ...)` at the top of the test module, while
these run from the repository root without it. If both land, that is worth
reconciling, and I am happy to adopt whichever convention maintainers prefer or
to rebase onto #648 once it merges.

Incidentally, #648 also fixes the unbalanced return in `compute()` that made
`bwt_func` raise `ValueError` instead of returning NaN. I ran into that
separately while reading the lifelong learning paradigm; good to see it already
covered there.

PR #824, opened 2026-08-17
fix(yaoba): restore paradigm dispatch and remove hard-coded config pathsFix/yaoba restoration

31groot:
/kind bug

**What this PR does / why we need it**:

This PR restores the Yaoba `singletask_learning_tta` and
`singletask_learning_acboost` examples by reconnecting their custom
paradigm types to the Ianvs algorithm dispatch layer and removing
machine-specific paths from the active-learning boost configuration.

Specifically:

- Registers `singletasklearning_tta` and
  `singletasklearning_acboost` in `ParadigmType`.
- Wires `SingleTaskLearningTTA` and `SingleTaskLearningACBoost` into
  the singletask-learning paradigm exports.
- Adds dispatch branches in `Algorithm.paradigm()` for both Yaoba
  paradigm types.
- Moves the Boost example's `requirements.txt` to the example root.
- Replaces hard-coded Yaoba dataset/config paths in
  `testenv.yaml` and the Boost model configurations with
  project-relative paths.
- Uses the local Yaoba JSON files and dataset image directory expected
  by the example.

Before these changes, the custom Yaoba `paradigm_type` values were not
recognized by the framework, causing `Algorithm.paradigm()` to return
`None` and eventually producing:

    AttributeError: 'NoneType' object has no attribute 'run'

This PR addresses that framework-level dispatch problem and restores
the example configuration to use portable project-relative paths.

The remaining dataset availability and end-to-end execution work is
tracked as part of the broader restoration described in #823.

**Which issue(s) this PR fixes**:

Related to #823

Issue #823, opened 2026-08-17
Yaoba Example Restoration: Paradigm Dispatch and Configuration Fixes


31groot:
**LFX Mentorship Fall 2026 (Term 3), Comprehensive Example Restoration for KubeEdge Ianvs: Phase IV**
**Related:** #230

---

## 1. Selected Example

`examples/yaoba/` - two sub-examples:

* `singletask_learning_yolox_tta` (test-time augmentation for defect detection)
* `singletask_learning_boost` (active-learning instance-boosting for defect detection)

I selected `yaoba` because, unlike several other examples restored in prior Phase IV pre-tasks (`PCB-AoI`, `cityscapes-synthia` variants, `MOT17`, `cloud-edge-collaborative-inference-for-llm`, `robot-cityscapes-synthia`), it has not yet been picked up by any previous LFX pre-task submission.

My investigation shows it is currently **completely non-functional**, and more importantly, non-functional in a way that current documentation gives no hint of.

---

## 2. Investigation

I cloned `kubeedge/ianvs` and installed the core package (`pip install -e .`) to trace the failure from first principles rather than relying on the example's own README.

### Bug 1 (Root Cause, Both Sub-examples): Unregistered `paradigm_type` values

Both `algorithm.yaml` files declare non-standard paradigm types:

* `singletask_learning_yolox_tta/testalgorithms/algorithm.yaml` -> `paradigm_type: "singletasklearning_tta"`
* `singletask_learning_boost/testalgorithms/algorithm.yaml` -> `paradigm_type: "singletasklearning_acboost"`

`core/common/constant.py`'s `ParadigmType` enum only defines: `singletasklearning`, `incrementallearning`, `multiedgeinference`, `lifelonglearning`, `federatedlearning`, `federatedclassincrementallearning`, `jointinference`.

Neither yaoba value is a member.

In `core/testcasecontroller/algorithm/algorithm.py`, `Algorithm.paradigm()` string-matches `self.paradigm_type` against this enum to decide which `Paradigm` subclass to instantiate, and falls through to `return None` when nothing matches, silently, with no error or warning.

That `None` is then handed to `core/testcasecontroller/testcase/testcase.py`, which calls `paradigm.run()` on it unconditionally.

The result is:

```text
AttributeError: 'NoneType' object has no attribute 'run'
```

which is subsequently wrapped by the generic pipeline error handler as:

```text
[...]

PR #891, opened 2026-08-23
fix(core): declare the dataset fields that core paradigms read

akshita317:
**What type of PR is this?**

/kind bug

`Dataset._parse_config` (`core/testenvmanager/dataset/dataset.py:71-73`) assigns only keys that already exist on the instance:

```python
    if attr in self.__dict__:
        self.__dict__[attr] = value
```

Four fields are read off the `Dataset` instance by core paradigm code, but are not declared in `Dataset.__init__`. Any value a `testenv.yaml` supplies for them is therefore discarded without warning, and the later attribute access raises:

| field | read by |
|---|---|
| `val_url` | `core/.../singletask_learning_tta.py:77` |
| `known_dataset_url` | `core/.../singletask_learning_active_boost.py:61` |
| `unknown_dataset_url` | `core/.../singletask_learning_active_boost.py:62` |
| `image_folder_url` | `core/.../singletask_learning_active_boost.py:49,63` and `singletask_learning_tta.py:46,76` |

Reproduced on `main` against the real classes, using `examples/yaoba/singletask_learning_boost/testenv/testenv.yaml` unmodified:

```
declared in YAML : ['image_folder_url', 'known_dataset_url', 'test_url',
                    'train_url', 'unknown_dataset_url', 'val_url']
present on object: ['label', 'test_data', 'test_data_info', 'test_index',
                    'test_url', 'train_data', 'train_data_info',
                    'train_index', 'train_url']

dataset.image_folder_url    -> AttributeError: 'Dataset' object has no attribute 'image_folder_url'
dataset.known_dataset_url   -> AttributeError: 'Dataset' object has no attribute 'known_dataset_url'
dataset.unknown_dataset_url -> AttributeError: 'Dataset' object has no attribute 'unknown_dataset_url'
```

`git log --all -S image_folder_url -- core/testenvmanager/dataset/dataset.py` returns nothing, so these were never declared. The paradigms that read them and the example that configures them both arrived in `7111198`, without the matching registration.

**The change** is four attribute declarations defaulting to `""`, consistent with the surrounding fields. It is purely additive — examples that do not set these keys keep the same empty-string default and behave exactly as before.

**Verification**

| check | result |
|---|---|
| the attribute reads above | resolve to the configured paths (3/3 → 0/3 failing) |
| `Dataset` constructed for every `examples/**/testenv*.yaml` | 48 ok, 0 failed |
| `Rank` constructed for every `rank:` block | 43 ok, 0 failed |
| `pylint --max-positional-arguments=10 core/` | 10.00/10, unchanged from `main` |

*Not verified:* a full `ianvs -f` run of `examples/yaoba`. That example pins `mmcv-full 1.7.1` and `mmdet 2.28.2` and needs the YaoBa image corpus, which has no public download, so I could not execute it end to end and am not claiming it passes.

Related to #742 and #823.

This does **not** on its own make `examples/yaoba` run — the paradigm dispatch registration in #824 (@31groot) is also needed, and the two changes are independent and do not conflict.

What it does do is make the path corrections in #796 (@hibaa23) and in #824 effective. Both correct `image_folder_url` to a repo-relative path, and at present that value cannot reach core because the key is dropped here. @hibaa23 confirmed on #796 that they are happy to defer the config-line edit to whichever PR carries the core registration; this PR deliberately touches **no** example file, so it should not conflict with either.

@Nikhil-VS1811, who is assigned to #742, and I split the work on that issue: this registration is the piece I picked up, and they are looking at the broader silent-drop pattern across the other example families.

Scope is deliberately narrow. Auditing all 48 `examples/**/testenv*.yaml` against the fields `Dataset.__init__` declares, 12 configs declare at least one dropped key, but cross-checking every `self.dataset.<attr>` read in `core/` against those declarations closes the set at exactly these four:

```
undeclared but read: ['image_folder_url', 'known_dataset_url', 'unknown_dataset_url', 'val_url']
```

The dropped keys in the other 10 configs (`cifar100`, `federated-llm`, `PIPL`, `industrialEI`) have no consumer anywhere in `core/` or `examples/`, so they are dead config or human-readable metadata rather than this defect, and registering them would be speculative. Full audit and script in [#742](https://github.com/kubeedge/ianvs/issues/742#issuecomment-5387742210).

```
Dataset configuration keys read by the single-task learning TTA and active-boost paradigms (val_url, known_dataset_url, unknown_dataset_url, image_folder_url) are now honoured instead of being silently discarded.
```


akshita317:
/assign @MooreZheng
/cc @hsj576

Small Core change (10 lines, additive) unblocking the yaoba path fixes in #796 and #824. Verified against all 48 example testenv configs and all 43 rank blocks with no regression; details in the description.

Shikhar-Kesharwani:
### Review — supplementary configuration-validation finding

The four missing `Dataset` declarations are correctly restored, and I agree that this fixes the current silent-drop/`AttributeError` failure.

I found one additional edge case introduced by the new declarations: the newly added fields default to an empty string, but `Dataset._check_fields()` does not validate these fields after parsing.

This means a `testenv.yaml` can now omit `image_folder_url`, `known_dataset_url`, `unknown_dataset_url`, or `val_url`, and `Dataset` construction will still succeed with the corresponding attribute set to `""`. That is different from the pre-PR failure, where a configured field was silently discarded and later raised `AttributeError`.

These fields are not merely metadata: the TTA path uses `image_folder_url` to construct the actual inference image paths, while Active Boost consumes `known_dataset_url`, `unknown_dataset_url`, and `image_folder_url` as training/inference inputs. An omitted value can therefore defer a configuration error into a later filesystem/model-processing failure, or produce incorrect relative paths, rather than failing at configuration time.

I recommend either validating these fields when the corresponding paradigm requires them, or explicitly documenting them as optional and adding the appropriate downstream guards. At minimum, a regression test should cover a configuration that omits each newly registered field and verify that the resulting behavior is intentional and fails with a clear configuration error when the field is required.

This is separate from the issue fixed by this PR: the current change ensures configured values are no longer discarded; the remaining concern is that missing values are now silently accepted as empty strings.

PR #896, opened 2026-08-23
fix(simulation): make host resource checks cgroup aware

### LINKED ISSUE #895: Host resource checks read the host, not the container: a 1 CPU / 512 MB container passes the 4 core / 4 GB gate
**What happened**:

`check_host_memory()` and `check_host_cpu()` in `core/testcasecontroller/simulation_system_admin/simulation_system_admin.py` are meant to stop a simulation build on a machine with less than 4 GB free memory or fewer than 4 CPU cores. They read those numbers from `lscpu` (line 113) and `/proc/meminfo` (line 83).

Neither of those is namespaced by cgroups. A container gets its own filesystem, network and process list, but `/proc/meminfo` is still answered by the host kernel, and `lscpu` reads `/sys/devices/system/cpu`, which is also the host's. So inside a container both report the whole machine rather than the slice the container is allowed to use.

The result is that a container holding 1 CPU and 512 MB passes a check that says it requires 4 cores and 4 GB.

I ran the two probe functions inside a container limited to exactly that, alongside every replacement that has been proposed for them so far:

```console
$ docker run --rm --cpus 1.0 --memory 512m ... python /w/run2.py
PR #656 fixed CPU parser       -> 12 cores      (gate needs >=4) => PASS
os.cpu_count()                 -> 12 cores      (gate needs >=4) => PASS
sched_getaffinity              -> 12 cores      (gate needs >=4) => PASS
MemAvailable                   -> 6416236 kB (6.12 GB) (gate needs >=4GB) => PASS

ground truth from the cgroup:
cpu.max    -> 100000/100000 = 1.00 CPU
memory.max -> 536870912 bytes = 512 MB
```

All four pass. They differ in how the host number is obtained, not in which machine is measured, so none of them changes the outcome.

The check should measure what the process is actually allowed to use, so a container with 1 CPU and 512 MB is rejected before a cluster build starts, and a bare metal host with 12 cores and 6 GB still passes.

On any Linux host with more than 4 cores and more than 4 GB free, run the repository's own probe functions inside a smaller container. `simulation_system_admin.py` is copied unmodified, with only `core.common.log` stubbed:

```bash
docker run --rm --cpus 1.0 --memory 512m -v /path/to/repro:/w:ro -w /w python:3.11-slim \
  sh -c 'apt-get -qq update >/dev/null && apt-get -qq install -y util-linux >/dev/null && python run.py'
```

The shorter version, without any Ianvs code, shows the same thing:

```console
$ docker run --rm --cpus 1.0 --memory 512m ubuntu:24.04 bash -c \
    'lscpu | grep "^CPU(s):"; grep MemAvailable /proc/meminfo; echo ---; cat /sys/fs/cgroup/cpu.max; cat /sys/fs/cgroup/memory.max'
CPU(s):                               12
MemAvailable:    6413004 kB
---
100000 100000
536870912
```

The first two lines are what the check reads. The last two are what the container actually has.

Why this matters for Ianvs specifically rather than as a general observation.

`sandbox-engine.md` section 6.6 states that `check_host_enviroment()` "verifies Docker daemon reachability, kind availability, ≥4 CPU cores, and ≥4 GB free memory before cluster provisioning begins", and names "a laptop or CI host" as the target. Container based CI is the normal case: GitHub Actions container jobs, GitLab CI with the Docker executor and Codespaces all run inside a resource limited container. In those environments that stated guarantee does not hold.

Exceeding a cgroup memory limit is also not a graceful failure. The kernel OOM killer terminates the process, so the check that exists to prevent an underpowered run is inert in exactly the situation where the failure is least readable.

There is a second order point worth flagging. The Simulation Sandbox work is about running test cases inside resource bounded containers, so once that lands these checks would be running inside a boundary they cannot currently see.

Proposed direction: read the cgroup v2 files (`cpu.max`, `memory.max`, `memory.current`) first, fall back to the cgroup v1 files, and fall back to the host figure only when no limit applies, so behaviour on bare metal is unchanged. I have a patch that does this and verifies in three environments (bare host still passes, a 1 CPU / 512 MB container is now rejected, an unlimited container still passes, which matters because a careless fix rejects that one too). It keeps `pylint --max-positional-arguments=10` at 10.00/10. I will open a PR against this issue.

Related but separate: PR #656 fixes the `lscpu | grep CPU:` parse failure in the same function, and PR #820 replaces this module entirely. This is a different problem from either, since it concerns which machine is measured rather than how the value is parsed, and PR #820's replacement reads `os.cpu_count()` and CPU affinity, both of which report host values as shown above.

Found while working through the Simulation Sandbox pre-test in #348. Environment: Ubuntu 24.04.3 LTS on WSL2, kernel 6.6.87.2, Docker 28.3.0, cgroup v2 unified hierarchy, Ianvs `main` at `37a9c60`.

RudraDudhat2509:
**What type of PR is this?**

/kind bug

`check_host_memory()` and `check_host_cpu()` are supposed to stop a simulation build on a machine that has less than 4 GB free or fewer than 4 cores. They read those numbers from `lscpu` (line 113) and `/proc/meminfo` (line 83), and neither of those is namespaced by cgroups. In a container both report the host, not the container's limit, so a container holding 1 CPU and 512 MB passes the check.

This is a different problem from the parsing bugs already reported on the same two functions. Those are about how the host number is read. This is about which machine is measured, so it survives all of them. Inside a container limited to exactly 1 CPU and 512 MB:

| what reads the value | reports | gate needs | result |
|---|---|---|---|
| PR #656's fixed `lscpu` parser | 12 cores | >= 4 | PASS |
| `os.cpu_count()` | 12 cores | >= 4 | PASS |
| `sched_getaffinity` (used by PR #820) | 12 cores | >= 4 | PASS |
| `MemAvailable` | 6.12 GB | >= 4 GB | PASS |
| **`/sys/fs/cgroup/cpu.max`** | **1.00 CPU** | | |
| **`/sys/fs/cgroup/memory.max`** | **512 MB** | | |

The fix reads the cgroup v2 files first, then the cgroup v1 files, and falls back to the host figure only when no limit applies, so bare metal behaviour does not change.

It also swaps `MemFree` for `MemAvailable` on that fallback path. `MemFree` excludes reclaimable page cache and understates what is usable on any warm host, and leaving it would have left a second wrong answer inside a function this PR is already touching. That half was suggested by others on #348 and is not my finding.

**Why this matters here rather than as a general observation**

`sandbox-engine.md` section 6.6 says `check_host_enviroment()` "verifies Docker daemon reachability, kind availability, >=4 CPU cores, and >=4 GB free memory before cluster provisioning begins", and names "a laptop or CI host" as the target. Container based CI is the normal case, so that stated guarantee does not currently hold there. Going over a cgroup memory limit is also not a graceful failure, the kernel OOM killer terminates the process, so the check is inert exactly where the failure is hardest to read.

**Testing**

Three environments, because a fix that only reads the cgroup files breaks the third one. In an unlimited container `cpu.max` reads `max`, and treating that as a number rejects a container that should pass.

| environment | cpus | free mem | check_host_cpu | check_host_memory |
|---|---|---|---|---|
| bare host, no limit | 12 | 7041956 kB | PASSED | PASSED |
| container, `--cpus 1.0 --memory 512m` | 1 | 502128 kB | REJECTED | REJECTED |
| container, no limits | 12 | 7033888 kB | PASSED | PASSED |

The module under test is the file from this branch, with only `core.common.log` stubbed.

`pylint --max-positional-arguments=10 core/testcasecontroller/simulation_system_admin/` reports 10.00/10, the same as before this change.

**Notes for review**

- `cpu.max` can express a fraction, so `--cpus 0.5` floors to 1 rather than 0. A request below one core should be rejected by the gate rather than crash it.
- `_read_first_line()` returns `None` instead of raising, so a host with no cgroup hierarchy, for example macOS, falls through to the existing behaviour.
- PR #820 replaces this module wholesale. If that lands this change is moot, but its replacement reads `os.cpu_count()` and CPU affinity, so the same issue would carry over and is worth fixing there instead.

Fixes #895


31groot:
### Request changes :  no-cgroup CPU fallback is not portable

This is a separate portability issue from the cgroup handling addressed by this PR.

The PR documentation says that when no cgroup hierarchy applies, the implementation falls through to the existing host behaviour, and specifically notes macOS as an example.

However, `get_host_number_of_cpus()` now ends with:

```python
return len(os.sched_getaffinity(0))
```

when neither cgroup v2 nor cgroup v1 provides a CPU quota.

That fallback is not portable to platforms where `os.sched_getaffinity()` is unavailable.

I reproduced the exact PR head `81b1578` by simulating:

1. no available cgroup CPU hierarchy; and
2. a platform without `os.sched_getaffinity`.

The result was:

```text
FAILED
AttributeError: module 'os' has no attribute 'sched_getaffinity'
```

So the current fallback is effectively:

```text
no cgroup CPU limit
        ↓
os.sched_getaffinity(0)
        ↓
AttributeError on platforms without that API
```

This is especially relevant because the PR's own description treats the no-cgroup case as a supported fallback rather than as an unsupported platform.

Please guard the affinity-based fallback with capability detection and retain a platform-independent CPU-count fallback for systems without `sched_getaffinity` (for example, the previous host CPU-count path).

A regression test should cover the combination of:

* no cgroup quota;
* no `os.sched_getaffinity`;
* successful CPU-count fallback.

**Verification boundary:** reproduced the exact fallback failure against PR head `81b1578` with the cgroup reader disabled and `os.sched_getaffinity` unavailable. No real macOS host or simulation cluster was required.

RudraDudhat2509:
fixed in df347ec. guarded with hasattr + os.cpu_count() fallback, verified on windows since it has no sched_getaffinity either.

core/ has no test setup (no tests dir, no pytest, CI is just pylint) so a regression test wouldn't run anywhere. thx for the find

PR #908, opened 2026-08-24
docs: fix §6.3 config contract YAML — correct block key and leaf key names

PragnyaKhandelwal:
/kind documentation

Fixes a documentation error in §6.3 "Configuration contract" of the Simulation Sandbox proposal.

The YAML block in §6.3 used `cluster_simulation:` as the block key with `cloud_nodes:`/`edge_nodes:` as leaf keys. However, `BenchmarkingJob._parse_config()` dispatches only on the key `simulation:` (line 107 of benchmarkingjob.py), so `cluster_simulation:` silently produces `self.simulation = None` and `build_simulation_environment()` is never called. Additionally, `cloud_nodes` and `edge_nodes` are not attributes of the `Simulation` class — the correct keys are `cloud_number` and `edge_number`.

This was identified by Yung-Cheng Chen (#881) and verified against the current codebase.

Related to #348

### DIFF
--- a/docs/proposals/simulation/sandbox-engine/sandbox-engine.md
+++ b/docs/proposals/simulation/sandbox-engine/sandbox-engine.md
@@ -133,6 +133,10 @@ When `sandbox_profile` is absent, execution falls through to `testcase.run(works
 
 ### 6.3 Configuration contract (opt-in)
 
+The cluster simulation is configured under the existing `simulation:` key,
+parsed by `BenchmarkingJob._parse_simulation_config()` into `self.simulation`.
+The `sandbox_profile:` key (new, this proposal) coexists with it:
+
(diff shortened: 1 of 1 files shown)


MooreZheng (commented):
As discussed, the code will pull Sedna Docker images tagged as latest which will not use the sedna wheel that other contributors is working on

MooreZheng (changes requested):
Conclusion on the meeting: pending for the whole picture of proposal revision

MooreZheng (changes requested):
It is already close to the final version. As discussed, 
1. Add more metrics on GPU and hard-disk stroage for AI data training;
2. Add plans for documentations, especially Readme.md, LFX and Kubeedge blog.

PR #942, opened 2026-08-26
fix(examples): correct broken path references in Cloud_Robotics singletask learning example

azaynul10:
### What type of PR is this?

/kind bug

### What this PR does / why we need it

Every intra-repository path reference in
`examples/Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation/` points at a
directory that does not exist. The example cannot be run as shipped, and the run command
in its own `Readme.md` names a fourth nonexistent path.

The directory is `Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation`. It is
spelled four different wrong ways across six references in four files, and two of those
also misspell the sub-path (`single_task_learing`, `sinlge_task_learning`):

| File | Line | Declared | Resolves |
|---|---|---|---|
| `benchmarkingjob.yaml` | 9 | `./examples/cloud_robot/single_task_learning/testenv/testenv.yaml` | no |
| `benchmarkingjob.yaml` | 22 | `./examples/cloud-robotics/single_task_learning/testalgorithm/rfnet/algorithm.yaml` | no |
| `testenv/testenv.yaml` | 16 | `./examples/cloud-robotics/single_task_learing/testenv/accuracy.py` | no |
| `testenv/testenv.yaml` | 31 | `./examples/cloud_robotics/sinlge_task_learning/testenv/accuracy.py` | no |
| `testalgorithms/algorithm.yaml` | 8 | `./examples/cloudrobotics/singletask_learning_bench/testalgorithms/rfnet/basemodel.py` | no |
| `Readme.md` | 133 | `examples/cloud-robotics/single_task_learning/semantic-segmentation/benchmarkingjob.yaml` | no |

Each target exists exactly once inside the example's own directory, so every correction is
unambiguous. The complete file list is:

```
examples/Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation/
├── Readme.md
├── benchmarkingjob.yaml
├── testalgorithms/
│   ├── algorithm.yaml
│   └── basemodel.py
└── testenv/
    ├── accuracy.py
    └── testenv.yaml
```

This is the failure mode described in #230: an example that stopped matching the tree
around it and was never caught, because nothing runs it.

### Verification

Paths are resolved by `core.common.utils.is_local_file` with the working directory set to
the repository root, which is how Ianvs resolves a config path. Before this change, 6 of 6
references fail to resolve. After it, 6 of 6 resolve.

```
$ python - <<'PY'
from core.common import utils
for p in [
  "./examples/Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation/testenv/testenv.yaml",
  "./examples/Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation/testalgorithms/algorithm.yaml",
  "./examples/Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation/testenv/accuracy.py",
  "./examples/Cloud_Robotics/singletask_learning_bench/Semantic_Segmentation/testalgorithms/basemodel.py",
]:
    print(utils.is_local_file(p), p)
PY
```

No behaviour outside this example is affected. The change is confined to string values in
three YAML files and one line in a Readme.

### Two blockers this uncovers, deliberately not fixed here

Correcting the paths lets the config parse further than it previously could, which exposes
two problems that need a maintainer's decision rather than my guess. I would rather name
them than pick a value and have it be wrong.

**1. The dataset paths point at a contributor's home directory.**

```yaml
# testenv/testenv.yaml:5-7
train_url: "/home/aryan/ianvs/dataset/train-index.txt"
test_url:  "/home/aryan/ianvs/dataset/test-index.txt"
```

`Readme.md:117` says the user sets this themselves, so an absolute path here is expected by
design, but shipping one that names a specific machine is not. `Readme.md:27-28` and
`:112-114` describe a layout that would make `/ianvs/dataset/` the value consistent with
this example's own instructions. Happy to change it to that, or to whatever placeholder
convention the maintainers prefer, in this PR or a follow-up.

**2. There is no `rank:` block, so the job crashes after the run completes.**

This config has no `benchmarkingjob.rank` section at all. Its `selected_dataitem` and
`save_mode` keys sit under `test_object:` instead. Of the 48 configs under `examples/`
carrying a top-level `benchmarkingjob:` section, this is the only one without a `rank:`
block.

The consequence is not a config error. `BenchmarkingJob.__init__` sets `self.rank = None`
(`core/cmd/obj/benchmarkingjob.py:47`), `_parse_rank_config` only fires on a literal `rank`
key (`:105-106`), and `_check_fields` (`:53-74`) validates `name`, `workspace` and
`test_object` but never `rank`. So validation passes, the benchmark runs to completion, and
then:

```python
# core/cmd/obj/benchmarkingjob.py:96-98
if test_results:
    self.rank.save(succeed_testcases, test_results, output_dir=self.workspace)
```

`AttributeError: 'NoneType' object has no attribute 'save'`, after the compute is spent.

Adding a `rank:` block is not a typo fix: `Rank._check_fields` requires `sort_by`
(`core/storymanager/rank/rank.py:64`), and choosing a sort key for this example is a design
decision. I can open a follow-up PR with a proposed block, or file it as an issue, whichever
the maintainers prefer.

### Which issue(s) this PR fixes

Part of #230.

### Relationship to existing work

I checked the tracker before opening this. #744 covers validator work on path breakage
generally; this is a concrete instance in one example rather than a change to the validator.
The missing `rank:` block relates to the `_parse_config` silent-discard class discussed on
#829 and #828, but is a distinct mechanism: here the key is absent entirely rather than
misindented, so no `Rank` is ever constructed and value validation cannot help.


pratikpanda2006:
**Supplementary review:** Fixes relative path references in the Cloud_Robotics single-task-learning benchmarking YAML. The patch corrects the immediate broken-path symptom but hardcodes paths as static strings rather than resolving them relative to a detected project root — this will break again if the benchmark script is invoked from outside the repo root. Suggest using `os.path.dirname(__file__)`-relative resolution instead of literal relative strings for durability.

azaynul10:
Thanks for looking at it @pratikpanda2006  . I think the suggestion is aimed at a different layer than this PR touches.

The diff is three YAML files and one Markdown file, with no Python in it:
```
Readme.md                      | 2 +-
benchmarkingjob.yaml           | 4 ++--
testalgorithms/algorithm.yaml  | 2 +-
testenv/testenv.yaml           | 4 ++--
```
os.path.dirname(__file__) is a Python expression, so it is not available inside a YAML value where these strings live.

On the underlying point: repo-root-relative strings are Ianvs's existing convention for example configs rather than a choice this PR makes. 123 of the 151 YAML files under examples/ reference paths as ./examples/.... This PR corrects six paths that pointed at directories which do not exist, moving the example from "cannot load" to "loads", using the same convention as the rest of the tree.

Your observation that configs assume invocation from the repo root is accurate, and it applies repo-wide rather than to this example. Changing it would mean changing how Core resolves config-relative paths across all of those files, which seems worth its own issue rather than folding into an example path fix. Happy to see it raised separately.

pratikpanda2006:
Thanks for clarifying, @azaynul10  sir  That makes complete sense—I missed that the diff was scoped strictly to the .yaml config values rather than the Python loader logic.

You're right that keeping it consistent with the existing repo-root relative path convention across the 123+ other example configs is the right call for this PR. I'll open a separate issue to track general path-resolution robustness in the core framework controller.

LGTM for this fix!

PR #957, opened 2026-08-27
[PHASE IV] feat(core): device capability contract for bfloat16 and declared floors

vjuhhii:
**What type of PR is this?**

/kind bug

Implements items A, B, C, D and E from #888. Item F (the CI check) is deliberately left out and
offered in that thread.

`ParadigmBase` decides device *visibility* from `use_gpu` (#767). That answers whether a GPU is
used. It does not answer whether the GPU can execute what an algorithm is about to ask of it — a
separate question with a worse failure mode. Below compute capability 8.0 PyTorch **emulates**
`bfloat16` rather than refusing it, so a benchmark completes and reports timings produced by
emulated arithmetic. Measured on a GTX 1080 (CC 6.1), that path is 1.85× slower than float32 in a
4096² matmul, and nothing in the output says so.

Three commits, reviewable in order:

| Commit | Item |
|---|---|
| `feat(core): add device capability contract and guard bfloat16 selection` | A, C, D |
| `docs(examples): document the compute capability floors these examples impose` | E |
| `feat(core): let testenv.yaml declare a minimum compute capability` | B |

### A — `core/common/device.py`

`device_profile()` reports the resolved device and what it natively executes; `supports_bf16()`
answers the question the example sites need.

**It lives in `core/common/`, not `core/testenvmanager/`.** #888 sketched the latter. `core/common/`
is right because examples already import from it — `from core.common.log import LOGGER` appears in
`huggingface_llm.py`, the same directory as one of the sites fixed here. No new plumbing needed.

**torch is imported lazily.** `core` does not depend on torch and `requirements.txt` does not list
it, so a module-level import would break `import core` for anyone without it, including the CI lint
job. A checkout without torch gets the cpu profile rather than an `ImportError`.

**`bf16_native` comes from the capability major version, not `torch.cuda.is_bf16_supported()`.**
That function defaults to `including_emulation=True` and returns `True` on hardware with no native
`bfloat16` — verified on a GTX 1080, where it reports `True` while `including_emulation=False`
reports `False`. Using the obvious guard would have reproduced the bug inside the fix.

### C — the three sites

```python
# eagle_llm.py
torch_dtype=torch.bfloat16 if supports_bf16() else torch.float32,

# block/drafter.py, block/verifier.py
self.dtype = (
    torch.bfloat16 if self.device == "cuda" and supports_bf16() else torch.float32
)
```

The `self.device == "cuda"` half is kept deliberately. Dropping it would let an explicitly
CPU-configured run select `bfloat16` merely because an Ampere card exists in the machine — a new
defect in place of the fixed one.

### D — provenance warning

Silence is the defect, so the fallback is logged once per process, naming the device and its
capability. Placing it inside `supports_bf16()` rather than at each call site means it fires
exactly when a caller asked about `bfloat16` and the answer was "not natively", and it needs no
logger plumbing in the example files.

### B — declared floors in `testenv.yaml`

```yaml
testenv:
  min_compute_capability: "8.0"
```

Parsed into `(major, minor)` and checked in `TestEnv.prepare()`, which
`benchmarkingjob.py:89` calls before `build_testcases()`. That is the earliest point at which the
requirement can be evaluated, and it runs before the dataset is processed, so a run that cannot
proceed reports why before anything is downloaded. `ParadigmBase.__init__` would also have worked
but would have collided with #767; this placement touches neither.

Only a **cuda device below the floor** is an error. A run resolving to cpu is logged and allowed to
continue, because `use_gpu: false` asks for exactly that and a declared floor should not veto it.

**No example declares a floor in this PR, on purpose.** After item C the `bfloat16` path degrades to
float32 rather than failing, so declaring `8.0` for those examples would convert a safe fallback
back into a hard stop. The genuine floor also depends on the chosen backend — `vllm` needs 7.5, the
huggingface backend does not. The mechanism belongs in core; the values belong to whoever owns each
example.

### E — documentation

`cloud-edge-collaborative-inference-for-llm` listed only quantities under "Required Resources" —
gigabytes, core counts, a Python version — and no capability, so an 8 GB GTX 1080 satisfied every
line and still could not produce a valid result. Both floors are now stated, along with a note that
the reference results further down the page were produced with `tensor_parallel_size: 4` while the
section asks for one GPU.

`cloud-edge-speculative-decoding-benchmark` now states the `bfloat16` floor and the fallback, and
its results caveat mentions the compute dtype.

### Tests

31 cases, no GPU required. They stand in a fake torch module so every tier is exercised on any
runner: torch absent, no CUDA device, pre-Ampere (6.1), Turing (7.5), Ampere and newer
(8.0 / 8.6 / 9.0), warn-once behaviour, capability parsing including malformed input, and the
`testenv.yaml` wiring end to end.

One of them found a real bug during development: `parse_capability("")` reached `int("")` and
raised `invalid literal for int()` instead of the intended message. Fixed before this was pushed.

Verified against real hardware too — on a GTX 1080 `device_profile()` returns:

```json
{
  "device": "cuda",
  "name": "NVIDIA GeForce GTX 1080",
  "capability": [6, 1],
  "bf16_native": false,
  "total_memory_gb": 8.0
}
```

`pylint core` reports 9.97/10, unchanged from `main`; the only findings are the pre-existing
`R0917` in `dataset.py` that CI already works around. `TestEnv` crossed pylint's default
instance-attribute limit with the new field, suppressed in the same style the file already uses for
`too-few-public-methods`.

### Things reviewers may want to push back on

1. `cloud-edge-speculative-decoding-benchmark` previously imported nothing from `core`; this makes
   it the first such import there. It resolves in every supported execution path since ianvs runs
   from the repository root, and `cloud-edge-collaborative-inference-for-llm` already relies on the
   same thing. If you would rather that example stayed free of `core` imports, I can invert it so
   the capability arrives through `kwargs`.
2. `TestEnv.prepare()` as the validation point is a choice, not the only option. If you would prefer
   it in `ParadigmBase` alongside the `use_gpu` handling, that is a small move once #767 lands.

Addresses #888 items A, B, C, D and E. Item F (static CI check in the example validator) remains
open. Related: #767, #890.

### DIFF
--- a/core/common/device.py
+++ b/core/common/device.py
@@ -1,0 +1,127 @@
+# Copyright 2026 The KubeEdge Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Compute device capability reporting.
+
+Ianvs decides device *visibility* in `ParadigmBase` from the `use_gpu` setting.
+That answers whether a GPU is used. It does not answer whether the GPU can
+execute what an algorithm is about to ask of it, which is a separate question
+with its own failure mode: on hardware below compute capability 8.0, PyTorch
+emulates `bfloat16` instead of refusing it, so a benchmark completes and reports
+numbers produced by emulated arithmetic.
+
+This module reports the facts an algorithm needs to make that decision. It
+returns plain data rather than torch objects, so that `core` keeps no torch
+dependency of its own.
+
+See issue #888.
+"""
+
+from core.common.log import LOGGER
+
+BF16_MIN_MAJOR = 8
+"""First CUDA compute capability major version with native bfloat16 (Ampere)."""
+
+_EMULATION_REPORTED = False
+
+
+def device_profile():
+    """Report the resolved compute device and what it can natively execute.
+
+    torch is imported lazily because it is not a dependency of `core`; a
+    checkout without torch installed gets the cpu profile rather than an
+    ImportError.
+
+    Returns
+    ------
+    dict
+        device: str
+            "cuda" or "cpu".
+        name: str or None
+            Device name as reported by the driver, None on cpu.
+        capability: tuple of (int, int) or None
+            CUDA compute capability, None on cpu.
+        bf16_native: bool
+            Whether bfloat16 executes natively rather than through emulation.
+        total_memory_gb: float or None
+            Device memory in GiB, None on cpu.
+
+    """
+    # pylint: disable=import-outside-toplevel
+    # `core` does not depend on torch; see the module docstring.
+    try:
+        import torch
+    except ImportError:
+        return _cpu_profile()
+
+    if not torch.cuda.is_available():
+        return _cpu_profile()
+
+    index = torch.cuda.current_device()
+    properties = torch.cuda.get_device_properties(index)
+    major, minor = torch.cuda.get_device_capability(index)
+
+    return {
+        "device": "cuda",
+        "name": properties.name,
+        "capability": (major, minor),
+        "bf16_native": major >= BF16_MIN_MAJOR,
+        "total_memory_gb": round(properties.total_memory / 1024 ** 3, 2),
+    }
+
+
+def _cpu_profile():
+    """Profile reported when no CUDA device is usable."""
+    return {
+        "device": "cpu",
+        "name": None,
+        "capability": None,
+        "bf16_native": False,
+        "total_memory_gb": None,
+    }
+
+
+def supports_bf16():
+    """Report whether bfloat16 executes natively on the resolved device.
+
+    Deliberately derived from the compute capability rather than from
+    `torch.cuda.is_bf16_supported()`. That function defaults to
+    `including_emulation=True` and so answers True on hardware with no native
+    bfloat16, which is the trap this module exists to avoid.
+
+    Warns once per process when the answer is False on a CUDA device, because
+    that is the case where a caller would otherwise have selected an emulated
+    dtype and reported the resulting timings as a measurement.
+
+    Returns
+    ------
+    bool
+
+    """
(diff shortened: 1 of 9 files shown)


Hiteshsai007:
Splitting device **visibility** (`use_gpu`) from device **capability** (bf16 / compute-capability
floor) is the right model, and returning plain data keeps `core` torch-free. The `bf16_native`
decision being derived from the compute-capability major version — rather than
`torch.cuda.is_bf16_supported()` (which defaults to `including_emulation=True` and returns True on
hardware with no native bf16) — correctly avoids the emulated-arithmetic trap this module exists to
flag. The "declares a floor but resolved to cpu → warn and continue" choice is also sensible.

One issue, on the **`_EMULATION_REPORTED` flag**: it's a module-global that is set once and never
reset, i.e. "warn once per process." But one ianvs process runs **many TestCases in sequence**. After
the first emulated-bf16 case logs the warning, every later case on the same sub-capability GPU is
silently suppressed — nothing resets the flag between test cases. For a benchmark that's precisely
the silent failure this warning exists to prevent (emulated timings reported as native hardware
numbers), so warn-once-per-process can hide it for cases 2…N in a multi-example run.

Options:
- emit the warning **per TestCase/paradigm construction** rather than per process; or
- reset `_EMULATION_REPORTED` at the per-case boundary (e.g. around `testenv.prepare()`); or
- if once-per-process is intentional to cut log noise, also record the emulation event in the
  **per-case system-metric/leaderboard output**, so the per-case signal survives even when the log
  line doesn't repeat.

This is the same process-global-across-TestCases pattern as the `CUDA_VISIBLE_DEVICES` leak I
reported on #767 — process-level state that assumes a single construction.

vjuhhii:
You are right, and thank you — this is a real defect and an embarrassing one, since it reproduced in miniature exactly what the issue is about. TestCaseController.run_testcases() iterates self.test_cases in one process, so a once-per-process warning would announce the first affected measurement and pass silently over every one after it.

Fixed in e35bee7. I went simpler than any of your three options and removed the suppression outright: supports_bf16() now warns on every call that resolves to the emulated path. The call sits in model loading, not in a hot loop, so the volume is one warning per affected model load — which is the right granularity, because each one corresponds to a measurement the caveat applies to. Suppressing repeats was solving a noise problem I had invented.

The test that asserted "exactly one warning" now asserts one per call, with the reason in the docstring so nobody reintroduces the suppression as a tidy-up.

Your third option is the better long-term answer and I did not take it here: recording the emulation event in the per-case metrics puts the provenance next to the numbers instead of in a log that a reader may never see. That belongs in the results pipeline rather than bolted onto this PR, and I would rather propose it separately than grow a size/XL change further. If you would like to take that piece, it is unclaimed.

Issue #938, opened 2026-08-26
Misplaced rank children are dropped by BenchmarkingJob._parse_config, so PIPL's curated selected_dataitem never reaches the leaderboard


azaynul10:
## Summary

`Rank._parse_config` assigns only config keys that already exist in `self.__dict__` and
drops everything else without a warning. Every field `_check_fields()` guards has a
truthy default assigned before it runs, so a config whose `rank` keys are nested at the
wrong level validates cleanly and runs to completion on defaults.

`examples/PIPL/edge-cloud_collaborative_learning_bench/benchmarkingjob.yaml` is a live
instance in the shipped tree. Its `selected_dataitem` curates 3 modules and 9 metrics,
none of which reach `Rank`, so the published leaderboard silently contains every
discovered module and metric column instead.

**Corrected after review.** For this config the misplaced keys never reach
`Rank._parse_config` at all. They are dropped one level higher, by
`BenchmarkingJob._parse_config` at `benchmarkingjob.py:110`. The mechanism and the impact
are unchanged; what moves is where a fix has to go. Detail in the update section below.

This was found while reviewing #829 and is split out of it deliberately. #829 validates
the *value* of `save_mode`; this is about a validly-valued key never being delivered.
Diagnosis is shared with @Abhishek86798, who confirmed the misindentation independently
on that thread.

## Relationship to #742, #891 and #937

The same `if attribute in self.__dict__` filter exists in several `_parse_config`
implementations, and one of them is already under discussion, so it is worth separating
the two cases before anything here is read as a duplicate.

**#742 and #891, and @Nikhil-VS1811's analysis in #937, cover the `Dataset` class.** There
the key *is* present in `config.items()` but is not declared as an attribute in
`Dataset.__init__`, so it is skipped and the downstream paradigm code raises
`AttributeError` on the missing attribute. The signal is a crash, and the fix is to
declare the field.

**This issue is about the `Rank` class, and about the other half of the filter.** Here the
key is spelled correctly and *is* declared on the class, but it is nested at the wrong
level, so it never appears in `config.items()` at all and the filter never sees it. There
is no crash, because the class default is already assigned. The signal is a silently
wrong output artifact.

Those differ in the three ways that matter for a fix: **declaring the field does not help
here** because the field is already declared; **the failure is silent rather than an
`AttributeError`**, so nothing surfaces it; and the damage lands in a **published CSV**
rather than at load time. A patch for #742 leaves this open, and vice versa.

[...]

PR #998, opened 2026-09-03
fix(pipl): restore rank configuration nesting

diegodevxd:
/kind bug

## What this PR does / why we need it

Restores the PIPL benchmark rank configuration so its curated leaderboard columns are passed correctly to `Rank`.

## Root cause

`visualization` and `save_mode` were siblings of `rank`, while `selected_dataitem` was nested under `visualization`. `BenchmarkingJob._parse_config()` therefore did not forward these settings to `Rank`, causing it to use the default leaderboard selection.

## Changes

Moves `visualization`, `selected_dataitem`, and `save_mode` under the existing `rank` block without changing their configured values.

This PR intentionally scopes the fix to the shipped PIPL configuration. Core-level validation for misplaced or unknown configuration keys is left out of scope and can be addressed separately.

## Testing

- `python .github/workflows/validator/validation_runner.py --static --example examples/PIPL/edge-cloud_collaborative_learning_bench`
- YAML regression assertion comparing baseline and patched rank key sets
- `git diff --check`

Full PIPL end-to-end execution was not performed because its model/data environment is not available locally.

Addresses #938


diegodevxd:
Ready for review — the patch is unchanged since it was opened.

This is a config-only fix for #938: `visualization`, `selected_dataitem` and `save_mode` are restored as direct children of `rank` in `examples/PIPL/edge-cloud_collaborative_learning_bench/benchmarkingjob.yaml`, matching the structure used by the other examples (e.g. `examples/cifar100/fci_ssl/fed_ci_match/benchmarkingjob.yaml`). No values, ordering or unrelated YAML were changed — 1 file, +4 −4, indentation only.

Verified with PyYAML against the patched file:

```
rank keys        : [sort_by, visualization, selected_dataitem, save_mode]
metrics (9)      : 9
modules (3)      : 3
hyperparams (4)  : 4
```

This confirms the two keys `BenchmarkingJob._parse_config` was silently dropping now reach `Rank`, and that the curated `selected_dataitem` is delivered intact. `git diff --check` is clean and DCO passes.

/assign @moorezheng

